PR #19, opened 2019-03-19
Bounded In memory checkpointer

emmanueldenloye:


slpopejoy:
Closing. It's too bad we didn't do this in hindsight as it would have caught the problem with consensus that is plaguing OnDisk.

PR #67, opened 2019-03-26
[mempool] validation errors

slpopejoy:
@gregorycollins can you get this in on low priority? I gave up on the socket-based mempool, but it seems like `pInsert` in `toBackend` should first validate before handing it off to `issueMvCmd`


slpopejoy:
closing to revisit later

PR #72, opened 2019-03-26
[mempool,pactEx] mark mempool txs as confirmed when we call pactValidate

gregorycollins:


src/Chainweb/WebPactExecutionService.hs
```diff
@@ -52,3 +59,19 @@ mkPactExecutionService q = PactExecutionService
...
+-- consensus instead of just killing every tx that ever made it into a valid
+-- block
+markAllConfirmed
+    :: MempoolBackend ChainwebTransaction
+    -> PayloadWithOutputs
+    -> IO ()
+markAllConfirmed mempool payload = mempoolMarkConfirmed mempool txHashes
+  where
+    txcfg = mempoolTxConfig mempool
+    decodeTx = codecDecode $ txCodec txcfg
```
slpopejoy: This is fine I guess, but this is perhaps why Pact service should be doing this, as it has the decoded transactions and could straight give you the hashes. Up to you

src/Chainweb/WebPactExecutionService.hs
```diff
@@ -37,13 +44,13 @@ mkWebPactExecutionService hm = WebPactExecutionService $ PactExecutionService
...
-mkPactExecutionService :: TQueue RequestMsg -> PactExecutionService
-mkPactExecutionService q = PactExecutionService
+mkPactExecutionService :: MempoolBackend ChainwebTransaction -> TQueue RequestMsg -> PactExecutionService
+mkPactExecutionService mempool q = PactExecutionService
   { _pactValidateBlock = \h pd -> do
       mv <- validateBlock h pd q
       r <- takeMVar mv
       case r of
-        Right pdo -> return pdo
+        Right pdo -> markAllConfirmed mempool pdo >> return pdo
```
larskuhtz: So in case of a fork, the user would have to resubmit? Forks at depth 0 are not that rare. I guess, it is good enough for testnet?

src/Chainweb/WebPactExecutionService.hs
gregorycollins: I don't want this code to live in here past next week -- this was the smallest change set I could muster

src/Chainweb/WebPactExecutionService.hs
gregorycollins: It is good enough for test net. Will fix reintroduction tomorrow

larskuhtz: Would it be possible to change the is to something like `markAllConfirmedForBlock` and tag the confirmed transactions with the BlockHeader. 

That would allow to track when a transaction has to be re-submitted because the block ends up on a fork. It would be relatively straight forward to add the event of a block moving to a fork to the `WebBlockHeaderDB` BlockHeader Stream.

larskuhtz: Currently `WebBlockHeaderDB` emits a stream of `BlockHeader`. We could change that to a stream of something like `BlockHeaderEvent = NewBlockHeader BlockHeader | ForkBlockHeader BlockHeader`.

gregorycollins: I don't have much time but I have an algorithm for the mempool update already -- basically mempool has to watch the `Cut` tvar. When it sees an update, it walks both sides of the old and new cut back to the `greatestCommonAncestor`, accumulating tx lists as it goes. You reintroduce all of the tx hashes that were on the losing fork but didn't make it on the winning one.

As a result of doing this, the mempool consensus engine will publish a `TVar (Cut, Map ChainId (Vector ChainwebTransaction))`, and this is what miners will mine on. Does that sound ok to you

larskuhtz: Consensus is already doing this internally, computing forks. So, it would make sense to just publish this as an "ConsensusEvent" instead of expecting each client to recompute it. It's also useful information for other consumers and not just the mem-pool.

I am still not comfortable with the mempool deciding on which cut to mine on. Deciding the winner is inherently the job of consensus. Mempool should provide transactions for a given parent cut, but not chose the parent cut itself.

larskuhtz: Currently the flow is that the consensus publishes a new cut. The miner sees the cut and asks pact/mempool to create a block for that particular cut. I think, that's a simple and intuitive flow.

larskuhtz: It's the cut-processing pipeline that drives the consensus control flow. I think we shouldn't change that.

larskuhtz: But I see the benefit of allowing the mem-pool to preempt mining if it wants to update the payload for the currently mind block. Let's talk about that later.

PR #77, opened 2019-03-27
Doug/slow tests on master

mightybyte:


test/SlowTests.hs
```diff
@@ -29,7 +29,7 @@ main = defaultMain suite
 
 suite :: TestTree
 suite = testGroup "ChainwebSlowTests"
-    [ Chainweb.Test.MultiNode.test Warn (TestWithTime petersonChainGraph) 10 120 Nothing
+    [ Chainweb.Test.MultiNode.test Warn (TestWithTime petersonChainGraph) 10 300 Nothing
```
larskuhtz: Let's try and hope it's not flaky and too annoying.

If it turns out to be too long or flaky, we can just lower thresholds radically and run only for, say, 60 or 120 seconds.

larskuhtz: Eventually, we should also make the time and thresholds configurable. Unfortunately that's a bit tricky with tasty. But we could always create a separate non-tasty version of it or just drop the tasty version.

.gitlab-ci.yml
```diff
@@ -61,19 +64,23 @@ nix-linux:
...
 
+slow-tests:
+  stage: test
+  only:
+    - master
+  tags:
+    - linux
+  script:
+    - echo This is a master slow test run.
+    - result/ghc/chainweb/bin/slow-tests
```
larskuhtz: slow-tests does also contain ssl-certificate related tests that caused issues in nix-sandboxes in the past. Are we sure they work here? Otherwise, one can pass `--pattern=/consensus/` to slow-tests to only run the consensus tests.

mightybyte: This won't be a problem because it's being run directly, not in a nix sandbox.

test/SlowTests.hs
mightybyte: :+1:  I'll give this a try and we can re-adjust later.

PR #79, opened 2019-03-27
enable simple cors on all chainweb endpoints

larskuhtz:


eskimor:
It builds now - I have not yet checked whether it actually works. That part is not straight forward without an actual deployment. @larskuhtz - I will try to check locally with an http build, do you have the means to deploy this to some server for testing?

eskimor:
@larskuhtz I was able to test and what I was afraid of, we need to add header "content-type" to the list of allowed headers.

eskimor:
Actually 'content-type' is in `simpleHeaders`, but simpleHeaders is not included in `simpleCorsResourcePolicy` which seems incorrect.

larskuhtz:
If I remember correctly not all headers and in particular content types are allowed in simple cors. So we may need a non-simple cors policy in that case.

eskimor:
I dug a little deeper. simpleHeaders include content-type - but the simple policy does not include simpleHeaders, but none - relying on a default that is simple headers excluding content-type.

I still haven't read the standard yet, so there might be a reason for this unintuitive implementation.

larskuhtz:
Simple CORS requests are subject to the following constraint:

> If the request is a POST request the content type is constraint to simple content types (application/x-www-form-urlencoded, multipart/form-data, text/plain),

CORS is weird. In the standard they try to not break existing common HTTP scenarios, so their are a lot of special cases.

The complete list of constraints is in the documentation of wai-cors:

http://hackage.haskell.org/package/wai-cors-0.2.6/docs/Network-Wai-Middleware-Cors.html#v:simpleCorsResourcePolicy

larskuhtz:
This is the haddock documentation for cors headers:

> `corsRequestHeaders :: ![HeaderName]`	
> Field names of HTTP request headers that are allowed in CORS requests. Header names that are included in simpleHeaders, except for content-type, are implicitly included and thus optional in this list.

I didn't make that stuff up. It's actually in the standard. :-)

larskuhtz:
I think the easiest is if you take look at the documentation here http://hackage.haskell.org/package/wai-cors-0.2.6/docs/Network-Wai-Middleware-Cors.html#v:CorsResourcePolicy and define a CORS policy that meets your needs.

Once we know what we need I can either update the PR or you can push directly to it.

PR #80, opened 2019-03-27
Clarify README

fosskers:
This also removes all information regarding network orchestration, including Docker info and the `docker.nix` itself.


README.md
```diff
@@ -6,12 +6,16 @@ Read our whitepaper: [Chainweb: A Proof-of-Work Parallel-Chain Architecture for
 
 ### Building with Nix
```
slpopejoy: I do question this. 
- Our readme is basically a Nix advertisement
- it's a novel that noobs will absolutely mess up

README.md
```diff
@@ -20,6 +24,9 @@ substituters = http://nixcache.kadena.io https://pact.cachix.org https://nixcach
 trusted-public-keys = kadena-cache.local-1:8wj8JW8V9tmc5bgNNyPM18DYNA1ws3X/MChXh1AQy/Q= pact.cachix.org-1:cg1bsryGrHnQzqEp52NcHq4mBBL+R25XbR2Q/I/vQ8Y= ryantrinkle.com-1:JJiAKaRv9mWgpVAz8dwewnZe0AzzEAzPkagE9SP5NWI= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
 ```
 
+This will allow you to connect to Kadena's Nix cache, and pull pre-built
```
slpopejoy: Do we really want to be on the hook for supporting the world's chainweb nix builds?

README.md
```diff
@@ -28,13 +35,24 @@ cd chainweb-node
...
-Requirements:
+#### Dependencies
 
-- Cabal >= 2.2
-- GHC >= 8.4
-- (optional) Stack >= 1.9
+- `ghc >= 8.4` (Haskell compiler) and `cabal >= 2.2` (Haskell build-tool)
+  - [Linux / Mac](https://www.haskell.org/ghcup/)
+  - [Windows](https://chocolatey.org/search?q=ghc) (via *chocolatey*)
+- (optional) `stack >= 1.9` (Alternate Haskell build-tool, which installs GHC for you)
```
slpopejoy: Not sure why this is "optional" when lead developers on our team rely on it.

slpopejoy: In terms of spending time supporting non-kadena devs, stack is strictly superior.

README.md
```diff
@@ -151,204 +167,10 @@ chainweb-nodes and collecting the logs from all nodes:
 mkdir -p tmp/run-nodes-logs
 
 # the first argument is the path to the chainweb-node binary
-./scripts/run-nodes.sh ./chainweb-node 20 ./tmp/run-nodes-logs
+./scripts/run-nodes.sh ./chainweb-node 10 ./tmp/run-nodes-logs
```
slpopejoy: Why is `run-nodes` here? The heading is "Configuring a chainweb-node"

README.md
slpopejoy: Also, maybe not for this PR, we should mention binaries right about now. I'm going to look at throwing something up now.

fosskers: Yeah I'm also not sure about saying *Do Nix* at the very top.

README.md
fosskers: You tell me :)

README.md
fosskers: I'm with you on this one.

README.md
fosskers: You're right, `run-nodes.sh` should be mentioned in "Running Chainweb".

README.md
mightybyte: Yes, I think this is perfectly reasonable.  Supplying a nix cache is basically equivalent to supplying binaries.

README.md
mightybyte: I don't have a strong opinion on wording, but stack is certainly not mandatory.

README.md
fosskers: And with the CDN plan, this should be even easier.

PR #82, opened 2019-03-28
Feat/local execution

slpopejoy:
Done, but currently hanging after the first request ...


src/Chainweb/Pact/PactService.hs
```diff
@@ -175,7 +175,7 @@ initPactService' ver cid chainwebLogger spv act = do
     let !pd = P.PublicData def (unsafeGetChainId cid) def def
     let !pse = PactServiceEnv Nothing checkpointEnv spv pd
 
-    evalStateT (runReaderT act pse) theState
+    evalStateT (runReaderT act pse) (PactServiceState theState Nothing)
```
larskuhtz: This uses the lazy state monad. Is this intentional? Could it increase the risk of space leaks?

slpopejoy: @marklnichols any comment here -- both `PactService` and `Chainweb.Pact.Types` use lazy state. Was there a reason for this? I'm assuming it was an oversight.

Fixing.

slpopejoy: Great catch @larskuhtz

slpopejoy: Note doing strict state across the board @emmanueldenloye @marklnichols @emilypi . 

Also cleaning up some other warnings -- uninitialized variables in tests, and Lars' `putLog` (underscoring).

node/Utils/Logging.hs
```diff
@@ -507,7 +507,7 @@ withElasticsearchBackend mgr esServer ixName inner = do
         , HTTP.requestHeaders = [("content-type", "application/json")]
         }
 
-    putLog a = HTTP.defaultRequest
+    _putLog a = HTTP.defaultRequest
```
larskuhtz: 👍 Thanks @slpopejoy

PR #86, opened 2019-03-29
add 429 throttling to chainweb-node

larskuhtz:
This will most likely require some tweaking to find a good value. The current default may be too strict for running tests on low-latency local networks.


src/Chainweb/Chainweb.hs
```diff
@@ -325,6 +338,21 @@ withChainwebInternal conf logger peer payloadDb inner = do
                 mConf = _configMiner conf
                 mCutDb = _cutResCutDb cuts
 
+            -- initialize throttler
+            throttler <- initThrottler (defaultThrottleSettings $ TimeSpec 4 0)
+                { throttleSettingsRate = int $ _configThrottleRate conf
+                , throttleSettingsPeriod = 1 / micro
+                , throttleSettingsBurst = int $ 3 * _configThrottleRate conf
```
fosskers: > Number of concurrent requests allowed - should be greater than rate / period, defaults to 1

3,000 concurrent requests here?

larskuhtz: Oh, good point, I think I misread the documentation.

larskuhtz: Not sure, this is the default:

```Haskell
defaultThrottleSettings :: TimeSpec -> ThrottleSettings
defaultThrottleSettings expirationInterval = ThrottleSettings
  { throttleSettingsRate = 1
  , throttleSettingsPeriod = 1000000
  , throttleSettingsBurst = 1
  , throttleSettingsCacheExpiration = expirationInterval
  , throttleSettingsIsThrottled = const True
  , throttleSettingsOnThrottled = const $
      responseLBS status429 [("Content-Type", "application/json")] "{\"message\":\"Too many requests.\"}"
  }
```

I think, base on that the current code is correct and the documentation is miss-leading

fosskers: It seems to me that the `Burst` (3000 for us) shouldn't be higher than the `Rate` (1000 for us).

https://hackage.haskell.org/package/wai-middleware-throttle-0.3.0.0/docs/Network-Wai-Middleware-Throttle.html#t:ThrottleSettings

PR #87, opened 2019-03-29
Mark/remote test

marklnichols:


test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -0,0 +1,272 @@
...
+
+    putStrLn $ "Server listening on port: " ++ show thePort ++ ". Sending client request..."
+
+    env <- getClientEnv thePort cwVersion
+    tts <- sendTest env
+    return $ testGroup "PactRemoteTest" tts
+
+sendTest :: ClientEnv -> IO [TestTree]
+sendTest env = do
+    let msb = A.decode $ toS escapedCmd :: Maybe SubmitBatch
```
slpopejoy: use `decodeStrictOrThrow` from Chainweb.Utils

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -0,0 +1,272 @@
...
+tests :: IO TestTree
+tests = do
+    peerInfoVar <- newEmptyMVar
+    let cwVersion = TestWithTime petersonChainGraph
+    theAsync <- async $ runTestNodes Warn cwVersion nNodes Nothing peerInfoVar
+    link theAsync
+    newPeerInfo <- readMVar peerInfoVar
+    let thePort = _hostAddressPort (_peerAddr newPeerInfo)
+
+    putStrLn $ "Server listening on port: " ++ show thePort ++ ". Sending client request..."
```
larskuhtz: Let's avoid chatty tests. 

In other places we use HUnit progress settings as backend for the Chainweb-logger. In multi node tests there is some CPP setup to log to the console when compiled with `-dDEBUG`. So, those are options. 

But tests run in parallel and if each tests prints to the console it's not helpful always.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -0,0 +1,272 @@
...
+                        Right rsp -> do
+                            tt1 <- checkResponse "command-0" rks rsp
+                            return (tt0 : [tt1])
+
+getClientEnv :: Port -> ChainwebVersion -> IO ClientEnv
+getClientEnv thePort cwVersion = do
+    let mgrSettings = HTTP.mkManagerSettings (HTTP.TLSSettingsSimple True False False) Nothing
+    mgr <- HTTP.newTlsManagerWith mgrSettings
+    let url = testUrl thePort cwVersion "0.0" 8
+    -- putStrLn $ "URL: " ++ showBaseUrl url
```
larskuhtz: remove?

test/Chainweb/Test/Pact/RemotePactTest.hs
marklnichols: The test shuts down when a response is retrieved from poll, or after a fixed number of retries.  The time it takes running local is 2.1752 seconds with a 2 second wait before the poll.  Any lesser wait time actually makes it take longer due to poll failing, etc.

So everything else is inside the 0.1752 seconds...

Ok, I'll add the decodeStrictOrThrow and add the other stuff to another PR.

PR #88, opened 2019-03-29
skip redundant initialization of pact web-service

larskuhtz:


src/Chainweb/Pact/Service/PactInProcApi.hs
```diff
@@ -70,7 +70,7 @@ withPactService' ver cid logger memPoolAccess mv action = do
     reqQ <- atomically (newTQueue :: STM (TQueue RequestMsg))
     a <- async (PS.initPactService ver cid logger reqQ memPoolAccess mv)
     link a
-    initWebService reqQ (return ()) -- web service for 'local' requests not yet implemented
+    -- initWebService reqQ (return ()) -- web service for 'local' requests not yet implemented
```
slpopejoy: just delete :)

slpopejoy: also please delete `initWebService`

slpopejoy:
This is already addressed in #82

PR #90, opened 2019-04-01
Testnet stability improvements

larskuhtz:
Testnet stability:

* [x] fix POW miner
* [x] add an efficient POW miner that avoids memory allocations in the inner loop.
* [x] bound size of cut-queue

Logging and Debugging

* [x] add more logging to CutDb (we make it less chatty again once testnet is stable)
* [x] separate testing and production use in `Cut.hs` to help with debugging the code in that module.
* [x] add telemetry for the sizes of various internal queues.
* [x] make mem-pool less chatty
* [x] fix some log scope labels
* [x] log session statistics less often

House Keeping:

* [x] support recent version of `massiv`
* [x] fix TaskQueue unit tests


src/Data/DiGraph/FloydWarshall.hs
```diff
@@ -153,10 +166,10 @@ floydWarshallInternal
     -> Array U Ix2 (Double,Int)
 floydWarshallInternal a = foldl' go a [0..n-1]
   where
-    (n :. _) = size a
+    Sz (n :. _) = size a
```
fosskers: There is an incomplete pattern match here.

src/Data/DiGraph/FloydWarshall.hs
```diff
@@ -189,10 +202,10 @@ floydWarshall_
     -> Array U Ix2 Double
 floydWarshall_ a = foldl' go a [0..n-1]
   where
-    (n :. _) = size a
+    Sz (n :. _) = size a
```
fosskers: Here too.

src/Chainweb/CutDB.hs
```diff
@@ -274,25 +275,38 @@ processCuts
...
         -- ignore left values for now
     & S.scanM
-        (\a b -> give (_webBlockHeaderStoreCas headerStore) $ joinIntoHeavier_ (_cutMap a) b)
+        (\a b -> joinIntoHeavier_ (_webBlockHeaderStoreCas headerStore) (_cutMap a) b
+        )
         (readTVarIO cutVar)
-        (\c -> atomically (writeTVar cutVar c) >> logFun @T.Text Debug "write new cut")
+        (\c -> do
+            atomically (writeTVar cutVar c)
+            s <- pQueueSize queue
```
fosskers: This `s` is unused.

src/Data/DiGraph/FloydWarshall.hs
larskuhtz: I don't think it's incomplete. It think the pattern definition is just missing the `complete` pragma. Let me see if there is another backward compatible way to patch `n` out of the size...

larskuhtz: Well, it's also only a problem with `massiv-0.2.7`. Version 0.3 doesn't seem to have that problem, but `cabal2nix` fails on version 0.3 :-(

scripts/run-nodes.sh
```diff
@@ -96,11 +96,13 @@ function run-node () {
             --hostname=127.0.0.1 \
             --node-id=$NID \
             --test-miners=$N \
-            --chainweb-version=testWithTime \
+            --chainweb-version=testnet00 \
```
fosskers: Likely problematic, here and below.

src/Chainweb/CutDB.hs
```diff
@@ -189,8 +188,10 @@ _cut = readTVarIO . _cutDbCut
 cut :: Getter (CutDb cas) (IO Cut)
 cut = to _cut
 
+-- | FIXME: make limit configurable
```
fosskers: Should this be done?

src/Chainweb/CutDB.hs
```diff
@@ -274,25 +275,37 @@ processCuts
     -> TVar Cut
     -> IO ()
 processCuts logFun headerStore payloadStore queue cutVar = queueToStream
+    & S.chain (\_ -> logFun @T.Text Info "start processing new cut")
```
fosskers: How much spam is this likely to produce?

src/Chainweb/Miner/POW.hs
```diff
@@ -132,120 +156,245 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
...
 
-    -- | INVARIANT: A new cut, chain, and parent header is reselected after each
-    -- hash failure. This ensures that the Cut the miner is working on doesn't
-    -- grow stale, and cause forks. Without this condition (or a similar one),
-    -- the miners cause forks quite aggressively.
+awaitNextCut :: CutDb cas -> Cut -> IO Cut
+awaitNextCut cutDb c = atomically $ do
+    c' <- _cutStm cutDb
+    when (c' == c) retry
+    return c'
```
fosskers: So, this allows the Miner to continue mining only so long as the `Cut` he has is the newest one. Instead of polling for a new one all the time like before, this approach is "event" based. Okay, cool.

src/Chainweb/Miner/POW.hs
```diff
@@ -132,120 +156,245 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
...
+    --
+    case getAdjacentParents c p of
+
+        Nothing -> mineCut logfun conf nid cutDb c nonce adjustments
+            -- spin until a chain is found that isn't blocked
+
+        Just adjParents -> do
+
+            -- get payload
+            let mokPact = False
```
fosskers: This line can be removed by now.

src/Chainweb/Miner/POW.hs
```diff
@@ -132,120 +156,245 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
...
+                        [ (Transaction "testTransaction", TransactionOutput "testOutput")
+                        ]
+
+            -- get target
+            --
+            T2 target adjustments' <- getTarget cid p adjustments
+
+            -- Assemble block without Nonce and Timestamp
+            --
+            let candidateHeader = testBlockHeader'
```
fosskers: The `testBlockHeader*` functions should have their names updated.

src/Chainweb/Miner/POW.hs
```diff
@@ -132,120 +156,245 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
...
+                    (Nonce 0) -- preliminary
+                    target
+                    epoche -- preliminary
+                    p
+
+            newHeader <- (usePowHash v mineFast) candidateHeader nonce
+
+            -- create cut with new block
+            --
+            -- This is expecte to succeed, since the cut invariants should
```
fosskers:
```suggestion
            -- This is expected to succeed, since the cut invariants should
```

src/Chainweb/Miner/POW.hs
```diff
@@ -132,120 +156,245 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
...
+    tryAdj h b
+        | _blockHeight b == h = Just $! _blockHash b
+        | _blockHeight b == h + 1 = Just $! _blockParent b
+        | otherwise = Nothing
+
+-- | Run a mining loop. Updates creation time and nonce.
+-- The result is guaranteed to have a valid nonce.
+--
+mine :: BlockHeader -> Nonce -> IO BlockHeader
+mine h nonce = do
```
fosskers: Unused?

src/Data/DiGraph/FloydWarshall.hs
```diff
@@ -46,6 +51,14 @@ import GHC.Generics
 
 import Numeric.Natural
 
+-- -------------------------------------------------------------------------- --
+-- Massiv backward compatibility
+
+#if ! MIN_VERSION_massiv(0,3,0)
+maximum' :: (Source r ix e, Ord e) => Array r ix e -> e
+maximum' = M.maximum
+#endif
```
fosskers: Shouldn't need this anymore.

src/P2P/TaskQueue/Test.hs
```diff
@@ -47,7 +48,7 @@ newtype TestRunnerException = TestRunnerException Int
 instance Exception TestRunnerException
 
 testRunner :: AttemptsCount -> PQueue (Task Int a) -> IO ()
-testRunner limit q = forM_ [0..] $ session limit q (\_ _ -> return ())
+testRunner limit q = forM_ [0..] $ session limit q (\_ _ -> yield >> return ())
```
fosskers: `void yield`?

scripts/run-nodes.sh
larskuhtz: Oh, good catch! this was only for debugging, it wasn't meant to be submitted.

src/Chainweb/CutDB.hs
larskuhtz: I wish it would produce a lot :-). Unfortunately, this is where the testnet nodes stall. So, I put in more logging here. I'll remove it again once Testnet is stable.

Overall, it doesn't produce too much logs. Generally, logging isn't a concern at this point. A concern might be how much data we store in ES, but that should be taken care of by proper configuration of ES. As long as logging doesn't slow down the Chainweb-nodes themself too much, we shouldn't care.

src/Chainweb/Miner/POW.hs
larskuhtz: Yes. Should we delete it and only keep the fast miner? I was wondering if this function may still be useful for testing and debugging.

src/Chainweb/Miner/POW.hs
larskuhtz: Yes, I think, for testing one should provide a mock `PactExcecutionService`.

src/Chainweb/Miner/POW.hs
fosskers: I think we should just keep the fast miner.

PR #91, opened 2019-04-03
request logger for chainweb-node

larskuhtz:


src/Chainweb/Utils/RequestLog.hs
```diff
@@ -0,0 +1,138 @@
...
+    , _requestResponseLogDurationMicro = d
+    }
+
+-- | NOTE: this middleware should only be used for APIs that don't stream. Otherwise
+-- the logg may be delayed for indefinite time.
+--
+requestResponseLogger :: Logger l => l -> Middleware
+requestResponseLogger logger app req respond = do
+    let !reqLog = logRequest req
+    reqTime <- getTime Monotonic
```
gregorycollins: Use the coarse monotonic clock: it's ~6x faster (as of the last time I benchmarked it), and you don't need nanosecond-level accuracy here

src/Chainweb/Utils/RequestLog.hs
```diff
@@ -0,0 +1,138 @@
...
+    app req $ \res -> do
+        resTime <- getTime Monotonic
+        logFunctionJson logger Info
+            $ logRequestResponse reqLog res
+            $ (int $ toNanoSecs $ diffTimeSpec resTime reqTime) `div` 1000
+        respond res
```
gregorycollins: So the time here is just processing time but doesn't cover the amount of time necessary to send the response body. Should you collect another time delta after `respond` finishes?

src/Chainweb/Utils/RequestLog.hs
larskuhtz: Accuracy is sufficient (~100 micro would be more than enough)?

I was just to lazy to check :-)

src/Chainweb/Utils/RequestLog.hs
larskuhtz: Oops, good catch. I'll fix it.

Usually, one wants both, "time to first byte" and "total time". One would also want to count the response bytes, but that requires some special casing on the different response body types. I don't think we have long-running streaming endpoints or very data heavy endpoints. So, let's keep it simple and add features as we need them.

src/Chainweb/Utils/RequestLog.hs
larskuhtz: It's seems that the `MonotonicCoarse` constructor is only available on Linux. Let's just use `Monotonic` for now.

PR #92, opened 2019-04-03
Age-aware Mining

fosskers:
Currently, if fresh/out-of-date nodes are introduced to the network, they spend a fair amount of time getting caught up. Meanwhile, they continue to mine old blocks and yield `Cut` information into the network. This is wasted work, as any blocks they produce will never be considered by the main "Consensus Pack". 

This PR allows the POW miner to detect its distance from the main Consensus. If too far, it will halt mining until caught up.


src/Chainweb/BlockHeader.hs
```diff
@@ -174,6 +174,16 @@ instance IsMerkleLogEntry ChainwebHashTag BlockHeight where
...
 
+instance Semigroup BlockHeight where
+    BlockHeight a <> BlockHeight b = BlockHeight $ a + b
+    {-# INLINE (<>) #-}
+
+instance Monoid BlockHeight where
+    mempty = 0
+
+    mappend = (<>)
+    {-# INLINE mappend #-}
```
fosskers: Potentially dishonest to force the `Semigroup` instance to be additive, but Lens requires a `Monoid`, and we don't have `deriving via` available to us yet.

src/Chainweb/CutDB.hs
```diff
@@ -146,6 +148,7 @@ defaultCutDbConfig v = CutDbConfig
 data CutDb cas = CutDb
     { _cutDbCut :: !(TVar Cut)
     , _cutDbQueue :: !(PQueue (Down CutHashes))
+    , _cutNetworkCutHeight :: !(TVar (Maybe BlockHeight))
```
fosskers: `Maybe`, so that anyone listening to this `TVar` can avoid special logic during initial network conditions (i.e. before any `Cut` information has come in).

src/Chainweb/CutDB.hs
```diff
@@ -297,31 +303,41 @@ processCuts logFun headerStore payloadStore queue cutVar = queueToStream
         )
     & S.effects
   where
+    -- | Broadcast the newest estimated `BlockHeight` of the network to other
+    -- components of this node.
+    --
+    updateNetworkHeight :: CutHashes -> IO ()
+    updateNetworkHeight = atomically . writeTVar networkHeight . Just . cutHashesHeight
```
fosskers: @larskuhtz is there a chance that a `Cut` could come in that would significantly drop the agreed-upon Cut Height?

src/Chainweb/BlockHeader.hs
larskuhtz: Normally, one would use a `newtype` wrapper in cases like this. In this case an additive Semigroup seems fine. It measure the amount of work that went into a block. Addition seems the natural operation for this. Alternatives might be `Min` or `Max`, but definitely not multiplication. I think, addition is fine.

src/Chainweb/CutDB.hs
larskuhtz: Would `TMVar` work here?

src/Chainweb/Cut/CutHashes.hs
```diff
@@ -93,3 +101,7 @@ cutToCutHashes p c = CutHashes
     , _cutHashesChainwebVersion = _chainwebVersion c
     }
 
+-- | The "Cut Height" represented by the given `CutHashes`.
+--
+cutHashesHeight :: CutHashes -> BlockHeight
+cutHashesHeight = view (cutHashes . each . _1)
```
larskuhtz: There is already a `_cutHashesHeight` field in `CutHashes. Also, do you need a `Lens` or would a `Getter` be enough?

src/Chainweb/CutDB.hs
larskuhtz: Should the name of this field somehow indicate that this is just an *expected* value, or an educated guess of what the consensus is. For instance `approxNetworkCutHeight` or `expectedMinNetworkCutHeight`?

larskuhtz: While using this information can greatly improve performance, it has also great potential for being abused. It gives an attacker the ability to remotely switch off mining, by sending invalid cuts of large block height.I think, the name should somehow reflect that it should be used with care.

Going forward, we may factor in the expected block height based on block rate and time of genesis to predict if the expectation is reasonable. And, of course, we should add some sort of reputation management, but that alone wouldn't be enough. Once we have persistence implemented, we'll probably want a configuration option to turn this feature on and of manually.

larskuhtz: What about using the average (or some regression) of all cuts in the queue? That could mitigate some of the risk. Well, we would still have to make sure that the cuts in the queue represent a good mix of origin, for which we would need some Kademlia-style partitioning/classification of nodes (both for GET and PUT).

src/Chainweb/CutDB.hs
```diff
@@ -297,31 +303,41 @@ processCuts logFun headerStore payloadStore queue cutVar = queueToStream
...
     queueToStream = do
         Down a <- liftIO (pQueueRemove queue)
         S.yield a
         queueToStream
 
+    isVeryOld :: CutHashes -> IO Bool
     isVeryOld x = do
         h <- _cutHeight <$> readTVarIO cutVar
-        let r = int (_cutHashesHeight x) <= (int h - threshold)
+        let !r = int (_cutHashesHeight x) <= (int h - threshold)
```
larskuhtz: I don't think that the BangPattern is needed here, since we discriminate on the value in next line. But it doesn't hurt neither.

src/Chainweb/CutDB.hs
```diff
@@ -370,6 +386,32 @@ cutHashesToBlockHeaderMap headerStore payloadStore hs = do
...
+-- implied by the given `ChainwebVersion`. In essence, if our current Cut is...:
+--
+--   * below the threshold: Stop mining.
+--   * above the threshold: We have joined the pack of honest miners, and can mine.
+--   * even higher than the network: We are either a "superior" fork, or we are in
+--     initial network conditions where there is no real consensus yet. In this
+--     case, we keep mining.
+--
+hasJoinedConsensus :: ChainwebVersion -> CutDb cas -> IO Bool
+hasJoinedConsensus v cutdb = readTVarIO (_cutNetworkCutHeight cutdb) >>= \case
```
larskuhtz: As lined out above, eventually, we should base our assessment of the current consensus on more than just the tip of the queue. For now it's fine.

src/Chainweb/Cut/CutHashes.hs
fosskers: Woops I'm an idiot.

src/Chainweb/CutDB.hs
fosskers: I've done the renaming.

src/Chainweb/CutDB.hs
larskuhtz: There will be incoming cuts with very low height, but the agreed upon height should never decrease. Cuts of very low height should be filtered out immediately by the cut processor.

larskuhtz: There is some interference with #161 here: with the changes from that PR, if a cuts that are far ahead, would be dropped, too. So we would see in the processor only cuts that are at most 1000 blocks ahead. Once we get close to those we would start mining, even though the network is much further ahead.

One solution could be to sample the current height after we drop old cuts and before we drop far-ahead cuts.

larskuhtz:
What about keeping a little bit of state and compute the current height by more than a single incoming cut. E.g. we could keep track of the maximum for each node/origin and take some percentile of that. For example: the 90 percentile or the minimum of the max of the top 5 nodes, whatever, is less.

fosskers:
@larskuhtz Yes I think something like that will be better for preventing attacks / being more accurate. I'm thinking about design now.

src/Chainweb/CutDB.hs
```diff
@@ -266,26 +284,28 @@ startCutDb
 startCutDb config logfun headerStore payloadStore = mask_ $ do
     cutVar <- newTVarIO (_cutDbConfigInitialCut config)
     queue <- newEmptyPQueue
-    cutAsync <- asyncWithUnmask $ \u -> u $ processor queue cutVar
+    peerHeights <- newTVarIO . PeerHeights $ BQ.singleton 16 (BlockHeight 0)
```
fosskers: Unsure how big the Queue should be.

src/Chainweb/CutDB.hs
```diff
@@ -329,6 +351,18 @@ processCuts logFun headerStore payloadStore queue cutVar = queueToStream
...
+    -- components of this node.
+    --
+    -- Ignores any Cuts that came into the pipeline from /this/ node, i.e.
+    -- `CutHashes` whose origin field is `Nothing`.
+    --
+    updateNetworkHeight :: CutHashes -> IO ()
+    updateNetworkHeight ch = case _cutOrigin ch of
+        Nothing -> pure ()
+        Just _ -> atomically . modifyTVar' peerHeights $
+            \(PeerHeights q) -> PeerHeights $ BQ.cons (_cutHashesHeight ch) q
```
fosskers: `O(1)` consing + "forgetting" of oldest values.

stack.yaml
```diff
@@ -5,7 +5,9 @@ packages:
 
 extra-deps:
   - aeson-1.4.3.0
+  - base-orphans-0.8.1@sha256:defd0057b5db93257528d89b5b01a0fee9738e878c121c686948ac4aa5dded63,2927
```
fosskers: What are you doing there...

src/Chainweb/CutDB.hs
emilypi: Why not use `TBQueue` in this case?

src/Chainweb/Miner/POW.hs
```diff
@@ -102,22 +102,23 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
...
 
-    go
-        :: Given WebBlockHeaderDb
-        => Given (PayloadDb cas)
-        => MWC.GenIO
-        -> Int
-        -> Adjustments
-        -> IO ()
+    go :: Given WebBlockHeaderDb
+       => Given (PayloadDb cas)
```
emilypi: Could we purge these `Given` values?

src/Chainweb/CutDB.hs
```diff
@@ -480,6 +519,33 @@ cutHashesToBlockHeaderMap headerStore payloadStore hs = do
...
+        !thresh = int . catchupThreshold $ _chainwebVersion cutdb
+        !mini = bool (nh - thresh) 0 $ thresh > nh
+    when (currentHeight < mini) retry
+    pure cur
+
+-- | The distance from the true Cut within which the current node could be
+-- considered "caught up".
+--
+catchupThreshold :: ChainwebVersion -> Natural
+catchupThreshold = (2 *) . diameter . _chainGraph
```
emilypi: It'd probably be easier to inline this here or as a data entry in the chainweb version instead of calculating this every time we  need to establish  a consensusCut. That way, we can think of chainweb versions as having a tuneable "catchup threshold" built in.

fosskers: Memoize?

src/Chainweb/Miner/POW.hs
fosskers: Likely? @larskuhtz ?

src/Chainweb/CutDB.hs
fosskers: There's no way to fold over a `TBQueue`.

fosskers:
Closing for now - the stand-alone mining client will likely make all of this defunct.

PR #96, opened 2019-04-05
Cut Id

larskuhtz:


src/Chainweb/CutDB.hs
```diff
@@ -300,10 +300,13 @@ processCuts logFun headerStore payloadStore queue cutVar = queueToStream
         (readTVarIO cutVar)
         (\c -> do
             atomically (writeTVar cutVar c)
-            logFun @T.Text Info "write new cut"
+            loggc Info c "published cut"
         )
     & S.effects
   where
+    loggc :: HasCutId c => LogLevel -> c -> T.Text -> IO ()
+    loggc l c msg = logFun @T.Text l $  "cut " <> cutIdToTextShort (_cutId c) <> ": " <> msg
```
fosskers: The `_cutId` call is repeated a handful of times for the same `Cut`.

PR #100, opened 2019-04-07
Remove unsafe use of unsafeChainId

larskuhtz:
This PR removes almost all uses of `unsafeChainId`. In most cases the unsafe use was not needed, because the context already supported safe creation of `ChainId` values. This indicates that from the code it was not obvious how to safely construct of `ChainId` values.

This PR addresses this in two ways: It make safe construction of `ChainId` values more convenient and it improves the documentation.

The following new functions are added:

* [x] `chainIds :: HasChainwebVersion v => v -> HS.HashSet ChainId`
* [x] `mkChainId :: MonadThrow m => HasChainwebVersion v => Integral i => v -> i -> m ChainId`
* [x] `someChainId :: HasCallStack => HasChainwebVersion v => v -> ChainId`

The type of the function `randomChainId` is changed to

* [x] `randomChainId :: HasChainwebVersion v => v -> IO ChainId`

All uses of `unsafeChainId` in tests and examples are replaced by above functions.

Updated documentation of `ChainId`:

```Haskell
-- | ChainId /within the context a Chainweb instance/.
--
-- The set of valid ChainIds is determined by the 'ChainwebVersion'. In almost
-- all use cases there should be a context that is an instance of
-- 'HasChainwebVersion' and `chainIds' can be used get the set of chain ids.
--
-- In the context of a particular chain the 'ChainId' can be obtained via
-- instances of 'HasChainId'.
--
-- /How to create values of type 'ChainId'/
--
-- * To fold or traverse over all chain ids use 'chainIds'
-- * To deserialize a chain id use 'mkChainId'.
-- * For random chain id consider using 'randomChainId'.
-- * For some arbitrary but fixed chain id consider using 'someChainId'.
```

### DIFF
--- a/examples/TransactionGenerator.hs
+++ b/examples/TransactionGenerator.hs
@@ -68,6 +68,7 @@ import Chainweb.Simulate.Contracts.HelloWorld
 import Chainweb.Simulate.Contracts.SimplePayments
 import Chainweb.Simulate.Utils
 import Chainweb.Pact.RestAPI
+import Chainweb.Version
 
 data TransactionCommand = NoOp | DeployContracts [FilePath] | RunStandardContracts | RunSimpleExpressions | PollRequestKeys [ByteString]
   deriving (Show, Eq, Read, Generic)
@@ -141,7 +142,7 @@ defaultTransactionConfig :: TransactionConfig
 defaultTransactionConfig =
   TransactionConfig
     { _scriptCommand      = DeployContracts []
-    , _nodeChainId        = unsafeChainId 0
+    , _nodeChainId        = someChainId Testnet00
     , _serverRootPrefix   = "https://us1.chainweb.com"
     , _isChainweb         = True
     , _chainwebNodePort   = ChainwebPort 443
--- a/src/Chainweb/ChainId.hs
+++ b/src/Chainweb/ChainId.hs
@@ -17,7 +17,7 @@
 -- Maintainer: Lars Kuhtz <lars@kadena.io>
 -- Stability: experimental
 --
--- TODO
+-- The defininitions in this module are also exported via "Chainweb.Version".
 --
 module Chainweb.ChainId
 ( ChainIdException(..)
--- a/src/Chainweb/ChainId.hs
+++ b/src/Chainweb/ChainId.hs
@@ -89,18 +89,21 @@ instance Exception ChainIdException
 -- -------------------------------------------------------------------------- --
 -- ChainId
 
--- | ChainId /within a Chainweb/.
+-- | ChainId /within the context a Chainweb instance/.
 --
--- Generally a block chain is /globally/ uniquely identified by its genesis hash.
--- This type only uniquely identifies the chain /locally/ within the context of
--- a chainweb.
+-- The set of valid ChainIds is determined by the 'ChainwebVersion'. In almost
+-- all use cases there should be a context that is an instance of
+-- 'HasChainwebVersion' can be used get the set of chain ids.
 --
--- However, the chainweb context is globally uniquely identified by the
--- 'ChainwebVersion', which in turn is determined by the identities of the
--- chains in the chainweb. Since the chainweb topology is statically represented
--- on the type level, while the Chainweb version is a runtime value, we break
--- the cycle by using the Chainweb version as globally unique identifier and
--- include the 'ChainwebVersion' and the 'ChainId' into the genesis hash.
+-- In the context of a particular chain the respective 'ChainId' can be obtained
+-- via instances of 'HasChainId'.
+--
+-- /How to create values of type 'ChainId'/
+--
+-- * To fold or traverse over all chain ids, use 'chainIds'.
+-- * To deserialize a chain id, use 'mkChainId'.
+-- * For random chain id consider using 'randomChainId'.
+-- * For some arbitrary but fixed chain id consider using 'someChainId'.
 --
 newtype ChainId :: Type where
     ChainId :: Word32 -> ChainId
@@ -233,8 +236,8 @@ instance SingKind ChainIdT where
 -- -------------------------------------------------------------------------- --
 -- Testing
 
--- | Generally, the 'ChainId' is determined by the genesis block of a chain for
--- a given 'Chainweb.Version'. This constructor is only for testing.
+-- | This function should be be rarely needed. Please consult the documentation
+-- of 'ChainId' for alternative ways for obtain 'ChainId' values.
 --
 unsafeChainId :: Word32 -> ChainId
 unsafeChainId = ChainId
@@ -245,3 +248,4 @@ unsafeGetChainId (ChainId cid) = cid
 
 instance Arbitrary ChainId where
     arbitrary = unsafeChainId <$> arbitrary
+
--- a/src/Chainweb/ChainId.hs
+++ b/src/Chainweb/ChainId.hs
@@ -89,7 +89,7 @@ instance Exception ChainIdException
 -- -------------------------------------------------------------------------- --
 -- ChainId
 
--- | ChainId /within the context a Chainweb instance/.
+-- | ChainId /within the context of a Chainweb instance/.
 --
 -- The set of valid ChainIds is determined by the 'ChainwebVersion'. In almost
 -- all use cases there should be a context that is an instance of
@@ -102,7 +102,7 @@ instance Exception ChainIdException
 --
 -- * To fold or traverse over all chain ids, use 'chainIds'.
 -- * To deserialize a chain id, use 'mkChainId'.
--- * For random chain id consider using 'randomChainId'.
+-- * For a random chain id consider using 'randomChainId'.
 -- * For some arbitrary but fixed chain id consider using 'someChainId'.
 --
 newtype ChainId :: Type where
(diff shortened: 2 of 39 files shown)


examples/TransactionGenerator.hs
```diff
@@ -141,7 +142,7 @@ defaultTransactionConfig :: TransactionConfig
 defaultTransactionConfig =
   TransactionConfig
     { _scriptCommand      = DeployContracts []
-    , _nodeChainId        = unsafeChainId 0
+    , _nodeChainId        = someChainId Testnet00
```
larskuhtz: Should the chainweb version be configurable? If so, the default settings for `_serverRootPrefix` and `_chainwebNodePort` could be derived from the boots.trap nodes of the respective version.

test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -65,7 +66,7 @@ pactTestSetup :: IO PactTestSetup
 pactTestSetup = do
     let loggers = alwaysLog
     let logger = newLogger loggers $ LogName "PactService"
-    let pactCfg = pactDbConfig (Test petersonChainGraph)
+    let pactCfg = pactDbConfig testVersion
```
larskuhtz: `pactExecTests` below uses `Testnet00` as chainweb version. I assume that mixing different chainweb version in throughout these tests was not intended.

fosskers:
This is great. Reviewing soon.

examples/TransactionGenerator.hs
fosskers: Almost certainly. @emmanueldenloye uses `TestWithTime` locally on his machine.

src/Chainweb/ChainId.hs
```diff
@@ -233,8 +236,8 @@ instance SingKind ChainIdT where
 -- -------------------------------------------------------------------------- --
 -- Testing
 
--- | Generally, the 'ChainId' is determined by the genesis block of a chain for
--- a given 'Chainweb.Version'. This constructor is only for testing.
+-- | This function should be be rarely needed. Please consult the documentation
+-- of 'ChainId' for alternative ways for obtain 'ChainId' values.
```
fosskers:
```suggestion
-- of 'ChainId' for alternative ways to obtain 'ChainId' values.
```

src/Chainweb/Version.hs
```diff
@@ -350,3 +360,38 @@ class HasChainwebVersion a where
...
+    cid = unsafeChainId (fromIntegral i)
+{-# INLINE mkChainId #-}
+
+-- | Sometimes, in particular for testing and examples, some fixed chain id is
+-- needed, but it doesn't matter which one. This function provides some valid
+-- chain id.
+--
+someChainId :: HasCallStack => HasChainwebVersion v => v -> ChainId
+someChainId = head . toList . chainIds
+    -- 'head' is guaranteed to succeed because the empty graph isn't valid chain
```
fosskers:
```suggestion
    -- 'head' is guaranteed to succeed because the empty graph isn't a valid chain
```

src/Chainweb/Version.hs
```diff
@@ -350,3 +360,38 @@ class HasChainwebVersion a where
...
+    -- graph.
+{-# INLINE someChainId #-}
+
+-- | Uniformily get a random ChainId
+--
+randomChainId :: HasChainwebVersion v => v -> IO ChainId
+randomChainId v = (!!) (toList cs) <$> randomRIO (0, length cs - 1)
```
fosskers: Naughty `O(n)` lookup once we've expanded to 1,000 chains ;)

test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -45,16 +45,17 @@ import Pact.Types.Logger
...
 import Chainweb.Pact.PactService
 import Chainweb.Pact.Types
 import Chainweb.Test.Pact.Utils
-import Chainweb.Version (ChainwebVersion(..))
-import Chainweb.ChainId
+import Chainweb.Version (ChainwebVersion(..), someChainId)
 import Chainweb.BlockHash
 
+testVersion :: ChainwebVersion
+testVersion = Testnet00
```
fosskers: Are you sure?

tools/ea/Ea.hs
```diff
@@ -170,7 +166,7 @@ genPayloadModule v txFiles = do
 
     let logger = genericLogger Warn TIO.putStrLn
 
-    payloadWO <- initPactService' Testnet00 (unsafeChainId 0) logger noSPVSupport $
+    payloadWO <- initPactService' Testnet00 (someChainId Testnet00) logger noSPVSupport $
```
fosskers: Hm I wonder when `Testnet00` snuck in here, since the `ChainwebVersion` is specifiable on the command line.

src/Chainweb/ChainId.hs
```diff
@@ -89,18 +89,21 @@ instance Exception ChainIdException
...
--- chains in the chainweb. Since the chainweb topology is statically represented
--- on the type level, while the Chainweb version is a runtime value, we break
--- the cycle by using the Chainweb version as globally unique identifier and
--- include the 'ChainwebVersion' and the 'ChainId' into the genesis hash.
+-- In the context of a particular chain the respective 'ChainId' can be obtained
+-- via instances of 'HasChainId'.
+--
+-- /How to create values of type 'ChainId'/
+--
+-- * To fold or traverse over all chain ids, use 'chainIds'.
```
slpopejoy: Add a note that all of these functions are in `Chainweb.Version`.

Better would be to consider how to make a single import, since at this point you'll be importing `Chainweb.ChainId` just for the type and `Chainweb.Version` for everything useful. Consider 
- combining `Chainweb.ChainId` and `Chainweb.Version` modules
- exporting all of ChainId's API out of `Chainweb.Version`
- Renaming `Version` (kind of uninspiring name for an entire Chainweb network) to justify including the `ChainId` API. `NetworkId`or `ChainwebId` or something

test/Chainweb/Test/Pact/PactExec.hs
slpopejoy: Correct

test/Chainweb/Test/Pact/PactExec.hs
slpopejoy: This is correct

tools/ea/Ea.hs
slpopejoy: Doubt it snuck in, my initial payload gen was hardcoded to Testnet00.

In truth we'll be generating payloads uniquely for each chain at some point, so this code will evolve.

test/Chainweb/Test/Pact/PactInProcApi.hs
```diff
@@ -171,9 +173,9 @@ checkBlockTransactions filePrefix bTrans = do
 
     return $ testGroup "BlockTransactions" $ ttTrans : [ttTransHash]
 
-getBlockHeaders :: Int -> [BlockHeader]
-getBlockHeaders n = do
-    let gbh0 = genesis
+getBlockHeaders :: ChainId -> Int -> [BlockHeader]
+getBlockHeaders cid n = do
```
slpopejoy: I know this isn't your code but can you clean this up? The `do` is driving me crazy.

src/Chainweb/ChainId.hs
larskuhtz: > exporting all of ChainId's API out of Chainweb.Version

Yes, that's a good idea.

> Renaming Version (kind of uninspiring name for an entire Chainweb network)

That's a longstanding issue. It was brought up during the very first Chainweb-code review, but it never was high priority. `NetworkId` is already taken by the P2P network (we would have to rename the type there first to avoid confusion). What about `ChainwebInstance`?

Changing the name would be straight forward, but would touch a lot of code. So, maybe we should do it during vacation time or some holiday when there are not a lot of open PRs?

test/Chainweb/Test/Pact/PactInProcApi.hs
larskuhtz: :-) will do

src/Chainweb/ChainId.hs
```diff
@@ -89,18 +89,21 @@ instance Exception ChainIdException
 -- -------------------------------------------------------------------------- --
 -- ChainId
 
--- | ChainId /within a Chainweb/.
+-- | ChainId /within the context a Chainweb instance/.
```
emilypi:
```suggestion
-- | ChainId /within the context of a Chainweb instance/.
```

src/Chainweb/ChainId.hs
slpopejoy: OK let's just do the exports for now.

emilypi:
@larskuhtz I just noticed there was no change made to `PactService.hs`. Do you mind making sure your script is catching all files with `unsafeChainId` in them? It seems to be off.

src/Chainweb/Version.hs
larskuhtz: I am afraid that's the best we can to with a `HashSet`. Well, traversing a list of 1000 elements isn't that bad.

src/Chainweb/ChainId.hs
```diff
@@ -89,18 +89,21 @@ instance Exception ChainIdException
...
 
--- | ChainId /within a Chainweb/.
+-- | ChainId /within the context of a Chainweb instance/.
 --
--- Generally a block chain is /globally/ uniquely identified by its genesis hash.
--- This type only uniquely identifies the chain /locally/ within the context of
--- a chainweb.
+-- The set of valid ChainIds is determined by the 'ChainwebVersion'. In almost
+-- all use cases there should be a context that is an instance of
+-- 'HasChainwebVersion' can be used get the set of chain ids.
```
slpopejoy: Add a note to import Chainweb.Version instead of this module.

PR #101, opened 2019-04-08
sync mempool with consensus' active cut

gregorycollins:


fosskers:
Closing after talking to @gregorycollins .

PR #115, opened 2019-04-15
speed up mining with faster target check

larskuhtz:
The performance of the inner mining loop is roughly doubled be checking whether the pow hash meets the target directly on the buffers without first converting to `PowHashNat`.


fosskers:
Approved, with a tiny patch handed off to Lars to be manually added.

PR #65, opened 2019-03-26
To Pact, Or Not To Pact

fosskers:
Have you read Hamlet?

This PR introduces a new `ChainwebVersion`, `PactWithTime`, which features simulated mining delay as well as proper Pact validation and execution. `TestWithTime` has been reverted to its previous meaning, which in turn restores the efficiency of `slow-tests`:

```
 Number of logs: 2136
 Expected BlockCount: 300
    {"_statMinHeight":252,"_statMaxHeight":253,"_statBlockCount":278,"_statAvgHeight":252.9,"_statMedHeight":253}
    {"avgEfficiency%":90.97122302158273,"medEfficiency%":91.00719424460432,"minEfficiency%":90.64748201438849,"maxEfficiency%":91.00719424460432}
```

PR #135, opened 2019-04-16
`run-nodes` in Haskell

fosskers:
This PR replaces the Bash-based `run-nodes` script with a Haskell-based binary of the same name. This simplifies the code, allows non-Bash-savvy devs to edit it, and allows more flexibility with CLI args.

Note that usage has slightly changed. Previously:
```
./scripts/run-nodes path/to/chainweb-node 10 es:oures.com:123
```
where things like "loglevel" were set with environment variables, and argument order was fixed. Now we have:
```
colin@yumi ~/c/h/chainweb> stack exec -- run-nodes --help
run-nodes - Run a local cluster of chainweb-node binaries

Usage: run-nodes --exe PATH [--nodes COUNT] [--version VERSION] [--config PATH]
                 [CHAINWEB-FLAGS]

Available options:
  --exe PATH               Full path to a chainweb-node binary
  --nodes COUNT            Number of Nodes to run (default: 10)
  --version VERSION        Chainweb Version to run the Nodes with (default:
                           testWithTime-peterson)
  --config PATH            Path to Chainweb config YAML file
  CHAINWEB-FLAGS           Native flags that a chainweb-node accepts
  -h,--help                Show this help text
```
Note the `--version` flag. This allows us to change the `ChainwebVersion` used on the fly, which will help #65 . It also provides a sensible default (`testWithTime-peterson`) for when the user doesn't care. Recall that a single `chainweb-node` otherwise uses `testnet00` by default.

Most basic usage:
```
run-nodes --exe=path/to/chainweb-node
```
and output will be sent to STDOUT.

image


fosskers:
Using `optparse-parse` generic would have simplified the code even more, but I wanted special parsing logic for the ElasticSearch strings and the `ChainwebVersion`.

fosskers:
Oh, perhaps I need to update the README as well.

larskuhtz:
Would it make sense to keep the original shell script? It's still easier to make ad-hoc changes for special test setups.

larskuhtz:
What about passing any extra options (or options after `--`) unchanged to the Chainweb nodes? That would make it easy to pass additional settings.

scripts/RunNodes.hs
```diff
@@ -0,0 +1,128 @@
...
+logDir :: Log -> IO ()
+logDir (ES _ _) = pure ()
+logDir (Path p) = shelly $ mkdir_p (fromText p)
+
+main :: IO ()
+main = do
+  env@(Env e ns _ lt _) <- execParser opts
+  print env
+  traverse_ logDir lt
+  whenM (doesFileExist e) $ do
```
larskuhtz: Check if file is executable?

scripts/RunNodes.hs
```diff
@@ -0,0 +1,128 @@
...
+    -- Launch Bootstrap Node
+    boot <- async $ runNode 0 (Just "scripts/test-bootstrap-node.config") env
+    threadDelay 200000  -- 0.2 seconds
+    -- Launch Common Nodes
+    mapConcurrently_ (\n -> runNode n Nothing env) [1 .. ns - 1]
```
larskuhtz: Should we add some delay between starting nodes? Maybe based on `n`?

scripts/RunNodes.hs
```diff
@@ -0,0 +1,128 @@
...
+    void $ wait boot
```
larskuhtz: If the bootstrap nodes didn't start or crashes, we wouldn't find out, because `mapConcurrently` blocks and we never get to this line. What about either running the bootstrap node along with the other nodes (using `mapConcurrently` or `runConcurrently`) or using something like `withAsync ... $ \boot -> link boot >> ...`?

scripts/run-nodes.sh
```diff
@@ -1,139 +0,0 @@
...
-
-        # Run without LOG_DIR
-        $RUN \
-            --hostname=127.0.0.1 \
-            --node-id=$NID \
-            --test-miners=$N \
-            --chainweb-version=testWithTime \
-            --interface=127.0.0.1 \
-            --log-level=$LOGLEVEL \
-            $PORT_ARG \
```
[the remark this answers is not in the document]
larskuhtz: I think it's a leftover from some specialized test-setup and was checked in accidentally.

src/Chainweb/Version.hs
```diff
@@ -208,6 +209,15 @@ instance IsMerkleLogEntry ChainwebHashTag ChainwebVersion where
     {-# INLINE toMerkleNode #-}
     {-# INLINE fromMerkleNode #-}
 
+-- | Display the `ChainwebVersion` without any Chain Graph information.
```
larskuhtz: Why is this needed? It may cause confusion when two different versions are displayed equal.

scripts/RunNodes.hs
```diff
@@ -0,0 +1,128 @@
...
+
+    logText :: Log -> T.Text
+    logText (ES h p) = sformat ("es:" % stext % ":" % int) h p
+    logText (Path p) = sformat ("file:" % stext) p
+
+    ops :: [T.Text]
+    ops = [ "--hostname=127.0.0.1"
+          , sformat ("--node-id=" % int) nid
+          , sformat ("--test-miners=" % int) ns
+          , sformat ("--chainweb-version=" % stext) $ plainVersionFromText v
```
larskuhtz: I am somewhat concerned about stripping the graph information from the version here. We'll soon start working on hard-forks. So, will probably have to deal more often with different graphs. It looks like this code will reset the version to the default graph, even if the user asked for a different graph.

scripts/RunNodes.hs
```diff
@@ -0,0 +1,128 @@
...
+      where
+        host = A.string "es:" *> A.takeWhile1 (/= ':') <* A.char ':'
+        port = A.decimal
+
+    plainPath :: A.Parser Log
+    plainPath = Path <$> (A.string "local:" *> A.takeText)
+
+pLogLevel :: Parser T.Text
+pLogLevel = strOption
+  (long "log-level" <> metavar "info|warn|error|debug" <> value "info"
```
larskuhtz:
```suggestion
  (long "log-level" <> metavar "debug|info|warn|error" <> value "info"
```

scripts/RunNodes.hs
larskuhtz: Should this print a warning and exit with non-zero code in case `e` isn't executable?

scripts/RunNodes.hs
fosskers: Woops, you're right.

scripts/RunNodes.hs
fosskers: I had assumed that `--chainweb-version` would not accept a version string that had a graph indicated. Let me double-check that.

scripts/RunNodes.hs
fosskers: Actually, leaving out the delay doesn't seem to make a difference.

src/Chainweb/Version.hs
fosskers: I've removed it.

scripts/RunNodes.hs
fosskers: Done.

fosskers: Done.

scripts/RunNodes.hs
fosskers: Done.

scripts/RunNodes.hs
fosskers: Fixed.

fosskers:
@larskuhtz I've enabled "passthrough", so that it's easy to feed `chainweb-node`'s real flags through to each instance.

tools/run-nodes/RunNodes.hs
```diff
@@ -0,0 +1,94 @@
...
+  env@(Env e ns _ c _) <- execParser opts
+  print env
+  canExec <- (executable <$> getPermissions e) `catch` (\(_ :: SomeException) -> pure False)
+  if | not canExec -> error $ e <> " is not executable, or does not exist."
+     | otherwise -> do
+         putStrLn "Starting cluster..."
+         -- Launch Bootstrap Node
+         withAsync (runNode 0 (Just c) env) $ \boot -> do
+           link boot
+           threadDelay 1000000  -- 1 second
```
fosskers: `link` is key here. We wait for 1 second here, which ensures that any problems in starting the Bootstrap (including bad flags) will kill the parent thread before any of the commons start.

fosskers:
If there were an easy way to provide default values with `optparse-generic`, this code would be even shorter.

PR #153, opened 2019-04-16
Statically-known Text representations for `ChainwebVersion`

fosskers:
This PR allows the usage of `ChainwebVersion` text representations like `testWithTime-peterson` instead of `testWithTime-4075880449`. This applies to CLI flags, as well as JSON.

In turn, this fixes the "version `Map` priming" problem, which blocks further work elsewhere (#135).

The new accepted postfixes are:
- `-singleton`
- `-pair`
- `-triangle`
- `-peterson`
- `-twenty`
- `-hoffman-singleton`

### DIFF
--- a/src/Chainweb/Version.hs
+++ b/src/Chainweb/Version.hs
@@ -82,7 +82,7 @@ import Control.DeepSeq
 import Control.Lens
 import Control.Monad.Catch
 
-import Data.Aeson
+import Data.Aeson hiding (pairs)
 import Data.Bits
 import Data.Bytes.Get
 import Data.Bytes.Put
@@ -92,6 +92,7 @@ import qualified Data.HashMap.Strict as HM
 import qualified Data.HashSet as HS
 import Data.Proxy
 import qualified Data.Text as T
+import Data.Tuple (swap)
 import Data.Word
 
 import GHC.Generics (Generic)
@@ -205,10 +206,14 @@ instance IsMerkleLogEntry ChainwebHashTag ChainwebVersion where
     {-# INLINE fromMerkleNode #-}
 
 chainwebVersionToText :: ChainwebVersion -> T.Text
-chainwebVersionToText v@Test{} = "test-" <> sshow (chainwebVersionId v)
-chainwebVersionToText v@TestWithTime{} = "testWithTime-" <> sshow (chainwebVersionId v)
-chainwebVersionToText v@TestWithPow{} = "testWithPow-" <> sshow (chainwebVersionId v)
-chainwebVersionToText Testnet00 = "testnet00"
+chainwebVersionToText v =
+    case HM.lookup v prettyVersions of
+        Just t -> t
+        Nothing -> case v of
+            Test{} -> "test"
+            TestWithTime{} -> "testWithTime"
+            TestWithPow{} -> "testWithPow"
+            Testnet00 -> "testnet00"
 {-# INLINABLE chainwebVersionToText #-}
 
 -- | Read textual representation of a `ChainwebVersion`.
@@ -216,8 +221,12 @@ chainwebVersionToText Testnet00 = "testnet00"
 chainwebVersionFromText :: MonadThrow m => T.Text -> m ChainwebVersion
 chainwebVersionFromText t =
     case HM.lookup t chainwebVersions of
-        Nothing -> throwM $ TextFormatException $ "Unknown Chainweb version: \"" <> t
         Just v -> pure v
+        Nothing -> case t of
+            "test" -> pure $ Test petersonChainGraph
+            "testWithTime" -> pure $ TestWithTime petersonChainGraph
+            "testWithPow" -> pure $ TestWithPow petersonChainGraph
+            _ -> throwM . TextFormatException $ "Unknown Chainweb version: " <> t
 
 instance HasTextRepresentation ChainwebVersion where
     toText = chainwebVersionToText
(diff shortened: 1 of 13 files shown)


src/Chainweb/Version.hs
```diff
@@ -209,109 +206,76 @@ instance IsMerkleLogEntry ChainwebHashTag ChainwebVersion where
...
+    <> [ ("testnet00", Testnet00) ]
+  where
+    f v p = map (\(k, g) -> (p <> k, v g)) pairs
+    pairs = [ ("-singleton", singletonChainGraph)
+            , ("-pair", pairChainGraph)
+            , ("-triangle", triangleChainGraph)
+            , ("-peterson", petersonChainGraph)
+            , ("-twenty", twentyChainGraph)
+            , ("-hoffman-singleton", hoffmanSingletonGraph)
+            ]
```
fosskers: Anything special I need to do to make sure these CAFs stay in unGC'd?

slpopejoy: CAFs get GCd???

larskuhtz: It seems that this map is consulted each time a chainweb version is serialized or deserialized.

I am a bit concerned about mixing test and production versions here. In the previous version of the code only test versions uses special code paths. The lookup for production version was very straight forward and efficient.

What about keeping the old implementation of `chainwebVersionFromText` and just adapt the case

```
-- Generic test versions
	--
	chainwebVersionFromText t = ... 
```

for test versions?

src/Chainweb/Version.hs
```diff
@@ -209,109 +206,76 @@ instance IsMerkleLogEntry ChainwebHashTag ChainwebVersion where
...
+            _ -> throwM . TextFormatException $ "Unknown Chainweb version: " <> t
 
 instance HasTextRepresentation ChainwebVersion where
     toText = chainwebVersionToText
     {-# INLINE toText #-}
     fromText = chainwebVersionFromText
     {-# INLINE fromText #-}
 
+chainwebVersions :: HM.HashMap T.Text ChainwebVersion
+chainwebVersions = HM.fromList $
```
larskuhtz: This builds a map for every possible combination of version stem and graph. Would it be possible to reduce the map to just the choices of graphs and build the combined test chainweb version on the fly?

larskuhtz: hmm, maybe it doesn't matter that much... it's only a factor of three...

src/Chainweb/Version.hs
```diff
@@ -209,109 +206,76 @@ instance IsMerkleLogEntry ChainwebHashTag ChainwebVersion where
...
-chainwebVersionToText v@TestWithPow{} = "testWithPow-" <> sshow (chainwebVersionId v)
-chainwebVersionToText v@Simulation{} = "simulation-" <> sshow (chainwebVersionId v)
+chainwebVersionToText v =
+    case HM.lookup v prettyVersions of
+        Just t -> t
+        Nothing -> case v of
+            Test{} -> "test"
+            TestWithTime{} -> "testWithTime"
+            TestWithPow{} -> "testWithPow"
+            Testnet00 -> "testnet00"
```
larskuhtz: Same as below: I don't see a reason to move the production versions and "default" versions into this case. In particular the production cases should be separate cases, which is more efficient and more readable.

src/Chainweb/Version.hs
```diff
@@ -209,109 +206,76 @@ instance IsMerkleLogEntry ChainwebHashTag ChainwebVersion where
...
+            , ("-twenty", twentyChainGraph)
+            , ("-hoffman-singleton", hoffmanSingletonGraph)
+            ]
+
+prettyVersions :: HM.HashMap ChainwebVersion T.Text
+prettyVersions = HM.fromList . map swap $ HM.toList chainwebVersions
+
+chainwebVersionIds :: HM.HashMap Word32 ChainwebVersion
+chainwebVersionIds =
+    HM.fromList . map (chainwebVersionId &&& id) $ HM.elems chainwebVersions
```
larskuhtz: Same here: `chainwebVersionId Testnet00 = 0x00000001` was much more readable and efficient.

src/Chainweb/Version.hs
```diff
@@ -209,109 +206,76 @@ instance IsMerkleLogEntry ChainwebHashTag ChainwebVersion where
...
-    case HM.lookup h m of
-        Just _ -> return ()
-        Nothing -> atomically
-            $ modifyTVar' testChainwebVersionMap $ HM.insert h v
-    return h
-  where
-    h = i .|. (testChainwebVersionMask .&. int (hash v))
+toTestChainwebVersion Testnet00 _ =
+    error "toTestChainwebVersion must not be called for a production instance"
+toTestChainwebVersion v i = i .|. (testChainwebVersionMask .&. int (hash v))
```
larskuhtz: I propose that we don't use the `hash` function any more but just hard-code the version-id component for each know graph.

src/Chainweb/Version.hs
fosskers: This is my paranoia, I personally have no answer to that. Do we know that CAFs stay around forever, and are only evaluated once?

src/Chainweb/Version.hs
fosskers: I believe the production versions are already caught somewhere higher up the call chain. Adding the case here was only to prevent exhaustive pattern match warnings.

fosskers:
I'll see what I can rework.

src/Chainweb/Version.hs
fosskers: > It seems that this map is consulted each time a chainweb version is serialized or deserialized.

That old code also consulted the `TVar`'d Map for each serialization, too.

src/Chainweb/Version.hs
fosskers: That should give us a small perf boost :+1:

src/Chainweb/Version.hs
```diff
@@ -205,20 +203,15 @@ instance IsMerkleLogEntry ChainwebHashTag ChainwebVersion where
...
-    case HM.lookup v prettyVersions of
-        Just t -> t
-        Nothing -> case v of
-            Test{} -> "test"
-            TestWithTime{} -> "testWithTime"
-            TestWithPow{} -> "testWithPow"
-            Testnet00 -> "testnet00"
+chainwebVersionToText :: HasCallStack => ChainwebVersion -> T.Text
+chainwebVersionToText Testnet00 = "testnet00"
+chainwebVersionToText v = fromJuste $ HM.lookup v prettyVersions
```
fosskers: Hm, this will be a source of silent pattern match failures. When someone creates a new production `ChainwebVersion` value, they won't be warned about the missing match here.

fosskers: Hopefully quickcheck would catch it.

src/Chainweb/Version.hs
larskuhtz: Yes, that's my point: It did so *only* for test versions.

PR #157, opened 2019-04-18
Initial Testnet01 Configuration

fosskers:
The main different between this and `Testnet00` is the associated `ChainGraph`, which for now has been set to `twentyChainGraph`.

It's annoying to have to add new hard-coded block headers, but with soon-to-come future work, these will be removed. For now, they're necessary for everything to compile.

**Note:** The `chainweb-node` binary has been set to use `Testnet01` by default, but be aware that until I change the config files living on our running machines,  binary pushes to our current network will ignore this new default, as they set `testnet00` explicitly in their config files.


src/Chainweb/Difficulty.hs
```diff
@@ -316,6 +316,8 @@ blockRate TestWithTime{} = Just $ BlockRate 4
 blockRate TestWithPow{} = Just $ BlockRate 10
 -- 120 blocks per hour, 2,880 per day, 20,160 per week, 1,048,320 per year.
 blockRate Testnet00 = Just $ BlockRate 30
+-- 120 blocks per hour, 2,880 per day, 20,160 per week, 1,048,320 per year.
+blockRate Testnet01 = Just $ BlockRate 30
```
mightybyte: Hmmm, this pattern is making me wonder if we should bundle all these parameters up in a data structure so you can see all the params for a given testnet in one place.  Not for now though.  Maybe later.

fosskers: I understand the desire for that. The main thing opposing it is the usage patterns. We don't have an "application Monad", so there's no common `Env` running through the entire app. What most things have is a `ChainwebVersion` on hand (or they can easily find it via a nearby `Cut` or `BlockHeader`). Then with that, they can `O(1)` dispatch to any of these special values.

mightybyte: Yeah, that makes sense.  Probably the right call.

fosskers: Putting all the possible configuration values (which sometimes are lambdas!) into one place would almost certainly violate some rule about separation-of-concerns.

PR #158, opened 2019-04-18
Add upper-bound parameter to cut GET Rest API

larskuhtz:


src/Chainweb/Cut.hs
```diff
@@ -203,6 +204,26 @@ cutAdjs c cid = HM.intersection
...
+limitCut wdb h c = c & (cutHeaders . itraverse) f
+  where
+    ch = h `div` int (order (_chainGraph wdb))
+    f cid x = do
+        db <- give wdb $ getWebBlockHeaderDb cid
+        a <- S.head_ $ branchEntries db
+            Nothing (Just 1)
+            Nothing (Just $ int ch)
+            mempty (HS.singleton $ UpperBound $ _blockHash x)
+        return $ fromJuste a
```
fosskers: If using `fromJuste`, the top-level function should include `HasCallStack`.

PR #159, opened 2019-04-18
Transaction Generator (Now with logging)

emmanueldenloye:
The transaction generator is now at a good point for a pull request. There is a small logging bug that happens when a listen is issued after receipt of the request key of a send. The bug is that the log of this event is not shown. 

Besides that, there are several things to note:

1) The generator checks that the chain id is valid with the given version. (The script bails out if the chain id is invalid).
2) There is documentation with instructions upon how to use each of the commands. This file is in examples/transaction-generator.txt. (For the emacs folks, this is an org file masquerading as a text file. It will open in org-mode).
3) There is another script (it is scripts/deploy-and-send.sh) which combines deploying a standard set of contracts (coin, simple-payments, hello-world) and sending transactions from these contracts.


src/Chainweb/HostAddress.hs
```diff
@@ -68,6 +68,7 @@ module Chainweb.HostAddress
 , portFromText
 , pPort
 , readPortBytes
+, unsafePortFromText
```
mightybyte: It doesn't look like this is used.  Can we remove it?

examples/TransactionGenerator.hs
```diff
@@ -299,92 +555,194 @@ mainInfo =
...
 numContracts :: Int
 numContracts = 2
 
-send :: SubmitBatch -> ClientM RequestKeys
-poll :: Poll -> ClientM PollResponses
-listen :: ListenerRequest -> ClientM ApiResult
-local :: Command Text -> ClientM (CommandSuccess Value)
-
-send :<|> poll :<|> listen :<|> local = client (Proxy :: Proxy (PactApi_))
+type TheApi
```
mightybyte: Why do we need to redeclare the API here?  We should be able to re-use the upstream definition and avoid the repetition.

examples/TransactionGenerator.hs
```diff
@@ -299,92 +555,194 @@ mainInfo =
...
-      RunSimpleExpressions -> sendSimpleExpressions config
-      PollRequestKeys requestKeys -> pollRequestKeys (RequestKeys (map (RequestKey . Hash) requestKeys)) config
-
--- TOOD: This is here for when a user wishes to deploy their own
+    if HS.member (_nodeChainId config) $ chainIds_ $ _chainGraph (_nodeVersion config)
+       then startup config
+       else error $ "This chain: " <> show (_nodeChainId config) <> " is invalid given this chainweb version: " <> show (_nodeVersion config)
+  where
+    startup config = do
+      let transHandle = config ^. logHandleConfig
```
mightybyte: Don't worry about changing it now, but for future reference...in general if you're only reading, then the accessor will be more efficient and optimizable by the compiler, and easier to understand than lens operator soup.  In other words, don't use a lens when a function call will do.

examples/TransactionGenerator.hs
```diff
@@ -48,246 +66,484 @@ import Servant.API
...
+
+transactionCommandFromText :: MonadThrow m => Text -> m TransactionCommand
+transactionCommandFromText = readTransactionCommandBytes . T.encodeUtf8
+{-# INLINE transactionCommandFromText #-}
+
+readTransactionCommandBytes :: MonadThrow m => B8.ByteString -> m TransactionCommand
+readTransactionCommandBytes = parseBytes "transaction-command" transactionCommandParser
+{-# INLINE readTransactionCommandBytes #-}
+
+transactionCommandParser :: A.Parser TransactionCommand
```
mightybyte: Is there a reason you couldn't leverage the JSON instances here instead of having to maintain these parsers?

examples/Chainweb/Simulate/Contracts/SimplePayments.hs
```diff
@@ -88,111 +91,52 @@ simplePaymentsContractLoader adminKeyset = do
...
-  | RequestPay Account Account Amount
-  | CreateAccount Account Balance [SomeKeyPair]
-
-parens :: String -> String
-parens s = "(" ++ s ++ ")"
+  = SPRequestGetBalance Account
+  | SPRequestPay Account Account Amount
+  | SPCreateAccount Account Balance [SomeKeyPair]
 
 instance Show SimplePaymentRequest where
```
mightybyte: Is the existential in `SomeKeyPair` the reason this isn't being auto-derived?  If so, we're using `PublicMeta` in lieu of `SomeKeyPair` in other places.  Could we do that here?

chainweb.cabal
```diff
@@ -712,7 +724,17 @@ executable transaction-generator
...
+        , stm
+        , stm-chans
         , text >= 1.2
+        , time >= 1.8
+        , unordered-containers >= 0.2
+        , wai >= 3.2
+        , wai-app-static >= 3.1
+        , wai-extra >= 3.0
+        , wai-cors >= 0.2
+        , warp >= 3.2
```
mightybyte: Why are we depending on warp and all these wai packages?  The transaction generator shouldn't need to run a web server.

mightybyte:
Also, this branch has conflicts that need to be resolved.

chainweb.cabal
emmanueldenloye: I needed the `Utils.Logging`  and `Utils.Logging.Config` modules from the `node` folder. Requiring those modules pulled in those extra dependencies, unfortunately.

examples/TransactionGenerator.hs
emmanueldenloye: While I could use `somePactApi` which has type `somePactApi :: ChainwebVersion -> ChainId -> SomeApi`, that would require unpacking the data type SomeApi and it does not seem like it is possible to retrieve the appropriate Proxy value when unpacking it. Hopefully, someone can chime in and point out if I'm wrong.

src/Chainweb/HostAddress.hs
emmanueldenloye: It is used. Look at line 334 of `TransactionGenerator.hs`.

examples/TransactionGenerator.hs
emmanueldenloye: Already changed.

examples/TransactionGenerator.hs
```diff
@@ -48,246 +66,484 @@ import Servant.API
...
+makeLenses ''TransactionGeneratorState
+
+defaultScriptConfig :: ScriptConfig
+defaultScriptConfig =
+  ScriptConfig
+    {
+      _scriptCommand       = RunSimpleExpressions def def
+    , _nodeChainId         = unsafeChainId 0
+    , _isChainweb          = True
+    , _chainwebHostAddress = HostAddress (unsafeHostnameFromText "127.0.0.1") (unsafePortFromText "1789")
```
larskuhtz: There is `unsafeHostAddressFromText` wouldn't it work here?

src/Chainweb/HostAddress.hs
larskuhtz: I think `unsafeHostAddressFromText` could be used in `TransactionGenerator.hs:334`

chainweb.cabal
slpopejoy: Generally you only need the deps if your code directly, and for instance you are not directly issuing CORS configurations so I'm sure you don't need `wai-cors` declared here.

examples/Chainweb/Simulate/Contracts/CoinContract.hs
```diff
@@ -0,0 +1,196 @@
...
+            object
+              [ "admin-keyset" .= fmap formatB16PubKey adminKeyset
+              , "receiver-guard" .= fmap formatB16PubKey guard
+              ]
+      mkExec theCode theData meta adminKeyset Nothing
+  where
+    _functionErrPrefix :: String
+    _functionErrPrefix = "createCoinContractRequest:"
+
+-----------------------------------
```
slpopejoy: You will never be issuing buy gas or coinbase, so you can delete the commented code here.

examples/Chainweb/Simulate/Contracts/CryptoCritters.hs
```diff
@@ -7,7 +7,7 @@
 module Chainweb.Simulate.Contracts.CryptoCritters where
```
slpopejoy: As discussed can we ditch the extra contracts and just go with coin-contract stuff and something like simple payments? We don't want to maintain any more pact code than totally necessary.

examples/Chainweb/Simulate/Utils.hs
```diff
@@ -29,22 +33,40 @@ testSomeKeyPairs = do
 formatB16PubKey :: SomeKeyPair -> Text
 formatB16PubKey = toB16Text . formatPublicKey
 
+-- | note this is "sender00"'s key
 someED25519Pair :: (PublicKeyBS, PrivateKeyBS, Text, PPKScheme)
 someED25519Pair =
     ( PubBS $ getByteString
-        "ba54b224d1924dd98403f5c751abdd10de6cd81b0121800bf7bdbdcfaec7388d"
+        "368820f80c324bbc7c2b0610688a7da43e39f91d118732671cd9c7500ff43cca"
```
slpopejoy: We will definitely want to exercise more than just one gas payer in the future. As I mentioned elsewhere, the ApiReq yaml format has the ability to specify keys so I would prefer we used that.

examples/TransactionGenerator.hs
```diff
@@ -48,246 +66,484 @@ import Servant.API
...
+    , _logHandleConfig     = U.StdOut
     }
 
-transactionConfigParser :: MParser TransactionConfig
-transactionConfigParser = id
-  <$< scriptCommand .:: option auto
-      % long "script-command" <> short 'c'
-      <> help "The specific command to run (DeployContracts|RunStandardContracts|RunSimpleExpressions)."
-  <*< nodeChainId .::option auto
+scriptConfigParser :: MParser ScriptConfig
```
slpopejoy: Could this stuff be used to do script parsing instead of the above manual parsing?

transaction-generator.txt
```diff
@@ -0,0 +1,108 @@
+* Deploying Contracts
```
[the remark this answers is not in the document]
emmanueldenloye: It is org-mode from emacs. Fortunately, org makes it easy to change to another format. I'll make the change.

I initially chose this format because I thought it would not too confusing for the reader if I only stuck to fairly basic features from org-mode (mostly).

chainweb.cabal
```diff
@@ -679,30 +678,43 @@ executable transaction-generator
     ghc-options:
         -threaded
         -with-rtsopts=-N
-    hs-source-dirs: examples
+    hs-source-dirs: examples node
```
mightybyte: Oh, my below comment is being caused by this line.  What Stuart is saying is correct.  If you remove `node` here, your package will still have access to everything exported from the `node` directory.  This is the case because of the internal build depend below on line 696.

examples/TransactionGenerator.hs
mightybyte: ...and `Pact.Server.API`

src/Chainweb/HostAddress.hs
emmanueldenloye: I've addressed this.

examples/TransactionGenerator.hs
emmanueldenloye: I've addressed this by making `send`, `poll` and `listen` top level commands (they're out of the `Reader`). They, of course, still must the chainweb version and the chain id as arguments in the call (in addition to their usual arguments). I think this solution is cleaner than mucking around with the `BaseUrl` object.

examples/TransactionGenerator.hs
emmanueldenloye: I have addressed this issue.

examples/TransactionGenerator.hs
```diff
@@ -48,246 +66,484 @@ import Servant.API
...
+instance FromJSON (ScriptConfig -> ScriptConfig) where
+  parseJSON = withObject "ScriptConfig" $ \o -> id
+    <$< scriptCommand       ..: "scriptCommand"       % o
+    <*< nodeChainId         ..: "nodeChainId"         % o
+    <*< isChainweb          ..: "isChainweb"          % o
+    <*< chainwebHostAddress ..: "chainwebHostAddress" % o
+    <*< nodeVersion         ..: "chainwebVersion"     % o
+    <*< logHandleConfig     ..: "logging"             % o
+
+data ServantRecord = ServantRecord
```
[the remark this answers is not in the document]
emmanueldenloye: I have addressed this in an earlier comment.

examples/TransactionGenerator.hs
emmanueldenloye: I've actually addressed this in another manner.

examples/TransactionGenerator.hs
```diff
@@ -48,246 +66,484 @@ import Servant.API
...
-    rkeys <-
-      ExceptT $
-      runClientM (send (SubmitBatch [cmd])) cenv
-    ExceptT $ runClientM (poll (Poll (_rkRequestKeys rkeys))) cenv
-
-
-sendTransactionGetRequest :: Command Text -> TransactionGenerator (PrimState IO) (Either ServantError RequestKeys)
-sendTransactionGetRequest cmd = do
-  cenv <- view genClientEnv
+  send <-  views genServantRecord apiSend
```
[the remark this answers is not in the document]
emmanueldenloye: I've addressed this.

examples/Chainweb/Simulate/Contracts/CoinContract.hs
emmanueldenloye: This has been taken care of!

examples/Chainweb/Simulate/Contracts/CryptoCritters.hs
emmanueldenloye: Should we also keep `HelloWorld`?

chainweb.cabal
emmanueldenloye: I've addressed this issue in a new commit.

examples/Chainweb/Simulate/Contracts/SimplePayments.hs
emmanueldenloye: This show instance has been taken out. Done.

chainweb.cabal
emmanueldenloye: I've resolved this.

src/Chainweb/Utils/Gexf.hs
```diff
@@ -0,0 +1,327 @@
+{-# LANGUAGE DataKinds #-}
```
mightybyte: What is this file for?  I'm guessing it's an artifact of a merge?

emmanueldenloye: I just copied the file `chainweb-node/node/Utils/Gexf.hs`. Nothing more, nothing less.

examples/Chainweb/Simulate/Contracts/SimplePayments.hs
```diff
@@ -88,111 +91,52 @@ simplePaymentsContractLoader adminKeyset = do
...
-    "RequestPay: " ++ parens (show accountA) ++ " " ++ parens (show accountB) ++ " " ++ parens (show amount)
-  show (CreateAccount account balance _) = "CreateAccount: " ++ parens (show account) ++ " " ++ parens (show balance)
+  show (SPRequestGetBalance account) = "SPRequestGetBalance: " ++ parens (show account)
+  show (SPRequestPay accountA accountB amount) =
+    "SPRequestPay: " ++ parens (show accountA) ++ " " ++ parens (show accountB) ++ " " ++ parens (show amount)
+  show (SPCreateAccount account balance _) = "SPCreateAccount: " ++ parens (show account) ++ " " ++ parens (show balance)
 
-createSimplePaymentRequest :: SimplePaymentRequest -> Maybe [SomeKeyPair] -> IO (Command Text)
-createSimplePaymentRequest (CreateAccount (Account account) (Balance initialBalance) somekeyset) Nothing = do
+createSimplePaymentRequest :: PublicMeta -> SimplePaymentRequest -> Maybe [SomeKeyPair] -> IO (Command Text)
```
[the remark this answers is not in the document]
emmanueldenloye: I think this is something that we could evaluate in a later PR.

examples/Chainweb/Simulate/Utils.hs
emmanueldenloye: Ok, I can address in a later PR.

src/Chainweb/Utils/Gexf.hs
emmanueldenloye: This has been moved elsewhere as per Lars' request.

examples/Chainweb/Simulate/Contracts/SimplePayments.hs
emmanueldenloye: I've resolved this by removing the show instances.

examples/TransactionGenerator.hs
```diff
@@ -48,344 +67,565 @@ import Servant.API
...
 instance ToJSON TimingDistribution
 
-data TransactionConfig = TransactionConfig
-  { _scriptCommand :: TransactionCommand
-  , _nodeChainId :: !ChainId
-  , _serverRootPrefix :: String
-  , _isChainweb :: Bool
-  , _chainwebNodePort :: !ChainwebPort
-  , _timingDistribution :: !TimingDistribution
+data ScriptConfig = ScriptConfig
```
mightybyte: Minor comment on these changes to ScriptConfig and the subsequent instances (not to be addressed now)...in general I'm trending more and more away from this kind of vertical alignment.  It definitely makes things more readable, but at the cost of significantly more friction with respect to increased diff size, more merge conflicts, etc.  On individual projects this is not a big deal, but with teams it can be quite costly.

PR #160, opened 2019-04-22
Minimum Difficulty Adjustment Thresholds

fosskers:
From the Haddocks:

> Spikes in *HashRate* may occur as the mining network grows and shrinks. To ensure that adjustment does not occur too quickly or with too much granularity, we enforce a "minimum factor of change" (Z) for `adjust`. If `adjust` notices that the change for the given window is not large enough, then it will not occur. The overall pattern of difficulty then becomes less of a rippling pond, and more of a series of plateaus with distinct jumps. Analysis has been shown that Z should be greater than a factor of e=2.71828⋯ (source needed).

In particular, this change softens the effect of sudden addition / loss of mining power. For our current production `ChainwebVersion`s we set the minimum required change to be `3`. Therefore, on average (when we're already half way through a mining window), 4/5 of the miners would need to suddenly quit the network in order to decrease the difficulty at the next adjustment interval.

### DIFF
--- a/src/Chainweb/Difficulty.hs
+++ b/src/Chainweb/Difficulty.hs
@@ -62,8 +62,8 @@ module Chainweb.Difficulty
 , blockRate
 , WindowWidth(..)
 , window
-, MaxAdjustment(..)
-, maxAdjust
+, MinAdjustment(..)
+, minAdjust
 , prereduction
 -- ** Adjustment
 , adjust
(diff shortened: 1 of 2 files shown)


src/Chainweb/Difficulty.hs
```diff
@@ -338,22 +338,22 @@ window Testnet00 = Just $ WindowWidth 120
...
-maxAdjust TestWithPow{} = Just $ MaxAdjustment 3
+minAdjust :: ChainwebVersion -> Maybe MinAdjustment
+minAdjust Test{} = Nothing
+minAdjust TestWithTime{} = Nothing
+minAdjust TestWithPow{} = Just $ MinAdjustment 3
 -- See `adjust` for motivation.
-maxAdjust Testnet00 = Just $ MaxAdjustment 3
-maxAdjust Testnet01 = Just $ MaxAdjustment 3
+minAdjust Testnet00 = Just $ MinAdjustment 3
+minAdjust Testnet01 = Just $ MinAdjustment 3
```
fosskers: We dispatch on `ChainwebVersion` here, instead of hard-coding. This opens the possibility of transparently updating the value for future `ChainwebVersion`s. 

Should we hard-code this instead?

slpopejoy: LGTM

PR #161, opened 2019-04-22
faster consensus catchup

larskuhtz:
* [x] Do catch up in steps of 1000 blocks
* [x] Include origin header into pulled cut hashes

### DIFF
--- a/src/Chainweb/CutDB.hs
+++ b/src/Chainweb/CutDB.hs
@@ -283,6 +283,7 @@ processCuts
 processCuts logFun headerStore payloadStore queue cutVar = queueToStream
     & S.chain (\c -> loggc Info c $ "start processing")
     & S.filterM (fmap not . isVeryOld)
+    & S.filterM (fmap not . farAHead)
     & S.filterM (fmap not . isOld)
     & S.filterM (fmap not . isCurrent)
     & S.chain (\c -> loggc Info c $ "fetch all prerequesites")
@@ -311,10 +312,23 @@ processCuts logFun headerStore payloadStore queue cutVar = queueToStream
     threshold :: Int
     threshold = int $ 2 * diameter graph * order graph
 
+    -- NOTE: THIS NUMBER MUST BE LARGER THAN THE RESPECTIVE LIMIT
+    -- IN 'CutDB.Sync'
+    farAHeadThreshold :: Int
+    farAHeadThreshold = 2000
+
     queueToStream = do
         Down a <- liftIO (pQueueRemove queue)
         S.yield a
         queueToStream
+
+    -- FIXME: this is problematic. We should drop these before they are
+    -- added to the queue, to prevent the queue becoming stale.
+    farAHead x = do
+        h <- _cutHeight <$> readTVarIO cutVar
+        let r = (int (_cutHashesHeight x) - farAHeadThreshold) >= int h
+        when r $ loggc Info x "skip far ahead cut"
+        return r
 
     isVeryOld x = do
         h <- _cutHeight <$> readTVarIO cutVar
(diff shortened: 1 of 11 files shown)


src/Chainweb/CutDB.hs
```diff
@@ -311,11 +312,24 @@ processCuts logFun headerStore payloadStore queue cutVar = queueToStream
     threshold :: Int
     threshold = int $ 2 * diameter graph * order graph
 
+    -- NOTE: THIS NUMBER MUST BE LARGER THAN THE RESPECTIVE LIMIT
+    -- IN 'CutDB.Sync'
+    farAHeadThreshold :: Int
+    farAHeadThreshold = 2000
```
fosskers: This makes me nervous.

src/Chainweb/CutDB/Sync.hs
```diff
@@ -95,9 +99,12 @@ syncSession v useOrigin p db logg env = do
...
 
     receive = do
-        c <- getCut cenv
-        logg @T.Text Debug $ "got cut " <> sshow c
-        addCutHashes db c
+        -- Query cut that is at most 1000 blocks ahead
+        h <- _cutHeight <$> _cut db
+        c <- getCut cenv (h + 1000)
+
+        logg @T.Text Info $ "receivecd cut " <> sshow c
```
fosskers:
```suggestion
        logg @T.Text Info $ "received cut " <> sshow c
```

src/P2P/TaskQueue.hs
```diff
@@ -140,12 +143,22 @@ instance Ord (Task env a) where
 -- enough to be useful in other contexts, too.
 --
 session
+    :: AttemptsCount
+    -> PQueue (Task env a)
+    -> LogFunction
+    -> env
+    -> PeerInfo
+    -> IO Bool
+session li q lo e _ = session_ li q lo e
```
fosskers: Why not consolidate on a single `session` function, if it's just redirecting like this anyway?

fosskers: Ah, it's that the `P2pSession` type changed. Hm.

src/Chainweb/Utils/Paging.hs
```diff
@@ -129,6 +132,15 @@ data NextItem k
     | Exclusive k
     deriving stock (Eq, Show, Ord, Functor, Foldable, Traversable)
 
+_getNextItem :: NextItem k -> k
+_getNextItem (Inclusive k) = k
+_getNextItem (Exclusive k) = k
+{-# INLINE _getNextItem #-}
+
+getNextItem :: Getter (NextItem k) k
+getNextItem = to _getNextItem
```
fosskers: This throws a redundant constraint warning.

src/Chainweb/CutDB.hs
larskuhtz: This will be fixed in the Rocksdb PR #163

src/P2P/TaskQueue.hs
larskuhtz: The `session_` function is there mostly to limit the amount of code changes and not have to wrap `PeerInfo` in a `Maybe` for situations when it's not available.

src/Chainweb/Utils/Paging.hs
larskuhtz: Yes. Didn't the team agree sometime ago to remove the redundant-constraint warning flag from the cabal file?

larskuhtz: I guess removing it from the cabal file should be in a separate PR. So I'll add the the pragma to suppress it.

larskuhtz: accidentally committed this into #163. So, won't fix it in the PR.

PR #162, opened 2019-04-22
Mempool: Transaction Reintroduction

marklnichols:


src/Chainweb/BlockHeaderDB.hs
```diff
@@ -24,7 +24,7 @@ module Chainweb.BlockHeaderDB
 (
 -- * Chain Database Handle
   Configuration(..)
-, BlockHeaderDb
+, BlockHeaderDb(..)
```
fosskers: Almost certainly naughty. I imagine that the constructor was hidden for a reason.

src/Chainweb/Mempool/Consensus.hs
```diff
@@ -0,0 +1,94 @@
...
+        nxt <- S.next stream
+        case nxt of
+            -- end of the stream, last item is common branch point of the forks
+            -- removing the common branch block with tail -- the lists should never be empty
+            Left _ -> return (V.tail oldBlocks, V.tail newBlocks)
+
+            Right (LeftD blk, strm) -> go strm (V.cons blk oldBlocks, newBlocks)
+            Right (RightD blk, strm) -> go strm (oldBlocks, V.cons blk newBlocks)
+            Right (BothD lBlk rBlk, strm) -> go strm ( V.cons lBlk oldBlocks,
+                                                       V.cons rBlk newBlocks )
```
fosskers: `cons`ing like this gets expensive quickly.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -264,55 +274,67 @@ data InMemoryMempoolData t = InMemoryMempoolData {
...
-
+    return $! InMemoryMempool cfg dataLock txB tid blockHeaderDb payloadDb
   where
     newData = InMemoryMempoolData <$> newIORef PSQ.empty
-                                  <*> newIORef HashMap.empty
-                                  <*> newIORef HashSet.empty
-
+                                          <*> newIORef HashMap.empty
+                                          <*> newIORef HashSet.empty
+                                          <*> return Nothing
```
fosskers: Funky alignment?

src/Chainweb/Orphans.hs
```diff
@@ -0,0 +1,35 @@
...
+    parseJSON v = do
+        s <- T.unpack <$> parseJSON v
+        return $! read s
+
+-- orphan, needed for mock -- remove once this instance makes it upstream into pact
+instance ToJSON GasPrice where
+    toJSON (GasPrice d) = toJSON d
+
+instance FromJSON GasPrice where
+    parseJSON v = GasPrice <$> parseJSON v
```
fosskers: Do  these live in Pact yet?

src/Chainweb/Payload/PayloadStore/InMemory.hs
```diff
@@ -57,8 +57,7 @@ newPayloadCache = PayloadCache <$> newBlockOutputsStore
...
 
-newPayloadDb :: IO (PayloadDb HashCAS.HashMapCas)
-newPayloadDb = PayloadDb
-    <$> newTransactionDb
-    <*> newPayloadCache
-
+newPayloadDb :: IO (Maybe (PayloadDb HashCAS.HashMapCas))
+newPayloadDb = do
+    pdb <- PayloadDb <$> newTransactionDb <*> newPayloadCache
+    return $ Just pdb
```
fosskers: Why does this need to return in `Maybe`?

test/Chainweb/Test/Mempool/InMem.hs
```diff
@@ -5,24 +5,38 @@ module Chainweb.Test.Mempool.InMem
...
     -- run the reaper @100Hz for testing
     cfg = InMemConfig txcfg mockBlockGasLimit (hz 100)
     hz x = 1000000 `div` x
     hashmeta = chainwebTestHashMeta
     hasher = chainwebTestHasher . codecEncode mockCodec
+----------------------------------------------------------------------------------------------------
+
+-- copied from Chainweb.Test.Utils
+toyVersion :: ChainwebVersion
+toyVersion = Test singletonChainGraph
```
fosskers: Why copy? There should probably be a relatively top-level test version for one-off tests (if that's not what's in `Chainweb.Test.Utils` already).

test/Chainweb/Test/Mempool/Socket.hs
```diff
@@ -8,14 +8,36 @@ import Control.Concurrent
...
 import Chainweb.Test.Mempool (MempoolWithFunc(..))
 import qualified Chainweb.Test.Mempool
+import Chainweb.Test.Utils (toyChainId)
 import Chainweb.Utils (Codec(..))
+import Chainweb.Version
+import Data.CAS.RocksDB
+
 ------------------------------------------------------------------------------
+tests :: TestTree
+tests = withResource (newPool cfg) Pool.destroyAllResources $
```
fosskers: This resource allocation should probably happen higher up.

test/Chainweb/Test/Mempool/Socket.hs
```diff
@@ -24,6 +46,10 @@ data TestServer = TestServer {
   , _tsServerThread :: !ThreadId
 }
 
+-- copied from Chainweb.Test.Utils
+toyVersion :: ChainwebVersion
+toyVersion = Test singletonChainGraph
```
fosskers: Here too.

test/Chainweb/Test/Mempool/Sync.hs
```diff
@@ -39,6 +53,10 @@ tests = mempoolProperty "Mempool.syncMempools" gen propSync withFunc
       pre (not (Set.null xss || Set.null yss || Set.null zss) && length ys < 10000 && length zs < 10000)
       return (xss, yss, zss)
 
+-- copied from Chainweb.Test.Utils
+toyVersion :: ChainwebVersion
+toyVersion = Test singletonChainGraph
```
fosskers: Here too.

src/Chainweb/Mempool/Consensus.hs
marklnichols: This is list is just the depth of a fork, so it should be extremely short.

src/Chainweb/Mempool/InMem.hs
marklnichols: fixed

src/Chainweb/Orphans.hs
marklnichols: No, I don't see them.  I can move them over in a quick follow-up PR.  I don't want to add a Pact version bump on top of this already giant PR.

src/Chainweb/Payload/PayloadStore/InMemory.hs
marklnichols: It was done as a Maybe because none of the mempool tests (except for the one I added) needed it, and it looked like a lot of work to contruct dummy instance for all the other tests.  I'll look at it again, I maybe have an easier time after the past few days experiences.

test/Chainweb/Test/Mempool/InMem.hs
marklnichols: It was there, I just had to add it to the exports.

test/Chainweb/Test/Mempool/Socket.hs
marklnichols: Same as above, done.

test/Chainweb/Test/Mempool/Sync.hs
marklnichols: Same as above, done.

src/Chainweb/BlockHeaderDB.hs
marklnichols: I just took a look, if there is a reason its not obvious, but Pact needs the three types contained within.

test/Chainweb/Test/Mempool/Socket.hs
marklnichols: I started to change this, but I'm going to have to put it on the list for another PR.  I didn't write this code and its a bit complicated to pull out the various IO resources.

src/Chainweb/Mempool/InMem.hs
marklnichols: Fixed

src/Chainweb/BlockHeaderDB.hs
fosskers: @larskuhtz ?

src/Chainweb/Mempool/Consensus.hs
fosskers: Is `Vector` being used for something efficient later in the pipeline?

marklnichols: The use of Vector here is due to it being the collection used by the mempool, e.g.
    mempoolInsert :: Vector t -> IO ()  
    mempoolGetBlock :: Int64 -> IO (Vector t)

I assume Greg made an appropriate choice here, but I don't know if there were any particular concerns about performance that led to it.

fosskers: One would think that the consing here could be done via a list, with a `V.fromList` at the very end. But if the list is always™  going to be short, then the change might not be worth it either.

chainweb.cabal
```diff
@@ -388,6 +392,7 @@ test-suite chainweb-tests
         , extra >= 1.6
         , fgl >= 5.6
         , hashable >= 1.2
+        , hashtables >= 1.2
```
gregorycollins: Just use `IORef` + `HashMap`? This is for the testsuite and the additional dependency doesn't seem worth it to gain speedup there.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -245,15 +256,16 @@ data InMemConfig t = InMemConfig {
 
 
 ------------------------------------------------------------------------------
-data InMemoryMempool t = InMemoryMempool {
+data InMemoryMempool t cas = InMemoryMempool {
     _inmemCfg :: InMemConfig t
   , _inmemDataLock :: MVar (InMemoryMempoolData t)
   , _inmemBroadcaster :: TxBroadcaster t
   , _inmemReaper :: ThreadId
+  , _inmemBlockHeaderDb :: BlockHeaderDb
```
gregorycollins: This stuff doesn't belong in here? This data is only needed by the mempool consensus layer, and so you should stuff it in there. Adding these functions here pollutes the api, when you could have

```
data MempoolConsensus t = MempoolConsensus {
    _mpcMempool :: MempoolBackend t,
    _mpcLastNewBlockParent :: Maybe (IORef BlockHeader),
    ...
}
```

Nothing that the consensus fork processing algorithm needs to do should require knowledge of mempool internals beyond that which is published by the api, so it's better to wrap it.

src/Chainweb/Mempool/Consensus.hs
```diff
@@ -0,0 +1,133 @@
...
+
+chainwebTxsFromPWO :: PayloadWithOutputs -> IO (Set ChainwebTransaction)
```
gregorycollins: This should be using `HashSet`

src/Chainweb/Mempool/Consensus.hs
```diff
@@ -0,0 +1,133 @@
...
+
+-- | Collect the blocks on the old and new branches of a fork.  The old blocks are in the first
+--   element of the tuple and the new blocks are in the second.
+collectForkBlocks
+    :: S.Stream (Of (DiffItem BlockHeader)) IO ()
+    -> IO (Vector BlockHeader, Vector BlockHeader)
+collectForkBlocks theStream =
+    go theStream (V.empty, V.empty)
+  where
+    go stream (oldBlocks, newBlocks) = do
```
gregorycollins: need bangs on the accumulators `oldBlocks` and `newBlocks`

src/Chainweb/Mempool/InMem.hs
```diff
@@ -340,59 +372,118 @@ reaperThread cfg dataLock restore = forever $ do
...
+                 -> Maybe (PayloadDb cas)
+                 -> LogFunction
+                 -> BlockHeader
+                 -> IO (Vector ChainwebTransaction)
+processForkInMem lock blockHeaderDb payloadStore logFun newHeader = do
+    theData <- readMVar lock
+    -- convert: Maybe (IORef BlockHeader) -> Maybe BlockHeader
+    lastHeader <- traverse readIORef (_inmemLastNewBlockParent theData)
+
+    MPCon.processFork logFun blockHeaderDb newHeader lastHeader
```
gregorycollins: ...and this proves you don't need any specific knowledge of mempool internals to implement this, since the only field you're pulling out of it is the new one.

src/Chainweb/Mempool/Mempool.hs
```diff
@@ -110,6 +115,14 @@ data MempoolBackend t = MempoolBackend {
     -- TODO: move this inside TransactionConfig or new MempoolConfig ?
   , mempoolBlockGasLimit :: Int64
 
+    -- | keeps track of the PARENT of the last newBlock request - used to re-introduce txs
+    --   in the case of forks
+  , mempoolLastNewBlockParent :: Maybe (IORef BlockHeader)
```
gregorycollins: Definitely doesn't belong here in this api

src/Chainweb/Mempool/Mempool.hs
```diff
@@ -130,7 +143,8 @@ data MempoolBackend t = MempoolBackend {
   , mempoolMarkConfirmed :: Vector TransactionHash -> IO ()
 
     -- | These transactions were on a losing fork. Reintroduce them.
-  , mempoolReintroduce :: Vector TransactionHash -> IO ()
+
+  , mempoolReintroduce :: Vector t -> IO ()
```
gregorycollins: :+1:

src/Chainweb/Transaction.hs
```diff
@@ -42,6 +38,9 @@ data PayloadWithText = PayloadWithText
     deriving (Show, Eq, Generic)
     deriving anyclass (NFData)
 
+instance Ord PayloadWithText where
```
gregorycollins: As I said over in the other spot, `Hashable` is much better here -- rather than ord-comparing payload bytes (yuck), there is a hash code inside `Payload` that you can just project out and use when implementing `hashWithSalt`. Come to think of it, the derived `Eq` instance is also pretty bad here, we should just make this an `==` on `payloadBytes` because otherwise `payloadObj` is going to be compared for equality also.

src/Chainweb/Chainweb/ChainResources.hs
```diff
@@ -221,8 +224,10 @@ mempoolSyncP2pSession chain logg0 env _ = newIORef False >>= go
     errorHandler (e :: SomeException) = do
         logg Warn ("mempool sync session failed: " <> sshow e)
         throwM e
-
-    peerMempool = MPC.toMempool v cid txcfg gaslimit env
+    peerMempool = do
+        -- no sync needed / wanted for lastNewBlockParent attribute:
```
gregorycollins: More proof this field doesn't belong here :)

src/Chainweb/Mempool/Consensus.hs
```diff
@@ -0,0 +1,133 @@
...
+              logFun @(JsonLog ReintroducedTxs) Info $ JsonLog reIntro
+              return results
+  where
+    f trans header = S.union trans <$> payloadLookup header
+
+    logg :: LogLevel -> T.Text -> IO ()
+    logg = logFun
+
+
+data ReintroducedTxs = ReintroducedTxs
```
gregorycollins: Maybe add a suffix to indicate to the reader that this is intended for log messages?

test/Chainweb/Test/Mempool/Consensus.hs
```diff
@@ -0,0 +1,406 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Chainweb.Test.Mempool.Consensus
```
gregorycollins: :+1: tests look really good.

PR #165, opened 2019-04-23
Explicit Testing Scenarios

fosskers:
This PR introduces a new `ChainwebVersion` value, `TimedCPM` (i.e. "timed consensus/pact/mempool") which is to be the new default for `run-nodes`. `TestWithTime` and `TestWithPow` have been renamed, and the original meaning of `TestWithTime` has been restored (i.e. just consensus testing).

The meaning of each `ChainwebVersion` value has also been clarified in docstrings.

Closes #65


src/Chainweb/Pact/Types.hs
```diff
@@ -150,6 +147,12 @@ instance FromJSON MinerInfo where
   parseJSON = withObject "MinerInfo" $ \o ->
     MinerInfo <$> o .: "m" <*> (KeySet <$> o .: "ks" <*> o .: "kp")
 
+emptyPayload :: PayloadWithOutputs
```
emmanueldenloye: Is it better for discoverability that you return emptyPayload back to where? It seems like it should be there instead of here.

fosskers: I would, except that it's used in multiple modules (and hence exported from here).

PR #166, opened 2019-04-23
Display raw `ChainwebVersion` in `info` Object

fosskers:
As can be seen, for instance, here: https://us1.chainweb.com/swagger.json


eskimor:
Thanks a lot @fosskers!

Unfortunately, I just found the definition of Info: https://hackage.haskell.org/package/swagger2-2.3.1.1/docs/Data-Swagger.html#t:Info

So this is a fixed field data type, implemented in swagger itself. The information contained, especially `description` does not seem to be meant to be machine readable. We could abuse the description field for containing convenient machine readable information, together with a comment "This is machine read, please don't change!", but it might be a better option to just recover the desired condensed information, from the highly redundant representation in the paths in the `paths` object.

So, I am closing this, as it seems a bit too fragile, to depend on information in the description field.

PR #167, opened 2019-04-23
Early Adjustment Windows

fosskers:
The value `prereduction` found in `Chainweb.Difficulty` is used to artificially slow mining during initial conditions of the network. Its specific value is mostly a guess based on behaviour observed during testing. 

The downsides to introducing `prereduction` are multifold:

- The Genesis blocks need to be mined, since only certain `Nonce` values will allow validation to succeed.
- The Genesis blocks need to be hard-coded, since re-mining them on-the-fly isn't realistic.
- Hard-coding the blocks is boilerplate, and will only get worse with time as more `ChainwebVersion`s are invented.
- A side tool needed to be written to perform said hard-coding.
- Unit tests slow down, since the genesis blocks need to be re-mined in order to be validated.

This PR adds early adjustment windows for `BlockHeight`s lower than the first "official" adjustment.


fosskers:
Rebased onto `master`.

fosskers:
Rebased again.

fosskers:
Rebased again.

fosskers:
Yet another rebase.

src/Chainweb/TreeDB/Difficulty.hs
```diff
@@ -61,14 +61,23 @@ hashTarget db bh
...
     ww :: Natural
     ww = case window ver of
-      Just (WindowWidth n) -> n
+      Just (WindowWidth n)
+          -- Intent: Initial network conditions are chaotic. `HashTarget`s are
+          -- so easy that mining progresses extremely quickly. In order to not
+          -- overwhelm the Cut network, we perform earlier adjustments before
+          -- the first "official" adjustment epoch is reached.
+          | height < n -> n `div` 10
+          | otherwise -> n
```
emilypi: If we derive `Num` for `WindowWidth`, alot of this unwrapping and rewrapping goes away, ya?

src/Chainweb/TreeDB/Difficulty.hs
```diff
@@ -61,14 +61,23 @@ hashTarget db bh
...
 
     ww :: Natural
     ww = case window ver of
-      Just (WindowWidth n) -> n
+      Just (WindowWidth n)
+          -- Intent: Initial network conditions are chaotic. `HashTarget`s are
+          -- so easy that mining progresses extremely quickly. In order to not
+          -- overwhelm the Cut network, we perform earlier adjustments before
+          -- the first "official" adjustment epoch is reached.
+          | height < n -> n `div` 10
```
emilypi: Where are we deriving the div by 10 adjustment from? Not that it's wrong - I just don't know how to tell _why_ this is correct just from looking at it

src/Chainweb/TreeDB/Difficulty.hs
emmanueldenloye: I second this!

src/Chainweb/TreeDB/Difficulty.hs
fosskers: I'll add this.

src/Chainweb/TreeDB/Difficulty.hs
fosskers: I'll give it a shot.

PR #169, opened 2019-04-24
Mark/mempool test retries

marklnichols:


test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -108,7 +108,7 @@ tests rdb = do
     tt1 <- testPoll cmds cwEnv rks
     let tConfig = mempoolTxConfig noopMempool
     let mPool = toMempool version cid tConfig 10000 cwEnv :: MempoolBackend ChainwebTransaction
-    tt2 <- testMPValidated mPool rks
+    let tt2 = testCase "mempoolValidationCheck" $ testMPValidated mPool rks
 
     return $ testGroup "PactRemoteTest" $ tt0 : (tt1 : [tt2])
```
fosskers:
```suggestion
    return $ testGroup "PactRemoteTest" [tt0, tt1, tt2] 
```

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -108,9 +108,9 @@ tests rdb = do
     tt1 <- testPoll cmds cwEnv rks
     let tConfig = mempoolTxConfig noopMempool
     let mPool = toMempool version cid tConfig 10000 cwEnv :: MempoolBackend ChainwebTransaction
-    tt2 <- testMPValidated mPool rks
+    let tt2 = testCase "mempoolValidationCheck" $ testMPValidated mPool rks
 
-    return $ testGroup "PactRemoteTest" $ tt0 : (tt1 : [tt2])
+    return $ testGroup "PactRemoteTest" $ [tt0, tt1,tt2]
```
fosskers: My instincts still tell me that this function can be made to not return `IO TestTree`. All environment-like values can be brought in from the outside. Any `IO` things that need to occur within tests can do so within a `testCase`'s `Assertion`.

marklnichols: Based on what we were discussing yesterday, I now can probably restructure the whole thing and hopefully achieve that goal.  I'd rather get this in now, though, as it IO situation is no worse and it fixes the breaking build.  I'll make an issue for the reorg of those tests

PR #172, opened 2019-04-25
Bump Pact Pin

fosskers:
This will probably need help from @emilypi or @marklnichols to fix the API breakage.


src/Chainweb/Pact/Backend/Orphans.hs
```diff
@@ -46,137 +47,106 @@ import Pact.Types.Server
...
 
-deriving instance Generic ModuleData
-
-deriving instance Generic (ModuleDef a)
-
-deriving instance Generic Ref
-
-deriving instance Generic RefStore
-
+deriving instance Generic (Governance g)
```
fosskers: I'd love to get rid of all of these. Orphans are an indication of a design flaw.

emmanueldenloye: If you talk to either @emilypi or @slpopejoy, they will tell you there are plans to move these instances back into Pact; a place I think everyone would agree they should be. Until then, expect more of these type of PRs. Sorry...

fosskers: Luckily it seems like a lot of the types did get `Generic` instances in recent Pact, so at least we were able to cut down on some boilerplate here.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -226,16 +224,21 @@ checkResponse filePrefix rks (PollResponses theMap) =
...
-
 getCwBaseUrl :: Port -> BaseUrl
 getCwBaseUrl thePort = BaseUrl
     { baseUrlScheme = Https
     , baseUrlHost = "127.0.0.1"
     , baseUrlPort = fromIntegral thePort
     , baseUrlPath = "" }
 
-escapedCmd :: BS.ByteString
-escapedCmd = [r|{"cmds":[{"hash":"d0613e7a16bf938f45b97aa831b0cc04da485140bec11cc8954e0509ea65d823472b1e683fa2950da1766cbe7fae9de8ed416e80b0ccbf12bfa6549eab89aeb6","sigs":[{"addr":"368820f80c324bbc7c2b0610688a7da43e39f91d118732671cd9c7500ff43cca","sig":"71cdedd5b1305881b1fd3d4ac2009cb247d0ebb55d1d122a7f92586828a1ed079e6afc9e8b3f75fa25fba84398eeea6cc3b92949a315420431584ba372605d07","scheme":"ED25519","pubKey":"368820f80c324bbc7c2b0610688a7da43e39f91d118732671cd9c7500ff43cca"}],"cmd":"{\"payload\":{\"exec\":{\"data\":null,\"code\":\"(+ 1 2)\"}},\"meta\":{\"gasLimit\":100,\"chainId\":\"0\",\"gasPrice\":1.0e-4,\"sender\":\"sender00\"},\"nonce\":\"2019-03-29 20:35:45.012384811 UTC\"}"}]}|]
```
fosskers: @marklnichols might take offense to the removal of this ;)

@emmanueldenloye and I figured it would be safer / more clear / more maintainable to carry around the Haskellized version of the same structure (`testBatch`). It also helped fix a bug.

marklnichols: That's fine as long as it produces the same Pact code I had pasted from Stuarts utility.

src/Chainweb/Pact/Backend/Orphans.hs
```diff
@@ -298,23 +267,24 @@ instance Serial1 Def where
         _dDefBody <- deserializeWith m
         _dMeta <- deserialize
         _dInfo <- deserialize
-        return $ Def {..}
+        return $ Def{..}
 
+-- FIXME: This should be autoderivable, since `Generic1` is autoderivable.
```
emmanueldenloye: I'm just leaving this note as a reminder of our discussion in which we agreed upon there needing to be a separate PR to replace these instances with the auto-derivable ones.

PR #173, opened 2019-04-26
Consolidate binaries

mightybyte:
This PR prunes some unused binaries and consolidates many of the others into a single `cwtool` binary.  On my machine this speeds up a full `cabal build` by ~20% and reduces our total generated binary size by ~20% which will save additional time both for developers downloading cached artifacts and for testnet deployments.


exe/CwTool.hs
```diff
@@ -0,0 +1,68 @@
...
+
+main :: IO ()
+main = do
+    cmd <- customExecParser p opts
+    case cmd of
+      C2Gexf -> Chain2Gexf.main
+      Ea -> Ea.main
+      GenTransactions -> TransactionGenerator.main
+      RunNodes -> RunNodes.main
+      SlowTests -> SlowTests.main
```
larskuhtz: Will the existing option parsing still work for the sub-commands? 

Does `customExecParser` update the argument environment? 

Otherwise I would have expected something like `withArgs` that would strip the command from the options list and pass only the remaining options.

mightybyte: With the exception of ea (which Colin and I are handling), I did basic testing of each one and they appear to work.   I didn't do exhaustive checking because some are more involved, but I at least ran things like `--help`.

test/SlowTests.hs
```diff
@@ -25,7 +25,9 @@ import Chainweb.Version
 import qualified Network.X509.SelfSigned.Test
 
 main :: IO ()
-main = defaultMain suite
+main = do
+  putStrLn "In SlowTests.main"
```
fosskers: Was this for debugging?

exe/CwTool.hs
```diff
@@ -1,68 +1,90 @@
...
+      ["-h"] -> printHelp topLevelCommands
+      ["--help"] -> printHelp topLevelCommands
+      (cmd:restOfArgs) -> do
+        case filter (\cs -> cmd == csCmd cs) topLevelCommands of
+          [] -> do
+            printf "Error: \"%s\" is not a valid command\n\n" cmd
+            printHelp topLevelCommands
+            exitFailure
+          [cs] -> withArgs restOfArgs $ csAction cs
+          _ -> error "Duplicate command encountered.  This shouldn't happen!"
```
fosskers: I'm surprised that `optparse-applicative` can't "just do" what we were trying to do here.

fosskers:
Slight issue with `--help` output on subcommands:
```
colin@yumi ~/c/h/chainweb> cwtool run-nodes --help
Usage: cwtool --exe PATH [--nodes COUNT] [--version VERSION] [--config PATH]
              [CHAINWEB-FLAGS]
  Run a local cluster of chainweb-node binaries
```

exe/CwTool.hs
```diff
@@ -0,0 +1,90 @@
...
+    padding = replicate (max 0 (n - length s)) ' '
+
+
+cmdHelpLine :: CommandSpec -> String
+cmdHelpLine cs = printf "  %s%s" (padRight 25 $ csCmd cs) (csDescription cs)
+
+topLevelCommands :: [CommandSpec]
+topLevelCommands =
+  [ CommandSpec
+      "bigbang"
```
fosskers: `"ea"` works again.

test/SlowTests.hs
mightybyte: Ahh, yes.  Good catch.

exe/CwTool.hs
mightybyte: Yeah, I had the same thought.

https://github.com/pcapriotti/optparse-applicative/issues/340

I think we'd be able to do it if each subcommand was also an optparse-applicative parser that we had full control over (although I still think it would not be as composable as I was hoping).  But alas, that is not the case.  We don't have control over the `SlowTests` main because it just calls out to `defaultMain` from `tasty`.

exe/CwTool.hs
mightybyte: Fixed

mightybyte:
> Slight issue with `--help` output on subcommands:
> 
> ```
> colin@yumi ~/c/h/chainweb> cwtool run-nodes --help
> Usage: cwtool --exe PATH [--nodes COUNT] [--version VERSION] [--config PATH]
>               [CHAINWEB-FLAGS]
>   Run a local cluster of chainweb-node binaries
> ```
> 
> These are all the suboptions for `run-nodes`.

Yes, that's the intended behavior.  If the first argument is a subcommand, then you get the help for that subcommand.  If you want top-level help, `--help` needs to be your first argument.  That seems like the right behavior to me, otherwise how do you get help for the subcommands?

PR #174, opened 2019-04-27
integrate pact tests into tasty test-suite

larskuhtz:


test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -87,154 +201,61 @@ pactTestSetup = do
...
-testReq4 = TestRequest
-    { _trCmd = Code "(test1.create-global-accounts)"
-    , _trEval = fileCompareTxLogs "create-accounts-expected.txt"
-    , _trDisplayStr = "Creates two accounts" }
+-- | A test runner for golden tests.
+--
+fileCompareTxLogs :: String -> IO TestResponse -> ScheduledTest
+fileCompareTxLogs label respIO = pactGoldenSch label $ do
+    resp <- respIO
+    return $ toS $ Y.encode
```
larskuhtz: Usage of `Y.encode` is problematic. There is no standardized normal form for Yaml or JSON encoded data and the encoding may change between different version of `aeson`/`yaml`, any dependency, `ghc`, platform, etc.

If we find the output to be sufficiently stable it may be fine for golden tests. Otherwise we should implement a normalized JSON encoding.

test/Chainweb/Test/Pact/PactInProcApi.hs
```diff
@@ -8,170 +10,129 @@
...
+    reqQ <- reqIO
+    locVar0c <- testLocal >>= \t -> local t reqQ
+    goldenBytes "local" =<< takeMVar locVar0c
+
+goldenBytes :: Y.ToJSON a => Exception e => String -> Either e a -> IO BL.ByteString
+goldenBytes label (Left e) = assertFailure $ label ++ ": " ++ show e
+goldenBytes label (Right a) = return $ BL.fromStrict $ Y.encode $ object
+    [ "test-group" .= label
+    , "results" .= a
+    ]
```
larskuhtz: See comment above. Usage of non-normalized Yaml/JSON for golden file encodings can cause flaky test results.

marklnichols: I think the use of encode is not limited to the tests -- if it needs changing we'll have to find all the occurrences using in the various places conversions are done.

test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -87,154 +201,61 @@ pactTestSetup = do
...
+    gasLimit = fromMaybe 0 (_ccGasLimit cmdConfig)
+    gasRate = fromMaybe 0 (_ccGasRate cmdConfig)
+    gasEnv = GasEnv (fromIntegral gasLimit) 0.0 (constGasModel (fromIntegral gasRate))
+
+execTest :: (forall a . PactServiceM a -> IO a) -> TestRequest -> ScheduledTest
+execTest runPact request = _trEval request $ do
+    cmdStrs <- mapM getPactCode $ _trCmds request
+    trans <- mkPactTestTransactions $ V.fromList cmdStrs
+    results <- runPact $ execTransactions (Just $ nullBlockHash) defaultMiner trans
+    let outputs = V.toList $ snd <$> _transactionPairs results
```
marklnichols: $ not needed: (Just nullBlockHash)

test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -87,154 +201,61 @@ pactTestSetup = do
...
+    cmdConfig = toCommandConfig pactCfg
+    gasLimit = fromMaybe 0 (_ccGasLimit cmdConfig)
+    gasRate = fromMaybe 0 (_ccGasRate cmdConfig)
+    gasEnv = GasEnv (fromIntegral gasLimit) 0.0 (constGasModel (fromIntegral gasRate))
+
+execTest :: (forall a . PactServiceM a -> IO a) -> TestRequest -> ScheduledTest
+execTest runPact request = _trEval request $ do
+    cmdStrs <- mapM getPactCode $ _trCmds request
+    trans <- mkPactTestTransactions $ V.fromList cmdStrs
+    results <- runPact $ execTransactions (Just $ nullBlockHash) defaultMiner trans
```
larskuhtz:
```suggestion
    results <- runPact $ execTransactions (Just nullBlockHash) defaultMiner trans
```

test/Chainweb/Test/Pact/PactInProcApi.hs
larskuhtz: My understanding is that these tests are not concerned with cryptographic consistency of the data but changes in the APIs. Since the different JSON/YAML encodings are semantically equal, the tests would still be valid. 

It's just about spurious test failures. So we may just move on and add a normalized encoding down the road if needed.

larskuhtz: If the tests were concerned about the cryptographic identity of the the data base the test should just compare the Merkle roots for the respective payload data structures.

PR #177, opened 2019-04-30
Golden Reorg

fosskers:
This PR relocates some of our utilities related to `tasty-golden`, and also stores test data in directories with clearer names.


test/Chainweb/Test/Golden.hs
```diff
@@ -0,0 +1,40 @@
...
+-- internal modules
+
+import Chainweb.Test.Utils (ScheduledTest(..))
+
+---
+
+goldenFilesDir :: FilePath
+goldenFilesDir = "test/golden/"
+
+golden
```
fosskers: @larskuhtz should these just be in `Chainweb.Test.Utils` instead?

larskuhtz: Maybe rename the module to `Chainweb.Test.Utils.Golden` to make it clear that it's test utils. Currently, it's sometimes a bit difficult to find utils because there is no distinction between utility and test module namespaces.

larskuhtz: Actually, I think an internal library for test utils would make sense, but let's not discuss this here.

fosskers: I'll do a quick renaming.

PR #181, opened 2019-05-01
Data model - create all necessary tables

emmanueldenloye:


emmanueldenloye:
That is an agreeable request. I'll modify this PR and just keeping making commits. I'll mark the Asana tickets appropriately going forward.

PR #182, opened 2019-05-01
support for servant 0.16

larskuhtz:
This PR makes the code base ready for servant >= 0.15. It doesn't require it.

The code paths for servant > 0.14 aren't executed by our CI, so don't get any testing. I tested the code in the context of testnet and it worked.

I plan to do a followup PR that exchanges the use of io-streams with the servant builtin stream type for servant >= 0.15, which would simplify the code.

With this PR the streaming clients for servant >= 0.15 don't buffer. Depending on whether the streams are consumer or producer bound, this could lead to connections being kept open longer than needed or the processing being delayed longer than needed. Most likely it's fine.


src/Chainweb/Mempool/RestAPI.hs
```diff
@@ -98,7 +103,12 @@ type MempoolGetBlockApi v c t =
     QueryParam "blockSize" Int64 :> Post '[JSON] [t]
 
 #if MIN_VERSION_servant(0,15,0)
-#error TODO: need to support servant >= 0.15
+type MempoolGetPendingApi v c t =
```
mightybyte: This CPP macro is unnecessary.  The two if clauses are exactly the same.

src/Chainweb/Mempool/RestAPI/Client.hs
```diff
@@ -217,7 +243,29 @@ subscribeClient v c = runIdentity $ do
 
 ------------------------------------------------------------------------------
 #if MIN_VERSION_servant(0,15,0)
-#error TODO: need to support servant >= 0.15
+
+-- TODO: the code in this module could be simplfied by replacing the use of
+-- io-streams with servant's build-in SourceIO stream type.
+--
+instance Show a => FromSourceIO a (Streams.InputStream a) where
+    fromSourceIO (SourceT src) = unsafePerformIO $ src $ \step ->
```
mightybyte: 😬

larskuhtz: This also was in the original code for servant 0.14. 

My understanding is that it is due to the design of io-streams to implement the underlying stream loop with mutable state (`IORef`). So, unlike with other streaming frameworks, initializing a stream requires an IO action that isn't an effect in the stream itself.

mightybyte: Ahh yes.  I missed that.  Carry on.  👍

larskuhtz:
I tried pinning servant-0.16. That's a rabbit whole, because it also requires QuickCheck >= 1.12, which itself generates a lot of pain. 

Things would be easier if nix would be able to deal with Hackage revisions when pulling directly from Hackage. So, even if we'd upgrade all our packages to a recent Hackage snapshot, we would have to pull many packages from GitHub, because the tar-balls in Hackage don't include package revisions.

PR #183, opened 2019-05-02
SQLite versioned checkpointer

emmanueldenloye:
Mostly the latter part of  the statement in the quotes.


src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+  , ReorgVersion(..)
+  , BlockVersion(..)
+  , bvVersion
+  , bvBlock
+  , detectVersionChange
+  , onVersionChange
+  , OnVersionChange(..)
+  ) where
+
+-- import Control.Concurrent.MVar
```
slpopejoy: Just delete imports, don't leave them around.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+  deriving newtype Num
+
+data BlockVersion = BlockVersion
+  { _bvBlock :: !BlockHeight
+  , _bvVersion :: !ReorgVersion
+  }
+
+makeLenses ''BlockVersion
+
+_checkTableExist :: Database -> Utf8 -> IO [[SType]]
```
slpopejoy: This is useless now as is the test below, we should start moving to more meaningful tests soon.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+        ]
+      ]
+    ans2 = [[SInt 0,SInt 0]]
+    ans3 = [[SText "user1",SInt 0,SInt 0]]
+
+
+createBlockHistoryTable :: Database -> IO ()
+createBlockHistoryTable c =
+  exec_
+    c
```
slpopejoy: This is excessive linebreaking. Please go to "exec_ c \n ..."

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+-}
+
+detectVersionChange ::
+     Database
+  -> BlockHeight
+  -> BlockHash
+  -> IO (Either String OnVersionChange)
+detectVersionChange c bRestore hsh  = do
+  let sqlBtoBH [[SInt bh]] = BlockHeight (fromIntegral bh)
+      sqlBtoBH _ = error "sqlBtoBH: wat"
```
slpopejoy: This is a bad style, you should never `fmap` a conversion function that can fail. See below for more idiomatic code.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+      sqlVtoV [[SInt version]] = ReorgVersion version
+      sqlVtoV _ = error "sqlVtoV: wat"
+  bCurrent <- sqlBtoBH <$>
```
slpopejoy: As noted above, the conversion function can fail so never fmap something like this. Also there is no reason to spread around the code so much, use a more sequential style for clarity, like:
```haskell
bCurrent <- do
  r <- qry_ c "SELECT max(blockheight) AS current_block_height FROM BlockHistory;" [RInt]
  case r of 
    [[SInt bh]] -> BlockHeight (fromIntegral bh)
    _ -> die $ "detectVersionChange: expected single row int response, got: " ++ show r
```

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+
+
+
+-}
+
+detectVersionChange ::
+     Database
+  -> BlockHeight
+  -> BlockHash
+  -> IO (Either String OnVersionChange)
```
slpopejoy: Let's move to an exception-based/`MonadThrow` model, either returns are a drag for complex code like this.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+    qry_ c "SELECT max(version) AS current_version FROM VersionHistory;" [RInt]
+
+  -- enforce invariant that the history has (B_restore-1,H_parent).
+  historyInvariant <-
+    qry c "SELECT COUNT(*)\
+           \ FROM BlockHistory\
+           \ WHERE blockheight = (?)\
+           \ AND hash = (?);" [SInt $ fromIntegral $ pred bRestore, SBlob (encode hsh)] [RInt]
+
+  if historyInvariant /= [[SInt 1]]
```
slpopejoy: You're probably going to want a function like `expectSing` from pact: https://github.com/kadena-io/pact/blob/fa27df6c325da037c4fae0ff2217912ac58c1fa5/src-ghc/Pact/Persist/SQLite.hs#L113

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+    then return $ Left "HistoryInvariantViolation"
+  -- enforce invariant that B_restore is not greater than B_current + 1
+    else return $ case compare bRestore (bCurrent + 1) of
```
slpopejoy: Why is this here and on line 333

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+mkSnapshot c = do
+  s@(SnapshotName name) <- mkSnapshotName
+  exec' c "SAVEPOINT (?)" [SBlob name]
+  return s
+
+newtype SnapshotName = SnapshotName BS.ByteString
+  deriving (Eq, Ord, IsString)
+  deriving newtype Show
+
+mkSnapshotName :: IO SnapshotName
```
slpopejoy: snapshots don't actually need unique names, and re-using one can help enforce that you're not overlapping them. Also they're not snapshots, they're savepoints, so name accordingly:
- for the block savepoint, just call it "BLOCK"
- for each transaction savepoint, call it "TX"
Again, neither of these can overlap at their respective levels, so unique names are unnecessary.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,375 @@
...
+  -> IO (Either String (BlockVersion, SnapshotName))
+onVersionChange c bRestore (BlockVersion bCurrent vCurrent) =
+  case compare bRestore (bCurrent + 1) of
+
+    -- Probably better to throw an exception here.
+    GT -> return $ Left "bRestore should not be greater than bCurrent + 1"
+
+    -- normal operation
+    EQ -> do
+      name <- mkSnapshot c
```
slpopejoy: You'll probably want to start this elsewhere, plus you'll want to run the version maintenance case in it's own transaction (outside of a tx, sqlite will autocommit on each stmt which is slow).

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: That was left there by accident. I've updated the function `onVersionChange`.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: I'm not sure how snapshot slipped in there.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: Deleted.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: Deleted.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: Fixed.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: Understood. Maybe I could use `expectSing` like in the suggestion below?

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: My hunch is that it might be better to use `throwDbError` (from `Pact.Types.Runtime`)?

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: Ok.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: My hunch is that it might be better to use throwDbError (from Pact.Types.Runtime) instead of die from System.Exit?

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+                  [SText key
+                  , SInt (fromIntegral bh)
+                  , SInt (fromIntegral version)
+                  , SInt (fromIntegral txid)
+                  , SBlob (toStrict (Data.Aeson.encode v))]
+            _ -> throwM $ userError $ "writeRow: key " <>  toS kk <> " was found in table."
+        Update -> do
+          res <- qry_ db ("SELECT rowkey FROM " <> domainTableName d <> " WHERE rowkey=" <> key) [RText]
+          case res of
+            [] -> throwM $ userError $ "writeRow: key " <>  toS kk <> " was not found in table."
```
slpopejoy: this is potentially user facing so indicate that _update_ failed here.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -0,0 +1,116 @@
...
+    CheckpointEnv
+      { _cpeCheckpointer = undefined $ checkpointer
+      , _cpeLogger = loggr
+      , _cpeGasEnv = gasEnv
+      }
+
+type Db = MVar (CWDbEnv SQLiteEnv)
+
+innerRestore :: Db -> BlockHeight -> BlockHash -> IO (Either String (SQLiteEnv, TxId, Maybe ExecutionMode))
+innerRestore dbenv bh hsh = runCWDb dbenv $ runExceptT $ withExceptT ((mappend "restore :") . show) $ do
```
slpopejoy: ditch the ExceptT and just encase everything in a `tryAny`

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -0,0 +1,116 @@
...
+  txid <- gets _bsTxId
+  mode <- gets _bsMode
+  return (sqlenv, txid, mode)
+
+-- Assumes that BlockState has been initialized properly.
+innerRestoreInitial :: Db -> IO (Either String (SQLiteEnv, TxId, Maybe ExecutionMode))
+innerRestoreInitial dbenv = runCWDb dbenv $ do
+    r <- callDb (\db ->
+              qry db
+              "SELECT * FROM BlockHistory WHERE blockheight = ? AND hash = ?;"
```
slpopejoy: if we don't need the data do `count(*)`

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -0,0 +1,116 @@
...
+    result <- tryAny $ runCWDb dbenv $ do
+      commitSavepoint "BLOCK"
+      systemInsert (BlockHistory bh hsh)
+      bs <- gets _bsBlockVersion
+      systemInsert (VersionHistory bs)
+    return $ case result of
+      Left err -> Left $ "save: " <> show err
+      Right _ -> Right ()
+
+-- We could remove this function entirely.
```
slpopejoy: remove it

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -83,6 +110,59 @@ instance A.FromJSON PactDbConfig
...
+
+-- data PactDbStateNew = PactDbStateNew
+--   { _pdsPactDb :: P.PactDb (CWDbEnv SQLiteEnv)
+--   , _pdsdbEnv :: CWDbEnv SQLiteEnv
+--   }
+
+-- makeLenses ''PactDbStateNew
+
+data CheckpointerNew p = CheckpointerNew
+    { restoreNew :: BlockHeight -> BlockHash -> IO (Either String (p, P.TxId, Maybe P.ExecutionMode))
```
slpopejoy: We don't need TxId as an output anymore, and ExecutionMode is an input to restoreXXX

src/Chainweb/Pact/Utils.hs
```diff
@@ -9,15 +9,14 @@
 --
 -- Pact service for Chainweb
 
-module Chainweb.Pact.Utils
-    ( toEnv'
-    , toEnvPersist'
-    ) where
+module Chainweb.Pact.Utils where
```
slpopejoy: looks like we don't need this module anymore

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+withTempSQLiteConnection ps ml action =
+  withTempFile (\file -> withSQLiteConnection 0 0 file ps ml False action)
+
+data SystemInsert
+  = BlockHistory BlockHeight
+                 BlockHash
+  | VersionHistory BlockVersion
+  | VersionedTables TableName BlockVersion
+
+systemInsert  :: DbConstraints m SQLiteEnv => SystemInsert -> m ()
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+  let bvEnv = BlockVersion bRestore vNext
+  tableMaintenanceRowsVersionedTables bvEnv
+  tableMaintenanceVersionedTables bvEnv
+  deleteHistory bvEnv
+  -- beginSavepoint "BLOCK"
+
+tableMaintenanceRowsVersionedTables :: DbConstraints m SQLiteEnv => BlockVersion -> m ()
+tableMaintenanceRowsVersionedTables (BlockVersion (BlockHeight bh) _) = do
+  tblNames <- callDb $ \db ->
+    qry_ db ("SELECT tablename FROM VersionedTables WHERE blockheight < " <> Utf8 (toS (show bh))) [RText]
```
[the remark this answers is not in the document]
emmanueldenloye: I've changed this.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+  forM_ tblNames $ \case
+      [SText t] -> callDb $ \db -> exec_ db $ "DELETE FROM " <> t <> " WHERE blockheight >= " <> Utf8 (toS (show bh))
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+      _ -> throwDbError "An error occured\
+                        \ while querying the\
+                        \ VersionedTables table for table names."
+
+tableMaintenanceVersionedTables :: DbConstraints m SQLiteEnv => BlockVersion -> m ()
+tableMaintenanceVersionedTables (BlockVersion (BlockHeight bh) _) = do
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+deleteHistory :: DbConstraints m SQLiteEnv => BlockVersion -> m ()
+deleteHistory (BlockVersion bh _) =  callDb $ \db ->
+  exec' db
+    "DELETE FROM BlockHistory\
+    \ WHERE BlockHistory.blockheight >= ?"
+    [SInt (fromIntegral bh)]
+
+beginSavepoint :: DbConstraints m SQLiteEnv => SavepointName ->  m ()
+beginSavepoint (SavepointName name) =
+  callDb $ \db -> exec_ db ("SAVEPOINT " <> Utf8 name <> ";")
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+
+commitSavepoint :: DbConstraints m SQLiteEnv => SavepointName -> m ()
+commitSavepoint (SavepointName name) =
+  callDb $ \db -> exec_ db ("RELEASE SAVEPOINT " <> Utf8 name <> ";")
+
+rollbackSavepoint :: DbConstraints m SQLiteEnv => SavepointName ->  m ()
+rollbackSavepoint (SavepointName name) =
+  callDb $ \db -> exec_ db ("ROLLBACK TRANSACTION TO SAVEPOINT " <> Utf8 name <> ";")
+
+preBlock :: DbConstraints m SQLiteEnv => m a -> m a
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+
+_preBlockTest :: IO ()
+_preBlockTest = undefined
+
+initSchema :: DbConstraints m SQLiteEnv => m ()
+initSchema = do
+  createBlockHistoryTable
+  createVersionHistoryTable
+  createVersionedTablesTable
+  createVersionedTable "SYS:keysets" (BlockVersion (BlockHeight 0) (ReorgVersion 0))
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+      Namespaces -> callDbWithKey (convNamespaceName k)
+      (UserTables _) -> callDbWithKey (convRowKey k)
+      Pacts -> callDbWithKey (convPactId k)
+ where
+   getRow :: DbConstraints m SQLiteEnv => FromJSON v => [[SType]] -> m (Maybe v)
+   getRow =
+     \case
+       [] -> return Nothing
+       [[SBlob a]] -> return $ decode (fromStrict a)
+       err@_ -> throwM $ userError $ "readRow: Expected (at most) single result, but got: " <> show err
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+            [] -> exec' db
+                  ("INSERT "
+                  <> domainTableName d
+                  <> " ('rowkey','blockheight','version','txid','tabledata')\
+                     \ VALUES (?,?,?,?,?);")
+                  [SText key
+                  , SInt (fromIntegral bh)
+                  , SInt (fromIntegral version)
+                  , SInt (fromIntegral txid)
+                  , SBlob (toStrict (Data.Aeson.encode v))]
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+    VersionedTables (TableName name) (BlockVersion (BlockHeight bh) (ReorgVersion bv)) -> do
+      let s = "INSERT INTO VersionedTables ('tablename','blockheight','version') VALUES (?,?,?);"
+          theTableName = Utf8 (toS name)
+      callDb $ \db -> exec' db s
+        [ SText theTableName
+        , SInt (fromIntegral bh)
+        , SInt (fromIntegral bv)
+        ]
+
+insertDomain :: DbConstraints m SQLiteEnv => RowKey -> TableName -> BlockHeight -> ReorgVersion -> TxId -> Value -> m ()
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+    \txid UNSIGNED BIGINT,\
+    \tabledata BLOB);")
+  systemInsert (VersionedTables name b)
+
+detectVersionChange :: DbConstraints m SQLiteEnv => BlockHeight -> BlockHash -> m VersionUpdate
+detectVersionChange bRestore hsh  = do
+  bCurrent <- do
+    r <- callDb $ \ db -> qry_ db "SELECT max(blockheight) AS current_block_height FROM BlockHistory;" [RInt]
+    case r of
+      [[SInt bh]] -> return $ BlockHeight (fromIntegral bh)
```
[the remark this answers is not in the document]
emmanueldenloye: addressed.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,532 @@
...
+  | SameVersion !BlockVersion
+  deriving Show
+
+versionUpdate :: DbConstraints m SQLiteEnv => VersionUpdate -> m ()
+versionUpdate (SameVersion bv) =
+  assign bsBlockVersion bv -- TODO: remove if tests pass since version has not changed.
+  -- beginSavepoint "BLOCK"
+versionUpdate (VersionChange (BlockVersion bRestore vNext)) = do
+  assign bsBlockVersion (BlockVersion bRestore vNext)
+  let bvEnv = BlockVersion bRestore vNext
```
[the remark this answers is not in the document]
emmanueldenloye: done

chainweb.cabal
```diff
@@ -197,8 +197,11 @@ library
         , P2P.TaskQueue.Test
 
         -- pact
+        , Chainweb.Pact.Backend.ChainwebPactDb
+        , Chainweb.Pact.Backend.Utils
```
gregorycollins: Sort

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,482 @@
...
+         [] -> return Nothing
+         [[SBlob a]] -> return $ decode $ fromStrict a
+         err -> internalError $ "readRow: Expected (at most) a single result, but got: " <> T.pack (show err)
+
+writeSys :: (AsString k, ToJSON v) => WriteType -> Domain k v -> k -> v -> BlockHandler SQLiteEnv ()
+writeSys wt d k v = do
+        BlockVersion bh version <- gets _bsBlockVersion
+        txid <- gets _bsTxId
+        callDb (write (getKeyString k) bh version txid)
+        let toTableName (Utf8 str) = TableName $ toS str
```
gregorycollins: move to where block

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,482 @@
...
+    [SText key
+    , SInt (fromIntegral bh)
+    , SInt (fromIntegral version)
+    , SInt (fromIntegral txid)
+    , SBlob (toStrict (Data.Aeson.encode v))]
+
+-- don't need the initial select, it will just fail. Look at pact
+backendWriteWrite :: ToJSON v => Utf8 -> Utf8 -> BlockHeight -> ReorgVersion -> TxId -> v -> Database -> IO ()
+backendWriteWrite key tn bh version txid v db =
+  exec' db ("INSERT OR REPLACE INTO [" <> tn <> "] ('rowkey','blockheight','version','txid','tabledata') VALUES (?,?,?,?,?) ;")
```
gregorycollins: Can't we do this for all the cases? I was wondering for them "why isn't this a `REPLACE`"

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,482 @@
...
+    [SText key, SInt (fromIntegral bh), SInt (fromIntegral version), SInt (fromIntegral txid), SBlob (toStrict (Data.Aeson.encode v))]
+
+writeUser :: WriteType -> Domain RowKey (ObjectMap PactValue) -> RowKey -> ObjectMap PactValue -> BlockHandler SQLiteEnv ()
+writeUser wt d k row = do
+      case d of
+        (UserTables _) -> do
```
gregorycollins: You're calling this function based on already having `case`d on the `Domain` in `writeRow` -- just pass the table name instead

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,482 @@
...
+          callDb (backendWriteUpdate (Utf8 $ toS $ asString $ key) tn bh version txid row')
+          record (prependToTableName "USER:" (toTableName $ tn)) d k row'
+        ins = do
+          let tn = domainTableName d
+          callDb $ backendWriteInsert (Utf8 $ toS $ asString $ key) tn bh version txid row
+          record (prependToTableName "USER:" (toTableName $ domainTableName d)) d k row
+
+prependToTableName :: Text -> TableName -> TableName
+prependToTableName txt (TableName name) =
+  TableName (txt <> name)
```
gregorycollins: Write `TableName` instances for `IsString` and `Monoid` instead?

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,482 @@
...
+     deleteHistory bvEnv
+
+tableMaintenanceRowsVersionedTables :: BlockVersion -> BlockHandler SQLiteEnv ()
+tableMaintenanceRowsVersionedTables bv = do
+  tableMaintenanceRowsVersionedSystemTables bv
+  tableMaintenanceRowsVersionedUserTables bv
+
+tableMaintenanceRowsVersionedSystemTables :: BlockVersion -> BlockHandler SQLiteEnv ()
+tableMaintenanceRowsVersionedSystemTables (BlockVersion bh _) = callDb $ \db -> do
+  exec' db "DELETE FROM [SYS:KeySets] WHERE blockheight >= ?" [SInt (fromIntegral bh)]
```
gregorycollins: Should this be `>` instead of `>=`? You're `restore`ing to `bh` right, not to one before it?

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,482 @@
...
+  exec' db "DELETE FROM [SYS:Modules] WHERE blockheight >= ?" [SInt (fromIntegral bh)]
+  exec' db "DELETE FROM [SYS:Namespaces] WHERE blockheight >= ?" [SInt (fromIntegral bh)]
+  exec' db "DELETE FROM [SYS:Pacts] WHERE blockheight >= ?" [SInt (fromIntegral bh)]
+
+{- rename VersionedTables to UserTables -}
+tableMaintenanceRowsVersionedUserTables :: BlockVersion -> BlockHandler SQLiteEnv ()
+tableMaintenanceRowsVersionedUserTables (BlockVersion (BlockHeight bh) _) = do
+  tblNames <- callDb $ \db ->
+    qry db "SELECT tablename FROM VersionedTables WHERE blockheight < ?;" [SInt (fromIntegral bh)] [RText]
```
gregorycollins: This is gonna update every table always, rather than just touching the tables that have been updated on the fork. Here's what I think you should do instead:

```SQL
CREATE TABLE VersionedTableCreation(
    tablename TEXT NOT NULL,
    blockheight BIGINT NOT NULL,
    CONSTRAINT foo UNIQUE(tablename));
CREATE TABLE VersionedTableMutation(
    tablename TEXT NOT NULL,
    blockheight BIGINT NOT NULL,
   CONSTRAINT foo2 UNIQUE(tablename, blockheight));
CREATE INDEX foo3 ON VersionedTableMutation(blockheight);
```

Every time you write a row you have to twiddle the mutation table (with a `REPLACE`). Then during this 
maintenance:

getting list of tables to `DROP`:
```
SELECT DISTINCT tablename FROM VersionedTableCreation WHERE blockheight > ?
```

getting list of tables you have to run `DELETE` on:

```
SELECT DISTINCT tablename FROM VersionedTableMutation WHERE blockheight > ?
```

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
slpopejoy: no for insert/update we want corresponding failure semantics

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
slpopejoy: Please review the algorithm

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
slpopejoy: Also shouldn't this be warning/not building (per `-Werror` on incomplete pattern match??)

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
slpopejoy: these are from `pact` and it already has `IsString`, not `Monoid` unfort.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
slpopejoy: Let's hold off on this until we do on-demand, the specification for now is "pessimistic" which means no dirty tracking.

chainweb.cabal
```diff
@@ -245,6 +248,9 @@ library
         , data-dword >= 0.3
         , deepseq >= 1.4
         , directory >= 1.3
+        -- add version bounds for direct-sqlite
```
slpopejoy: either do it or get rid of the comment

chainweb.cabal
```diff
@@ -245,6 +248,9 @@ library
...
+        , direct-sqlite
+        , ekg >= 0.4
```
slpopejoy: why?

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -83,6 +110,59 @@ instance A.FromJSON PactDbConfig
...
+type VersionHandler p a = ReaderT p (StateT BlockState IO) a
+
+-- data PactDbStateNew = PactDbStateNew
+--   { _pdsPactDb :: P.PactDb (CWDbEnv SQLiteEnv)
+--   , _pdsdbEnv :: CWDbEnv SQLiteEnv
+--   }
+
+-- makeLenses ''PactDbStateNew
+
+data CheckpointerNew p = CheckpointerNew
```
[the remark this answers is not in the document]
slpopejoy: It will replace Checkpointer in the next PR

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -34,19 +47,21 @@ import Test.Tasty.HUnit
...
     (applyContinuation', applyExec', buildExecParsedCode)
 import Chainweb.Pact.Utils (toEnv', toEnvPersist')
+import Chainweb.Pact.Backend.Utils
 import Chainweb.Test.Pact.Utils
 import Chainweb.Test.Utils
 
 tests :: ScheduledTest
 tests = testGroupSch "Checkpointer"
-  [ testCase "testInMemory" testInMemory ]
+  [ testCase "testInMemory" testInMemory, testCase "testRelationalKeyset" testKeyset, testCase "testRelational" testRelational ]
```
slpopejoy: break to lines

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -310,3 +326,409 @@ testCheckpointer loggers CheckpointEnv{..} dbState00 = do
...
+testRelational =
+  withTempSQLiteConnection []  $ \sqlenv -> do
+    let initBlockState = BlockState 0 Nothing (BlockVersion 0 0) M.empty (newLogger loggers "BlockEnvironment")
+        loggers = pactTestLogger False
+    msql <- newMVar (BlockEnv sqlenv initBlockState)
+    runBlockEnv msql initSchema
+    cpEnv <- initRelationalCheckpointerNew msql (newLogger loggers "RelationalCheckpointer") freeGasEnv
+    relationalTest cpEnv
+
+relationalTest :: CheckpointEnvNew SQLiteEnv -> Assertion
```
slpopejoy: add a comment that this copypastes the entire in-memory test.

Also we probably need to migrate the in-memory to match the `NewCheckpointer` in the subsequent PR. Having an in-memory version is useful for performance tests. Thus the comment should be a TODO to have a single test at that point.

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -310,3 +326,409 @@ testCheckpointer loggers CheckpointEnv{..} dbState00 = do
...
+          (BlockEnv sqlenv initBlockState)
+          (\v -> runBlockEnv v initSchema)
+
+begin :: PactDb e -> Method e (Maybe TxId)
+begin pactdb = _beginTx pactdb Transactional
+
+commit :: PactDb e -> Method e [TxLog Value]
+commit pactdb = _commitTx pactdb
+
+runRegression ::
```
slpopejoy: comment that this should be migrated back to Pact

emmanueldenloye: I actually did this before you wrote the comment. I should have uploaded that commit before you started this review.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: @slpopejoy, I had also thought that there should be a case warning, HOWEVER, GHC is able to figure out that there is only ONE conceivable value for the argument `d` because of the type signature of `writeUser`. Which does imply that I don't need the case statement. I'l get rid of the case statement.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: I had considered that, but then I would have to make changes to Pact. I would need the permission of @slpopejoy to do that.

chainweb.cabal
emmanueldenloye: I'm not sure how that got there, I don't remember adding it. It has been removed. Chainweb still builds.

chainweb.cabal
emmanueldenloye: done

chainweb.cabal
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: done

test/Chainweb/Test/Pact/Checkpointer.hs
emmanueldenloye: done

test/Chainweb/Test/Pact/Checkpointer.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,482 @@
...
+  callDb $ \db -> exec_ db
+    "CREATE TABLE VersionedTables (tablename TEXT\
+  \ , blockheight UNSIGNED BIGINT\
+  \ , version UNSIGNED BIGINT\
+  \ , CONSTRAINT versionTableConstraint UNIQUE(version, blockheight));"
+
+createVersionedTable :: TableName -> BlockVersion -> BlockHandler SQLiteEnv ()
+createVersionedTable name b = do
+  callDb $ \db -> exec_ db $
+                  "CREATE TABLE "
```
gregorycollins: Does this table also need an index?

slpopejoy: yes

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
```diff
@@ -13,59 +13,112 @@ module Chainweb.Pact.Backend.InMemoryCheckpointer
...
+--               <> ", known=" <> pack (show (HMS.keys store))
+--           Nothing -> "Something went wrong at genesis. There is nothing here!"
+
 
-initInMemoryCheckpointEnv :: P.Logger -> P.GasEnv -> IO CheckpointEnv
-initInMemoryCheckpointEnv logger gasEnv = do
-    inmem <- newMVar mempty
+
+data Store = Store
+  { _theStore :: HashMap (BlockHeight, BlockHash) (DbEnv PureDb)
```
slpopejoy: use uniform prefixes, or don't bother with `_`.

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
```diff
@@ -13,59 +13,112 @@ module Chainweb.Pact.Backend.InMemoryCheckpointer
...
-            Nothing -> return $ Left $
-              "InMemoryCheckpointer: Restore not found: height=" <> show height
-              <> ", hash=" <> show hash
-              <> ", known=" <> show (HMS.keys store)
+doRestore :: MVar Store ->  Maybe (BlockHeight, ParentHash) -> IO PactDbEnv'
+doRestore lock (Just (height, hash)) = do
+    pactdbenv <- withMVarMasked lock $ \store -> do
+        case HMS.lookup (pred height, hash) (_theStore store) of
+            Just dbenv -> do
+              mdbenv <- newMVar dbenv
```
slpopejoy: you need to store `mdbenv` in `_dbenv` here, for retrieval in `save` (and `discard` should probably reset it, so maybe `_dbenv` should be `Maybe (MVar (DbEnv PureDb))`

slpopejoy: eh, skip the Maybe stuff. I think if you store it here that should fix it

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
```diff
@@ -13,59 +13,112 @@ module Chainweb.Pact.Backend.InMemoryCheckpointer
...
+              <> ", hash=" <> pack (show hash)
+              <> ", known=" <> pack (show (HMS.keys (_theStore store)))
 
-restoreInitial' :: MVar Store -> IO (Either String PactDbState)
-restoreInitial' lock = do
-    let bh = nullBlockHash
-    restore' lock (BlockHeight 0) bh
+    modifyMVar_ lock (\s ->  pure $ s { _tempSaveHeight = Just height })
+    return pactdbenv
+doRestore lock Nothing = withMVarMasked lock $ \store -> return $ PactDbEnv' (PactDbEnv pactdb (_dbenv store))
```
slpopejoy: store _dbenv here too

src/Chainweb/Pact/Backend/Utils.hs
```diff
@@ -110,10 +112,10 @@ withSQLiteConnection :: String -> [Pragma] -> Bool -> (SQLiteEnv -> IO c) -> IO
 withSQLiteConnection file ps todelete action = bracket opener closer action
   where
     closer c = do
-      void $ close $ _sConn c
+      void $ close_v2 $ _sConn c
       when todelete (removeFile file)
     opener = do
-      e <- open $ fromString file
+      e <- open_v2 (fromString file) (0x00000002 .|. 0x00000004 .|. 0x00010000) "unix"
```
gregorycollins: Do you have to pass in the vfs name here, or does `nullPtr` work as default?

emmanueldenloye: It seems I have to pass a definitive value here.

slpopejoy: Is anybody supposed to understand what the opaque bit values mean? Please comment what each is and why we should care.

src/Chainweb/Pact/Backend/Utils.hs
```diff
@@ -0,0 +1,204 @@
...
+instance StringConv String Utf8 where
+  strConv l = Utf8 . strConv l
+
+instance StringConv Utf8 String where
+  strConv l (Utf8 bytestring) = strConv l bytestring
+
+fastNoJournalPragmas :: [Pragma]
+fastNoJournalPragmas = [
+  "synchronous = NORMAL",
+  "journal_mode = WAL",
```
slpopejoy: Why WAL and not MEMORY?

src/Chainweb/Pact/PactService.hs
```diff
@@ -465,16 +450,16 @@ applyPactCmd isGenesis (Env' dbEnv) cmdIn miner = do
...
+        then applyGenesisCmd logger dbenv pd spv cmd
+        else applyCmd logger dbenv miner gasModel pd spv cmd
     pure $! toHashCommandResult result
 
 toHashCommandResult :: P.CommandResult [P.TxLog A.Value] -> HashCommandResult
 toHashCommandResult = over (P.crLogs . _Just) (P.pactHash . encodeToByteString)
 
-updateState :: PactDbState  -> PactServiceM ()
-updateState = assign psStateDb
+updatePactDbEnv :: PactDbEnv'  -> PactServiceM ()
```
slpopejoy: Per other comments, remove.

src/Chainweb/Pact/Types.hs
```diff
@@ -35,7 +35,8 @@ module Chainweb.Pact.Types
   , psCheckpointEnv
   , psSpvSupport
   , psPublicData
-  , psStateDb
+  -- , psStateDb
```
slpopejoy: remove and all other commented code

src/Chainweb/Pact/Utils.hs
```diff
@@ -20,10 +20,13 @@ module Chainweb.Pact.Utils
 import Data.Aeson
 
 import Control.Concurrent.MVar
+-- import Control.Lens
```
slpopejoy: Remove commented code

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -274,3 +309,48 @@ withPactCtx v cutDB f
         evalPactServiceM ctx pact
   where
     start = testPactCtx v (someChainId v)
+
+initializeSQLite :: IO (IO (), SQLiteEnv)
+initializeSQLite = do
+      (file, del) <- newTempFile
+      -- TODO: Should change what is passed as the vfs_module.
+      e <- open_v2 (fromString file) (0x00000002 .|. 0x00000004 .|. 0x00010000) "unix"
```
slpopejoy: Needless to say, we can't be copying and pasting opaque values like this around.

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
```diff
@@ -11,62 +11,76 @@ module Chainweb.Pact.Backend.InMemoryCheckpointer
...
+  { _theStore :: HashMap (BlockHeight, BlockHash) (DbEnv PureDb)
+  , _tempSaveHeight :: Maybe BlockHeight
+  , _dbenv :: MVar (DbEnv PureDb)
+  }
 
-restoreInitial' :: MVar Store -> IO (Either String PactDbState)
-restoreInitial' lock = do
-    let bh = nullBlockHash
-    restore' lock (BlockHeight 0) bh
+initInMemoryCheckpointEnv :: Loggers -> Logger -> GasEnv -> IO (PactDbEnv', CheckpointEnv)
```
slpopejoy: don't return `PactDbEnv'` (see other comments)

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -0,0 +1,100 @@
...
+
+-- chainweb
+import Chainweb.BlockHash
+import Chainweb.BlockHeader
+import Chainweb.Pact.Backend.ChainwebPactDb
+import Chainweb.Pact.Backend.Types
+import Chainweb.Pact.Backend.Utils
+import Chainweb.Pact.Service.Types (internalError)
+
+initRelationalCheckpointer :: BlockState -> SQLiteEnv -> Logger -> GasEnv -> IO (PactDbEnv', CheckpointEnv)
```
slpopejoy: Don't return PactDbEnv' (see other comments)

src/Chainweb/Pact/Types.hs
```diff
@@ -152,11 +153,20 @@ data PactServiceEnv = PactServiceEnv
...
 
+-- data PactServiceState = PactServiceState
+--   {_psStateValidated :: Maybe BlockHeader
+--   }
+
 data PactServiceState = PactServiceState
-  { _psStateDb :: !PactDbState
-  , _psStateValidated :: !(Maybe BlockHeader)
+  { _psPactDbEnv :: PactDbEnv'
+  , _psStateValidated :: Maybe BlockHeader
```
slpopejoy: Let's move `_psPactDbEnv` out of state, it's dangerous as we could accidentally use the wrong instance. Instead, use the `PactDbEnv'` returned from `restore`, threaded through the various functions in `PactService`, and don't return this from the various initializer calls.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -0,0 +1,100 @@
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE FlexibleContexts  #-}
+{-# LANGUAGE ViewPatterns      #-}
+
+-- |
+-- Module: Chainweb.Pact.RelationalCheckpointer
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/PactService.hs
```diff
@@ -73,21 +75,20 @@ import Chainweb.BlockHeader
 import Chainweb.ChainId (ChainId)
 import Chainweb.CutDB (CutDb)
 import Chainweb.Logger
-import Chainweb.Pact.Backend.InMemoryCheckpointer (initInMemoryCheckpointEnv)
-import Chainweb.Pact.Backend.MemoryDb (mkPureState)
+-- import Chainweb.Pact.Backend.InMemoryCheckpointer (initInMemoryCheckpointEnv)
```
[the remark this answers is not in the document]
emmanueldenloye: done

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -63,15 +73,20 @@ import Pact.Types.Logger
...
 import Chainweb.ChainId (chainIdToText)
 import Chainweb.CutDB
-import Chainweb.Pact.Backend.Types
+-- import Chainweb.Pact.Backend.Types
 import Chainweb.Pact.PactService
 import Chainweb.Pact.Backend.InMemoryCheckpointer (initInMemoryCheckpointEnv)
-import Chainweb.Pact.Backend.MemoryDb (mkPureState)
+import Chainweb.Pact.Backend.RelationalCheckpointer (initRelationalCheckpointer)
+import Chainweb.Pact.Backend.Utils
+-- import Chainweb.Pact.Backend.MemoryDb (mkPureState)
```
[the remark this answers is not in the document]
emmanueldenloye: done

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -63,15 +73,20 @@ import Pact.Types.Logger
...
 import Pact.Types.SPV (noSPVSupport)
 import Pact.Types.Util (toB16Text)
+import Pact.Types.SQLite hiding (fastNoJournalPragmas)
 
 -- internal modules
 
 import Chainweb.ChainId (chainIdToText)
 import Chainweb.CutDB
-import Chainweb.Pact.Backend.Types
+-- import Chainweb.Pact.Backend.Types
```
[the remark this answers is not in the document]
emmanueldenloye: done

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -80,7 +95,7 @@ import qualified Chainweb.Version as V
 import Chainweb.WebPactExecutionService
 
 import Pact.Gas
-import Pact.Interpreter
+-- import Pact.Interpreter
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/MemoryDb.hs
```diff
@@ -24,3 +24,6 @@ mkPureState env = do
     return $
         PactDbState
             { _pdbsDbEnv = envPersist' }
+
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -29,75 +32,172 @@ module Chainweb.Pact.Backend.Types
...
+             )
+
+data PactDbEnv' = forall e. PactDbEnv' (PactDbEnv e)
+
+instance Logging (BlockHandler p) where
+  log c s = view logger >>= \l -> liftIO $ logLog l c s
+
+type ParentHash = BlockHash
+
+-- TODO: pass (Maybe blockHash) (only populated on validateblock) to
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -29,75 +32,172 @@ module Chainweb.Pact.Backend.Types
...
 
--- functions like the ones below need to be implemented internally
--- , prepareForValidBlock :: BlockHeight -> BlockHash -> IO (Either String PactDbState)
--- , prepareForNewBlock :: BlockHeight -> BlockHash -> IO (Either String PactDbState)
 data CheckpointEnv = CheckpointEnv
-    { _cpeCheckpointer :: Checkpointer
-    , _cpeLogger :: P.Logger
-    , _cpeGasEnv :: P.GasEnv
+    { _cpeCheckpointer :: !Checkpointer
+    , _cpeLogger :: Logger
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+  callDb "tableMaintenanceRowsVersionedSystemTables" $ \db -> do
+    exec' db "DELETE FROM [SYS:keysets] WHERE blockheight >= ?" [SInt (fromIntegral bh)]
+    exec' db "DELETE FROM [SYS:modules] WHERE blockheight >= ?" [SInt (fromIntegral bh)]
+    exec' db "DELETE FROM [SYS:namespaces] WHERE blockheight >= ?" [SInt (fromIntegral bh)]
+    exec' db "DELETE FROM [SYS:pacts] WHERE blockheight >= ?" [SInt (fromIntegral bh)]
+
+tableMaintenanceRowsVersionedUserTables :: BlockHandler SQLiteEnv ()
+tableMaintenanceRowsVersionedUserTables = do
+  (BlockVersion bh _) <- gets _bsBlockVersion
+  tblNames <- callDb "tableMaintenanceRowsVersionedUserTables" $ \db -> qry db
```
gregorycollins: Can we float the `callDb` out so that there's one `callDb` and everything happens under one lambda? Repeatedly using ` callDb` is going to waste time pointlessly twiddling the mutex

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+          ("DELETE FROM [" <> tbl <> "] WHERE blockheight >= ?")
+          [SInt (fromIntegral bh)]
+      _ -> internalError "tableMaintenanceRowsUserTables: An error occured\
+                        \ while querying the\
+                        \ VersionedTables table for table names."
+
+dropUserTables :: BlockHandler SQLiteEnv ()
+dropUserTables = do
+  (BlockVersion (BlockHeight bh) _) <- gets _bsBlockVersion
+  tblNames <- callDb "dropUserTables" $ \db -> qry db
```
gregorycollins: Same here (callDb)

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+      [SText name] -> do
+        callDb "dropUserTables" $ \db -> exec_ db $ "DROP TABLE [" <> name <> "];"
+      _ -> internalError "dropUserTables: An error occured\
+                        \ while querying the\
+                        \ VersionedTables table for table names."
+
+deleteHistory :: BlockHandler SQLiteEnv TxId
+deleteHistory = do
+  (BlockVersion bh v) <- gets _bsBlockVersion
+  callDb "Deleting from BlockHistory" $ \db -> exec' db
```
gregorycollins: One call to `callDb`

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+      log "DDL" $ "initSchema: "  ++ show name
+      callDb "initSchema" $ \db ->
+        exec_ db $
+        "CREATE TABLE "
+        <> Utf8 (toS (flip T.snoc ']' $ T.cons '[' $ asString name))
+        <> " (rowkey TEXT\
+           \, blockheight UNSIGNED BIGINT\
+           \, version UNSIGNED BIGINT\
+           \, txid UNSIGNED BIGINT\
+           \, rowdata BLOB)"
```
gregorycollins: Index here too

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -0,0 +1,100 @@
...
+               [SText t] -> exec_ db ("DROP TABLE [" <> t <> "];")
+               _ -> internalError "Something went wrong when resetting tables."
+        beginSavepoint Block
+      _ -> internalError "restoreInitial: The genesis state cannot be recovered!"
+  return $ PactDbEnv' $ PactDbEnv chainwebpactdb dbenv
+
+doSave :: Db -> BlockHash -> IO ()
+doSave dbenv hash = runBlockEnv dbenv $ do
+  commitSavepoint Block
+  t <- gets _bsTxId
```
gregorycollins: rename `t` to something like `nextTxId`

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/SQLite/V2.hsc
```diff
@@ -0,0 +1,14 @@
+{-# LANGUAGE ForeignFunctionInterface #-}
+
+module Chainweb.Pact.Backend.SQLite.V2 where
+
+import Database.SQLite3.Bindings.Types
+
+import Foreign
+import Foreign.C
+
+foreign import ccall "sqlite3_open_v2"
```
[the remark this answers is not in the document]
gregorycollins: Needed for most-paranoid sqlite mutex locking semantics (as discussed offline)

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
emmanueldenloye: done

test/Chainweb/Test/Pact/Utils.hs
emmanueldenloye: moved to utils, and documented

src/Chainweb/Pact/Utils.hs
emmanueldenloye: done

src/Chainweb/Pact/Types.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -29,75 +32,172 @@ module Chainweb.Pact.Backend.Types
...
     , _pdbcGasRate :: Maybe Int
     } deriving (Eq, Show, Generic)
 
-instance A.FromJSON PactDbConfig
+instance FromJSON PactDbConfig
 
 makeLenses ''PactDbConfig
 
+data SQLiteEnv = SQLiteEnv
+  { _sConn :: Database
```
gregorycollins: add bang on `_sConn` too

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -29,75 +32,172 @@ module Chainweb.Pact.Backend.Types
...
+  { _bsTxId :: !TxId
+  , _bsMode :: !(Maybe ExecutionMode)
+  , _bsBlockVersion :: !BlockVersion
+  , _bsTxRecord :: !(Map TableName [TxLog Value])
+  }
+
+makeLenses ''BlockState
+
+data BlockDbEnv p = BlockDbEnv
+  { _bdbenvDb :: p
```
gregorycollins: Add bangs

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -29,75 +32,172 @@ module Chainweb.Pact.Backend.Types
...
+  , _logger :: Logger
+  }
+
+makeLenses ''BlockDbEnv
+
+
+data BlockEnv p = BlockEnv
+  { _benvDb :: BlockDbEnv p
```
gregorycollins: add bang

src/Chainweb/Pact/Backend/Utils.hs
```diff
@@ -0,0 +1,204 @@
+{-# LANGUAGE FlexibleInstances #-}
```
gregorycollins: Sort this list

src/Chainweb/Pact/Backend/Utils.hs
```diff
@@ -0,0 +1,204 @@
...
+  case res of
+    Left err -> internalError $ "callDb (" <> callerName <> "): " <> (pack $ show err)
+    Right r -> return r
+
+withSavepoint :: SavepointName -> BlockHandler SQLiteEnv a -> BlockHandler SQLiteEnv a
+withSavepoint name action = do
+  beginSavepoint name
+  result <- try action
+  case result of
+    Right r -> do
```
gregorycollins: add bang to `r` or change `return r` to `liftIO $ evaluate r`

src/Chainweb/Pact/Backend/Utils.hs
gregorycollins: From the sqlite docs:

> `If the fourth parameter is a NULL pointer then the default sqlite3_vfs object is used.`

So you definitely want to prefer `nullPtr` to `"unix"` here

src/Chainweb/Pact/Backend/Utils.hs
gregorycollins: `MEMORY` isn't safe (if machine dies, process crashes or is unsafely stopped, etc., the DB will be corrupt.) We do need correct transactional semantics here

src/Chainweb/Pact/PactService.hs
```diff
@@ -140,22 +141,17 @@ initPactService' cid chainwebLogger spv act = do
     let loggers = pactLoggers chainwebLogger
     let logger = P.newLogger loggers $ P.LogName ("PactService" <> show cid)
     let gasEnv = P.GasEnv 0 0.0 (P.constGasModel 1)
+    let blockstate = BlockState 0 Nothing (BlockVersion 0 0) M.empty
 
-    checkpointEnv <- initInMemoryCheckpointEnv logger gasEnv
-
-    env <- P.mkPureEnv loggers
-    theState <- mkPureState env
+    -- TODO: The file and pragmas should come from a config file
```
gregorycollins: N.B. this is v. important to fix and we need to get to it in the next changeset

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+-- WE SHOULD FLUSH THE IN-BLOCK CACHE HERE
+
+doCommit :: BlockHandler SQLiteEnv [TxLog Value]
+doCommit = use bsMode >>= \mm -> case mm  of
+  Nothing -> doRollback >> internalError "doCommit: Not in transaction"
+  Just m -> do
+    txrs <- use bsTxRecord
+
+    if m == Transactional then do
+      bsTxId += 1
```
gregorycollins: Would prefer not to allow a thunk in here at all, although I don't think you can accumulate a thunk chain of length >= 2. (We always force txid at each row write and block history insert). Prefer `Control.Monad.State.Strict.modify'` if possible.

test/Chainweb/Test/Pact/Utils.hs
gregorycollins: :+1: at least name them

chainweb.cabal
```diff
@@ -215,18 +215,23 @@ library
         , P2P.TaskQueue.Test
 
         -- pact
+        , Chainweb.Pact.Backend.ChainwebPactDb
         , Chainweb.Pact.Backend.InMemoryCheckpointer
         , Chainweb.Pact.Backend.MemoryDb
+        , Chainweb.Pact.Backend.RelationalCheckpointer
         , Chainweb.Pact.Backend.Types
+        , Chainweb.Pact.Backend.Utils
+        , Chainweb.Pact.Backend.SQLite.DirectV2
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
slpopejoy: Are you sure you want to go against Ed lens code here? I mean I would probably prefer `.= succ` but we use lens all over the place. Given we're IO bound here I can't really justify this low-level of a change.

chainweb.cabal
```diff
@@ -263,6 +268,7 @@ library
         , deepseq >= 1.4
         , digraph >= 0.1.0.2
         , directory >= 1.3
+        , direct-sqlite >= 2.3
```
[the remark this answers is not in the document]
emmanueldenloye: done

default.nix
```diff
@@ -334,7 +334,13 @@ in
     shellToolOverrides = ghc: super: {
       stack = pkgs.stack;
       cabal-install = pkgs.haskellPackages.cabal-install;
-      ghcid = pkgs.haskellPackages.ghcid;
+      # ghcid = pkgs.haskellPackages.ghcid;
```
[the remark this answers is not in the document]
emmanueldenloye: I've taken this out in a commit that is to come.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+-- |
+-- Module: Chainweb.Pact.ChainwebPactDb
+-- Copyright: Copyright © 2019 Kadena LLC.
+-- License: MIT
+-- Maintainer: Emmanuel Denloye-Ito <emmanuel@kadena.io>
+-- Stability: experimental
+--
+
+module Chainweb.Pact.Backend.ChainwebPactDb
+  ( chainwebpactdb
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+            [SInt (fromIntegral version)]
+            [RText]
+  forM ks $ \row -> do
+    case row of
+      [SText (Utf8 k)] -> return $ fromString $ toS k
+      _ -> internalError "doKeys: The impossible happpened."
+{-# INLINE doKeys #-}
+
+doTxIds :: TableName -> TxId -> BlockHandler SQLiteEnv [TxId]
+doTxIds (TableName tn) (TxId tid) = do
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+  Just m -> do
+    txrs <- use bsTxRecord
+
+    if m == Transactional then do
+      bsTxId += 1
+      -- commit
+      commitSavepoint PactDbTransaction
+      resetTemp
+    else doRollback
+    return $ fromdlist $ foldr (\a b -> todlist a . b) id txrs
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+doGetTxLog :: FromJSON v => Domain k v -> TxId -> BlockHandler SQLiteEnv [TxLog v]
+doGetTxLog d txid = do
+  BlockVersion bh version <- gets _bsBlockVersion
+  rows <- callDb "doGetTxLog" $ \db -> qry db
+        ("SELECT rowkey, rowdata FROM [" <> domainTableName d <> "] WHERE txid = ? AND version = ? AND blockheight = ?")
+        [SInt (fromIntegral txid), SInt (fromIntegral bh), SInt (fromIntegral version)]
+        [RText, RBlob]
+  forM rows $ \case
+    [SText key, SBlob value] ->
+      case Data.Aeson.decode $ fromStrict value of
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+        Nothing -> internalError $ "doGetTxLog: Unexpected value, unable to deserialize log"
+        Just v -> return $ TxLog (toS $ unwrap $ domainTableName d) (toS $ unwrap $ key) v
+    err -> internalError $ "doGetTxLog: Expected single row with two columns as the result, got: " <> T.pack (show err)
```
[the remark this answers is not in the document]
emmanueldenloye: We cannot it seems.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+                      \, txid UNSIGNED BIGINT\
+                      \, rowdata BLOB)"
+  versionedTablesInsert name b
+
+handleVersion :: BlockHeight -> ParentHash -> BlockHandler SQLiteEnv TxId
+handleVersion bRestore hsh = do
+  bCurrent <- do
+    r <- callDb "handleVersion" $ \ db -> qry_ db "SELECT max(blockheight) AS current_block_height FROM BlockHistory;" [RInt]
+    SInt bh <- liftIO $ expectSingleRowCol "handleVersion: (block):" r
+    return $ BlockHeight (fromIntegral bh)
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+         [RInt]
+     newtxid <- if null result
+       then do
+         versionHistoryInsert bs (fromIntegral txid)
+         return (fromIntegral txid :: TxId)
+       else do
+        r <- liftIO $ expectSingleRowCol "handleVersion: block version check" result
+        case r of
+          SInt res -> return (fromIntegral res)
+          _ -> error "lol"
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -0,0 +1,495 @@
...
+         [SInt (fromIntegral bh), SInt (fromIntegral version)]
+         [RInt]
+     newtxid <- if null result
+       then do
+         versionHistoryInsert bs (fromIntegral txid)
+         return (fromIntegral txid :: TxId)
+       else do
+        r <- liftIO $ expectSingleRowCol "handleVersion: block version check" result
+        case r of
+          SInt res -> return (fromIntegral res)
```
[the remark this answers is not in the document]
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: Sure can do.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/Types.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/Types.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/Types.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/Utils.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/Utils.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: done

src/Chainweb/Pact/Types.hs
emmanueldenloye: done

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -29,75 +32,173 @@ module Chainweb.Pact.Backend.Types
...
 data Checkpointer = Checkpointer
-    { restore :: BlockHeight -> BlockHash -> IO (Either String PactDbState)
-    , restoreInitial ::IO (Either String PactDbState)
-    , save :: BlockHeight -> BlockHash -> PactDbState -> IO (Either String ())
-    , saveInitial :: PactDbState -> IO (Either String ())
-    , discard :: PactDbState -> IO (Either String ())
+    {
+      restore :: Maybe (BlockHeight, ParentHash) -> IO PactDbEnv'
+    , save :: BlockHash -> IO ()
+    , discard :: IO ()
```
slpopejoy: Wow this is a really clean interface now. Awesome

PR #186, opened 2019-05-07
Pass --hide-successes to test suite

mightybyte:
See this link for how this change looks in CI when there is a test failure:

https://gitlab.com/kadena-io/chainweb-node/-/jobs/208064271

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -54,6 +54,7 @@ in
           doCheck = runTests;
           doHaddock = runTests;
           doCoverage = runCoverage;
+          testTarget = "--test-option=--hide-successes";
         }));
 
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -54,6 +54,7 @@ in
           doCheck = runTests;
           doHaddock = runTests;
           doCoverage = runCoverage;
+          testTarget = "--test-option=--hide-successes";
```
fosskers: Which tool is expecting `--test-option`? Is it the underlying cabal?

mightybyte: Nix passes it to `Setup.hs` under the hood.

PR #193, opened 2019-05-08
Coin Faucet Contract

ggobugi27:
Coin-faucet module that initiates faucet. Faucet does the following: 
 - Gives out coin to account at request. Limits one request per hour. 
 - Receives coin return from accounts.
 - Records account's past behavior with faucet. (last-tx-time, total-coins-earned, total-coins-returned)


pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
...
+
+  (defun describe-behavior:string (address)
+    @doc "Describes past coin request and returns of the account"
+
+    (with-capability (HISTORY)
+        (with-read history-table address {
+          "total-coins-earned":= total-coins-earned,
+          "total-coins-returned":= total-coins-returned
+          }
+          (format "Address {} had earned {} and returned {}"
```
emilypi: small typo: "had" -> "has"

pact/coin-contract/coin-faucet.repl
```diff
@@ -0,0 +1,71 @@
...
+(expect-failure "Return fails due to insufficient funds in k1" (return-coin 'k1 10.0))
+
+(return-coin 'k2 1.0)
+(expect "Return succeeds and k2 balance decreases to 0.0" 0.0 (account-balance 'k2))
+(expect "Coin 1.0 was credited in faucet account" 999999999998.0 (account-balance 'faucet-account))
+
+;;;; TESTING describe-behavior
+
+(expect "Describes k2's behavior" "Address k2 had earned 1.0 and returned 1.0" (describe-behavior 'k2))
+(expect "Describes k1's behavior" "Address k1 had earned 2.0 and returned 0.0" (describe-behavior 'k1))
```
emilypi: These will need to change with the above comment ("had" -> "has")

pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
...
+  (defconst WAIT_TIME_PER_REQUEST 3600.0)
+  (defconst EPOCH  (time "1970-01-01T00:00:00Z"))
+
+  ; --------------------------------------------------------------------------
+  ; Coin Faucet Contract
+  ; --------------------------------------------------------------------------
+
+  (defun faucet-guard:guard () (create-module-guard 'faucet-admin))
+
+  (defun request-coin:string (address address-guard)
```
mightybyte: Can we extend this to allow people to request more than one coin?  I know doesn't seem like a very significant difference, but psychologically it feels better to have a few more coins.  It also scales better with evolving transaction fees.  A limit of 10 - 20 coins seems like a reasonable start.

pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
+(module coin-faucet FAUCET-GOVERNANCE
+
+  "'coin-faucet' represents Kadena's Coin Faucet Contract."
+
+  (defcap FAUCET-GOVERNANCE () true)
```
slpopejoy: add a comment that governance is TBD

pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
...
+
+  (use coin)
```
slpopejoy: add comment TODO to use hashed import

pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
...
+  (defconst FAUCET_ACCOUNT 'faucet-account)
+  (defconst COIN_PER_REQUEST 1.0)
+  (defconst WAIT_TIME_PER_REQUEST 3600.0)
+  (defconst EPOCH  (time "1970-01-01T00:00:00Z"))
+
+  ; --------------------------------------------------------------------------
+  ; Coin Faucet Contract
+  ; --------------------------------------------------------------------------
+
+  (defun faucet-guard:guard () (create-module-guard 'faucet-admin))
```
slpopejoy: It really makes more sense for this to be a `defconst` but I know that didn't work. Need to revisit in pact

pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
...
+
+  (defun request-coin:string (address address-guard)
+    @doc "Transfers COIN_PER_REQUEST coins from faucet account to requester.  \
+    \ Inserts or updates the record of the requester in history table to keep \
+    \ track of the record. Limits the number of coin giveout by time,         \
+    \ WAIT_TIME_PER_REQUEST."
+
+    (with-capability (HISTORY)
+      (with-default-read history-table address
+        {"last-tx-time": EPOCH, "total-coins-earned": 0.0, "total-coins-returned": 0.0}
```
slpopejoy: break to multiple lines for legibility

pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
...
+        (write history-table address {
+          "last-tx-time": (curr-time),
+          "total-coins-earned": (+ COIN_PER_REQUEST total-coins-earned),
+          "total-coins-returned": total-coins-returned }))))
+
+  (defun return-coin:string (address amount)
+    @doc "Returns the coin back to the faucet account after use. Updates the    \
+    \ history table to keep track of behavior. "
+
+    @model [(property (> amount 0.0))]
```
slpopejoy: wow it would appear this is enforced by coin's `transfer`. FV is a trip ...

pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
...
+    total-coins-returned:decimal
+    )
+
+  (deftable history-table: {history})
+
+  ; --------------------------------------------------------------------------
+  ; Capabilities
+  ; --------------------------------------------------------------------------
+
+  (defcap HISTORY ()
```
slpopejoy: As far as I can tell this is not required. Table access is already guarded and HISTORY is only used to duplicate this. Generally, the only reason to use a module guard is to make some _external_ operation controlled by the module, ie the ownership of the faucet tokens in the coin contract. Inside of a module, the module operates on its tables with the capability inherently. The `defcap` should be removed.

pact/coin-contract/coin-faucet.pact
```diff
@@ -0,0 +1,94 @@
...
+
+  (defun curr-time:time ()
+    @doc "Returns current chain's block-time in time type"
+    (add-time EPOCH (at 'block-time (chain-data)))
+  )
+
+)
+
+(create-table history-table)
+(coin.coinbase FAUCET_ACCOUNT (faucet-guard) 1000000000000.0)
```
slpopejoy: Following up on the HISTORY cap discussion above, this is the quintessential example of why the module guard exists, it's pretty difficult to come up with in-module _static_ module guard use case, as opposed to having a ledger where the module owns one row alongside other guards. As a reminder, capabilities guard in-module operations and cannot be exported outside a module, but guards can and should be used anywhere.

pact/coin-contract/coin-faucet.pact
ggobugi27: Changed request-coin to allow up to 20 maximum coins per request.

pact/coin-contract/coin-faucet.pact
ggobugi27: Added the comment

pact/coin-contract/coin-faucet.pact
ggobugi27: Added the comment

pact/coin-contract/coin-faucet.pact
ggobugi27: Changed to lines per each field.

pact/coin-contract/coin-faucet.pact
ggobugi27: Waiting for fv to see what happened..

pact/coin-contract/coin-faucet.pact
ggobugi27: Removed HISTORY

pact/coin-contract/coin-faucet.repl
```diff
@@ -0,0 +1,86 @@
...
+
+(return-coin 'k2 8.0)
+(expect "Return succeeds and k2 balance decreases from 10.0 to 2.0"
+  2.0 (account-balance 'k2))
+(expect "Coin 8.0 was credited in faucet account"
+  99999999999977.0 (account-balance 'faucet-account))
+
+;;;; TESTING describe-behavior
+
+(expect "Describes k2's behavior"
```
slpopejoy: I don't understand this operation. Why do we need `describe-behavior`? Can `txlog` or `keylog` suffice instead?

ggobugi27: The intention of describe-behavior was to show that total transaction behaviors (total-coins-earned, total-coins-returned) are saved in history-table for easy lookup, for later possibility of giving privileges to well-behaved accounts. The function itself serves no use.

slpopejoy: I get it -- I think my confusion is the formatted string -- it's better to directly expose the query and let the client decide what to do with it otherwise somebody might have to parse the string. Thus, it's a simple function `(defun read-history:object{history} (address) (read history-table address))` to provide access to the history.

ggobugi27: Replaced describe-behavior with read-history.

PR #194, opened 2019-05-09
TGX: Round-robin Chain Selection

fosskers:
This PR changes the behaviour of the Transaction Generator, such that it sends requests to all available chains in an even way (round-robin). Eventually  more schemes could be added.

It also reworks certain logging logic.

### DIFF
--- a/examples/TransactionGenerator.hs
+++ b/examples/TransactionGenerator.hs
@@ -92,7 +92,7 @@ import Chainweb.HostAddress
 import Chainweb.Pact.RestAPI
 import Chainweb.RestAPI.Utils
 import Chainweb.Simulate.Contracts.CoinContract
-import Chainweb.Simulate.Contracts.Common
+import qualified Chainweb.Simulate.Contracts.Common as Sim
 import Chainweb.Simulate.Contracts.HelloWorld
 import Chainweb.Simulate.Contracts.SimplePayments
 import Chainweb.Simulate.Utils
@@ -112,7 +112,7 @@ instance Default MeasureTime where
   def = MeasureTime False
 
 data TransactionCommand
-  = DeployContracts [ContractName] MeasureTime
+  = DeployContracts [Sim.ContractName] MeasureTime
   | RunStandardContracts TimingDistribution MeasureTime
   | RunSimpleExpressions TimingDistribution MeasureTime
   | PollRequestKeys [ByteString] MeasureTime
@@ -137,7 +137,7 @@ transactionCommandFromText = readTransactionCommandBytes . T.encodeUtf8
 {-# INLINE transactionCommandFromText #-}
 
 readTransactionCommandBytes :: MonadThrow m => B8.ByteString -> m TransactionCommand
-readTransactionCommandBytes = parseBytes "transaction-command" transactionCommandParser
+readTransactionCommandBytes = Sim.parseBytes "transaction-command" transactionCommandParser
 {-# INLINE readTransactionCommandBytes #-}
 
 transactionCommandParser :: A.Parser TransactionCommand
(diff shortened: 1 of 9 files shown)


examples/TransactionGenerator.hs
```diff
@@ -241,178 +263,196 @@ scriptConfigParser = id
...
-        void $ lift $ liftBaseDiscard forkIO $ do
-          (time,response) <- liftIO $ measureDiffTime (runClientM (listen version chain listenerRequest) clientEnv)
-          liftIO $ print response
-          -- withLabel ("component", "transaction-generator") $ -- add this back in later
-          logg Info $ toLogMessage (("It took " <> sshow time <> " seconds to get back the result.") :: Text)
-          -- withLabel ("component", "transaction-generator") $ -- add this back in later
-          logg Info $ toLogMessage (("The associated request is " <> sshow requestKey <> "\n" <> sshow response) :: Text)
+    forkedListen :: ClientEnv -> ChainwebVersion -> RequestKey -> IO ()
+    forkedListen ce v rk = do
+      print rk  -- TODO Should this be printed or logged?
```
emmanueldenloye: I think I had left that there by accident. I think it should be logged, so along with the metrics collected below.

examples/TransactionGenerator.hs
```diff
@@ -1,13 +1,15 @@
 {-# LANGUAGE BangPatterns                    #-}
 {-# LANGUAGE DataKinds                       #-}
-{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveAnyClass                  #-}
```
fosskers: Unfortunately something about this file doesn't allow `stylish-haskell` to succeed on it (otherwise these `#-}` would have been reformatted back).

exe/CwTool.hs
```diff
@@ -55,7 +55,7 @@ topLevelCommands =
       "Convert a persisted Chainweb into .gexf format for viewing"
       Chain2Gexf.main
   , CommandSpec
-      "gen-transactions"
+      "txg"
```
fosskers: Note!

examples/TransactionGenerator.hs
```diff
@@ -241,178 +256,188 @@ scriptConfigParser = id
...
+            ind <- state $ randomR (0, 2 :: Int)
             let operation = "+-*" !! ind
-            return (a, b, operation)
+            pure (a, b, operation)
       theCode = "(" ++ [op] ++ " " ++ show operandA ++ " " ++ show operandB ++ ")"
-  cid <- view genChainId
+
+  -- Choose a Chain to send this transaction to, and cycle the state.
+  cid <- uses gsChains NES.head
+  gsChains %= rotate
```
fosskers: The meat of this PR. Most other changes are to support this change, as well as fortify logging.

emmanueldenloye: Looking good!

examples/TransactionGenerator.hs
```diff
@@ -241,178 +256,188 @@ scriptConfigParser = id
...
       theData = object ["test-admin-keyset" .= fmap formatB16PubKey kps]
-  liftIO $ mkExec theCode theData publicmeta kps Nothing
+  cmd <- liftIO $ mkExec theCode theData publicmeta kps Nothing
+  pure (cid, cmd)
+
+-- | O(1). The head value is moved to the end.
+rotate :: NESeq a -> NESeq a
+rotate (h :<|| rest) = rest :||> h
 
+-- TODO What is this number?
```
emmanueldenloye: We might want to take this out so as to not confused anyone else who modifies this code in the future. The contracts we want to test are well established at this point.

exe/CwTool.hs
emmanueldenloye: Cool.

PR #195, opened 2019-05-11
Create per-day Elasticsearch indices

larskuhtz:
Implement the convention that Elasticsearch index names are suffixed with the date and a new index is created each day.

The implementation makes a `HEAD` call for the index before submitting a new `_bulk` POST requests to an Elasticsearch index. If the `HEAD` call fails a new index is created.

### DIFF
--- a/src/Utils/Logging.hs
+++ b/src/Utils/Logging.hs
@@ -92,6 +92,7 @@ import Configuration.Utils hiding (Error)
 import Control.Concurrent.Async
 import Control.Concurrent.Chan
 import Control.Concurrent.STM.TBQueue
+import Control.Concurrent.STM.TVar
 import Control.Lens hiding ((.=))
 import Control.Monad
 import Control.Monad.Catch
@@ -117,6 +118,8 @@ import qualified Network.Wai.Handler.Warp as W
 import Network.Wai.Middleware.Cors
 import Network.Wai.UrlMap
 
+import Numeric.Natural
+
 import System.IO
 import qualified System.Logger as L
 import System.Logger.Backend.ColorOption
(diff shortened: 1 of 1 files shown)


src/Utils/Logging.hs
```diff
@@ -472,7 +474,8 @@ withElasticsearchBackend
     -> (Backend a -> IO b)
     -> IO b
 withElasticsearchBackend mgr esServer ixName inner = do
-    void $ HTTP.httpLbs putIndex mgr
+    i <- curIxName
+    createIndex i
         -- FIXME Do we need failure handling? If this fails the exeption
         -- is propagated to the main applcation
```
fosskers: Can this `FIXME` be removed? I see that `createIndex` below does have some error handling.

src/Utils/Logging.hs
```diff
@@ -481,37 +484,50 @@ withElasticsearchBackend mgr esServer ixName inner = do
...
         void $ HTTP.httpLbs (putBulgLog msg) mgr
       where
-        go 0 !b !c = return (c, b)
-        go i !b !c = tryReadTBQueue queue >>= \case
+        go _ 0 !b !c = return (c, b)
+        go cix i !b !c = tryReadTBQueue queue >>= \case
             Nothing -> return (c, b)
-            Just x -> go i (b <> indexAction x) (pred c)
+            Just x -> do
+                go cix i (b <> indexAction cix x) (pred c)
```
fosskers: Is it not still an issue that `i` isn't being `pred`'d here? What stops the recursion?

larskuhtz: I think, the code is just wrong, and always was. Not just that `i` isn't decremented, but `c` is, which should be incremented.

Currently the recursion terminates only when the queue is empty, which works but defeats the purpose of the bound. However, just decrementing the bound wouldn't make sense neither, because we would start creating larger batches only after saturating the network link or es.

src/Utils/Logging.hs
larskuhtz: The error handling in in `createIndex` doesn't address this FIXME. I think we can remove it anyways.

PR #197, opened 2019-05-15
Minor Reorg

fosskers:
This PR moves `cwtool` to be with his friends in the `tools/` directory.


larskuhtz:
Do we want to remove the examples? An example doesn't have to be used in order to be useful. The purpose of examples is to show in an easy to understand way how a particular sub-system can be used. At least the P2P example may still be useful, I think.

fosskers:
I can unremove them if you think they're useful to keep around.

larskuhtz:
Generally, conciseness isn't a value on it self when it comes to code and code bases. Readability and maintainability are more important.

larskuhtz:
I think, the `TrivialSync` example can be removed:

```
-- | Adds some jitter to the P2P network and makes the example more interesting.
--
-- This demonstrates the inefficiency of the prototype implementation. After a
-- new connection is established all blockheaders are exchanged, which causes an
-- ever increasing overhead. Combined with a small value for msgBufferSize this
-- can quickly lead to a live-lock.
```

:-)

larskuhtz:
I think the p2p example is still useful, for writing code that tests the P2P network in isolation. Not sure about `examples/BlockHeaderChainDb.hs`. I guess it's not very useful, but not sure.

PR #200, opened 2019-05-16
TXG: Transaction Batches

fosskers:
This PR allows one to send transactions in batches, the size of which can be defined on the command line.

### DIFF
--- a/tools/txg/TXG.hs
+++ b/tools/txg/TXG.hs
@@ -34,6 +34,7 @@ import Control.Monad.Except
 import Control.Monad.Reader hiding (local)
 import Control.Monad.State.Strict
 
+import Data.Generics.Product.Fields (field)
 import qualified Data.HashSet as HS
 import qualified Data.List.NonEmpty as NEL
 import Data.LogMessage
@@ -90,8 +91,8 @@ import qualified Utils.Logging.Config as U
 
 generateDelay :: MonadIO m => TXG m Int
 generateDelay = do
-  distribution <- asks _confTimingDist
-  gen <- gets _gsGen
+  distribution <- asks confTimingDist
+  gen <- gets gsGen
   case distribution of
     Just (Gaussian gmean gvar) -> liftIO (truncate <$> normal gmean gvar gen)
     Just (Uniform ulow uhigh) -> liftIO (truncate <$> uniformR (ulow, uhigh) gen)
@@ -102,11 +103,11 @@ generateSimpleTransactions
   => TXG m (ChainId, [Command Text])
 generateSimpleTransactions = do
   -- Choose a Chain to send these transactions to, and cycle the state.
-  cid <- uses gsChains NES.head
-  gsChains %= rotate
+  cid <- NES.head <$> gets gsChains
+  field @"gsChains" %= rotate
   -- Generate a batch of transactions
   stdgen <- liftIO newStdGen
-  batch <- asks _confBatchSize
+  batch <- asks confBatchSize
   cmds <- liftIO . sequenceA . replicate (fromIntegral batch) $ f cid stdgen
   -- Delay, so as not to hammer the network.
   delay <- generateDelay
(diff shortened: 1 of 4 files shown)


tools/txg/TXG.hs
```diff
@@ -100,197 +100,168 @@ generateDelay = do
...
 rotate (h :<|| rest) = rest :||> h
 
 generateTransaction
-  :: (MonadIO m, MonadLog SomeLogMessage m)
-  => TXG m (ChainId, Command Text)
+  :: forall m. (MonadIO m, MonadLog SomeLogMessage m)
+  => TXG m (ChainId, [Command Text])
 generateTransaction = do
-  contractIndex <- liftIO $ randomRIO @Int (0, 2)
+  contractIndex <- liftIO $ randomRIO @Int (0, 0)
```
fosskers: I should probably change this back.

tools/txg/TXG.hs
```diff
@@ -100,197 +100,168 @@ generateDelay = do
...
+    case pollresponse of
+      Left e -> logg Error $ toLogMessage (sshow e :: Text)
+      Right _ -> pure ()
     let accounts = buildGenAccountsKeysets Sim.accountNames paymentKS coinKS
     pure (cid, accounts)
 
+  logg Info $ toLogMessage ("Real Transactions: Transactions are being generated" :: Text)
+
   -- Set up values for running the effect stack.
   gen <- liftIO createSystemRandom
```
larskuhtz: What about adding a config option for a seed create all generators from that, so that tests can be replayed?

tools/txg/TXG/Types.hs
```diff
@@ -83,118 +75,134 @@ data TimingDistribution
...
-
-defaultScriptConfig :: ScriptConfig
-defaultScriptConfig = ScriptConfig
-  { _scriptCommand   = RunSimpleExpressions def def
+  , _logHandleConfig :: !U.HandleConfig
+  , _batchSize :: !BatchSize
+  } deriving (Show, Generic)
+
+scriptCommand :: Functor f => (TXCmd -> f TXCmd) -> Args -> f Args
+scriptCommand f sc = (\tc -> sc { _scriptCommand = tc }) <$> f (_scriptCommand sc)
```
larskuhtz: Why not use TH to generate Lenses?

tools/txg/TXG/Types.hs
```diff
@@ -83,118 +75,134 @@ data TimingDistribution
...
   , _nodeVersion     = v
-  , _logHandleConfig = U.StdOut }
+  , _logHandleConfig = U.StdOut
+  , _batchSize       = BatchSize 1 }
   where
     v :: ChainwebVersion
     v = fromJuste $ chainwebVersionFromText "timedCPM-peterson"
 
-scriptConfigParser :: MParser ScriptConfig
+scriptConfigParser :: MParser Args
```
larskuhtz: In the context of `pLeftSemigroupalUpdate` it may be more convenient to define something along the lines of

```haskell
<*< hostAddresses %:: pListSemigroupalUpdate % textOption
    % long "host"
    <> short 'h'
    <> metavar "HOSTNAME:PORT"
    <> help "hostname and port of a host the receives transactions. This option can be used multiple times to send  transactions to more than one host"
```

i.e. bundle the hostname and port in a single option, which might be more intuitive for options that can be provided multiple times.

tools/txg/TXG.hs
```diff
@@ -100,197 +100,168 @@ generateDelay = do
...
   count <- liftIO $ readTVarIO countTV
-
-  liftIO . atomically $ do
-    -- writeTQueue tq $ "Sent transaction with request keys: " <> sshow requestKeys
-    -- when mtime $ writeTQueue tq $ "Sending it took: " <> sshow timeTaken
-    writeTQueue tq $ "Transaction count: " <> sshow count
+  liftIO . atomically . writeTQueue tq $ "Transaction count: " <> sshow count
 
   case requestKeys of
     Left servantError -> lift . logg Error $ toLogMessage (sshow servantError :: Text)
```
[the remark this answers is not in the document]
fosskers: Probably not.

tools/txg/TXG.hs
fosskers: Reasonable.

tools/txg/TXG/Types.hs
fosskers: That previously was the case. TH caused an annoyance during this round of changes, so I removed it as an experiment.

tools/txg/TXG/Types.hs
fosskers: That's currently already the case, is it not?

tools/txg/TXG.hs
fosskers: I will leave this as "future work".

PR #201, opened 2019-05-17
disable mempool subscriptions

larskuhtz:


fosskers:
A hopefully temporary measure, unless the original pubsub model can't be improved upon.

fosskers:
There seems to be a problem with the mempool tests.

PR #203, opened 2019-05-17
Remove socket mempool

slpopejoy:


slpopejoy:
Note Socket code is preserved on https://github.com/kadena-io/chainweb-node/tree/archive/socket-mempool

PR #206, opened 2019-05-22
Turn on -Werror on nix builds in CI

mightybyte:


slpopejoy:
This would seem to be strictly unnecessary after #207 so I propose we close.

slpopejoy:
Closing, reopen/merge if desired

PR #207, opened 2019-05-22
Add -Werror

slpopejoy:


chainweb.cabal
```diff
@@ -61,6 +61,7 @@ common debugging-flags
 common warning-flags
     ghc-options:
         -Wall
+        -Werror
```
mightybyte: Should we also add this to the other build sections (test suite, cw-tool, etc)?

fosskers: Already is! That's what `common` is for.

fosskers:
Is this to complement or supplant #206 ?

mightybyte:
@fosskers Complement

chainweb.cabal
mightybyte: Oh, right.  My bad.

PR #208, opened 2019-05-23
Pact API upgrade

### LINKED ISSUE #139: Integrate Pact API changes
Upgrade pact to incorporate #451 and related and basically get it compiling:
- Mempool/Pact RestAPI stuff 
- Pact return values in PactService -> RPC -> Mempool

slpopejoy:
Fixes #139


src/Chainweb/Pact/RestAPI.hs
```diff
@@ -23,14 +20,7 @@ import Chainweb.Version
...
 
-type ApiV1API = SendApi :<|> PollApi :<|> ListenApi :<|> LocalApi
-
-type SendApi = "send" :> ReqBody '[JSON] SubmitBatch :> Post '[JSON] RequestKeys
-type PollApi = "poll" :> ReqBody '[JSON] Poll :> Post '[JSON] PollResponses
-type ListenApi = "listen" :> ReqBody '[JSON] ListenerRequest :> Post '[JSON] ApiResult
-type LocalApi = "local" :> ReqBody '[JSON] (Command Text) :> Post '[JSON] (CommandSuccess Value)
-
-type PactApi_ = "pact" :> ApiV1API
+type PactApi_ = "pact" :> API.ApiV1API -- TODO unify with Pact versioning
```
slpopejoy: We should really unify these. Pact api is `api/v1/send` etc. Chainweb is `[chainweb stuff]/pact/send`. Pact docs refer to the former. Do we fully conform, ie `[chainweb stuff]/pact/api/v1/send`? Or do we update the pact docs to start with `v1/send` and say the prefix can vary? Or, I guess we could modify `pact -s` to be `pact/v1/send` and then chainweb is strictly a prefix.

src/Chainweb/Pact/RestAPI/Orphans.hs
```diff
@@ -3,34 +3,31 @@
 
 module Chainweb.Pact.RestAPI.Orphans () where
```
slpopejoy: Should we move these to Pact? Are they meaningful there?

src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -154,25 +156,26 @@ listenHandler
...
-    nullResponse = ApiResult (object []) Nothing Nothing
+    handleResult (Right r) = return r
+    handleResult _ = throwError $ err400 { errBody = "Timeout" }
 
     runListen timedOut = go Nothing
       where
         go !prevCut = do
             m <- waitForNewCut prevCut
             case m of
-                Nothing -> return nullResponse      -- timeout
```
slpopejoy: @gregorycollins I think I have the semantics right here, please confirm

tools/txg/TXG.hs
```diff
@@ -422,7 +422,7 @@ poll version chainid = go
   where
     _ :<|> go :<|> _ :<|> _ = api version chainid
 
-listen :: ChainwebVersion -> ChainId -> ListenerRequest -> ClientM ApiResult
+listen :: ChainwebVersion -> ChainId -> ListenerRequest -> ClientM (CommandResult H.Hash)
```
slpopejoy: Do we have any coverage of TXG? @emmanueldenloye @fosskers

src/Chainweb/Pact/RestAPI/Orphans.hs
marklnichols: I also have a couple of Aeson instances of GasPrice in Chainweb.Oprhans that I assume needs to be moved to Pact

tools/txg/TXG.hs
fosskers: Our tests don't test it, but perhaps they should be made to.

src/Chainweb/Pact/PactService.hs
```diff
@@ -196,39 +195,42 @@ serviceRequests memPoolAccess reqQ = do
             LocalMsg LocalReq{..} -> do
                 r <- try $ execLocal _localRequest
                 case r of
-                  Left e -> liftIO $ putMVar _localResultVar $ Left e
-                  Right r' -> liftIO $ putMVar _localResultVar r'
+                  Left (SomeException e) -> liftIO $ putMVar _localResultVar $ toPactInternalError e
```
marklnichols: I assume Left can only be of type SomeExecption ?

src/Chainweb/Pact/RestAPI/Server.hs
gregorycollins: Timeout is an expected condition here, it shouldn't return 400. Instead the CommandResult should be populated with a timeout code if possible. I guess it doesn't matter as long as the web client code does the right thing with respect to retry. If you do use an HTTP error code, properly it should be 500-series because 400-series responses are meant to indicate some problem with the request.

src/Chainweb/Pact/PactService.hs
```diff
@@ -449,11 +449,14 @@ applyPactCmd isGenesis (Env' dbEnv) cmdIn miner = do
...
-    (result, txLogs) <- liftIO $! if isGenesis
+    result <- liftIO $! if isGenesis
         then applyGenesisCmd logger dbEnv pd spv cmd
         else applyCmd logger dbEnv miner gasModel pd spv cmd
 
-    pure $! FullLogTxOutput (P._crResult result) txLogs
+    pure $! toHashCommandResult result
+
+toHashCommandResult :: P.CommandResult [P.TxLog A.Value] -> HashCommandResult
+toHashCommandResult = over (P.crLogs . _Just) (P.pactHash . encodeToByteString)
```
marklnichols: I doesn't seem that LamdaCase or TypeApplication are needed extension anymore

src/Chainweb/Pact/SPV.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE RecordWildCards #-}
```
marklnichols: This didn't need to be added

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -213,57 +217,56 @@ applyLocal
     -> PublicData
     -> SPVSupport
     -> Command (Payload PublicMeta ParsedCode)
-    -> IO (Either SomeException (CommandSuccess (Term Name)))
+    -> IO (CommandResult [TxLog Value])
 applyLocal logger dbEnv pd spv cmd@Command{..} = do
 
   -- cmd env with permissive gas model
   let pd' = set pdPublicMeta (publicMetaOf cmd) $ pd
```
marklnichols: can drop the $

src/Chainweb/Pact/Types.hs
```diff
@@ -157,19 +111,21 @@ emptyPayload = PayloadWithOutputs mempty miner coinbase h i o
...
 fromMinerData :: MonadThrow m => MinerData -> m MinerInfo
 fromMinerData = decodeStrictOrThrow . _minerData
 
-toCoinbaseOutput :: FullLogTxOutput -> CoinbaseOutput
-toCoinbaseOutput = CoinbaseOutput . encodeToByteString . toHashedLogTxOutput
+toCoinbaseOutput :: HashCommandResult -> CoinbaseOutput
+toCoinbaseOutput = CoinbaseOutput . encodeToByteString
 
-fromCoinbaseOutput :: MonadThrow m => CoinbaseOutput -> m  HashedLogTxOutput
+fromCoinbaseOutput :: MonadThrow m => CoinbaseOutput -> m (HashCommandResult)
```
marklnichols: Can drop the paren here

src/Chainweb/WebPactExecutionService.hs
```diff
@@ -29,13 +29,10 @@ import Chainweb.Payload
 import Chainweb.Transaction
 import Chainweb.Utils (codecDecode)
 
-import Data.Aeson (Value)
```
marklnichols: It looks like BangPatterns is no longer needed

test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -42,6 +41,8 @@ import Chainweb.Test.Pact.Utils
 import Chainweb.Test.Utils
 import Chainweb.Version (ChainwebVersion(..))
 
+import Pact.Types.Command
+
```
marklnichols: It looks like LamdaCase and TupleSections are no longer needed

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -125,7 +125,7 @@ responseGolden networkIO rksIO = golden "command-0-resp" $ do
     cwEnv <- _getClientEnv <$> networkIO
```
marklnichols: We don't need QuasiQuotes any more

src/Chainweb/Pact/RestAPI/Server.hs
slpopejoy: What should it return then? 500 "Timeout"? cloudflare does a "522 Timeout" https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#Cloudflare

larskuhtz: If the timeout is expected. What about `204 No Content` or `202 Accepted`?

src/Chainweb/Pact/PactService.hs
slpopejoy: This is an unusual one, where we don't think any exceptions will fire but "just in case". Also, the `SomeException` specializes `try` otherwise it wouldn't compile.

src/Chainweb/Pact/RestAPI.hs
slpopejoy: Based on discussion with @larskuhtz , going with fully specified url, since the "Pact API" is cross platform while chainweb might hang other services off of `pact/` (ie SPV queries etc). Thus `[chainweb stuff]/pact/api/v1/send`.

src/Chainweb/Pact/RestAPI/Server.hs
slpopejoy: I like `204`

slpopejoy: OTOH, maybe we should just do 200 with `{ "status": "timeout" }` body, just to avoid finicky header code docs. One thing I didn't do with this update is go for the "literate error responses", ie use `errorResponse` or whatever to make 402s etc have a parseable body with `{ "status": "failure", "code": "402", "message": ... }` as discussed, so that JS frontends can simply parse a body.

PR #210, opened 2019-05-23
[Bug] Fix SPV Test Nondeterminism

emilypi:
This should help with the non-determinism problem with SPV tests


src/Chainweb/Pact/SPV.hs
```diff
@@ -149,14 +149,3 @@ getTxIdx bdb pdb bh th = do
...
--- -------------------------------------------------------------------------- --
--- utilities
-
--- | Prepend "spvSupport" to any errors so we can differentiate messages
---
-spvError :: Text -> IO (Either Text a)
-spvError = return . Left . (<>) "spvSupport: "
-
-spvError' :: String -> IO (Either Text a)
-spvError' = spvError . pack
```
emilypi: The error strings already contain SPV data that looks like `SPV verify failed: <errstring>` coming from Pact. These functions are duplicate that work, so they're unnecessary.

test/Chainweb/Test/CutDB.hs
```diff
@@ -196,7 +200,34 @@ extendAwait cdb pact i p = race gen (awaitCut cdb p) >>= \case
...
+    => CutDb cas
+    -> WebPactExecutionService
+    -> Natural
+    -> Natural
+    -> (Cut -> Bool)
+    -> IO Cut
+extendAwaitRetryN _ _ _ 0 _ = throwM $
+    userError "Cut extension retry limit exceeded. This is a bug in Chainweb.Test.CutDb"
+extendAwaitRetryN cdb pact i n p = extendAwait cdb pact i p >>= \case
+    Nothing -> extendAwaitRetryN cdb pact 1 (pred n) p
```
emilypi: @larskuhtz in the case where we don't find any cuts that satisfies our predicate, this function just bumps the extension by 1 block until either we run out of retries (highly unlikely), or a block is found (very likely)

larskuhtz: Wouldn't this have the exact same effect as calling `extendAwait cab pact (i * n) p`?

src/Chainweb/Pact/SPV.hs
```diff
@@ -71,20 +71,20 @@ pactSPV cdbv l = SPVSupport $ \s o -> readMVar cdbv >>= go s o
...
+      x -> return . Left
         $ "TXIN or TXOUT must be specified to generate valid spv proofs: "
         <> x
 
     txOutputProofOf :: Object Name -> IO (Either Text (TransactionOutputProof SHA512t_256))
     txOutputProofOf o =
       case toPactValue $ TObject o def of
-        Left e -> spvError e
+        Left e -> return $ Left e
         Right t -> case fromJSON . toJSON $ t of
```
fosskers: This is `traverse`.

src/Chainweb/Pact/SPV.hs
```diff
@@ -119,7 +119,7 @@ getTxIdx bdb pdb bh th = do
     -- get BlockPayloadHash
     ph <- fmap (fmap _blockPayloadHash)
         $ entries bdb Nothing (Just 1) (Just $ int bh) Nothing S.head_ >>= \case
-            Nothing -> spvError "unable to find payload associated with transaction hash"
+            Nothing -> return $ Left "unable to find payload associated with transaction hash"
             Just x -> return $ Right x
```
fosskers: This is `pure . note "unable to find payload associated with transaction hash"`

src/Chainweb/Pact/SPV.hs
```diff
@@ -135,7 +135,7 @@ getTxIdx bdb pdb bh th = do
           & sindex (== th)
 
         case r of
-          Nothing -> spvError "unable to find transaction at the given block height"
+          Nothing -> return $ Left "unable to find transaction at the given block height"
           Just x -> return $ Right (int x)
```
fosskers: `pure . fmap int . note "..."`

src/Chainweb/Pact/SPV.hs
emilypi: well shoot. Don't tell anyone i missed that 🙏

src/Chainweb/Pact/SPV.hs
emilypi: do we have `errors` as a dependency?

src/Chainweb/Pact/SPV.hs
fosskers: I'm phoning Rob Norris as we speak.

src/Chainweb/Pact/SPV.hs
fosskers: Almost certainly.

larskuhtz: Why is `pure . note` better than `return $ Left`? 

The former might be cooler but the latter seems more readable, in particular for Haskell newcomers to me. And the conciseness argument doesn't apply either.

fosskers: `note` collapses the entire pattern match away.

larskuhtz: oh,  sorry, I missed that it saves the case distinction. Well, still not convinced that it's worth it, but I now see the benefit.

test/Chainweb/Test/CutDB.hs
emilypi: It's actually `i + n` for some `n` (note the 1 in `extendAwaitRetryN cdb pact 1 (pred n) p`!) . You said we'd want the minimum amount of blocks mined, so it makes sense extend the cut db by 1 block at a time. It seems (and I printed out the logs to double check), that we are only ever behind by 1 or 2 blocks at a time when we receive `Nothing`.

src/Chainweb/Pact/SPV.hs
fosskers: It's `note :: a -> Maybe b -> Either a b`, sibling to `hush :: Either a b -> Maybe b`, which I'd argue every Haskeller has written themselves in some `Utils` module :p 

I think they're handy to have around in general.

test/Chainweb/Test/CutDB.hs
larskuhtz: That would still be `n (n - 1) / 2`, thus quadratic, wouldn't it?

larskuhtz: Assuming no bugs, `extendAwait cdb pact n` mines up to `n` blocks, each on a randomly picked chain, until the predicate is satisfied. It seems to me that `extendAwaitRetryN cdb pact 1 (pred n)` should have the same effect as `extendAwait cdb pact (n + 1)`. If both have different semantics, it would mean that `extendAwait` is buggy.

src/Chainweb/Pact/SPV.hs
emilypi: added, thanks.

PR #212, opened 2019-05-24
Log Payload Sizes

fosskers:
An `Int` of the raw count of bytes.

### DIFF
--- a/src/Chainweb/Miner/POW.hs
+++ b/src/Chainweb/Miner/POW.hs
@@ -34,11 +34,11 @@ import Control.Monad.STM
 import Crypto.Hash.Algorithms
 import Crypto.Hash.IO
 
-import Data.Aeson (encode)
 import qualified Data.ByteArray as BA
 import Data.Bytes.Put
+import qualified Data.ByteString as BS
 import qualified Data.ByteString as B
-import qualified Data.ByteString.Lazy as BL
+import Data.Foldable (foldl')
 import qualified Data.HashMap.Strict as HM
 import Data.Int
 import Data.Proxy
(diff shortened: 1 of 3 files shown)


src/Chainweb/Miner/POW.hs
```diff
@@ -118,11 +120,13 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
                     Right !r -> return r
             go2 c
 
-        let !nmb = NewMinedBlock (ObjectEncoded newBh)
-                       . Seq.length
-                       $ _payloadWithOutputsTransactions payload
+        let bytes = BL.length $ encode payload
```
fosskers: Potentially naughty that I'm doing a "spurious" encoding here just to get the byte count. It's convenient because it lets me shove that data in with the `NewMinedBlock` logging data, which is already occurring. 

Should I be doing this elsewhere?

fosskers: Thoughts from a conversation with Doug: it's wasteful in the sense that it adds an extra serialization, but efficient in the sense that no extra logging is done - we just add the new data to the `NewMinedBlock` already being sent.

larskuhtz: I think, we should count the "actual" bytes instead of the JSON encoded bytes, since the latter is not well defined. Also JSON is just a transport encoding. Actually, I have an item in my backlog to add support for different transport encodings via HTTP `accept` headers. For `BlockHeader` we already support binary and JSON.

fosskers: I'm going to change this to count the size of `PayloadData` instead, since that is the actual data being sent across the network.

PR #213, opened 2019-05-25
Pact API improvements 

slpopejoy:
Ready once Pact PR goes in


src/Chainweb/Pact/Types.hs
```diff
@@ -140,7 +140,7 @@ data PactDbStatePersist = PactDbStatePersist
     , _pdbspPactDbState :: PactDbState
     }
 
-newtype GasSupply = GasSupply { _gasSupply :: Decimal }
+newtype GasSupply = GasSupply { _gasSupply :: GasPrice }
```
emilypi: `GasSupply` is our current working total. Why was `Decimal` replaced with `GasPrice`?

slpopejoy: Maybe not the best idea, it was in hopes of minimizing coercion. I'll look at this

slpopejoy: Ah the other reason is it needs a JSON representation to go into the message body, ie in `mkBuyGasCmd` and `GasPrice` has the right instances.

The larger issue is using dedicated newtypes for gas prices, when in reality these are simply money values which don't have a first class representation in the API. I propose leaving this for now and addressing in a Pact PR, ie a `Price` or `Money` API type.

slpopejoy: OK figured out a better solution, and one we should have done for `GasPrice`: newtype `ParsedDecimal` to grab its instances. Likewise, `GasLimit` should newtype `ParsedInteger`.

stack.yaml
```diff
@@ -40,6 +40,6 @@ extra-deps:
   - git: https://github.com/kadena-io/merkle-log.git
     commit: a7ae61d7082afe3aa1a0fd0546fc1351a2f7c376
   - git: https://github.com/kadena-io/pact.git
-    commit: 9456a750f67b62d48bed5ef2e60fbc7e9d2a4396
+    commit: b5601c601c3e497c938f6d804eb21f9a3fb30a2a
```
fosskers: This is the commit from the PR. The merged commit is: 0915aff9eed762b2ee58c648d900bc619e6fff96

cabal.project
```diff
@@ -9,7 +9,7 @@ package aeson
 source-repository-package
     type: git
     location: https://github.com/kadena-io/pact.git
-    tag: 9456a750f67b62d48bed5ef2e60fbc7e9d2a4396
+    tag: b5601c601c3e497c938f6d804eb21f9a3fb30a2a
```
fosskers: This is the commit from the PR. The merged commit is: 0915aff9eed762b2ee58c648d900bc619e6fff96

default.nix
```diff
@@ -200,8 +200,8 @@ in
         pact = dontCheck ( addBuildDepend (self.callCabal2nix "pact" (pkgs.fetchFromGitHub {
           owner = "kadena-io";
           repo = "pact";
-          rev = "9456a750f67b62d48bed5ef2e60fbc7e9d2a4396";
-          sha256 = "039fmgb5d4p524vagmkmmp3gajmvrly7r3iwy9mcdg6j7a6vqj9w";
+          rev = "b5601c601c3e497c938f6d804eb21f9a3fb30a2a";
```
fosskers: This is the commit from the PR. The merged commit is: 0915aff9eed762b2ee58c648d900bc619e6fff96

src/Chainweb/Pact/Types.hs
emilypi: Works for me

PR #215, opened 2019-05-28
Stack LTS 13

fosskers:
This PR:

- Bumps the LTS to 13 to match Pact
- Ensures future-compat to `servant-0.16`

Note that our source of truth for testing / releases is still a GHC 8.4-based `default.nix`, but this at least helps us catch issues with forward-compat before they hit us (it also detects warnings which 8.4 can't, which is more relevant now that we use `-Werror`).

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-12.26
+resolver: lts-13.26
 
 extra-deps:
   # --- Missing from Stackage --- #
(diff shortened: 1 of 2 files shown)


fosskers:
MacOS test being flakey.

stack.yaml
```diff
@@ -1,42 +1,36 @@
...
+  - aeson-1.4.3.0
+  - massiv-0.3.2.0
+  - servant-0.16.0.1
+  - servant-server-0.16
+  - servant-client-0.16
+  - servant-client-core-0.16
+
+  # --- Custom Pins --- #
   - git: https://github.com/kadena-io/merkle-log.git
     commit: a7ae61d7082afe3aa1a0fd0546fc1351a2f7c376
```
larskuhtz: Let's put this on Hackage. There are a few more features that we should implement eventually, but let's not wait for that.

fosskers:
Closing for now - this will be part of work I'll be revisiting in the next few weeks.

PR #220, opened 2019-05-29
Mark/mempool cons pr changes

marklnichols:
Closes #204


src/Chainweb/Mempool/Consensus.hs
```diff
@@ -25,67 +29,35 @@ import Control.Monad.Catch
...
-  where
-    f trans header = S.union trans <$> payloadLookup header
-
-    logg :: LogLevel -> T.Text -> IO ()
-    logg = logFun
+import Data.CAS
 
+----------------------------------------------------------------------------------------------------
+data MempoolConsensus t = MempoolConsensus
+    { mpcMempool :: MempoolBackend t
```
gregorycollins: Add bangs to `data` fields

src/Chainweb/Mempool/Consensus.hs
```diff
@@ -94,6 +66,95 @@ data ReintroducedTxs = ReintroducedTxs
...
+----------------------------------------------------------------------------------------------------
+processFork :: PayloadCas cas
+                 => BlockHeaderDb
+                 -> Maybe (PayloadDb cas)
+                 -> IORef (Maybe BlockHeader)
+                 -> LogFunction
+                 -> BlockHeader
+                 -> IO (Vector ChainwebTransaction)
+processFork blockHeaderDb payloadStore lastHeaderRef logFun newHeader = do
+    lastHeader <- readIORef lastHeaderRef
```
gregorycollins: Don't you need to write back to the ref after syncing?

src/Chainweb/Mempool/Consensus.hs
```diff
@@ -94,6 +66,95 @@ data ReintroducedTxs = ReintroducedTxs
...
+                n | n > 1  -> throwM $ MempoolConsensusException ("processFork -- height of new"
+                           ++ "block is more than one greater than the previous new block request")
+                  | otherwise -> do -- fork occurred, get the transactions to reintroduce
+                      oldTrans <- foldM f mempty oldBlocks
+                      newTrans <- foldM f mempty newBlocks
+
+                      -- before re-introducing the transactions from the losing fork (aka oldBlocks),
+                      -- filterout any transactions that have been included in the
+                      -- winning fork (aka newBlocks):
+                      let results = V.fromList $ S.toList $ oldTrans `S.difference` newTrans
```
gregorycollins: Force `results` with a bang pattern here

src/Chainweb/Mempool/InMem.hs
```diff
@@ -247,88 +202,43 @@ broadcasterThread broadcaster@(TxBroadcaster _ _ q doneMV) restore =
...
-    dataLock <- newData  >>= newMVar
-    tid <- forkIOWithUnmask (reaperThread cfg dataLock)
-    return $! InMemoryMempool cfg dataLock txB tid blockHeaderDb payloadDb
-  where
-    newData = InMemoryMempoolData <$> newIORef PSQ.empty
-                                  <*> newIORef HashMap.empty
-                                  <*> newIORef HashSet.empty
-                                  <*> return Nothing
+newInMemMempoolData :: IO (InMemoryMempoolData t)
+newInMemMempoolData = InMemoryMempoolData <$> newIORef PSQ.empty
```
gregorycollins: `<$>` -> `<$!>`

src/Chainweb/Pact/PactService.hs
```diff
@@ -273,12 +277,24 @@ liftCPErr = either internalError' return
...
+    -> PactServiceM PayloadWithOutputs
+execNewBlock mpAccess header miner = do
     let bHeight@(BlockHeight bh) = _blockHeight header
         bHash = _blockHash header
 
-    newTrans <- liftIO $! memPoolAccess bHeight bHash header
+    logInfo $ "execNewBlock, about to get call processFork: "
+           <> " (height = " <> sshow bHeight <> ")"
+           <> " (hash = " <> sshow bHash <> ")"
+    liftIO $ mpaProcessFork mpAccess header
```
gregorycollins: Why doesn't `getBlock` just call `processFork` internally? Not sure why you had to change the API here.

src/Chainweb/Pact/PactService.hs
```diff
@@ -349,16 +364,25 @@ logDebug = logg "DEBUG"
...
     let bHeight@(BlockHeight bh) = _blockHeight currHeader
-        bParent = _blockParent currHeader
         bHash = _blockHash currHeader
+        bParent = _blockParent currHeader
         isGenesisBlock = isGenesisBlockHeader currHeader
 
+    logInfo $ "execValidateBlock, about to get call setLastHeader: "
+        <> " (height = " <> sshow bHeight <> ")"
+        <> " (hash = " <> sshow bHash <> ")"
+    liftIO $ mpaSetLastHeader mpAccess currHeader
```
gregorycollins: You can do this inside `processFork` ?

src/Chainweb/Pact/PactService.hs
```diff
@@ -349,16 +364,25 @@ logDebug = logg "DEBUG"
...
+
+    miner <- decodeStrictOrThrow (_minerData $ _payloadDataMiner plData)
```
gregorycollins: Use `decodeStrictOrThrow'`

src/Chainweb/Pact/Service/PactInProcApi.hs
```diff
@@ -75,12 +84,53 @@ withPactService' ver cid logger memPoolAccess cdbv action = do
...
+   logFn = logFunction
+
+pactProcessFork
+    :: Logger logger
+    => MempoolConsensus ChainwebTransaction
+    -> logger
+    -> (BlockHeader -> IO ())
+pactProcessFork mpc theLogger bHeader = do
+    let forkFunc = (mpcProcessFork mpc) (logFunction theLogger)
     txHashes <- forkFunc bHeader
```
gregorycollins: Reintroduction could also happen directly inside `processFork`

src/Chainweb/Mempool/Consensus.hs
marklnichols: Ok, done

src/Chainweb/Mempool/Consensus.hs
marklnichols: No, because I'm updating that 'lastHeader' value at validate time.  Based on when this is called, the the 'newHeader' here is coming from a newBlock request.  

I considered updating the 'lastHeader' at both newBlock and validateBlock time, but that doesnt seem right, given that a "new Block" may likely not win and be validated.

src/Chainweb/Pact/PactService.hs
marklnichols: Mostly because I was not certain if they would always be called together (i.e., there might be a better place for processFork to be called)

src/Chainweb/Pact/PactService.hs
marklnichols: Not when processFork is being called only as part of the newBlock process.  This most importantly needs to happen at validate time

src/Chainweb/Pact/Service/PactInProcApi.hs
marklnichols: It could, I don't know if that is better or worse -- right now processFork is just calculating the list of transactions to reintroduce.

src/Chainweb/Mempool/Consensus.hs
gregorycollins: I think it should only happen at validate time? New blocks just get sent to front of consensus and are re-validated, right?

marklnichols: Yes, Its only at validate time.  And right, new blocks get validated in the same way blocks from other nodes are validated

PR #231, opened 2019-06-03
TXG: Single Transactions

fosskers:
This PR allows one to send a single TX and immediately wait for the result (via an internal `listen/` call).

Sample output, having sent `(+ 1 2)`:
```
ListenResponse 
    ( CommandResult 
        { _crReqKey = "O4pvMdZB9IzsvjrD4PfCsqZhOxa50ha98EFuONABvyo"
        , _crTxId = Just 120
        , _crResult = PactResult ( Right ( PLiteral ( LDecimal { _lDecimal = 3 } ) ) )
        , _crGas = Gas 1
        , _crLogs = Just "1sG131AkRe4pR0oMjpcvhYIfMP_VhYk5wTZ6sVoxhNI"
        , _crContinuation = Nothing
        , _crMetaData = Nothing
        } 
    )
```

### DIFF
--- a/tools/txg/TXG.hs
+++ b/tools/txg/TXG.hs
@@ -337,13 +337,16 @@ listenerRequestKey config host listenerRequest = do
 
 -- | Send a single transaction to the network, and immediately listen for its result.
 singleTransaction :: Args -> HostAddress -> SingleTX -> IO ()
-singleTransaction args host (SingleTX c cid) = do
-  cfg <- mkTXGConfig Nothing args host
-  kps <- testSomeKeyPairs
-  cmd <- mkExec (T.unpack c) (datum kps) (Sim.makeMeta cid) kps Nothing
-  runMaybeT (f cfg cmd) >>= \case
-    Nothing -> putStrLn "Single Transaction: failed" >> exitWith (ExitFailure 1)
-    Just res -> pPrintNoColor res
+singleTransaction args host (SingleTX c cid)
+  | not . HS.member cid . chainIds $ nodeVersion args =
+    putStrLn "Invalid target ChainId" >> exitWith (ExitFailure 1)
+  | otherwise = do
+      cfg <- mkTXGConfig Nothing args host
+      kps <- testSomeKeyPairs
+      cmd <- mkExec (T.unpack c) (datum kps) (Sim.makeMeta cid) kps Nothing
+      runMaybeT (f cfg cmd) >>= \case
+        Nothing -> putStrLn "Single Transaction: failed" >> exitWith (ExitFailure 1)
+        Just res -> pPrintNoColor res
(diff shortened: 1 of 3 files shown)


tools/txg/TXG.hs
```diff
@@ -332,6 +334,26 @@ listenerRequestKey config host listenerRequest = do
...
     cid = fromMaybe (unsafeChainId 0) . listToMaybe $ nodeChainIds config
 
+-- | Send a single transaction to the network, and immediately listen for its result.
+singleTransaction :: Args -> HostAddress -> SingleTX -> IO ()
+singleTransaction args host (SingleTX c cid) = do
+  cfg <- mkTXGConfig Nothing args host
+  kps <- testSomeKeyPairs
+  cmd <- mkExec (T.unpack c) (datum kps) (Sim.makeMeta cid) kps Nothing
+  runMaybeT (f cfg cmd) >>= \case
+    Nothing -> putStrLn "Single Transaction: failed" >> exitWith (ExitFailure 1)
```
emmanueldenloye: I think there should be some indication as to why the transaction failed. I could imagine someone easily forgetting a parentheses.

tools/txg/TXG/Types.hs
```diff
@@ -88,9 +89,14 @@ data TXCmd
   | RunSimpleExpressions TimingDistribution
   | PollRequestKeys ByteString
   | ListenerRequestKey ByteString
+  | SingleTransaction SingleTX
   deriving (Show, Eq, Generic)
   deriving anyclass (FromJSON, ToJSON)
 
+data SingleTX = SingleTX { _stxCmd :: Text, _stxChainId :: ChainId }
```
emmanueldenloye: Is there a check to make sure the chain id is a member of the peterson graph?

tools/txg/TXG.hs
fosskers: I'm writing a patch for Pact that will allow this to be cleaner. Notice the `ExceptT`s being hushed into `MaybeT` for now.

tools/txg/TXG/Types.hs
fosskers: Yes, let me do that.

tools/txg/TXG.hs
emmanueldenloye: Cool.

emmanueldenloye: You might want to open a new PR addressing this issue. I won't let this issue impede this PR from going through the tubes.

tools/txg/TXG/Types.hs
fosskers: As it turns out: at parse-time, no. It would need to be an extra manual check somewhere at runtime.

PR #232, opened 2019-06-04
Add missing files to cabal file and fix license field

larskuhtz:


chainweb.cabal
```diff
@@ -21,6 +21,17 @@ tested-with:
...
     CHANGELOG.md
     README.md
+    LICENSE
+
+    -- pact test data
+    pact/coin-contract/coin.repl
+    pact/coin-contract/coin.pact
+    pact/coin-contract/coin-sig.pact
+    pact/coin-contract/load-coin-contract.yaml
+    pact/coin-contract/spv-sig.pact
```
fosskers: You could also glob these too, as you did for the golden files.

larskuhtz: usage of `*` requires a common suffix. So it wouldn't help that much here. Also, I still prefer an explicit list, if it's not getting too long.

PR #233, opened 2019-06-04
Mark/reintro flag

marklnichols:


src/Chainweb/Chainweb.hs
```diff
@@ -164,13 +169,34 @@ instance FromJSON (TransactionIndexConfig -> TransactionIndexConfig) where
 pTransactionIndexConfig :: MParser TransactionIndexConfig
 pTransactionIndexConfig = pure id
 
+-- -------------------------------------------------------------------------- --
+-- ReintroTxsConfig
+
+data ReintroTxsConfig = ReintroTxsConfig
+    deriving (Show, Eq, Generic)
```
larskuhtz: Is there a chance that more fields are added to this constructor? Otherwise a simple boolean flag in `ChainwebConfiguration` would have been sufficient.

Introducing a dedicated type is more future proof / forward compatible if more settings are added.

src/Chainweb/Chainweb.hs
```diff
@@ -164,13 +169,34 @@ instance FromJSON (TransactionIndexConfig -> TransactionIndexConfig) where
...
+pReintroTxsConfig = pure id
+
 -- -------------------------------------------------------------------------- --
 -- Chainweb Configuration
 
 data ChainwebConfiguration = ChainwebConfiguration
     { _configChainwebVersion :: !ChainwebVersion
     , _configNodeId :: !NodeId
     , _configMiner :: !(EnableConfig MinerConfig)
+    , _configReintroTxs :: !(EnableConfig ReintroTxsConfig)
```
larskuhtz: I wonder if it would make sense to introduce a `MempoolConfig`data type to bundle all all Mempool related settings.

```Haskell
data MempoolConfig = MempoolConfig
    { _mempoolConfigP2p :: !MempoolP2pConfig
    , _mempoolConfigReintroTx :: !(EnableConfig ReintroTxsConfig) -- or, alternatively, just !Bool
    }
```

Maybe it's too early at this point, but if we keep adding mempool options we should do it.

src/Chainweb/Chainweb.hs
```diff
@@ -401,6 +432,7 @@ withChainwebInternal conf logger peer rocksDb inner = do
     v = _configChainwebVersion conf
     cids = chainIds v
     cwnid = _configNodeId conf
+    enableTxsReintro =  _enableConfigEnabled $ _configReintroTxs conf
```
larskuhtz: This could be moved into a where clause in the `go` function instead of passing it as an argument to `go`. However, I am not sure what is better performance wise. It depends on tail-recursion-optimization and common-subexpression-elimination play together in GHC. In the end, I may not make a difference at all.

marklnichols: I don't know if there are performance differences, but in general this situation or the "let vs where" question - I generally treat it like I was writing a proof -- is the main function more understandable if you define the sub-item first? or is it just noise that is better to fill in afterward?

src/Chainweb/Chainweb.hs
marklnichols: There already is an InMemConfig type (and at this point we are only using the InMem variant -- the other was removed) :

data InMemConfig t = InMemConfig {
    _inmemTxCfg :: {-# UNPACK #-} !(TransactionConfig t)
  , _inmemTxBlockSizeLimit :: {-# UNPACK #-} !Int64
  , _inmemReaperIntervalMicros :: {-# UNPACK #-} !Int
  , _inmemEnableReIntro :: !Bool
}

I didn't realize i didn't need the additional type for interfacing with the CLI.

PR #238, opened 2019-06-10
Servant 0.16

fosskers:
I found a relatively short set of deps that works with servant pinned to `>= 0.16`.


fosskers:
No longer necessary.

PR #240, opened 2019-06-12
Pact 3

fosskers:
This PR bumps our Pact dependency to Pact 3.0.1.

### DIFF
--- a/tools/txg/TXG.hs
+++ b/tools/txg/TXG.hs
@@ -30,15 +30,14 @@ import Configuration.Utils hiding (Error, Lens', (<.>))
 
 import Control.Concurrent.Async (mapConcurrently_)
 import Control.Concurrent.STM.TVar (modifyTVar')
-import Control.Error.Util (hushT, nothing)
 import Control.Lens hiding (op, (.=), (|>))
 import Control.Monad.Except
 import Control.Monad.Reader hiding (local)
 import Control.Monad.State.Strict
-import Control.Monad.Trans.Maybe (MaybeT(..))
 
 import Data.Generics.Product.Fields (field)
 import qualified Data.HashSet as HS
+import Data.List.NonEmpty (NonEmpty)
 import qualified Data.List.NonEmpty as NEL
 import Data.LogMessage
 import Data.Map (Map)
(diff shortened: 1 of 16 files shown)


tools/txg/TXG.hs
```diff
@@ -343,20 +344,18 @@ singleTransaction args host (SingleTX c cid)
...
-    -- TODO Make `ExceptT` once `RequestKeys` is a `NonEmpty`.
-    f :: TXGConfig -> Command Text -> MaybeT IO ListenResponse
-    f cfg@(TXGConfig _ _ ce v _) cmd =
-      hushT (ExceptT $ sendTransactions cfg cid [cmd]) >>= \case
-        RequestKeys [] -> nothing
-        RequestKeys (rk:_) -> hushT . ExceptT $ runClientM (listen v cid $ ListenerRequest rk) ce
+    f :: TXGConfig -> Command Text -> ExceptT ServantError IO ListenResponse
+    f cfg@(TXGConfig _ _ ce v _) cmd = do
+      RequestKeys (rk :| _) <- ExceptT . sendTransactions cfg cid $ pure cmd
+      ExceptT $ runClientM (listen v cid $ ListenerRequest rk) ce
```
fosskers: We got our wish, @emmanueldenloye  :)

fosskers:
Woops, gotta please Nix.

emmanueldenloye:
Let's wait until CI is done before you merge.

fosskers:
@emilypi looks like perhaps the Coin Contract needs to be updated?

emilypi:
@fosskers I pushed 2 changes: 

1. We were missing the `semialign` package to get us `Data.Align` for some tests. I've added it into the dependency list

2. The problem was not in the coin contract, but in the way we were generating `GasPrice`s. The new `GasPrice` parser is restrictive in that it now requires *non-negative Decimals*, and throws when you end up parsing a negative gasprice as part of the `FromJSON` parser - this is something we'll need to harden, as this is a bit of a surprise to me as well. A refinement type would be very much preferred instead of just throwing, but it's low priority since `GasPrice` is only ever set by us. Anyway, I've come up with the correct `Arbitrary GasPrice` instance and tests are good to go.

PR #242, opened 2019-06-13
[pact] Put pact tx bloom cache on disk.

gregorycollins:
- keep far fewer bloom filters in RAM
  - store blooms in rocksdb CAS as well as recent ones in ram, and fall back to CAS lookup if the in-mem lookup fails -- this should speed up startup

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -238,6 +238,7 @@ library
         , QuickCheck-GenT >=0.2
         , aeson >= 1.4.3
         , aeson-pretty >= 0.8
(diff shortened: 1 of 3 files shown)


larskuhtz:
It seems that this PR has a positive effect on memory consumption in testnet but a big negative effect on CPU usage:

image (1)

larskuhtz:
I propose that we put the PR on hold until we understand better it's effects on testnet performance.

fosskers:
Closing after talking with @gregorycollins .

PR #243, opened 2019-06-14
prevent pow miner from spinning

larskuhtz:


fosskers:
How should this change interact with #92 ?

larskuhtz:
I am in favor of getting #92 in. 

This PR is a really tiny patch that has shown a large effect. I think we want this behavior in any case. If it conflicts with #92 resolving it in a merge would be trivial.

src/Chainweb/Miner/POW.hs
```diff
@@ -134,6 +134,12 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
         --
         addCutHashes cutDb (cutToCutHashes Nothing c')
 
+        -- Wait for a new cut. We never mine twice on the same cut. If it stays
+        -- at the same cut for a longer time, we are most likely in catchup
+        -- mode.
+        --
+        void $ awaitNextCut cutDb c
```
fosskers: Is the check for `c' == c` in `awaitNextCut` sufficient for detecting what is new?

larskuhtz: It's supposed to interfere. Ideally, we are winning that block, so as soon as the newly minded block makes it through the (hopefully empty) cut queue, we want to start mining on it. 

The only thing we want to prevent is that we continue mining on the same cut again, while the new cut is stuck in the pipeline (or got rejected as too old) behind a cut with many missing dependencies in catchup-mode.

PR #245, opened 2019-06-14
Count Hash Attempts

fosskers:
For logging the overall hash power of the network.

### DIFF
--- a/src/Chainweb/BlockHeader.hs
+++ b/src/Chainweb/BlockHeader.hs
@@ -672,10 +672,10 @@ instance FromJSON (ObjectEncoded BlockHeader) where
     {-# INLINE parseJSON #-}
 
 data NewMinedBlock = NewMinedBlock
-    { _minedBlockHeader :: ObjectEncoded BlockHeader
-    , _minedBlockTrans :: Word
-    , _minedBlockSize :: Word   -- ^ Bytes
-    , _minedHashAttempts :: Word }
+    { _minedBlockHeader :: !(ObjectEncoded BlockHeader)
+    , _minedBlockTrans :: {-# UNPACK #-} !Word
+    , _minedBlockSize :: {-# UNPACK #-} !Word   -- ^ Bytes
+    , _minedHashAttempts :: {-# UNPACK #-} !Word }
     deriving (Eq, Show, Generic)
     deriving anyclass (ToJSON, NFData)
 
--- a/src/Chainweb/BlockHeader.hs
+++ b/src/Chainweb/BlockHeader.hs
@@ -673,8 +673,9 @@ instance FromJSON (ObjectEncoded BlockHeader) where
 
 data NewMinedBlock = NewMinedBlock
     { _minedBlockHeader :: ObjectEncoded BlockHeader
-    , _minedBlockTrans :: Int
-    , _minedBlockSize :: Int }  -- ^ Bytes
+    , _minedBlockTrans :: Word
+    , _minedBlockSize :: Word   -- ^ Bytes
+    , _minedHashAttempts :: Word }
     deriving (Eq, Show, Generic)
     deriving anyclass (ToJSON, NFData)
 
--- a/src/Chainweb/Miner/POW.hs
+++ b/src/Chainweb/Miner/POW.hs
@@ -46,7 +46,7 @@ import Data.Proxy
 import Data.Reflection (Given, give)
 import qualified Data.Sequence as Seq
 import qualified Data.Text as T
-import Data.Tuple.Strict (T2(..))
+import Data.Tuple.Strict (T2(..), T5(..))
 import Data.Word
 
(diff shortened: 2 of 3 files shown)


src/Chainweb/Miner/POW.hs
```diff
@@ -319,42 +325,46 @@ mine
...
+
+                        -- update the block creation time
+                        ct <- getCurrentTimeIntegral
+                        injectTime ct buf
+                        go 0 n
+
+                    go !i !n = do
+
+                        -- increment hash counter
+                        modifyIORef' counter (+ 1)
```
larskuhtz: I would move this out of this loop and increment in steps of 1000.

src/Chainweb/BlockHeader.hs
```diff
@@ -673,8 +673,9 @@ instance FromJSON (ObjectEncoded BlockHeader) where
 
 data NewMinedBlock = NewMinedBlock
     { _minedBlockHeader :: ObjectEncoded BlockHeader
-    , _minedBlockTrans :: Int
-    , _minedBlockSize :: Int }  -- ^ Bytes
+    , _minedBlockTrans :: Word
```
gregorycollins: add bang patterns and unpack pragmas

larskuhtz: All fields should be strict.

src/Chainweb/Miner/POW.hs
```diff
@@ -82,6 +83,10 @@ import Data.LogMessage (JsonLog(..), LogFunction)
 -- -------------------------------------------------------------------------- --
 -- Miner
 
+-- | The result of mining a new `Cut`.
+--
+data Ore = Ore !BlockHeader !PayloadWithOutputs !Cut !Adjustments !Word
```
larskuhtz: unpack fields?

src/Chainweb/Miner/POW.hs
```diff
@@ -114,7 +119,7 @@ powMiner logFun conf nid cutDb = runForever logFun "POW Miner" $ do
         -- Mine a new Cut
         --
         c <- _cut cutDb
-        T4 newBh payload c' adjustments' <- do
+        Ore newBh payload c' adjustments' hashAttempts <- do
```
larskuhtz: Could unboxed tuples be used here?

src/Chainweb/Miner/POW.hs
```diff
@@ -172,7 +178,7 @@ mineCut
     -> MWC.GenIO
     -> Cut
     -> Adjustments
-    -> IO (T4 BlockHeader PayloadWithOutputs Cut Adjustments)
+    -> IO Ore
```
larskuhtz: Use unboxed tuple?

src/Chainweb/Miner/POW.hs
fosskers: I don't think they can be, the contents aren't primitives.

src/Chainweb/Miner/POW.hs
fosskers: I moved it away from a tuple into a named data type to make things more explicit.

src/Chainweb/Miner/POW.hs
fosskers: I should add field names.

src/Chainweb/Miner/POW.hs
fosskers: Sure.

src/Chainweb/Miner/POW.hs
larskuhtz: I don't think that the content has to be primitive. The purpose of unboxed tuples is to allow functions to return more than one value without the overhead of creating a box for it.

```haskell
λ> :set -XUnboxedTuples
λ> let (# x, y #) = (# (1::Int), ("abc"::String) #)
λ> x
    1
λ> y
    "abc"
```

larskuhtz: The contents of the tuple are returned in registers (either as primitive values or pointers).

larskuhtz: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html?highlight=unboxed#extension-UnboxedTuples

src/Chainweb/Miner/POW.hs
larskuhtz: oops, sorry, I meant 100,000. Essentially, just moving up into the case `go 100000 !n`, avoiding any non-essential overhead in the inner loop.

src/Chainweb/Miner/POW.hs
larskuhtz: I'am not sure if `Ore` is more meaningful than `(# ... #)`. But, I think, the performance overhead doesn't matter here because `mineCut` is only called once per cut, which would be at most a few times per second.

PR #247, opened 2019-06-17
SPV-based Defpact Support

emilypi:
This finishes our work making `defpact`s SPV-aware, generalizing our previous work with SPV functionality in Pact. The new mechanism now supports cross-chain continuations (defpacts) as a *default*. 

We've provided an overload for the yield mechanism in `defpact`s to allow for the user stating a desired `chain-id` in each `yield`  on which they want to `resume`. The user then gives information in each `ContMsg` in the form of a `ContProof` object (bytestring-ified SPV proof) of outputs from the previous step. 

Then, with the latest PR in Pact, this proof is threaded through Haskell functionality similar to `verify-spv`, which decodes the proof object and enforces that the outputs of the previous yield match the proof object's record of what was yielded. In this way, SPV is used as a trustless witness to the proof that a yield has the correct data upon which to resume, which is exactly the shape of a cross-chain `defpacts`.

TODO: 

- [x] make cross-chain transfers a 2-step burn/create `defpact`
- [x] integrate the `verifyContinuation` functionality into SPV.hs in the pact service
- [x] add tests 

Note: `verify-spv` tests have been removed as we're thinking the previous `verify-spv`-driven work is probably obsolete with this spv defpact work.


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -187,25 +189,39 @@ applyCoinbase
...
-        coinbaseReq = RequestKey $ H.toUntypedHash (H.hash "COINBASE" :: H.PactHash)
+    let cenv = CommandEnv Nothing Transactional dbEnv logger freeGasEnv pd noSPVSupport
+        initState = initCapabilities [magic_COINBASE]
 
-    resultE <- catchesPactError $ coinbase cmdEnv minerInfo reward coinbaseReq
+    -- build the pure Exec so we can build the cmd
+    let cexec = Exec $! mkCoinbaseExec mid mks reward
 
-    case resultE of
+    ccmd <- mkCommand [] (_pdPublicMeta pd) (sshow ph) cexec
```
slpopejoy: 👍

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -187,25 +189,39 @@ applyCoinbase
...
 
-      Left e ->
-        jsonErrorResult cmdEnv coinbaseReq e [] (Gas 0)
-          "genesis tx failure for request key while running genesis"
+    cexec' <- case verifyCommand @PublicMeta ccmd of
```
slpopejoy: You should be using `applyExec'` like on #L375 instead of reproducing the `ProcSucc` stuff

emilypi: Amended

pact/coin-contract/coin.pact
```diff
@@ -7,19 +7,22 @@
...
     balance:decimal
-    guard:guard
-    )
+    guard:guard)
   (deftable coin-table:{coin-schema})
 
-  (defschema creates-schema
-    exists:bool
+  ; the shape of a cross-chain transfer (used for typechecking)
+  (defschema transfer-*
```
slpopejoy: kind of a weird/unidiomatic name, I would probably call it `transfer-schema` ...

pact/coin-contract/coin.pact
```diff
@@ -176,64 +171,51 @@
...
+         \source chain by the burn account. Note: must be called on the correct     \
+         \chain id as specified in the proof."
+
+    @model [ (property (> quantity 0.0))
+           , (property (not (= create-chain-id "")))
+           ]
+
+    (step
+      (with-capability (TRANSFER)
+        (debit delete-account quantity)
```
slpopejoy: I think we should consider that calling `cross-chain-transfer` with the current chain ID is a bug and should be enforced as such

src/Chainweb/Pact/SPV.hs
```diff
@@ -81,53 +89,78 @@ pactSPV cdbv l = SPVSupport (verifySPV cdbv l) (verifyCont cdbv l)
...
-        Just (CommandResult _ _ (PactResult (Right pv)) _ _ _ _) -> case fromPactValue pv of
-          (TObject !o _) -> return $! Right o
-          !o -> do
-            logLog l "ERROR" $ show o
-            return . Left $ pack "type error in associated pact transaction, should be object"
-        Just o -> do
-          logLog l "ERROR" $ show o
-          return . Left $ pack "Invalid command result in associated pact output"
+      t -> return . Left $!
+        "TXIN or TXOUT must be specified to generate valid spv proofs: " <> t
```
slpopejoy: we should probably ditch "TXIN" and just say `"unsupported SPV type: " <> t`

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -71,13 +73,14 @@ import Pact.Types.Term (DefName(..), ModuleName(..))
...
 
+import Chainweb.BlockHash
 import Chainweb.Pact.Service.Types (internalError)
-import Chainweb.Pact.Types (GasSupply(..), MinerInfo(..))
+import Chainweb.Pact.Types (GasSupply(..), MinerInfo(..), GasId(..))
 import Chainweb.Transaction (gasLimitOf, gasPriceOf)
+import Chainweb.Utils (sshow)
 
-------------------------------------------------------------------------------
+-- ------------------------------------------------------------------------ --
```
slpopejoy: weird spaces?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -94,66 +97,62 @@ applyCmd
...
-                    !finalResult = over (crLogs . _Just) (<> redeemLogs) cmdResult
-
-                logDebugRequestKey logger requestKey "successful gas redemption for request key"
-                pure finalResult
+    withGas l buyGasEnv cmd minerInfo supply userGasEnv $ \cenv cmd0 logs ->
+      runPayload cenv def cmd0 logs
 
+-- | A handle for wrapping 'runPayload' commands in a buy/redeem gas phase.
+--
+withGas
```
slpopejoy: This is fine but it didn't buy us much in terms of clarity ... you just moved 95% of the really confusing code from `applyCmd` to `withGas`

pact/coin-contract/coin.pact
emilypi: Changed

pact/coin-contract/coin.pact
emilypi: I've added the `enforce` and updated the docs

src/Chainweb/Pact/SPV.hs
emilypi: Done

src/Chainweb/Pact/TransactionExec.hs
emilypi: Discussed - this is a convention Lars set, and I was conforming to. I'm going to go ahead and remove.

src/Chainweb/Pact/TransactionExec.hs
emilypi: Discussed - I'll refactor `applyCmd` to be clearer about control flow instead of moving to `withGas`

PR #249, opened 2019-06-18
Experiment: Relaxed Servant Bounds

fosskers:
This PR and its CI output are a demonstration for the Servant developers that relaxing Servant's tight dependency bounds  has no adverse affect on Servant nor two large downstream projects, Pact and Chainweb. The dependency set represented by this project's pinned `nixpkgs` is *somewhere* around Stackage `lts-12.14` (GHC 8.4.3). Likewise, `stack.yaml` is pinned to `lts-12.26` and only requires new pins for the `servant-*` family.

This supersedes #238 , which also bumped Servant, but had to up-pin a number of other dependencies along with it.


default.nix
```diff
@@ -43,6 +43,42 @@ in
...
+              passthru = orig.passthru or {} // {
+                # When using callCabal2nix or callHackage, it is often useful
+                # to debug a failure by inspecting the Nix expression
+                # generated by cabal2nix. This can be accessed via this
+                # cabal2nixDeriver field.
+                cabal2nixDeriver = src;
+              };
+            });
+
+          callCabal2nixWithOptions = name: src: extraCabal2nixOptions: args:
```
fosskers: This function exists in `nixpkgs` since August of last year, but it seems to be missing from our particular pin. Without this, we cannot pin to GIthub-based dependencies that exist as an "uber repo" (i.e. no `.cabal` at the top level).

default.nix
```diff
@@ -43,6 +43,42 @@ in
                  url = "http://hackage.haskell.org/package/${pkgver}/${pkgver}.tar.gz";
                  inherit sha256;
                }) {};
+
+          # Adds a nix file as an input to the haskell derivation it
+          # produces. This is useful for callHackage / callCabal2nix to
+          # prevent the generated default.nix from being garbage collected
+          # (requiring it to be frequently rebuilt), which can be an
+          # annoyance.
+          callPackageKeepDeriver = src: args:
```
fosskers: Necessary for `callCabal2nixWithOptions` to work.

fosskers:
Closing for now.

PR #250, opened 2019-06-18
fix pruning of block header db

larskuhtz:
The existing implementation of pruning could in rare cases result in a block header that was missing its payload. This could happen if all blocks in a prefix of a fork were marked as false positives but the respective payloads were not marked. (False positive blocks with an unmarked predecessor are detected and deleted.)

The PR fixes the issue by detecting and deleting any falsely marked block for which the payload hasn't been marked. This correct, because the payloads for all blocks on the root forks are marked.

This guarantees that for each block header the respective payload exists. It also lowers the false positive rate for block headers even further. 

Note, that we don't require that for each payload there is a respective block header. We therefor don't do any effort to detect falsely marked payloads.

### DIFF
--- a/src/Chainweb/BlockHeaderDB.hs
+++ b/src/Chainweb/BlockHeaderDB.hs
@@ -226,13 +226,6 @@ pruneForks
     :: Logger logger
     => logger
     -> BlockHeaderDb
-    -> (BlockHeader -> Bool -> IO ())
-        -- ^ Deletion call back. This hook is called /after/ the entry is
-        -- deleted from the database. It's main purpose is to delete any
-        -- resources that were related to the deleted header and that are not
-        -- needed any more. The Boolean argument indicates whether the payload
-        -- of the block is shared with any block header that isn't marked for
-        -- deletion.
     -> Natural
         -- ^ The depth at which the roots are collected for marking block
         -- headers that are kept. Any fork that was active that the given depth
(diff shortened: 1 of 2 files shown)


src/Chainweb/BlockHeaderDB.hs
```diff
@@ -324,7 +330,7 @@ pruneForks logger cdb callback limit = do
             -- Delete nodes which parent isn't in the filter or is in the
             -- falsePositiveSet
             parentIsMarked <- stToIO $ BF.elem p marked
-            if not (isGenesisBlockHeader h) && (not parentIsMarked || HS.member p falsePositiveSet)
+            if not (isGenesisBlockHeader h) && (not parentIsMarked || HS.member p falsePositiveSet || not isPayloadMarked)
```
fosskers: Laziness question: if the `HS.member` call is true, will the `IO` involved in `isPayloadMarked` be ran? I'm trying to figure out which is better to place first.

larskuhtz: Binding in `IO` is generally strict. Lazy `IO` is only used for "true" input/output functions. I would expected that `stToIO` evaluates its argument strictly.

larskuhtz: I guess it depends on whether `BF.elem` is strict or returns a thunk. I would expect the former, but not sure.

fosskers: In that case, it might be a tiny bit more performant to put the `isPayloadMarked` check before the `HS.member`.

larskuhtz: I tried to refactor but my attempts led to more complex control flow. All of these checks should be pretty cheap (no IO involved), so I don't think optimizing this is worth the effort.

PR #253, opened 2019-06-19
Delete peers with unknown certificates from the p2p peer database

larskuhtz:


src/Network/X509/SelfSigned.hs
```diff
@@ -655,7 +656,13 @@ unsafeMakeCredential (X509CertChainPem cert chain) (X509KeyPem keyBytes) =
...
 data TlsPolicy
     = TlsInsecure
-    | TlsSecure Bool (ServiceID -> IO (Maybe Fingerprint))
+    | TlsSecure
+        { _tlsPolicyUseSystemStore :: !Bool
+            -- ^ whether to use the system certificate store.
+        , _tlsPolicyCallback :: !(ServiceID -> IO (Maybe Fingerprint))
+            -- ^ a callback for looking up certificate fingerprints for known
+            -- service endpoints.
+        }
```
fosskers: Does this not throw a "partial fields" warning?

src/Network/X509/SelfSigned.hs
```diff
@@ -716,6 +723,21 @@ certificateCache query = ValidationCache queryCallback (\_ _ _ -> return ())
                 <> " expected fingerprint: " <> T.unpack (fingerprintToText f)
                 <> " but got fingerprint: " <> T.unpack (fingerprintToText fp)
 
+-- | Check whether a connection failed due to an certificate missmatch
+--
+isCertificateMissmatchException :: HttpException -> Bool
```
fosskers: Missmatch -> Mismatch

src/P2P/Node.hs
```diff
@@ -318,6 +327,28 @@ syncFromPeer node info = runClientM sync env >>= \case
...
+    -- peer is removed from the database. That is, we allow only connection to
+    -- peers that we know through explicitly being added to the db.
+    --
+    -- We explicitly don't update the certificate fingerprint. The only we to
+    -- introduce a new peer into the network is by propagating it to the peer
+    -- databased. This allows to implement reputation management, gray-, and
+    -- black listing.
+    --
+#if MIN_VERSION_servant(0,16,0)
+    isCertMissmatch (ConnectionError e) = case fromException e of
```
fosskers: Missmatch -> Mismatch

src/Network/X509/SelfSigned.hs
larskuhtz: There haven't been any warnings in the nix build

PR #255, opened 2019-06-20
Chaindb pruning

larskuhtz:
* [x] fix a potential issue by including all recent blocks as roots,
* [x] temporarily disable payload pruning until block payload hashes are truly unique, and
* [x] add configuration flag for enabling and disabling pruning.


src/Chainweb/Chainweb.hs
```diff
@@ -202,6 +203,7 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configIncludeOrigin = True
     , _configThrottleRate = 1000
     , _configMempoolP2p = defaultMempoolP2pConfig
+    , _configPruneChainDatabase = True
```
fosskers: `True` by default?

larskuhtz: yes, that's the current behavior. It's fast and it can save a lot of time in pact replay.

PR #256, opened 2019-06-20
TLS 1.5

fosskers:
This fixes `tls-1.5.0` as a lower bound. This version came out quite recently to much fanfare.


chainweb.cabal
```diff
@@ -316,7 +310,7 @@ library
         , text >= 1.2
         , thyme == 0.3.6.0
         , time >= 1.8
-        , tls >=1.4
+        , tls ^>=1.5
```
emilypi: we are embracing the `^>=` boy? _nice_

fosskers: I do that for all my Haskelling now when I'm specifically pinned to the most recent major version. If the next version comes out and it still works, then I widen back to `>= x && < y`.

larskuhtz: I am not sure if I like those. It means each time each time tls raises their bound we'll have to react. In my experience these tight upper bounds lead to large amounts of PRs and commits that do nothing but raising upper bounds.

With Haskell, when a new version causes a break, the build breaks in any case, either because of a failing bound check, or because of a compile error. Except that the former is guaranteed to happen, if it's spurious, while the latter only happen if the new version actually causes a breakage.

PR #260, opened 2019-06-22
Reworked Hash Power Formula

fosskers:
The number is now the per-second hash rate.


fosskers:
Don't merge this yet.

fosskers:
Defunct.

PR #266, opened 2019-06-28
Add separate P2P network id for the mempool

larskuhtz:


src/Chainweb/RestAPI/Utils.hs
```diff
@@ -190,11 +191,13 @@ instance
 newtype NetworkEndpoint = NetworkEndpoint NetworkIdT
 
 type ChainEndpoint (c :: ChainIdT) = 'NetworkEndpoint ('ChainNetworkT c)
+type MempoolEndpoint (c :: ChainIdT) = 'NetworkEndpoint ('MempoolNetworkT c)
 type CutEndpoint = 'NetworkEndpoint 'CutNetworkT
 
 type family NetworkEndpointApi (net :: NetworkIdT) (api :: Type) :: Type where
     NetworkEndpointApi 'CutNetworkT api = "cut" :> api
     NetworkEndpointApi ('ChainNetworkT c) api = "chain" :> ChainIdSymbol c :> api
+    NetworkEndpointApi ('MempoolNetworkT c) api = "chain" :> ChainIdSymbol c :> "mempool" :> api
```
fosskers: GIven that this line is more specialized/specific than the one above it, should the order of these two be flipped?

larskuhtz: I don't think it matters

PR #268, opened 2019-07-02
optimize limit cut function

larskuhtz:
When getting new cuts nodes requests cuts that are only at most 1000 blocks ahead of the current cut of the node. The reason for that is to make nodes catching up more iteratively by going in chunks of 1000 blocks.

On the server side a limited cut is computed by using `branchEntries` which traverses the block header dbs backward until it reaches blocks at the target height. This traversal is currently not optimized. It follows the parent pointer with point wise lookups instead of using a reverse iterator, which has complexity `O(n)`. This works well for most cases because most of the time the number of traversed blocks is relatively small, and, thus, `n` is constant. For `limitCut` the `n` is in the order of the block height of the chain for each 1000-block chunk, and the overall complexity of catching up becomes `O(n*n)`. Since the node queries cuts from each active peer (testnet uses a peer count of 10 in the default configuration), this quadratic overhead incurs on 10 nodes. Moreover, for very large cut heights the `GET cut` queries timeout and the catch-up node spins making calls to peers, effectively running a DOS attack against the network.

Optimizing the `brachEntries` call by using a reverse iterator wouldn't resolve the overall problem.

This PR makes the assumption that for block heights further down in the history the history is linear and there is only a single block at that height. A fast code path is added that makes a point lookup for blocks at the given height. If there is only a single block it is used. Otherwise the expensive traversal is used as fall back to determine which block is on the main fork.

This approach is further refined, but by starting a backward traversal from all blocks at a constant number of blocks above the target height. By the time the target height is reached the set of ancestors has narrowed to a single block with high probability.

### DIFF
--- a/src/Chainweb/Chainweb.hs
+++ b/src/Chainweb/Chainweb.hs
@@ -85,7 +85,7 @@ module Chainweb.Chainweb
 
 import Chainweb.Payload.PayloadStore.RocksDB
 
-import Configuration.Utils hiding (Lens', (<.>))
+import Configuration.Utils hiding (Lens', (<.>), Error)
 
 import Control.Concurrent.Async
 import Control.Concurrent.MVar (newEmptyMVar, putMVar)
@@ -103,6 +103,7 @@ import qualified Network.HTTP.Client as HTTP
 import qualified Network.HTTP.Client.Internal as HTTP
 import Network.Socket (Socket)
 import Network.Wai
+import Network.Wai.Handler.Warp
 import Network.Wai.Middleware.Throttle
 
 import Numeric.Natural
@@ -464,7 +465,10 @@ runChainweb cw = do
         payloadDbsToServe = itoList $ const (view chainwebPayloadDb cw) <$> _chainwebChains cw
         pactDbsToServe = _chainwebPactData cw
 
-        serverSettings = peerServerSettings (_peerResPeer $ _chainwebPeer cw)
+        serverSettings
+            = setOnException
+                (\r e -> when (defaultShouldDisplayException e) (logg Error $ loggServerError r e))
+            $ peerServerSettings (_peerResPeer $ _chainwebPeer cw)
         serve = serveChainwebSocketTls
             serverSettings
             (_peerCertificateChain $ _peerResPeer $ _chainwebPeer cw)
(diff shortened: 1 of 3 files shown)


src/Chainweb/Chainweb.hs
```diff
@@ -512,3 +516,6 @@ runChainweb cw = do
   where
     logg = logFunctionText $ _chainwebLogger cw
     mempoolP2pConfig = _configMempoolP2p $ _chainwebConfig cw
+
+    loggServerError (Just r) e = "HTTP server error: " <> sshow e <> ". Request: " <> sshow r
+    loggServerError Nothing e = "HTTP server error: " <> sshow e
```
fosskers: Is this for debugging or prod usage?

src/Chainweb/Cut.hs
```diff
@@ -207,18 +207,84 @@ cutAdjs c cid = HM.intersection
...
+    -- expected to be linear. Also, generally, there aren't many forks.
+    -- Therefore, we expect this fast code path to be successful in most cases
+    -- where the requested height is further down in the history. If the
+    -- requested height is recent we don't care because the overhead of the
+    -- backward traversal is low.
+    --
+    fastRoute1 cid x = do
+        !db <- give wdb $ getWebBlockHeaderDb cid
+        let l = min (_blockHeight x) (int ch)
+        !a <- S.toList_ & entries db Nothing (Just 2) (Just $ int l) (Just $ int l)
```
fosskers: Does this need to be banged? Won't the pattern match below force the evaluation?

src/Chainweb/Chainweb.hs
larskuhtz: It is for production usage. The default behavior of wrap is to print any exception from the application to `stderr`. This emits on error log.

src/Chainweb/Cut.hs
larskuhtz: yeah, I think, the bang is meaningless here. Also `S.toList_` is strict. If it wasn't, that would be an issue on it's own, because the `entries` stream must not be used outside the `entries` bracket.

larskuhtz: I wonder how easy it is to leak the stream by using a non-strict accumulator. I guess, since the monad of the stream is fixed to `IO`, any straight forward fold of the stream would be strict in the stream. Not sure, though, if it is also strict in the values in the stream, so that evaluation may still access the underlying rocksdb iterator. If we want to be on the safe side, we should add an `NFData` constraint to the `TreeDB` methods that return streams.

larskuhtz:
```suggestion
        a <- S.toList_ & entries db Nothing (Just 2) (Just $ int l) (Just $ int l)
```

PR #269, opened 2019-07-02
PactDB checkpointer + Consensus reconciliation

gregorycollins:
Implement rewind + fast-forward across branches during pactdb `restore`.


src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -70,24 +73,25 @@ type Db = MVar (BlockEnv SQLiteEnv)
...
+    log "INFO" "dumping blockhistory here"
+    txts <- callDb "dumping" $ \db -> do
+        qry_ db "SELECT * FROM BlockHistory;" [RInt,RBlob,RInt] >>= mapM go
+    log "INFO" (intercalate ":" txts)
+      -- log "Info" "dumping versionhistory\n"
+      -- qry_ db "SELECT * FROM VersionHistory;" [RInt,RInt,RInt] >>= print
+  where
+    go :: [SType] -> IO String
+    go [SInt a, SBlob blob, SInt b] = do
+        let getshit = either error id
```
gregorycollins: Note to self: get rid of dump code

src/Chainweb/Pact/InternalTypes.hs
```diff
@@ -0,0 +1,85 @@
...
+{-# LANGUAGE TemplateHaskell #-}
+
+-- |
+-- Module: Chainweb.Pact.InternalTypes
+-- Copyright: Copyright © 2019 Kadena LLC.
+-- License: See LICENSE file
+-- Maintainer: Mark Nichols <mark@kadena.io>
+-- Stability: experimental
+--
+-- Pact Types module for Chainweb
```
slpopejoy: What is the rational for breaking this out; if justified then update the module header accordingly here and record yourself as maintainer

src/Chainweb/Payload/PayloadStore.hs
```diff
@@ -18,170 +18,20 @@
...
-, payloadCacheOutputTrees
-, payloadCacheTransactionTrees
-
--- * Payload Database
-
-, PayloadDb(..)
-, PayloadCas
-, payloadCache
-, transactionDb
+  module Chainweb.Payload.PayloadStore.Types
```
slpopejoy: This module has a single function in it now. Why create a types file?

src/P2P/Node.hs
```diff
@@ -295,9 +299,14 @@ peerClientEnv node = peerInfoClientEnv (_p2pNodeManager node)
 --
 syncFromPeer :: P2pNode -> PeerInfo -> IO Bool
 syncFromPeer node info = runClientM sync env >>= \case
-    Left e -> do
-        logg node Warn $ "failed to sync peers from " <> showInfo info <> ": " <> sshow e
-        return False
+    Left e
```
slpopejoy: Why are these changes in this PR?

src/Chainweb/Pact/InternalTypes.hs
```diff
@@ -0,0 +1,85 @@
...
+import Chainweb.Pact.Backend.Types
+import Chainweb.Payload.PayloadStore.Types
+import Chainweb.Transaction
+
+data PactServiceEnv cas = PactServiceEnv
+  { _psMempoolAccess :: !(Maybe MemPoolAccess)
+  , _psCheckpointEnv :: !CheckpointEnv
+  , _psSpvSupport :: !SPVSupport
+  , _psPublicData :: !PublicData
+  , _psPdb :: PayloadDb cas
```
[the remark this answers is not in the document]
larskuhtz: In other places we expose it to be able to switch between persisted db backends for production and in memory for testing.

Since we switched to RocksDB, we only use rocks db in production and testing. But we were planning to re-add an in-memory implementation of the storage API for testing. It just not high up on the priority list.

chainweb.cabal
```diff
@@ -42,12 +42,6 @@ source-repository head
     type: git
     location: https://github.com/kadena-io/chainweb.git
 
-flag tls13
```
larskuhtz: Hasn't this been removed already in master when we moved to tls-1.5?

larskuhtz: It seems this diff contains more changes that are already in master. Maybe rebase?

src/Chainweb/BlockHeader.hs
```diff
@@ -23,6 +23,7 @@
 {-# LANGUAGE TypeInType #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
```
larskuhtz: Where is this used?

src/Chainweb/BlockHeaderDB.hs
```diff
@@ -255,7 +168,11 @@ pruneForks logger cdb limit callback = do
 
     !roots <- keys cdb Nothing Nothing
         (Just $ MinRank $ Min rootHeight)
-        (Just $ MaxRank $ Max rootHeight)
+        Nothing
+            -- include all block headers in the root set, so that all headers
+            -- are included in the marking phase. PayloadBlockHashes can be
+            -- shared between blocks at different height, so we must make sure
+            -- that they are retained if a root depends on them.
```
larskuhtz: How is this related to the PR? Again, isn't this on master already?

src/Chainweb/BlockHeaderDB.hs
```diff
@@ -86,37 +83,11 @@ import Chainweb.Version
...
-
 -- -------------------------------------------------------------------------- --
 -- Internal
 
 type E = BlockHeader
 
--- -------------------------------------------------------------------------- --
--- Ranked Block Header
-
-newtype RankedBlockHeader = RankedBlockHeader { _getRankedBlockHeader :: BlockHeader }
```
larskuhtz: The need to separate these types from some of their basic properties, like serializations and textual representations, may hint at a more profound design issue. In my experience this often indicates that something is wrong with the overall design and abstractions in a project. The issue may have existed before and may have only been exposed by this PR.

Type modules are usually needed when data types of large structures (like an AST) are themself are mutually recursive, which is not the case in the Chainweb code base. So, I'd like to better understand why this is needed.

src/P2P/Node.hs
gregorycollins: Bad merge probably

src/Chainweb/BlockHeaderDB.hs
gregorycollins: There's a huge recursion that I had to cut. The module that deals with genesis blocks is included in several of the BlockHeaderDB / CutDB / WebBlockHeaderDB/etc files, but it wants to validate blocks, which requires a pact service. Now that pact service needs to look up block headers and payloads for rewind+fast-forward on `restore`, it needs a `BlockHeaderDB`.

To cut the cyclical dependency I had to move everything dealing with validation out. I may be able to fold some of them back together but we'll see.

src/Chainweb/BlockHeaderDB.hs
gregorycollins: Emmanuel's branch had a bad merge and this branch is relative to that. (Stu wanted to compare my changes to Emmanuel's that were already reviewed). If you want to diff against master I can pull up a different PR for that.

src/Chainweb/BlockHeaderDB.hs
```diff
@@ -12,6 +12,7 @@
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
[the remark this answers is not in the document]
gregorycollins: `instance TreeDb BlockHeaderDb` is orphan otherwise

src/Chainweb/Payload/PayloadStore.hs
gregorycollins: The single function needs `Chainweb.BlockHeader.Genesis (genesisBlockPayload)`, and this is the thing that's causing our circular dependency. (`Chainweb.BlockHeader.Genesis` depends on pact but pact now depends on payload store)

src/Chainweb/Pact/PactService.hs
```diff
@@ -336,78 +370,135 @@ execLocal cmd = do
...
+    playGenesisBlock = restoreCheckpointer Nothing >>=
+                       playOneBlock mpAccess currHeader plData
+
+    playFork bhdb payloadDb (_, lastHash) = do
+        lastHeader <- liftIO $ lookupM bhdb lastHash
+        (!_, _, newBlocks) <-
+            liftIO $ collectForkBlocks bhdb lastHeader currHeader
+        -- newBlocks cannot be empty
+        when (V.null newBlocks) $ fail "impossible: empty newBlocks"
+        V.foldM (fastForward payloadDb) undefined newBlocks
```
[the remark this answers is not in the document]
gregorycollins: mooted

src/Chainweb/Pact/PactService.hs
```diff
@@ -336,78 +370,135 @@ execLocal cmd = do
...
+        playOneBlock mpAccess block payload pdbenv
+
+
+execTransactions
+    :: Maybe BlockHash
+    -> MinerInfo
+    -> Vector ChainwebTransaction
+    -> PactDbEnv'
+    -> PactServiceM cas Transactions
+execTransactions nonGenesisParentHash miner ctxs (PactDbEnv' pactdbenv) = do
```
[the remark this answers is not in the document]
gregorycollins: I think I fixed this now (reworked the merge on emmanuel's branch so that diff items from master shouldn't show up here)

src/Chainweb/BlockHeaderDB.hs
larskuhtz: Maybe, you can add a special case for the genesis block validation? That one shouldn't require lookup of block headers.

Genesis blocks are validated only once at startup by pact internally.

larskuhtz: We run into this issue before. At that time we found a simpler solution that just moved a few functions into a submodule instead of moving all types. Not sure if the same is possible here.

src/Chainweb/TreeDB.hs
```diff
@@ -716,6 +721,51 @@ branchDiff_ db = go
...
     step = lift . lookupParentM GenesisParentThrow db
 
+-- -------------------------------------------------------------------------- --
+-- | Walks the ancestry of the given tree from the leaf back to the genesis.
+streamAncestors
+    :: forall db . TreeDb db
+    => db
+    -> DbEntry db
+    -> S.Stream (Of (DbEntry db)) IO ()
+streamAncestors db = go
```
larskuhtz: isn't this a special case of `branchEntries`? 

```Haskell
f (streamAncestors db e) == branchEntries db Nothing Nothing Nothing  Nothing mempty (HS.singleton $ UpperBound $ key e) f
```

The advantage of the implementation of `streamAncestors` in the PR is that it returns a stream that is safe to use outside of a bracket.

The disadvantage is that it prevents more efficient implementations for instances that require the bracket for resource management (like, for instance, implementations via rocksdb iterators, which is on our TODO list).

If the motivation is to provide a stream without a CPS bracket, maybe explain it in the haddock comment.

test/Chainweb/Test/Utils.hs
```diff
@@ -192,6 +195,22 @@ testRocksDb l = rocksDbNamespace (const prefix)
...
+    create = do
+      sysdir <- getCanonicalTemporaryDirectory
+      dir <- createTempDirectory sysdir "chainweb-rocksdb-tmp"
+      rocks <- openRocksDb dir
+      return (dir, rocks)
+    destroy (dir, rocks) = do
+        closeRocksDb rocks
+        destroyRocksDb dir
+        removeDirectoryRecursive dir
+          `catch` \(_ :: SomeException) -> return ()
```
larskuhtz: Omit catching Asynchronous exceptions? Use `catchAllSynchronous` from `Chainweb.Utils`?

src/Chainweb/BlockHeaderDB.hs
larskuhtz: I think (hope), that it should be possible refactor 'BlockHeaderDb` such that it doesn't depend on pact, without moving that much code and introducing orphans.

If the issue is payload validation, we should provide pact with a version/view of `BlockHeaderDb` that doesn't depend on it. Non-consensus components should should anyways be given a read-only version of BlockHeaderDb. Such a readonly version doesn't need to depend on validation/pact.

src/Chainweb/Pact/InternalTypes.hs
gregorycollins: I moved everything that was here to a different location instead

test/Chainweb/Test/Utils.hs
gregorycollins: Done

src/Chainweb/TreeDB.hs
gregorycollins: I ended up not needing this function so I removed it

src/Chainweb/BlockHeader.hs
gregorycollins: I don't recall but something was complaining that it was missing or I wouldn't have added it

src/Chainweb/BlockHeaderDB.hs
gregorycollins: Yes, that would be a fix for this issue, and we could re-unify the modules. We can clean up later after this changeset is in, the inverse operation of the refactor I did is trivial

larskuhtz: @gregorycollins @fosskers Let's do this soon. I created an issue here #272.

PR #270, opened 2019-07-04
A Young Miner, Cast Adrift

fosskers:
This PR provides a standalone `chainweb-miner` binary to be communicated with via HTTP.

### Follow-up Work

- Have `chainweb-node` use this HTTP interface
- GPU Support (via `accelerate` ?)

## Full Details
From the module header:

```haskell
-- == Purpose and Expectations ==
--
-- This tool is a mid-level, push-based, private, multicore CPU miner for
-- Chainweb. By this we mean:
--
--   * mid-level: the miner is aware of what `BlockHeader`s are, but does not
--     necessarily know how to construct them. It has no concept of `Cut`s or
--     the cut network used by Chainweb - it simply attempts `Nonce`s until a
--     suitable hash is found that matches the current `HashTarget`.
--
--   * push-based: Work is submitted to the miner in the form of a candidate
--     `BlockHeader`, in which the `Nonce` has been completely guessed. The
--     miner then automatically handles cancelling previous mining attempts,
--     hashing, cycling the `Nonce`, and caching the result. Note that we
--     maintain no "long connections", so work submissions yield an HTTP
--     response immediately, and the results of successful mining must be polled
--     manually.
--
--   * private: Work can be submitted (and results polled) by anyone. It is
--     therefore expected that a miner's HTTP interface is not exposed to the
--     open internet.
--     __It is the duty of the user (you) to ensure a correct configuration!__
--
--   * multicore: the miner uses 1 CPU core by default, but can use as many as
--     you indicate. GPU support will be added soon.
--
-- Since this miner is "mid-level", it is up to the user to submit valid
-- `BlockHeader`s derived from some suitable `Cut` (a full @chainweb-node@
-- handles this automatically).
--
-- When mining has completed for some given work, the process will sit idle. At
-- most, between 128 and 256 recently mined `BlockHeader`s per chain will be
-- kept in the cache. If the miner process exits, the cache is lost.
-- __Unpolled results thus cannot be recovered after a miner has been shutdown.__
--
-- == Usage ==
--
-- === submit/ endpoint ===
--
-- A POST call, expecting a single `BlockHeader` in its standard JSON format.
-- This will cancel any current mining. If you wish to mine on multiple chains
-- simultaneously, you can run multiple miner processes on different ports.
--
-- === poll/ endpoint ===
--
-- A GET call, given a `ChainId` and `BlockHeight` where we expect there to be a
-- result. /May/ return a single `BlockHeader`, if mining were successful there.
-- Failure indicates one of the following:
--
--   * New work cancelled the previous, thus no result was saved.
--   * Mining may still succeed before the work stales, but just hasn't yet.
--   * The given `ChainId` and `BlockHeight` were incorrect.
--
```


miner/Miner/Types.hs
```diff
@@ -0,0 +1,44 @@
...
+deriving instance ToSchema ChainNodeId
+deriving instance ToSchema HashDifficulty
+deriving instance ToSchema HashTarget
+deriving instance ToSchema Micros
+deriving instance ToSchema Nonce
+deriving instance ToSchema PowHashNat
+deriving instance ToSchema Word128
+deriving instance ToSchema Word256
+deriving instance ToSchema a => ToSchema (Time a)
+deriving instance ToSchema a => ToSchema (TimeSpan a)
```
fosskers: Perhaps these should live in `Chainweb.RestAPI.Orphans`?

stack.yaml
```diff
@@ -26,7 +26,11 @@ extra-deps:
   - libyaml-0.1.1.0      # Due to yaml
   - massiv-0.3.3.0       # Due to digraph
   - mwc-random-0.14.0.0  # Due to digraph
-  - scheduler-1.3.0      # Due to massiv
+  # - scheduler-1.3.0      # Due to massiv
+  - github: fosskers/haskell-scheduler
+    commit: 156210700afed37fc0a6f5b7a70d2f1e99675c05
+    subdirs:
+      - scheduler
```
fosskers: Temporary. The upstream PR has already been merged and released.

miner/Miner.hs
```diff
@@ -0,0 +1,250 @@
...
+
+--------------------------------------------------------------------------------
+-- Work
+
+data Env = Env
+    { cores :: Word16
+    , version :: ChainwebVersion
+    , work :: TMVar BlockHeader
+    , results :: TVar (HashMap (T2 ChainId BlockHeight) BlockHeader)
+    }
```
fosskers: `cores` and `version` will soon be pulled from the CLI.

fosskers:
```
colin@yumi ~/c/h/chainweb> chainweb-miner --help
Usage: chainweb-miner [--cores COUNT] [--version VERSION]
  The Official Chainweb Mining Client

Available options:
  --cores COUNT            Number of CPU cores to use (default: 1)
  --version VERSION        Chainweb Network Version (default: testnet01)
  -h,--help                Show this help text
```

fosskers:
```
colin@yumi ~> chainweb-miner --help
Usage: chainweb-miner [--cores COUNT] [--version VERSION] [--port PORT]
  The Official Chainweb Mining Client

Available options:
  --cores COUNT            Number of CPU cores to use (default: 1)
  --version VERSION        Chainweb Network Version (default: testnet01)
  --port PORT              Port on which to run the miner (default: 8081)
  -h,--help                Show this help text
```

miner/Miner.hs
```diff
@@ -0,0 +1,285 @@
...
+import qualified Data.HashMap.Strict as HM
+import qualified Data.Text as T
+import Data.Tuple.Strict (T2(..))
+import Foreign.Marshal.Alloc (allocaBytes)
+import qualified Network.Wai.Handler.Warp as W
+import Options.Applicative
+import Servant.API
+import Servant.Server (Application, Server, serve)
+import Servant.Swagger (toSwagger)
+import Servant.Swagger.UI (SwaggerSchemaUI, swaggerSchemaUIServer)
```
larskuhtz: I think, we should try to be consistent across all chainweb components. So, we should use swagger ui everywhere or nowhere.

Generally, do we want swagger ui? I am a bit concerned in overloading our tools and adding problematic dependencies. If a service exports just swagger json, one can still use a stand-alone swagger-ui client to get the same functionality.

miner/Miner.hs
```diff
@@ -0,0 +1,285 @@
...
+
+--------------------------------------------------------------------------------
+-- Servant
+
+type MinerAPI = "submit" :> ReqBody '[JSON] BlockHeader :> Post '[JSON] ()
+    :<|> "poll" :> Capture "chainid" ChainId
+                :> Capture "blockheight" BlockHeight
+                :> Get '[JSON] (Maybe BlockHeader)
+
+type API = MinerAPI :<|> SwaggerSchemaUI "help" "swagger.json"
```
larskuhtz: Let's use the common Chainweb UI prefix, which is available via `Chainweb.RestAPI.Utils`.

When we start having different components, we should make sure that are consistent and that the consistency is maintainable by using common utils.

miner/Miner.hs
```diff
@@ -0,0 +1,285 @@
...
+    (long "version" <> metavar "VERSION"
+     <> value Testnet01
+     <> help "Chainweb Network Version (default: testnet01)")
+  where
+    cver :: ReadM ChainwebVersion
+    cver = eitherReader $ \s ->
+        note "Illegal ChainwebVersion" . chainwebVersionFromText $ T.pack s
+
+pPort :: Parser Int
+pPort = option auto
```
larskuhtz: Not all `Int` are valid ports.

You'll also want to include the interface to listen on. In production, opften that will be only a local network or a VPN.
f
I propose using the exiting types and configuration tools parsers and that we use already in chainweb to keep the configuration UIs for all components consistent.

miner/Miner.hs
```diff
@@ -0,0 +1,285 @@
...
+    (long "port" <> metavar "PORT" <> value 8081
```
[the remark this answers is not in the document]
larskuhtz: It should align with the default ports for chainweb. Maybe, `default-chainweb-port + 1`?

miner/Miner.hs
```diff
@@ -0,0 +1,285 @@
...
+     <> help "Port on which to run the miner (default: 8081)")
+
+--------------------------------------------------------------------------------
+-- Work
+
+main :: IO ()
+main = do
+    env <- (opts <$> newEmptyTMVarIO <*> newTVarIO mempty) >>= execParser
```
larskuhtz: Let's use the same configuration system as for Chainweb-node. One motivation behind `configuration-tools` was that it makes it easy to use the same consistent configuration across different components of micro services. If we don't like it, we can use something else, but it should be consistent for all chainweb components. 

In particular when we compile the components into a single standalone node (for desktop users) we want the configuration file format and command line be consistent with the stand-alone components.

miner/Miner.hs
```diff
@@ -0,0 +1,285 @@
...
+-- | Submit a new `BlockHeader` to mine (i.e. to determine a valid `Nonce`).
+--
+submit :: Env -> BlockHeader -> IO ()
+submit (work -> w) bh = atomically $
+    isEmptyTMVar w >>= bool (void $ swapTMVar w bh) (putTMVar w bh)
+
+-- | For some `ChainId` and `BlockHeight`, have we mined a result?
+--
+poll :: Env -> ChainId -> BlockHeight -> IO (Maybe BlockHeader)
+poll (results -> tm) cid h = HM.lookup (T2 cid h) <$> readTVarIO tm
```
larskuhtz: Why/when is this API needed? How long is this information retained? What if we mined more than one block per chain and block height?

miner/Miner.hs
```diff
@@ -0,0 +1,285 @@
...
+
+-- | A supervisor thread that listens for new work and supervises mining threads.
+--
+mining :: Env -> IO ()
+mining e = do
+    bh <- atomically . takeTMVar $ work e
+    race newWork (go bh) >>= traverse_ miningSuccess
+    mining e
+  where
+    -- | Wait for new work to come in from a `poll` call. `readTMVar` has the
```
larskuhtz: s/poll/submit/ ?

miner/Miner.hs
```diff
@@ -0,0 +1,285 @@
...
+    newWork :: IO ()
+    newWork = void . atomically . readTMVar $ work e
+
+    -- | If the `go` call won the `race`, this functions saves the result of
+    -- that successful mining. If `newWork` won the race instead, then the `go`
+    -- call is automatically cancelled.
+    miningSuccess :: BlockHeader -> IO ()
+    miningSuccess new = do
+        let key = T2 (_blockChainId new) (_blockHeight new)
+        atomically $ modifyTVar' (results e) (HM.insert key new)
```
larskuhtz: shouldn't this call `PUT cut` on consensus for the newly minded block? Or are we planning to add another component around this that is sitting in-between this miner and consensus?

larskuhtz: I guess a more general question is: Mining is generally on a per-cut level. Since this component is aware of `BlockHeaders`, chains, and block heights (as opposed to a low-level miner that only finds a nonce), shouldn't it be rather work on a cut level?

Essentially, it would listen for new cuts (via submit) mine a new cut (by picking a chain and computing a block), and when successful submit the cut to consensus via `PUT cut`.

miner/Miner/Types.hs
larskuhtz: yes, I think so, let's keep all servant and swagger orphans in a single place.

I found the best way to deal with specialized, high idiomatic, instances (like the once for swagger and servant) is to keep them as orphans in a single place.

miner/Miner.hs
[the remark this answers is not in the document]
fosskers: The default chainweb port appears to be `0` :thinking:

miner/Miner.hs
fosskers: I'll remove it for now.

miner/Miner.hs
fosskers: My thought here was that `configuration-tools` was overkill: we only have three possible flags. Using raw `optparse-applicative` kept things simple.

miner/Miner.hs
fosskers: > Why/when is this API needed?

Since the miner is push-based, some higher level coordinator (a `chainweb-node`, custom pool software, something else, etc.) must poll for the results of mining.

> How long is this information retained?

No more than 256 of the most recently mined `BlockHeader`s per chain are kept in a local in-memory cache.

> What if we mined more than one block per chain and block height?

This will only happen if the Cut-level coordinator (or pool) instructed the miner to mine on the same parent twice. If so, the newest result would overwrite the older.

miner/Miner.hs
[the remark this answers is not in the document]
larskuhtz: 0 means it's assigned by the OS, which is the right thing to do for testing.

For the bootstrap node in run-nodes it is 1789. For public nodes, e.g. on testnet, it is 443.

But those are the "public" ports. I think for local APIs we want to use a different port and possibly a different interface.

miner/Miner.hs
larskuhtz: I don't think that this design will work well. Try to imagine how one would implement a client for that interface. That client would have to keep track of which blocks it submitted, it must keep state to remember to which cuts those belonged and what other blocks for other chains where submitted for that cut. Finally, it must loop in a tight loop to poll the results and push them to the cut pipeline, which requires a separate thread.

larskuhtz: Consider this example:

1. client pushes block candidate `a` for chain `x` at height `h` of cut `A` to miner.
2. miner starts working on `a`
3. client receives new cut and pushes block candidate `b` for chain `x` at height `h` of cut `B` to miner.
4. miner starts working on `b`.

Client poll for chain `x` at height `h` and receives a block. How does the client know for what cut the result is? It may be for cut `A` or cut `B`. The client could compare if the block matches the current cut (`B`) and reject the block if it doesn't fit. But what if the rejected block would actually provide a winning cut `A` that is better than `B`?

I think, all of this would be much simpler if the miner would be aware of cuts and would submit a new cut to the existing cut-pipeline API.

If the goal is that the mining logic, e.g. how to distribute hash power of different chains, should be implemented in the miner, the miner has to be aware of cuts in any case. So, a cut based API would be not only simpler but also more future proof.

larskuhtz: A cut aware miner could be implemented based on the existing Chainweb-consensus APIs. It could poll for new cuts on `GET cut` and push new cuts via `PUT cut`.

The only change in chainweb-consensus to support the new miner would be to allow to attach the block header to the cut, which is a feature that is already on our TODO, because it is needed for other things, too.

fosskers: > Finally, it must loop in a tight loop to poll the results and push them to the cut pipeline, which requires a separate thread.

You're right, and this is precisely how pool software works. The idea is that this "core miner" cares about "work" and nothing else. Consider this point, which tries to take a long view: it may not be a single `chainweb-node` instance that's curious about the mining results! In fact, it might not be a Haskell process written by us at all. With the current design, coordination between "Cut coordinators" and Miners is a potentially many-to-many relationship. If the Miner is instead made to "talk back" to some specific endpoint, that becomes one-to-one. 

Further, the fewer concerns a "core miner" has, the easier it will be to extend / reimplement. 

> How does the client know for what cut the result is?

This is solely an issue of how the miner handles its cache. We can make a decision now and be explicit about it in documentation. For instance, some rule like: if you submit work for chain `x` on height `h`, there will **never** be a result for it in the cache unless *that* specific job completed successfully. This implies that if the job for cut `A` had finished already, its result would be deleted from the cache when job `B` comes in for the same chain/height. Or we change the key type of the cache to be more granular.

PR #275, opened 2019-07-09
Static Nix Executables

fosskers:
CON: Doc generation has been disabled for the time being. Otherwise, it works! Deploying nodes is now at least 20x faster.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -50,12 +50,12 @@ in
           sha256 = "13lim8vv78m9lhn7qfjswg7ax825gn0v75gcb80hckxawgk8zxc1";
         };
 
-        chainweb = enableDWARFDebugging (overrideCabal super.chainweb (drv: {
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -50,12 +50,12 @@ in
           sha256 = "13lim8vv78m9lhn7qfjswg7ax825gn0v75gcb80hckxawgk8zxc1";
         };
 
-        chainweb = enableDWARFDebugging (overrideCabal super.chainweb (drv: {
+        chainweb = justStaticExecutables (enableDWARFDebugging (overrideCabal super.chainweb (drv: {
```
mightybyte: Can you put a conditional on this so we can leave everything the same in CI?

fosskers: It's CI that needs to be producing the static executable, no?

mightybyte: CI should do both.

larskuhtz: What is the benefit of dynamically linked binaries, if we are not going to deploy those?

PR #276, opened 2019-07-09
Optional static executables

mightybyte:
This should accomplish the same thing as https://github.com/kadena-io/chainweb-node/pull/275 without sacrificing the other stuff in CI.


fosskers:
Closing for now.

PR #278, opened 2019-07-10
Coin Contract Updates

emilypi:
Added enforce statements to make sure we only allow positive balances, actually implement interfaces (apparently they can be `use`d, but this doesn't make sense yet to me), and address governance bug


fosskers:
Can you squash the two same-named commits together?

pact/coin-contract/coin-sig.pact
```diff
@@ -6,14 +6,19 @@
...
 
-  (defun transfer:string (sender:string reciever:string receiver-guard:guard amount:decimal)
+  (defun transfer:string (sender:string receiver:string receiver-guard:guard amount:decimal)
     @doc "Transfer between accounts SENDER and RECEIVER on the same chain.    \
     \This fails if both accounts do not exist. Create-on-transfer can be      \
     \handled by sending in a create command in the same tx."
 
-    @model [ (property (> amount 0.0)) ])
+    @model [ (property (> amount 0.0))
+             (property (not (= sender reciever)))
```
fosskers: Variable name was changed, but use-site wasn't.

fosskers: Interestingly, `flycheck` isn't picking this up.

fosskers:
```
colin@yumi ~/c/h/c/p/coin-contract> pact coin-sig.pact 
Load successful
```
Despite the `reciever` symbol not being in scope.

emilypi: yikes. I think analysis might be broken

joelburget: I looked into this a bit more. Analysis actually does fail on this file (for the wrong reason, which I'll fix). But when invoked this way (with no flags, on a `.pact` file), Pact runs in "compile only" mode, meaning that analysis is never even called.

PR #286, opened 2019-07-12
Module Caching

emilypi:
This PR enables loaded module caching on a per-block basis


src/Chainweb/Pact/PactService.hs
```diff
@@ -466,10 +469,11 @@ applyPactCmd isGenesis (Env' dbEnv) cmdIn miner = do
...
     -- cvt from Command PayloadWithTexts to Command ((Payload PublicMeta ParsedCode)
     let !cmd = payloadObj <$> cmdIn
-    result <- liftIO $ if isGenesis
+    (!result, !mcache') <- liftIO $ if isGenesis
         then applyGenesisCmd logger dbEnv pd spv cmd
-        else applyCmd logger dbEnv miner gasModel pd spv cmd
-    pure $! toHashCommandResult result
+        else applyCmd logger dbEnv miner gasModel pd spv cmd mcache
+
+    pure $! (V.cons (toHashCommandResult result) v, mcache')
```
fosskers: Consider using a `T2` here, so that you won't need to bang the results above.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -104,64 +107,66 @@ applyCmd logger pactDbEnv minerInfo gasModel pd spv cmd = do
...
         case cmdResultE of
 
           Left e2 ->
 
-            jsonErrorResult payloadEnv requestKey e2 buyGasLogs (Gas 0)
+            jsonErrorResult payloadEnv requestKey e2 buyGasLogs (Gas 0) mcache'
               "tx failure for request key when running cmd"
 
-          (Right !cmdResult) -> do
+          Right !(cmdResult, mcache'') -> do
```
fosskers: Correct bang?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -104,64 +107,66 @@ applyCmd logger pactDbEnv minerInfo gasModel pd spv cmd = do
...
 
-              (Right !redeemResult) -> do
+              Right (!redeemResult, !mcache''') -> do
 
                 let !redeemLogs = fromMaybe [] $ _crLogs redeemResult
                     !finalResult = over (crLogs . _Just) (<> redeemLogs) cmdResult
 
                 logDebugRequestKey logger requestKey "successful gas redemption for request key"
-                pure finalResult
+                pure (finalResult, mcache''')
```
fosskers: Consider `T2`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -446,18 +447,21 @@ applyPactCmds
     -> Vector (P.Command PayloadWithText)
     -> MinerInfo
     -> PactServiceM (Vector HashCommandResult)
-applyPactCmds isGenesis env' cmds miner = V.mapM f cmds
+applyPactCmds isGenesis env' cmds miner =
+    fmap (\(T2 v _) -> v) $ V.foldM f (T2 mempty mempty) cmds
```
fosskers: Turns out:
```haskell
instance (Monoid a, Monoid b) => Monoid (T2 a b) where
  mempty = T2 mempty mempty
```

fosskers: You might also consider `<$>` (hlint would warn).

fosskers: Too bad `strict-tuple` doesn't have simple accessor functions eh?

src/Chainweb/Pact/PactService.hs
```diff
@@ -466,10 +470,11 @@ applyPactCmd isGenesis (Env' dbEnv) cmdIn miner = do
...
     -- cvt from Command PayloadWithTexts to Command ((Payload PublicMeta ParsedCode)
     let !cmd = payloadObj <$> cmdIn
-    result <- liftIO $ if isGenesis
+    T2 result mcache' <- liftIO $ if isGenesis
         then applyGenesisCmd logger dbEnv pd spv cmd
-        else applyCmd logger dbEnv miner gasModel pd spv cmd
-    pure $! toHashCommandResult result
+        else applyCmd logger dbEnv miner gasModel pd spv cmd mcache
+
+    pure $! T2 (V.cons (toHashCommandResult result) v) mcache'
```
fosskers: How "hot path" is this `V.cons`? Vector consing can get expensive.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -6,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TupleSections #-}
```
fosskers: Still needed?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -172,16 +178,13 @@ applyGenesisCmd logger dbEnv pd spv cmd = do
     let initState = initCapabilities [magic_FUND_TX, magic_COINBASE]
 
     resultE <- catchesPactError $! runPayload cmdEnv initState cmd []
-    case resultE of
-
+    fmap (flip T2 mempty) $! case resultE of
```
fosskers: Consider infixing with:
```haskell
`T2` mempty
```

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -104,64 +108,66 @@ applyCmd logger pactDbEnv minerInfo gasModel pd spv cmd = do
...
     case buyGasResultE of
 
       Left e1 ->
-        jsonErrorResult buyGasEnv requestKey e1 [] (Gas 0)
+        jsonErrorResult buyGasEnv requestKey e1 [] (Gas 0) mcache
           "tx failure for requestKey when buying gas"
-
-      Right buyGasResult -> do
-
+      Right (Left e) ->
```
fosskers: I... sense Monads...

Here and below with the `Right (Right ...`

src/Chainweb/Pact/PactService.hs
emilypi: We're only doing <blocksize> many of these a per block. Perhaps I could make these a list and `fromList` the result

src/Chainweb/Pact/PactService.hs
emilypi: Ah it does have a monoid instance! Okay, that's tricky though, because that's a pretty horrible monoid 😄 (no shortcircuiting!) . Tell you what. I'll write a module for `Data.Tuple.Strict` in `src` and write the convenience functions for us.

src/Chainweb/Pact/TransactionExec.hs
emilypi: No, thanks

src/Chainweb/Pact/TransactionExec.hs
emilypi: Here, we must fail explicitly on left with `internalError` as opposed to returning a command result, so we must case.

src/Chainweb/Pact/PactService.hs
fosskers: Just put it in `Chainweb.Util`.

emilypi: Sounds good

src/Chainweb/Pact/PactService.hs
fosskers: Where blocksize == TX count? That could be 1000.

emilypi: Yes

src/Chainweb/Pact/PactService.hs
```diff
@@ -446,18 +447,21 @@ applyPactCmds
     -> Vector (P.Command PayloadWithText)
     -> MinerInfo
     -> PactServiceM (Vector HashCommandResult)
-applyPactCmds isGenesis env' cmds miner = V.mapM f cmds
+applyPactCmds isGenesis env' cmds miner =
+    fmap (V.fromList . sfst) $ V.foldM f mempty cmds
```
fosskers:
```suggestion
    V.fromList . sfst <$> V.foldM f mempty cmds
```

src/Chainweb/Pact/PactService.hs
```diff
@@ -466,10 +470,11 @@ applyPactCmd isGenesis (Env' dbEnv) cmdIn miner = do
...
     -- cvt from Command PayloadWithTexts to Command ((Payload PublicMeta ParsedCode)
     let !cmd = payloadObj <$> cmdIn
-    result <- liftIO $ if isGenesis
+    T2 result mcache' <- liftIO $ if isGenesis
         then applyGenesisCmd logger dbEnv pd spv cmd
-        else applyCmd logger dbEnv miner gasModel pd spv cmd
-    pure $! toHashCommandResult result
+        else applyCmd logger dbEnv miner gasModel pd spv cmd mcache
+
+    pure $! T2 ((toHashCommandResult result):v) mcache'
```
fosskers:
```suggestion
    pure $! T2 (toHashCommandResult result : v) mcache'
```

PR #291, opened 2019-07-17
Weeding

fosskers:


stack.yaml
```diff
@@ -1,16 +1,16 @@
 resolver: lts-12.26
 
+ghc-options: {"$locals": -ddump-to-file -ddump-hi}
```
fosskers: This helps `weeder` function properly.

PR #292, opened 2019-07-17
Sending Block information to Amberdata

LindaOrtega:
Adds Amberdata logger to chainweb node.

`07/23/2019: Version 0`
Sends blocks information (timestamp, nonce, block height, block weight, ect.) to Amberdata dashboard from the mining component. 
Implications: won't send blocks if mining disabled; Amberdata expects block heights to be different and thus overrides a block if a new one received with same block height.

`07/26/2019: Version 1`
Sends (most) blocks information to Amberdata dashboard by monitoring cuts. Thus, its not dependent on the node's mining component. It is still possible that a cut was no read fast enough and its blocks are not logged. Rare that this happens, but currently working on fixing this. Something important to note is that number of transaction per block and block size are no longer being logged. This information was easily calculated in the mining component, but not so obvious using cuts. This is another item in my todo list. 

TESTING
**When starting from scratch**
```
stack exec -- chainweb-node --enable-reset-chain-databases --amberdata-config amberdata:<hostname>:443::<api-key>::<blockchain-id>::debug
```
**Other times**
```
stack exec -- chainweb-node --amberdata-config amberdata:<hostname>:443::<api-key>::<blockchain-id>::debug
```
@larskuhtz  also suggests using editing `--log-policy` to prevent congested log pipeline from dropping Amberdata logs.

**To clear out Amberdata dashboard**
```
curl -X DELETE "https://blockchains.amberdata.io/api/v1/blocks/0/gte" \
  -H  "accept: application/json" \
  -H  "x-amberdata-api-key:<APIKEY>" \
  -H  "x-amberdata-blockchain-id:<BLOCKCHAINID" \
  -H  "content-type: application/json" \
  -vvv
```
`07/28/2019:`
`stack exec -- chainweb-node --enable-reset-chain-databases` seems to be broken. Gives this output 
```chainweb-node: /Users/linda.ortega.cordoves/.local/share/chainweb-node/testnet01/0/sqlite/pactservice-sqlite-chain14.sqlite: removeDirectoryRecursive:removeContentsRecursive:removePathRecursive:getSymbolicLinkStatus: does not exist (No such file or directory)```

### CODE REVIEW

File: src/Chainweb/Miner/POW.hs [hunk 1]
```diff
@@ -131,9 +131,22 @@ powMiner lf conf nid cdb = runForever lf "POW Miner" $ do
...
+                   (_blockMiner newBh)
+                   (_minedBlockSize nmb)
+                   (_minedBlockTrans nmb)
+                   (_blockChainId newBh)
+                   (_blockWeight newBh)
 
         logg Info $! "POW Miner: created new block" <> sshow i
         lf @(JsonLog NewMinedBlock) Info $ JsonLog nmb
+        logg Info $! "POW Miner: sent AmberData block" <> sshow i
+        lf @(JsonLog AmberdataBlock) Info $ JsonLog adb
```
mightybyte: Am I reading this correctly that the miner is sending blocks to Amberdata?  If that's the case, then it seems like Amberdata won't receive blocks mined by any other miners?  That will work for now, but ultimately I think it would be better to have consensus send the blocks so that Amberdata will get all the blocks no matter who mines them.  @larskuhtz would it be hard to make this change?

LindaOrtega: Moved to use cut monitoring, so logging all blocks (not just the ones mined by the current node).

node/ChainwebNode.hs
```diff
@@ -161,6 +166,47 @@ runCutMonitor logger db = L.withLoggerLabel ("component", "cut-monitor") logger
             -- $ S.map _cutMap
             -- $ cutStream db
 
+
+runAmberdataBlockMonitor :: Logger logger => logger -> CutDb cas -> IO ()
+runAmberdataBlockMonitor logger db = L.withLoggerLabel ("component", "amberdata-block-monitor") logger $ \l -> do
+    go l `catchAllSynchronous` \e ->
```
larskuhtz: should this be wrapped into `runForever`. That function would take care of catching exceptions, long errors, and restart as needed.

src/Chainweb/BlockHeader.hs
```diff
@@ -694,6 +695,43 @@ data NewMinedBlock = NewMinedBlock
     deriving (Eq, Show, Generic)
     deriving anyclass (ToJSON, NFData)
 
+data AmberdataBlock = AmberdataBlock
```
larskuhtz: I think this data structure should go into a different module. What about maybe `Chainweb.Logging.Amberdata`?

`Chainweb.BlockHeader` is a core consensus data structure, and we should only include what is needed for consensus.

(The same holds for the `NewMinedBlock` type.)

node/ChainwebNode.hs
```diff
@@ -161,6 +166,47 @@ runCutMonitor logger db = L.withLoggerLabel ("component", "cut-monitor") logger
...
+                                           (_blockCreationTime bh)
+                                           (_blockParent bh)
+                                           (_blockNonce bh)
+                                           (_blockMiner bh)
+                                           -- undefined Blocksize
+                                           -- undefined blockTransNum
+                                           (_blockChainId bh)
+                                           (_blockWeight bh)
+                               )
+                               (HM.toList (_cutMap c))
```
larskuhtz: This approach may miss some blocks, while reporting other bocks more than once. Cuts don't increase strictly in chains, they can also skip over blocks, and are not even monotonic for individual chains.

We should write a function that takes a stream of new cuts and emits a stream all new blocks. Such a function would use the last observed cut and the current cut and compute the set of new blocks. It is easy to write using `Chainweb.TreeDB.getBranch`.

Are we fine with missing some blocks? Otherwise are we planning to address this for this PR or for a follow up PR?

mightybyte: I think I'd rather get this PR in now.  Missing blocks isn't ideal, but doing it in a follow-up PR is fine.

src/Chainweb/BlockHeader.hs
mightybyte: Agreed.

src/Utils/Logging.hs
```diff
@@ -453,6 +456,91 @@ withTextHandleBackend label mgr c inner = case _backendConfigHandle c of
 
 -- TODO: it may be more usefull to have a logger that logs all 'Right' messages
 
+-- -------------------------------------------------------------------------- --
+-- Amberdata Backend
```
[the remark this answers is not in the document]
mightybyte: Yeah, let's save that for later.

src/Utils/Logging.hs
```diff
@@ -453,6 +456,91 @@ withTextHandleBackend label mgr c inner = case _backendConfigHandle c of
...
+amberDataBatchSize :: Natural
+amberDataBatchSize = 5
+
+amberDataBatchDelayMs :: Natural
+amberDataBatchDelayMs = 1000
+
+-- | A backend for JSON log messages that sends all logs to the Amberdata /blocks
+-- endpoint.
+-- Messages are sent in a fire-and-forget fashion. If a connection fails, the
+-- messages are dropped without notice.
```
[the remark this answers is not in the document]
mightybyte: Good points.  I've also thought that we'll need this more sophisticated tracking, but that's for later.  This should be good enough for now.

src/Utils/Logging.hs
```diff
@@ -453,6 +456,91 @@ withTextHandleBackend label mgr c inner = case _backendConfigHandle c of
...
+                                 doDebug
+    errorLogFun _ _ = return ()
+
+    -- Collect messages. If there is at least one pending message, submit a
+    -- `blocks` request every second or when the batch size is {amberDataBatchSize} messages,
+    -- whatever happens first.
+    --
+    processor queue = do
+        -- set timer to 1 second
+        timer <- registerDelay (int amberDataBatchDelayMs)
```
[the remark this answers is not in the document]
mightybyte: We don't have significant delay requirements.  10 seconds should be fine.  Making it configurable should be done later though.

src/Utils/Logging/Config.hs
```diff
@@ -255,10 +262,67 @@ instance HasTextRepresentation ClusterId where
...
 
+
+-- | Backend for logging to Amberdata
+--
+data AmberdataConfig
+  = AmberdataConfig
+    { _ambredataConfigHost :: HostAddress
+    , _amberdataApiKey :: T.Text
+    , _amberdataBlockchainId :: T.Text
+    , _amberdataDebug :: Bool
```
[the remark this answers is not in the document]
mightybyte: Perhaps later.  I think this is fine for now.

node/ChainwebNode.hs
larskuhtz: I take care of this in a followup.

src/Utils/Logging.hs
```diff
@@ -453,6 +456,91 @@ withTextHandleBackend label mgr c inner = case _backendConfigHandle c of
...
 -- TODO: it may be more usefull to have a logger that logs all 'Right' messages
 
+-- -------------------------------------------------------------------------- --
+-- Amberdata Backend
+
+amberDataBatchSize :: Natural
+amberDataBatchSize = 5
+
+amberDataBatchDelayMs :: Natural
+amberDataBatchDelayMs = 1000
```
larskuhtz:
```suggestion
amberDataBatchDelayMs = 10000
```

PR #294, opened 2019-07-18
Gas Bug Fix + Coin Contract Updates

emilypi:
- Adds necessary `enforce`s to the coin contract
- calculate gas fees correctly
- bump gas in test transactions to cover costs


emilypi:
@slpopejoy i'll add a few more to cover the negative cases

PR #295, opened 2019-07-19
Give genesis accounts more money

mightybyte:
Make it rain!


fosskers:
CI passed, but none of the grant changes have actually been included in the payloads yet.

PR #296, opened 2019-07-19
Copy (presumably short) bytestrings decoded as merkle hashes

gregorycollins:
Try to make sure we don't accidentally pin a huge bytestring when decoding a merkle hash.

### DIFF
--- a/src/Chainweb/MerkleLogHash.hs
+++ b/src/Chainweb/MerkleLogHash.hs
@@ -83,7 +83,7 @@ newtype MerkleLogHash = MerkleLogHash (MerkleRoot (HashAlg ChainwebHashTag))
 -- | Smart constructor
 --
 merkleLogHash :: MonadThrow m => B.ByteString -> m MerkleLogHash
-merkleLogHash = fmap MerkleLogHash . decodeMerkleRoot
+merkleLogHash = fmap MerkleLogHash . decodeMerkleRoot . B.copy
 {-# INLINE merkleLogHash #-}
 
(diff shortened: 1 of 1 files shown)


src/Chainweb/MerkleLogHash.hs
```diff
@@ -83,7 +83,7 @@ newtype MerkleLogHash = MerkleLogHash (MerkleRoot (HashAlg ChainwebHashTag))
 -- | Smart constructor
 --
 merkleLogHash :: MonadThrow m => B.ByteString -> m MerkleLogHash
-merkleLogHash = fmap MerkleLogHash . decodeMerkleRoot
+merkleLogHash = fmap MerkleLogHash . decodeMerkleRoot . B.copy
```
larskuhtz: This is called internally

```Haskell
-- | Decode a 'MerkleHash' from bytes.
--
decodeMerkleHash
    :: forall a b m
    . MonadThrow m
    => HashAlgorithm a
    => BA.ByteArrayAccess b
    => b
    -> m (MerkleHash a)
decodeMerkleHash b
    | BA.length b /= hashSize @a = throwM e
    | otherwise = return $ MerkleHash $ BA.convert b
  where
    e = EncodingSizeException "MerkleHash"
        (Expected (hashSize @a @Int))
        (Actual (BA.length b))
{-# INLINE decodeMerkleHash #-}
```

`MerkleHash` is a new type over `BA.Bytes`. `BA.convert` internally uses ` copyByteArrayToPtr` from the `ByteArrayAccess` class. The default implementation uses C `memcpy`. Specialized implementation should still respect the "copy" in the name of the function. So we should be fine.

PR #300, opened 2019-07-25
fix and improve treedb documentation

larskuhtz:


src/Chainweb/TreeDB.hs
```diff
@@ -689,6 +699,11 @@ branchDiff
     -> S.Stream (Of (DiffItem (DbEntry db))) IO ()
 branchDiff db l r = branchDiff_ db l r >>= \i -> S.yield (BothD i i)
 
+-- | Compares two branches of a 'TreeDb'. The fork entry is returned as the
+-- result of the stream computation.
+--
+-- If you only need one branch of the fork you may use 'getBranch' instead.
+--
 branchDiff_
```
LindaOrtega: What's the difference between `branchDiff_` and `branchDiff`? The documentation is the same for both

larskuhtz: Oh, the difference is tiny: 

> The fork entry is included as *last item* in the stream.

vs.

> The fork entry is returned as the *result* of the stream computation.

larskuhtz: `branchDiff` returns the fork entry as part of the stream. `branchDiff_` excludes the fork entry from the stream and returns it as the result of the stream computation.

PR #301, opened 2019-07-25
fix Eq and Ord instances for payload types

larskuhtz:


larskuhtz:
The PR does two things:

1. Add missing `Eq` and `Ord` instances, and
2. Makes the implementations consistent.

For all these instances the field that is used is a primary key that uniquely identifies the whole structure. Limiting the instances to only use the key results in better performance, in particular in the case when comparing two equal values.

slpopejoy:
I would want to see a justification that the `Eq` "performance" improvement is actually meaningful ("we spend 40% of our time simply comparing these things with `==`"). 

My larger concern is non-obvious semantics, especially when the other parts of these objects are decidedly meaningful -- e.g. we have a bug where we're constructing them wrong "but the hash is fine". Somebody could write a unit test and expect `Eq` to do what the usual semantic is, and hide a bug.

For `Ord` it's fine.

larskuhtz:
The semantics of the instances as defined in this PR is the same as the semantics of the derived instances. The only difference is the efficiency of the implementation. If a value violated uniqueness invariant, the instances would be broken in both cases.

The should be no need for code outside of this module to construct values of these types directly. For each of the data types we provide smart constructors in this module. If we want to prevent accidental misuse we should force the usage of the smart constructors and only export readonly field accessors. I think the only reason we haven't done that is that it would introduce quite a bit of boilerplate, namely the definition of one function per constructor field.

larskuhtz:
We use this pattern for defining instances of `Eq`, `Ord`, and `Hashable`, in various other places in the code base.

Generally, I'd like to lock down more sensible (invariant carrying) data structure throughout the code base. For instance, currently all constructor fields of 'BlockHeader' are exported, which, when used directly to construct or manipulate block headers could cause great harm. Another example are missing read-only handles for the databases.

So, far this wasn't our highest priority and the team and code base was small enough to enforcement of these invariants by social contract. But, we should start moving the code more into a direction that enforces invariants like this on the programming language level.

When this invariants are properly enforced, the implementation of `Eq` instances shouldn't make any difference.

larskuhtz:
Regarding performance, for instance the `BlockTransactions` structure

```haskell
data BlockTransactions = BlockTransactions
    { _blockTransactionsHash :: !BlockTransactionsHash
        -- ^ Root of 'TransactionTree' of the block. Primary key of
        -- 'BlockTransactionsStore'. Foreign key into 'TransactionTreeStore'.

    , _blockTransactions :: !(S.Seq Transaction)
        -- ^ Ordered list of all transactions of the block.

    , _blockMinerData :: !MinerData
        -- ^ Miner data for rewards
    }
```

contains (with the currently defined limits) up to 1000 transactions, each of which is a 'ByteString' of considerable length, say 500 bytes per tx. In contrast the `BlockTransactionHash` is only 32 bytes long. When comparing two 'BlockTransactions',  if both are equal, the difference is 0.5MB (plus over head for the `Seq`) vs. 32 bytes. Also, the hash has a high entropy, when both values are different, the comparison of the hash will conclude after just a few bytes with high probability. If the derived instance starts with the comparison with the transactions (I am pretty sure, that it starts with the hash, but in the end we have no control over that), there is a possibility that a large prefix of the transaction sequence is equal.

slpopejoy:
I can give you a direct example of where this will decrease coverage: our golden tests for Pact, which use `Eq` heavily to test results including these datastructures. As it stands this change is unacceptable because of that alone.

I still need to see where `Eq` is a bottleneck -- where is production code using `Eq` on these datastructures? The safer approach is to have a "hashes equal" contract on these objects that gets used in critical bits. 

Lastly what is the motivation for this change? We've already spent a lot of time debating, what made you decide to do this now?

larskuhtz:
I don't see how this decreases coverage. Some of the `Eq` instances didn't exist before and have been added in this PR, which is the main motivation for this PR. Also, any two values that were equal before are still equal. And any two values that weren't equal before aren't equal with this PR.

larskuhtz:
Don't gold tests evaluate tests by comparing the contents of files? In that case these instances wouldn't be used in the tests. That would also explain why those tests worked even though some of these instance didn't exist.

slpopejoy:
It's a good point that our tests golden or otherwise were proceeding merrily without `Eq`, ie with `PayloadWithOutput`s, as justifying the primary-key based `Eq`. 

My ask then is that this PR force use of the smart constructors and just export accessors.

mightybyte:

PR #303, opened 2019-07-25
update encoding roundtrip tests

larskuhtz:
This PR depends on #301 to build.

### DIFF
--- a/src/Chainweb/Utils.hs
+++ b/src/Chainweb/Utils.hs
@@ -875,6 +875,11 @@ instance FromJSON (a -> a) => FromJSON (EnableConfig a -> EnableConfig a) where
         <$< enableConfigEnabled ..: "enabled" % o
         <*< enableConfigConfig %.: "configuration" % o
 
+instance FromJSON a => FromJSON (EnableConfig a) where
+    parseJSON = withObject "EnableConfig" $ \o -> EnableConfig
+        <$> o .: "enabled"
+        <*> o .: "configuration"
+
 -- | Command line parser for the configuration of a component that can be
 -- enabled or disabled.
 --
--- a/src/Chainweb/Utils.hs
+++ b/src/Chainweb/Utils.hs
@@ -877,7 +877,7 @@ instance FromJSON (a -> a) => FromJSON (EnableConfig a -> EnableConfig a) where
 
 instance FromJSON a => FromJSON (EnableConfig a) where
     parseJSON = withObject "EnableConfig" $ \o -> EnableConfig
-        <$> o .: "enabled"
+        <$!> o .: "enabled"
         <*> o .: "configuration"
 
 -- | Command line parser for the configuration of a component that can be
--- a/src/Network/X509/SelfSigned.hs
+++ b/src/Network/X509/SelfSigned.hs
@@ -859,6 +859,12 @@ validateX509CertChainPem (X509CertChainPem a b) =
 -- -------------------------------------------------------------------------- --
 -- Utils
 
+-- | Remove white space from the fron and end of a 'ByteString'.
+--
+-- Tested with
+--
+-- prop> T.decodeUtf8 (trim (T.encodeUtf8 x)) == T.strip x
+--
 trim :: B.ByteString -> B.ByteString
 trim = fst . B8.spanEnd isSpace . B8.dropWhile isSpace
 {-# INLINE trim #-}
--- a/src/Network/X509/SelfSigned.hs
+++ b/src/Network/X509/SelfSigned.hs
@@ -54,6 +54,7 @@ module Network.X509.SelfSigned
 , name
 , org
 , AltName(..)
+, DistinguishedName
 
 -- ** Certificate Fingerprints
 , FingerprintByteCount
@@ -119,6 +120,7 @@ import Data.Bifunctor
 import Data.ByteArray (ByteArray, convert)
 import qualified Data.ByteString as B (ByteString, length, pack)
 import qualified Data.ByteString.Char8 as B8
+import Data.Char
 import Data.Default (def)
 import Data.Foldable
 import Data.Foldable (toList)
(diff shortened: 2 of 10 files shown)


src/Network/X509/SelfSigned.hs
```diff
@@ -156,6 +158,23 @@ import System.X509 (getSystemCertificateStore)
 -- internal modules
 
 import Chainweb.Utils
+    ( HasTextRepresentation(..)
```
slpopejoy: this seems excessive

src/Network/X509/SelfSigned.hs
```diff
@@ -450,9 +470,21 @@ unsafeFingerprintFromText = fromJuste . fingerprintFromText . T.pack
 -- PEM Encoded Certificate
 
 newtype X509CertPem = X509CertPem B.ByteString
-    deriving (Show, Eq, Ord, Generic)
+    deriving (Show, Generic)
     deriving anyclass (Hashable, NFData)
 
+instance Eq X509CertPem where
```
slpopejoy: `trim` in Eq sounds bad from a perf POV plus, why do we expect these things to not be well behaved in the first place. A smart constructor sounds like a better approach.

larskuhtz: We don't own the PEM encoding. The standard (RFC 7468) requires parsers to ignore white space. In this PR in only trim at the front end the end (mostly to account for the optional '\n' at the end of files), and to make the tests pass. (For the line length, 64 characters seem to be the de facto standard.)

However, I fully standard compliant implementation would strip all white space. Should I update the PR to strip all white space?

larskuhtz: Performance doesn't really matter here, since in operation we only store and compare the fingerprints of certificates. It's possible that this instance is only used for the encoding roundtrip tests.

src/Network/X509/SelfSigned.hs
larskuhtz: I did this, because 'Network.X509.SelfSigned' is a candidate for being moved to a package on its own. `Chainweb.Utils` is the only internal dependency and I though I would be a good idea to keep track on what exactly the module depends on. I left it in the code so that I don't have to come up with the list again next time I work on this module.

src/Network/X509/SelfSigned.hs
slpopejoy: Would prefer one-time strip instead of constantly doing in `Eq`

src/Network/X509/SelfSigned.hs
slpopejoy: Fair enough

src/Network/X509/SelfSigned.hs
larskuhtz: I updated the code to strip on construction. It's still a little fiddly. A better solution would probably always decode into DER, but that seems overkill for the use case here.

Btw. my original implementation of `trim` in this PR was completely wrong, I hope, I got it right this time :-)

src/Network/X509/SelfSigned.hs
```diff
@@ -830,3 +855,11 @@ validateX509CertChainPem (X509CertChainPem a b) =
     case traverse_ decodePemX509Cert (a : b) of
         Left e -> throwError $ sshow e
         Right () -> return ()
+
+-- -------------------------------------------------------------------------- --
+-- Utils
+
+trim :: B.ByteString -> B.ByteString
```
slpopejoy: Unit test please (unless this is already covered in Quickcheck somehow)

gregorycollins: `isSpace` is slow, for ascii prefer the following from attoparsec:

```haskell
isSpace_w8 :: Word8 -> Bool
isSpace_w8 w = w == 32 || w - 9 <= 4
{-# INLINE isSpace_w8 #-}
```

If you want this function to be really fast:

```haskell
trim :: B.ByteString -> B.ByteString
trim s = t
  where
    predicate = not . isSpace_w8
    l = B.length s
    b = fromMaybe l $ B.findIndex predicate s
    -- mysteriously bytestring has `findIndex` but not `findIndexEnd`
    e = maybe 0 (+1) $ findIndexEnd predicate s
    !t = if e <= b then B.empty else B.unsafeTake (e - b) $! B.unsafeDrop b s


findIndexEnd :: (Word8 -> Bool) -> ByteString -> Maybe Int
findIndexEnd k (PS x s l) =
    accursedUnutterablePerformIO $ withForeignPtr x $ \f ->
    let beg = f `plusPtr` s
        e   = beg `plusPtr` (l - 1)
        go :: Ptr Word8 -> IO (Maybe Int)
        go p = if p < beg then return Nothing
               else do
                   w <- peek p
                   if k w
                     then return $! Just $! p `minusPtr` beg
                     else go (p `plusPtr` (-1))
    in go e
{-# INLINE findIndexEnd #-}
```

larskuhtz: I am not sure if this is worth it. PEM is a textual encoding that we only use for passing around configurations. This happens in the context of reading certificates from file or form the command line, which is slow in any case and where we don't care about perf. 

All more serious computations with X509 data is done either in DER encoding, or, actually, after parsing the ASN.1 data into proper data structures.

The motivation for this PR was only for checking roundtrip consistency of data structure that happens to contain PEM data. So we needed this to make those test pass. But we care more about the other contents of those structures.

src/Chainweb/Utils.hs
```diff
@@ -875,6 +875,11 @@ instance FromJSON (a -> a) => FromJSON (EnableConfig a -> EnableConfig a) where
         <$< enableConfigEnabled ..: "enabled" % o
         <*< enableConfigConfig %.: "configuration" % o
 
+instance FromJSON a => FromJSON (EnableConfig a) where
+    parseJSON = withObject "EnableConfig" $ \o -> EnableConfig
+        <$> o .: "enabled"
```
larskuhtz:
```suggestion
        <$!> o .: "enabled"
```

src/Network/X509/SelfSigned.hs
larskuhtz: I used quick check to test the implementation of `trim` and added a haddock property that documents the test.

mightybyte:

PR #308, opened 2019-07-27
Use GasLimit instead of Int64 for mempool limits

slpopejoy:


stack.yaml
```diff
@@ -1,6 +1,6 @@
 resolver: lts-12.26
 
-ghc-options: {"$locals": -ddump-to-file -ddump-hi}
+# ghc-options: {"$locals": -ddump-to-file -ddump-hi}
```
larskuhtz: Is this supposed to be part of this PR? I have no idea what it does, just asking :-)

slpopejoy: gets rid of .dump-hi files everywhere, just for stackers

PR #310, opened 2019-07-27
Restore newblock rewind error

slpopejoy:


slpopejoy:
Closing, as cut processing can create valid scenarios for new block rewind. Also, have an idea how to avoid validate races by implementing a priority queue to pact, or more correctly something that would peek into the queue and prevent mining when validate is trying to work.

PR #312, opened 2019-07-29
Replace Seq by Vector in Payloads and MerkleLog

larskuhtz:
This PR implements a long overdue change.

It also updates some documentation in WebBlockHeaderStore, adds an missing INLINE pragma for `encodeToByteString`, and adds some strictness annotations for the module cache.


src/Chainweb/Crypto/MerkleLog.hs
```diff
@@ -374,15 +374,15 @@ emptyBody = MerkleLogBody mempty
 mapLogEntries
     :: (forall x . IsMerkleLogEntry u x => x -> b)
     -> MerkleLogEntries u h s
-    -> S.Seq b
+    -> V.Vector b
 mapLogEntries f (MerkleLogBody s) = f <$> s
-mapLogEntries f (h :+: t) = f h S.<| mapLogEntries f t
+mapLogEntries f (h :+: t) = f h `V.cons` mapLogEntries f t
```
gregorycollins: Here the asymptotics have been damaged; `V.cons` is `O(n)` instead of seq's `O(log n)`. Better would be something like this:

```haskell
mapLogEntries f m = V.concat $ go id m
  where
    go dl (MerkleLogBody s) = dl . (V.map f s :)
    go dl (h :+: t) = go (dl . (V.singleton (f h))) t 
```

i.e. concat chunks of vectors collected via dlist, pushing the `O(n)` step to the end.

src/Chainweb/BlockHash.hs
```diff
@@ -236,21 +236,21 @@ decodeBlockHashRecordChecked ps = do
     hashes <- mapM decodeBlockHashWithChainIdChecked (Expected <$!> getExpected ps)
     return $! BlockHashRecord $! HM.fromList hashes
 
-blockHashRecordToSequence :: BlockHashRecord -> S.Seq BlockHash
-blockHashRecordToSequence = S.fromList . fmap snd . sort . HM.toList . _getBlockHashRecord
+blockHashRecordToVector :: BlockHashRecord -> V.Vector BlockHash
+blockHashRecordToVector = V.fromList . fmap snd . sort . HM.toList . _getBlockHashRecord
```
gregorycollins: Here it's quite a lot cheaper to sort vector-side with `vector-algorithms` (especially for small lists), and you can probably use `unsafeThaw`/`unsafeFreeze` to sort in place without copying the input. BTW part of my issue here is that upstream a lot of the things that are in hashmaps (like cuts, the mapping from chain id -> `ChainResources`), etc are `HashMap ChainId Foo` when they should be a newtype over sorted `Vector (ChainId, Foo)`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -586,7 +585,7 @@ applyPactCmd isGenesis dbEnv cmdIn miner mcache v = do
 
     -- cvt from Command PayloadWithTexts to Command ((Payload PublicMeta ParsedCode)
     let !cmd = payloadObj <$> cmdIn
-    T2 result mcache' <- liftIO $ if isGenesis
+    T2 !result !mcache' <- liftIO $ if isGenesis
```
gregorycollins: `T2` is a strict pair type so this change should be redundant?

src/Chainweb/Payload.hs
```diff
@@ -394,7 +395,7 @@ data BlockTransactions = BlockTransactions
         -- ^ Root of 'TransactionTree' of the block. Primary key of
         -- 'BlockTransactionsStore'. Foreign key into 'TransactionTreeStore'.
 
-    , _blockTransactions :: !(S.Seq Transaction)
+    , _blockTransactions :: !(V.Vector Transaction)
```
gregorycollins: :smiley: :smiley: :smiley: :smiley:

src/Chainweb/Payload.hs
```diff
@@ -425,11 +426,11 @@ instance HasMerkleLog ChainwebHashTag BlockTransactions where
...
     toLog a = merkleLog root entries
       where
-        BlockTransactionsHash (MerkleLogHash root) = _blockTransactionsHash a
-        entries = _blockMinerData a :+: MerkleLogBody (_blockTransactions a)
+        BlockTransactionsHash (MerkleLogHash (!root)) = _blockTransactionsHash a
+        !entries = _blockMinerData a :+: MerkleLogBody (_blockTransactions a)
 
     fromLog l = BlockTransactions
-        { _blockTransactionsHash = BlockTransactionsHash $ MerkleLogHash $ _merkleLogRoot l
+        { _blockTransactionsHash = BlockTransactionsHash $! MerkleLogHash $! _merkleLogRoot l
```
gregorycollins: I see my seq-paranoia is catching on. Force all the thunks

src/Chainweb/Crypto/MerkleLog.hs
larskuhtz: Yes, this place was the original reason to use `Seq`. However, the number of log entries is expected to be small.

How efficient is `V.concat`? Is it doing a single allocation?

src/Chainweb/BlockHash.hs
larskuhtz: Agreed, that this is the right thing to do. Since we currently don't have vector-algorithms in the list of dependencies, and this code here is already tested and doesn't seem to be a major performance concern at this point, I propose that we keep those ideas in mind for follow up work.

PR #313, opened 2019-07-30
Add telemetry backend for tracing runtime of computations

larskuhtz:
This PR adds some simple infrastructure for tracing calls and durations of functions.

The purpose is to 

1. be able to trace a few long-running, high-level, key function calls in production telemetry, and
2. help with temporarily instrumenting experimental code for debugging.

It is *not* a goal to implement a generic high-efficiency profiling framework or tracing framework. There are already specialized tools that cover different areas of this complex design space, that make different trade-offs. In particular, there is a reason why Haskell's profiling runtime isn't used by default.


src/Chainweb/Sync/WebBlockHeaderStore.hs
```diff
@@ -323,7 +325,8 @@ getBlockHeaderInternal headerStore payloadStore candidateHeaderCas candidatePayl
 
     validateAndInsertPayload :: BlockHeader -> PayloadData -> IO ()
     validateAndInsertPayload hdr p = do
-        outs <- pact hdr p
+        outs <- trace logfun "Chainweb.Sync.WebBlockHeaderStore.getBlockHeaderInternal.pact" (_blockHash hdr) (length (_payloadDataTransactions p))
```
gregorycollins: wrap long line

src/Utils/Logging/Trace.hs
```diff
@@ -0,0 +1,76 @@
...
+    :: MonadIO m
+    => ToJSON param
+    => LogFunction
+    -> T.Text
+    -> param
+    -> Int
+    -> m a
+    -> m a
+trace logg label param weight a = do
+    (!r, t) <- stopWatch a
```
gregorycollins: This could be expensive to add to every call. I have a couple of proposed knobs to add: first, make it #ifdef-ably no-op with a cabal flag. Second, allow for probabilistic tracing with hash-sampling:

  - add a `mwc-random` RNG to the trace function
  - add an argument to trace `sample :: Word32` -- if sample is zero, sample always, but when nonzero only collect a trace sample when random `x <= sample`.

larskuhtz: I would have used CPP+cabal flag, but I think, there is a policy in place, that usage of CPP and, generally, conditional compilation via cabal flags should be avoided.

Regarding the probabilistic sampling, I think, that would be the job of a more general profiling system. To me even calls to the PRNG + conditional branch in the code seem too much overhead to be used broadly in the code.

I'd rather keep this stuff very simple and discourage usage for anything but expensive, rage, high-level calls (the calls in the examples all already trigger several other log messages and involve expensive IO). Or make clear that it should be used only on temporary code branches for debugging.

larskuhtz: I guess, I should add a big warning to the is function, that clarifies how it should be used.

larskuhtz: I wonder if it would make sense to add your proposed optional probabilist sampling to the `LogFunction` type in general.

That could be an interesting feature for a logger, that would be orthogonal to log-levels for controlling logging volume.

src/Chainweb/Sync/WebBlockHeaderStore.hs
larskuhtz:
```suggestion
        outs <- trace 
            logfun
            "Chainweb.Sync.WebBlockHeaderStore.getBlockHeaderInternal.pact"
            (_blockHash hdr)
            (length (_payloadDataTransactions p))
```

PR #319, opened 2019-07-30
Improve Checkpointer benchmarks

slpopejoy:
# Fix indexes.
Index performance is critical to reads it turns out! (Shocker). 

# Benchmark with nested transactions
Faster than without


docs/PactBackendBenchmarks.md
```diff
@@ -62,6 +62,24 @@ variance introduced by outliers: 98% (severely inflated)
 
 # Phase two: comparing to Pact backends
 
+## Pact Pragmas
+
+```
+fastNoJournalPragmas :: [Pragma]
+fastNoJournalPragmas = [
+  "synchronous = OFF",
```
emmanueldenloye: Though I can see why setting `synchronous` is beneficial for speed, are we going to have worry about the safety of this pragma going forward?

docs/PactBackendBenchmarks.md
```diff
@@ -126,3 +144,31 @@ variance introduced by outliers: 78% (severely inflated)
...
+
+# Index problems and nested savepoints
+
+## Fixing index on versioned tables
+
+HUGE improvement: index on (rowkey, blockheight, txid), pact pragmas
+
+```
+benchmarking pact-backend/checkpointer/usertable
+time                 80.96 μs   (80.46 μs .. 81.97 μs)
```
emmanueldenloye: These numbers are awesome. Hopefully prepared statements (if possible) can bring this number down.

docs/PactBackendBenchmarks.md
slpopejoy: Until proven otherwise, I want us to be super-optimistic. You can always throw away your pactdb and rebuild.

PR #322, opened 2019-07-31
Benchmark coin contract

slpopejoy:


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -456,6 +461,9 @@ initCapabilities :: [Capability] -> EvalState
 initCapabilities cs = set (evalCapabilities . capGranted) cs def
 {-# INLINABLE initCapabilities #-}
 
+initModules :: ModuleCache -> EvalState -> EvalState
+initModules mcache = set (evalRefs . rsLoadedModules) mcache
+
```
emilypi: 👍

docs/PactBackendBenchmarks.md
```diff
@@ -172,3 +172,28 @@ mean                 70.07 μs   (69.10 μs .. 71.65 μs)
...
+                     0.998 R²   (0.995 R² .. 1.000 R²)
+mean                 27.93 ms   (27.76 ms .. 28.28 ms)
+std dev              523.9 μs   (255.0 μs .. 928.7 μs)
+```
+
+Module caching is important!
+
+```
+benchmarking pact-backend/coin-cp/transfer
+time                 97.43 μs   (97.25 μs .. 97.68 μs)
```
emilypi: That's a very significant speedup. Caching is important!

src/Chainweb/Pact/TransactionExec.hs
emilypi: You can `INLINABLE` this call, as well. There's no harm in inlining this as a pure function

slpopejoy: We're pretty close to having nice benchmarks for these functions so I'd probably wait in order to avoid knee-jerk inlining ...

mightybyte:

PR #325, opened 2019-08-01
Stricter Mempool

larskuhtz:


src/Chainweb/Mempool/InMem.hs
```diff
@@ -454,21 +455,22 @@ recordRecentTransactions maxNumRecent newTxs rlog = rlog'
     numNewItems = V.length newTxs
     oldNext = _rlNext rlog
     newNext = oldNext + fromIntegral numNewItems
-    newTxs' = reverse ([oldNext..] `zip` V.toList newTxs)
-    newL' = newTxs' ++ _rlRecent rlog
-    newL = force $ take maxNumRecent newL'
+    newTxs' = V.reverse (V.map (T2 oldNext) newTxs)
+    newL' = newTxs' <> _rlRecent rlog
+    newL = force $ V.take maxNumRecent newL'
```
gregorycollins: This change worsens asymptotics here but I don't suppose we care too much, the size of the vector is bounded

src/Chainweb/Mempool/InMem.hs
```diff
@@ -454,21 +455,22 @@ recordRecentTransactions maxNumRecent newTxs rlog = rlog'
...
-    | otherwise = Just $! txs
+    | oldHw == oldNext = Just mempty
+    | otherwise = Just $! V.toList txs
 
   where
     oldNext = _rlNext rlog
     oldestHw = oldNext - fromIntegral maxNumRecent
-    txs = map snd $ takeWhile pred $ _rlRecent rlog
-    pred x = fst x >= oldHw
+    txs = V.map (\(T2 _ b) -> b) $ V.takeWhile pred $ _rlRecent rlog
```
emilypi: I've added strict `fst` and strict `snd` combinators here: https://github.com/kadena-io/chainweb-node/blob/master/src/Chainweb/Utils.hs#L1099

```suggestion
    txs = V.map ssnd $ V.takeWhile pred $ _rlRecent rlog
```

src/Chainweb/Mempool/InMemTypes.hs
```diff
@@ -60,10 +62,10 @@ data InMemConfig t = InMemConfig {
 
 ------------------------------------------------------------------------------
 data InMemoryMempool t = InMemoryMempool {
-    _inmemCfg :: InMemConfig t
-  , _inmemDataLock :: MVar (InMemoryMempoolData t)
-  , _inmemReaper :: ThreadId
-  , _inmemNonce :: ServerNonce
+    _inmemCfg :: !(InMemConfig t)
+  , _inmemDataLock :: !(MVar (InMemoryMempoolData t))
```
emilypi: strict mvar?

src/Chainweb/Transaction.hs
```diff
@@ -42,8 +42,8 @@ import Chainweb.Utils (Codec(..))
 -- the Text that generated it, to make gossiping easier.
 --
 data PayloadWithText = PayloadWithText
-    { payloadBytes :: ByteString
-    , payloadObj :: Payload PublicMeta ParsedCode
+    { payloadBytes :: !ByteString
```
emilypi: The question on my mind is what are we doing with these payload bytes. Should this be a Short, since the only operations we do with this are decoding? (is that true?)

src/Chainweb/Mempool/InMem.hs
larskuhtz: yes, I am aware of it. There are a few `memcpy` going on here. However, for and array of 2048 pointers, it may still be faster than traversing the list. It also depends on how much inlining and fusion happens here.

The proper solution would be to use a mutable fixed size ring buffer.

src/Chainweb/Mempool/InMem.hs
larskuhtz: nice.

src/Chainweb/Transaction.hs
larskuhtz: Good point. I think, generally, when things are stored in memory across many major GC cycles, usage of `ShortByteString` can be a good idea.

emilypi: That's how I've been thinking about our memory problem. I wonder if the same could be said for `Text` swapping out for `Text.Short`.

larskuhtz: I am a bit careful with using it across the board. conversion is always a `memcpy`, so one should be a bit careful with large structure. Also, while currently, that overhead is minimal compared to other overheads, that may change in the future.

larskuhtz: Also my understanding is that the main benefit of short byte strings is that those are not pinned and live and the normal Haskell heap. Short text is mostly about saving a bit of memory, but not 100%.

PR #328, opened 2019-08-02
Increase block gas limit to allow blocks of up 1000 tx in devnet

larskuhtz:
NOTE: The function `Chainweb.Mempool.InMem.getBlockInMem` takes `InMemConfig t` as parameter, which contains `blockGasLimit` as configured in `Chainweb.Chainweb`. However, it ignores it and instead uses the hard-coded value `Chainweb.Pact.Service.PactInProcApi.maxBlockSize` that is provided as extra argument. That’s confusing. Which value is supposed to be used?

### DIFF
--- a/src/Chainweb/Chainweb.hs
+++ b/src/Chainweb/Chainweb.hs
@@ -342,7 +342,7 @@ mempoolConfig enableReIntro = Mempool.InMemConfig
     maxRecentLog
     enableReIntro
   where
-    blockGasLimit = 1000000               -- TODO: policy decision
+    blockGasLimit = 100000               -- TODO: policy decision
     mempoolReapInterval = 60 * 20 * 1000000   -- 20 mins
     maxRecentLog = 2048                   -- store 2k recent transaction hashes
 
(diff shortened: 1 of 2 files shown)


gregorycollins:
> NOTE: The function `Chainweb.Mempool.InMem.getBlockInMem` takes `InMemConfig t` as parameter, which contains `blockGasLimit` as configured in `Chainweb.Chainweb`. However, it ignores it and instead uses the hard-coded value `Chainweb.Pact.Service.PactInProcApi.maxBlockSize` that is provided as extra argument. That’s confusing. Which value is supposed to be used?

The idea was that you need to be able to take `gasLimit` as an argument, for the following reason -- after mempool returns a block of transactions, we need to validate them -- if any of the txs fail validation (e.g., not enough gas), we need to discard them from the candidate block (and the mempool) and go back to mempool to get residual txs. However I don't think that approach will even work, and the best thing to do is always just to return a fixed sized block with a value taken from the config (i.e. remove the argument).

src/Chainweb/Chainweb.hs
```diff
@@ -342,7 +342,7 @@ mempoolConfig enableReIntro = Mempool.InMemConfig
     maxRecentLog
     enableReIntro
   where
-    blockGasLimit = 1000000               -- TODO: policy decision
+    blockGasLimit = 100000               -- TODO: policy decision
```
emilypi: What are each of these numbers doing? This one has decreased, but the below has increased by an order of magnitude. Could you explain?

emilypi: Ah n/m you'e explained it.

PR #329, opened 2019-08-02
persist validated tx in the Mempool into RocksDB

larskuhtz:
This persists the validated transactions into RocksDB.

The main motivation for this PR is to store the validated transactions in encoded form which is more memory efficient than storing the parsed Command along with all its dependencies, and the encoded transaction in memory.

Well, from there it was just a tiny step to use `RocksDbTable` instead of an in memory table/map.


larskuhtz:
This code is meant as a quick an dirty fix for the memory issues with storing parsed TX. I am a bit surprised that performance is better than I expected. Still, if we keep any of this around, we should spent more time on the code.

PR #335, opened 2019-08-05
Amberdata logging: Adds payload tx and size & allows sending single chain

LindaOrtega:
Version 2 of Amberdata logging.

- Improves amberdata config flags.
- Amberdata can now be enabled/disabled via `--enable-amberdata-logger`. Disabled by default.
- Enables config validation
- Adds flag for only sending single chain information. If flag missing or null in the config file, then all chain information will be sent.
```
$stack exec -- chainweb-node --enable-amberdata-logger --amberdata-chain-id 0
```
- Adds payload transactions to blocks logging
- Adds payload size to blocks logging

To test:
```
$ stack exec -- chainweb-node --enable-amberdata-logger --amberdata-host <HOST>:<PORT> --amberdata-api-key <API-KEY> --amberdata-blockchain-id<BLOCKCHAIN-ID> --enable-amberdata-debug
```

### DIFF
--- a/src/Chainweb/Logging/Amberdata.hs
+++ b/src/Chainweb/Logging/Amberdata.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
 
 -- |
 -- Module: Chainweb.Logging.Amberdata
@@ -34,16 +35,21 @@ import Control.Concurrent.Async
 import Control.Concurrent.STM
 import Control.DeepSeq
 import Control.Lens.TH
+import Control.Lens (view)
 import Control.Monad
 import Control.Monad.Reader (liftIO)
 import Control.Monad.Error.Class (throwError)
 
 import Data.Bool
+import Data.CAS
 import qualified Data.ByteString.Builder as BB
+import qualified Data.ByteString as BS
 import qualified Data.Foldable as HM
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.IO as T
+import qualified Data.HashMap.Strict as HMS
+
 
 import GHC.Generics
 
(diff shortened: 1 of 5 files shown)


src/Utils/Logging/Config.hs
```diff
@@ -197,7 +198,8 @@ validateBackendConfig o = do
         validateHandleConfig $ _backendConfigHandle o
         case (_backendConfigHandle o, _backendConfigColor o) of
             (FileHandle _, ColorTrue) ->
-                tell ["log messages are formatted using ANSI color escape codes but are written to a file"]
+              liftIO $ throwError
+              $ error "log messages are formatted using ANSI color escape codes but are written to a file"
```
mightybyte: I think this `error` is unnecessary.

src/Chainweb/Logging/Amberdata.hs
```diff
@@ -112,94 +127,219 @@ instance ToJSON AmberdataBlock where
...
+amberdataBlockMonitor cid logger db = do
     logFunctionText logger Info "Initialized Amberdata Block Monitor"
+    case cid of
+      Nothing -> logFunctionText logger Info "Sending blocks from ALL chains"
+      Just cid' -> logFunctionText logger Info ("Sending blocks from chain " <> toText cid')
     void
-        $ S.mapM_ (logAllBlocks logger)
-        $ S.map cutToAmberdataBlocks
+        $ S.mapM_ logBlocks
         $ cutStream db
```
larskuhtz: Consider using `blockStream` from `Chainweb.CutDB` instead of `CutStream`. It will be easier to use and give better results. In particular, it doesn't skip blocks and it logs each block only once (except for when chain jumps back and force between the same forks).

src/Utils/Logging/Config.hs
larskuhtz: Yeah, don't use `error` here.

``` Haskell
λ> :t error
error :: [Char] -> a
```

So, `error` is like `undefined`, or an infinite loop (`let a = a in a`). It can have any type, so it is a valid argument to each function. But's usually not what you want.

src/Utils/Logging/Config.hs
```diff
@@ -197,7 +197,8 @@ validateBackendConfig o = do
         validateHandleConfig $ _backendConfigHandle o
         case (_backendConfigHandle o, _backendConfigColor o) of
             (FileHandle _, ColorTrue) ->
-                tell ["log messages are formatted using ANSI color escape codes but are written to a file"]
+              throwError
+              "log messages are formatted using ANSI color escape codes but are written to a file"
```
larskuhtz: You you remove this change? It is rare but happens that one wants to write the logs to a file including the ANSI codes. `tell` just prints a warning, but doesn't abort the program.

LindaOrtega: I reverted it in the latest commit.

PR #336, opened 2019-08-06
Miner Reward

emilypi:
This PR adds mining reward function to Pact Service derived from a discrete exponential decay function from our own @buckie. It is a function of `BlockHeight`, producing a pact type `ParsedDecimal`, which is the "value of mining up to some block height" in increments of 500,00 blocks per step. 

Additionally, I've split off `Miner` into its own `.hs`, because it accrued enough weight to be considered its own thing. 

N.B. for maintenance purposes, we decided to keep this as a CSV we read at startup for `PactService`, translating it to a map we hold in the environment.


src/Chainweb/Pact/PactService.hs
```diff
@@ -407,13 +421,54 @@ readAccountGuard
...
+        . V.toList
+        . V.map formatRow
+        $ vs
+  where
+    formatRow :: (Word64, Double) -> (BlockHeight, P.ParsedDecimal)
+    formatRow (!a,!b) =
+      let
+        !n = v ^. chainGraph . to (int . size)
+        !m = fromRational $ toRational b
+      in (BlockHeight a, P.ParsedDecimal $ m / n)
```
emilypi: Rewards in the csv are calculated as the total reward across all chains, so we must divide this total by the number of chains in the chainweb graph to get the reward for a single chain

src/Chainweb/Pact/PactService.hs
```diff
@@ -407,13 +421,54 @@ readAccountGuard
...
+    :: forall cas
+    . BlockHeight -> PactServiceM cas P.ParsedDecimal
+minerReward bh = do
+    m <- view $ psMinerRewards . at (roundBy bh 500000)
+    case m of
+      Nothing -> internalError
+          $ "block height outside of admissible range: "
+          <> sshow bh
+      Just r -> return r
+{-# INLINABLE minerReward #-}
```
emilypi: In 120 years, we'll start seeing the `Nothing` case

mightybyte: Is there some kind of closed form we could use for this at all further points in time?

emilypi:
@mightybyte The problem with a closed form is that in an exponential decay function, subsequent steps require knowledge of the previous steps to be calculated. That means we'd need to weave this into the state somehow, which is a longer conversation. It's doable, but it adds a layer of complexity.

PR #343, opened 2019-08-13
Miner Consolidation

fosskers:
This PR consolidates the former `Test` and `POW` mining modules and introduces a new moduler `Coordinator` module which will allows more generalized miners to be written.


fosskers:
Note that this is likely not the final form of the Coordinator, but it is a well-defined first step.

src/Chainweb/Logging/Config.hs
```diff
@@ -89,8 +89,8 @@ defaultLogConfig = LogConfig
     { _logConfigLogger = defaultLoggerConfig
     , _logConfigBackend = defaultBackendConfig
     , _logConfigTelemetryBackend = defaultEnableConfig defaultBackendConfig
+      -- Amberdata logging disabled by default
```
fosskers: This fixes broken Haddock generation.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -79,15 +80,15 @@ type Adjustments = HM.HashMap BlockHash (T2 BlockHeight HashTarget)
 
 newtype PrevBlock = PrevBlock BlockHeader
 
-powMiner
-    :: forall cas
-    . PayloadCas cas
-    => LogFunction
+mining
+    :: forall cas. PayloadCas cas
+    => (BlockHeader -> IO BlockHeader)
```
fosskers: This is the "inner loop", which can be virtually anything, and allows us to share this code between Prod and Test scenarios.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -256,18 +259,33 @@ mineCut logfun conf nid cdb g !c !adjustments = do
...
-        Nothing -> case blockDb cid of
+    getTarget v cid bh as = case v of
+        Test{} -> testTarget
+        TimedConsensus{} -> testTarget
+        TimedCPM{} -> testTarget
+        PowConsensus{} -> prodTarget
+        Development -> prodTarget
+        Testnet00 -> prodTarget
+        Testnet01 -> prodTarget
+        Testnet02 -> prodTarget
```
larskuhtz: could we move this case distinction into into `Chainweb.Version`? E.g. using something like

```Haskell
data TargetType = ProcTarget | TestTarget

getTargetType :: ChainwebVersion -> TargetType
```

src/Chainweb/Miner/Coordinator.hs
larskuhtz: I think this should really be independent of the `BlockHeader` type. That type is pretty high level with many internal dependencies and invariants attached. Given that (possibly external) core mining function operate on plain bytes (just injecting the nonce and computing the hash), something like the following seems more adequate:

```
newtype BlockHeaderBytes = BlockHeaderBytes MutableByteArray 
    -- this should be a pinned byte array so that pointer so it are stable

type CoreMininingFunc = BlockHeaderBytes -> IO BlockHeaderBytes
```

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -66,25 +76,61 @@ runMiner
...
+    conf :: MinerConfig
+    conf = _minerResConfig m
+
+    miners :: MinerCount
+    miners = _configTestMiners conf
+
     chooseMiner
         :: PayloadCas cas
-        => ChainwebVersion
+        => MWC.GenIO
```
larskuhtz: Why not move the initialization for the PRNG into `localTest`?

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -66,25 +76,61 @@ runMiner
...
-    chooseMiner Testnet01 = powMiner
-    chooseMiner Testnet02 = powMiner
+    chooseMiner g Test{} = mining (localTest g miners)
+    chooseMiner g TimedConsensus{} = mining (localTest g miners)
+    chooseMiner _ PowConsensus{} = mining (localPOW v)
+    chooseMiner g TimedCPM{} = mining (localTest g miners)
+    chooseMiner _ Development = mining (localPOW v)
+    chooseMiner _ Testnet00 = mining (localPOW v)
+    chooseMiner _ Testnet01 = mining (localPOW v)
+    chooseMiner _ Testnet02 = mining (localPOW v)
```
larskuhtz: Can this be moved to `Chainweb.Version`. E.g. with a predicate ala

```
useTestMiner :: ChainwebVersion -> Bool
```

src/Chainweb/Miner/Coordinator.hs
fosskers: This'll be addressed elsewhere.

PR #344, opened 2019-08-13
Pact 3.2.1 bump

emilypi:
Bumps the pact version to HEAD, which gets Greg his TTL + creation times


larskuhtz:
Is there an upper bound for the TTL? What about using one day, i.e. 86400 seconds?

emilypi:
No @larskuhtz - that's something you guys will have to enforce.

larskuhtz:
> No @larskuhtz - that's something you guys will have to enforce.

If it is going to be enforced to 1 day, the tests would fail. So maybe, we should use a smaller TTL in the tests. Say, 1,000 instead of 100,000?

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -255,7 +255,7 @@ testBatch = do
     pure $ SubmitBatch (pure c)
   where
     pm :: CM.PublicMeta
-    pm = CM.PublicMeta (CM.ChainId "0") "sender00" 100 0.1
+    pm = CM.PublicMeta (CM.ChainId "0") "sender00" 100 0.1 1000000 0
```
larskuhtz: Use a smaller constant in the tests? A few seconds should be enough. In any case it should be smaller than a possibly enforced upper limit. (As long as the test isn't to test the behavior for larger values.)

emilypi: We don't have the functionality in there yet, so I chose an arbitrary value. Changing it would be completely arbitrary at this point.

PR #346, opened 2019-08-14
Set an 10m default timeout on our tests.

gregorycollins:


larskuhtz:
Isn't this something that should be set on the command line when calling the tests?

fosskers:
@larskuhtz It still can be.

larskuhtz:
@fosskers I was just wondering. I think, it is common that the CI framework sets the timeout and not the test-suite itself.

But, I agree, that there shouldn't be a harm from a 10min timeout with the current test-suite.

gregorycollins:
> Isn't this something that should be set on the command line when calling the tests?

Yes but I don't know exactly how to teach the nix crap how to do it. This was easier

PR #347, opened 2019-08-14
Implement precise transaction indexing using SQLite.

gregorycollins:
- rip out Lars' temporary tx index using RocksDB
  - add tx index to the pact checkpointer
  - rip out half of mempool impl (it is no longer needed)
  - change mempool validation to take a block height+hash
  - configure mempool validator for chainweb node to use pact checkpointer to detect dupes. Since mempool is started first, this means pact service needs to export a mempool validator using MVar.


src/Chainweb/WebPactExecutionService.hs
```diff
@@ -53,13 +47,12 @@ mkPactExecutionService
     :: MempoolBackend ChainwebTransaction
     -> TQueue RequestMsg
     -> PactExecutionService
-mkPactExecutionService mempool q = PactExecutionService
+mkPactExecutionService _mempool q = PactExecutionService
```
larskuhtz: remove `_mempool` parameter?

src/Chainweb/WebPactExecutionService.hs
```diff
@@ -53,13 +47,12 @@ mkPactExecutionService
...
   { _pactValidateBlock = \h pd -> do
       mv <- validateBlock h pd q
       r <- takeMVar mv
       case r of
-        (Right !pdo) -> markAllValidated mempool pdo (_blockHeight h) (_blockHash h)
-                          >> return pdo
+        (Right !pdo) -> return pdo
```
larskuhtz: use `either`?

src/Chainweb/Chainweb/ChainResources.hs
```diff
@@ -127,11 +128,12 @@ withChainResources
...
 withChainResources v cid rdb peer logger mempoolCfg cdbv payloadDb prune dbDir nodeid resetDb inner =
-    withBlockHeaderDb rdb v cid $ \cdb ->
-      Mempool.withInMemoryMempool mempoolCfg rdb $ \mempool -> do
-        mpc <- MPCon.mkMempoolConsensus reIntroEnabled mempool cdb $ Just payloadDb
-        withPactService v cid (setComponent "pact" logger) mpc cdbv cdb payloadDb dbDir nodeid resetDb $
-          \requestQ -> do
+    withBlockHeaderDb rdb v cid $ \cdb -> do
+      vmvar <- newEmptyMVar
+      Mempool.withInMemoryMempool (mempoolCfg vmvar) $ \mempool -> do
+        mpc <- MPCon.mkMempoolConsensus True mempool cdb $ Just payloadDb
```
larskuhtz: Is the `True` parameter still needed?

src/Chainweb/Mempool/RestAPI/Server.hs
```diff
@@ -54,11 +53,9 @@ lookupHandler mempool txs = handleErrs (liftIO look)
     look = V.toList <$> mempoolLookup mempool txV
 
 
+-- TODO: should we re-implement get block for remote?
 getBlockHandler :: Show t => MempoolBackend t -> Maybe GasLimit -> Handler [t]
```
larskuhtz: What would be the reason for the function? Remove it?

src/Chainweb/WebPactExecutionService.hs
gregorycollins: Done

src/Chainweb/WebPactExecutionService.hs
gregorycollins: Done

src/Chainweb/Chainweb/ChainResources.hs
gregorycollins: No, I'll delete

src/Chainweb/Mempool/RestAPI/Server.hs
gregorycollins: It's not used at the moment so there's no downside in deleting it

larskuhtz:
Could we please stop pushing new feature to the is PR.

It's already very big, and I have trouble keeping up reviewing it. Each time I come back there are changes to files that I already reviewed.

I propose that we put things that are not directly related to the already exiting functionality of this PR into branches on to of this PR.

src/Chainweb/Chainweb/ChainResources.hs
```diff
@@ -155,7 +157,7 @@ withChainResources v cid rdb peer logger mempoolCfg cdbv payloadDb prune dbDir n
                 logg Info $ "finished pruning block header database. Deleted " <> sshow x <> " block headers."
 
             -- replay pact
-            let pact = pes mempool requestQ
+            let pact = pes requestQ
```
larskuhtz: This is used only once in line 168. Inline?

src/Chainweb/Chainweb/ChainResources.hs
```diff
@@ -155,7 +157,7 @@ withChainResources v cid rdb peer logger mempoolCfg cdbv payloadDb prune dbDir n
                 logg Info $ "finished pruning block header database. Deleted " <> sshow x <> " block headers."
 
             -- replay pact
```
larskuhtz: Remove comment?

larskuhtz:
```suggestion
```

src/Chainweb/Mempool/Mempool.hs
```diff
@@ -10,8 +10,64 @@
...
+-- transaction hashes.
+--
+-- Transaction lifecycle:
+--
+--   - a transaction is created and signed by a user, and sent (via the pact
+--     @/send@ endpoint) to the mempool, using 'mempoolQuarantine'
+--
+--   - the transaction enters a /quarantine pool/ until it can be validated.
+--     Mempool will not gossip a transaction to peers until it has been
+--     validated.
```
larskuhtz: What exactly is the goal of "validation" here. 

Pact validation is "per block" is with respect to block in the chain (not a tx) and is expensive. A tx that is valid in one block might be invalid in another block. A tx has to be re-evaluated for each block into which it is included.

A tx is gossiped around before it is included in block. So it it's not clear what "validation" here means. If it refers to some wellformedness checks, let's call it that way, to avoid confusion with block validation.

Also, why is expensive pact validation needed for a wellformedness check? It introduced delays and circular dependencies with pact validation. Overall it seems to make the control flow related to mempool-pact interaction much more complex.

larskuhtz: A propose that (as a followup to this PR) we assemble a list of wellformedness checks that a tx should pass in order to be included in the pool of pending/gossiped transactions.

Once we have that list it will be clear what resources are needed to perform those checks, in particular, whether pact-service is needed at all for that.

gregorycollins: You are correct that this is basically a well-formedness check, and is intended to be quick. I can rename it. The well-formedness criteria as they're sketched out right now:

  - the transaction has not already been played in a block (this requires the precise transaction index)
  - the transaction is not expired relative to the block's clock time (not yet implemented)
  - the transaction has a valid signature for a coin sender (Emily is still working on this)
  - the sender's account has at least the gas limit amount of coin (we think we have this working)

You are correct that we will drop transactions if (for example) consensus briefly lands on a fork where you don't have enough gas. Tough! This means that `sum(gasLimit)` for all your outstanding transactions must exceed your account balance. The reasons we are piggybacking the initial mempool well-formedness check onto newBlock time: a) performance (so that we avoid moving the pact db pointer, which is expensive), b) this gives us a convenient synchronization point with consensus.

larskuhtz: OK, that makes sense.  Some of the checks seem relatively strong for wellformedness:

> the transaction has not already been played in a block (this requires the precise transaction index)
> the sender's account has at least the gas limit amount of coin (we think we have this working)

I would expect that wellformedness does't depend on a context, but these two checks depend on the context of a particular block.

> the transaction is not expired relative to the block's clock time

This also depends on the context of a block. For wellformedness, wouldn't it be enough to check the weaker conditions `TTL < maxTTL` and `now <= creationTime tx + TLL <= now`?

> the transaction has a valid signature for a coin sender

What does this include? Just validity of the signature also authentication with respect to some data/account in the chain?

I would expect that wellformedness also includes more basic checks like:

* the tx has can be parsed and contains syntactically valid pact.
* the tx has the correct chainweb version.
* the tx is for the correct chain.
* the tx has a supported pact version.
* the tx fits into, yet to be defined, broad syntactic limits (e.g. overall size in bytes).
* the tx has acceptable (with respect to some broad constants) gas limits.
* the sender (public key) of the tx isn't blacklisted.

I assume some of these are probably implicit by successfully decoding the tx. It may still be good to document this.

Regarding piggybacking onto pact new block validation, I am not sure about the pros and cons. It seems to me that piggy packing adds some complexity to the code, and also causes tx to fail relatively late, after already causing some resource consumption. It also seems problematic that the gossip behavior depends on the checkpoint and the miner calling newBlock. But I also see how gossiping only after more rigorous validation helps to prevent DOS attacks.

There might be some middle ground here with (1) a cheap fail-fast wellformedness check for incoming tx and (2) an API for feedback from pact validation about tx validation failures that could be used for P2P reputation management for throwing and/or blacklisting senders of malformed txs or txs that consistently fail pact validation.

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -428,8 +440,12 @@ withPactCtxSQLite v cutDB bhdbIO pdbIO f =
...
       !ctx <- TestPactCtx
         <$!> newMVar (PactServiceState Nothing)
-        <*> pure (PactServiceEnv Nothing cpe spv pd pdb bhdb)
-      evalPactServiceM ctx (initialPayloadState v cid mempty)
+        <*> pure (PactServiceEnv Nothing cpe spv pd pdb bhdb adb)
+      evalPactServiceM ctx (initialPayloadState v cid)
       return (ctx, dbSt)
+
+withMVarResource :: a -> (IO (MVar a) -> TestTree) -> TestTree
+withMVarResource value = withResource (newMVar value) (const $ return ())
```
larskuhtz: This seems useful more generally. Maybe it can be moved to `Chainweb.Test.Utils`, so that it is available to other tests, too?

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -700,14 +744,17 @@ deleteHistory bh = do
...
-initSchema = withSavepoint DbTransaction $ do
-    createBlockHistoryTable
-    createTableCreationTable
-    createTableMutationTable
-    create (domainTableName KeySets)
-    create (domainTableName Modules)
-    create (domainTableName Namespaces)
-    create (domainTableName Pacts)
+initSchema = do
+    vacuumDb
```
larskuhtz: How expensive is this? In particular for databases that are more than 50% of the file system size and in the order of dozens of GB?

I wonder if this could be responsible for a performance regression that I observed on devnet when testing this PR.

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
```diff
@@ -53,21 +59,34 @@ initInMemoryCheckpointEnv loggers logger gasEnv = do
...
     modifyMVarMasked lock $ \store -> do
-      let bh = (pred height, hash)
+      let ph = pred height
+      let bh = (ph, hash)
+      let filt (_, mp) = (mempty, HMS.filterWithKey (\_ (v, _) -> v <= ph) mp)
       case HMS.lookup bh (_theStore store) of
             Just dbenv -> do
               mdbenv <- newMVar dbenv
-              let !store' = store { _lastBlock = Just bh, _dbenv = mdbenv }
+              played <- readMVar (_playedTxs store) >>= newMVar . filt
```
larskuhtz: Is the result of applying `filt` forced, including the hash set in the second tuple field?

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
```diff
@@ -37,14 +41,16 @@ data Store = Store
   { _theStore :: HashMap (BlockHeight, BlockHash) (DbEnv PureDb)
   , _lastBlock :: Maybe (BlockHeight, BlockHash)
   , _dbenv :: MVar (DbEnv PureDb)
+  , _playedTxs :: MVar (HashSet PactHash, HashMap PactHash (BlockHeight, BlockHash))
```
larskuhtz: Make the constructor fields strict?

larskuhtz: What about using strict Tuples?

src/Chainweb/Utils.hs
```diff
@@ -1074,9 +1076,22 @@ concurrentWith alloc inner params = do
...
 
+-- | Currying for functions of strict tuples
+--
+scurry :: forall a b c. ((T2 a b) -> c) -> a -> b -> c
+scurry k a b = k (T2 a b)
+
+-- | Uncurrying for functions of strict tuples
+--
+suncurry :: forall a b c. (a -> b -> c) -> T2 a b -> c
+suncurry k (T2 a b) = k a b
```
larskuhtz: 👍

larskuhtz: Should this (as well as `scurry`, `ssnd`, and `sfst`) be inlined, or do we rely on GHC doing it anyways?

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -299,6 +312,7 @@ data PactServiceEnv cas = PactServiceEnv
     , _psPublicData :: !PublicData
     , _psPdb :: PayloadDb cas
     , _psBlockHeaderDb :: BlockHeaderDb
+    , _psActiveDbHandle :: MVar (Maybe PactDbEnv')
```
larskuhtz: Make all constructor fields strict?

src/Chainweb/Mempool/Mempool.hs
gregorycollins: I'll rip out the quarantine later today since Stu doesn't want it. Will add a TODO to add the pre-gossip check

test/Chainweb/Test/Pact/Utils.hs
gregorycollins: Maybe but I'm not doing it before I go on vacation :) :)

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
gregorycollins: I took it out

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
gregorycollins: We don't care -- the in memory checkpointer is not in use. I'll fix it anyways

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -773,5 +772,7 @@ getEndingTxId bh = callDb "getEndingTxId" $ \db -> do
     convertInt (SInt thing) = fromIntegral thing
     convertInt _ = error "impossible"
 
+-- Careful doing this! It's expensive and for our use case, probably pointless.
+-- We should reserve vacuuming for an offline process
```
larskuhtz: Maybe a command line switch, so that users know what they can expect and aren't surprised when startup takes a long time?

larskuhtz: Actually, should we just provide a separate tool (or mode of Chainweb-node) that only does off-line db maintenance. Like consistency checks, repairs, pruning, GC, etc?

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
gregorycollins: Another place we don't care

gregorycollins: Don't care :)

src/Chainweb/Utils.hs
gregorycollins: I'll do it

src/Chainweb/Pact/Backend/Types.hs
gregorycollins: `StrictData` is on here

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
gregorycollins: We'll just delete it for now

src/Chainweb/Chainweb/ChainResources.hs
gregorycollins: Done

Issue #297, opened 2019-07-19
Increase nonce size


mightybyte:
We're currently using 64 bits which is definitely not enough.  Probably need to bump it to 256 or 512.

mightybyte:
After discussion, the current 64 bits should be sufficient.  The miners also control the miner address (account to which they want to send their block reward) and the timestamp, which should give enough uniqueness that 64 bits should be sufficient for a long time to come.

PR #348, opened 2019-08-14
Simple Remote Miner Communication

fosskers:
This PR introduces the `--remote-miner` flag. If specified at least once, `chainweb-node` will forego running its own mining "inner loop", and instead make remote calls to hosts provided by the aforementioned flag.

Since the existing `chainweb-miner` only provides the tight inner loop and knows nothing else about Chainweb, we follow the "push-work-pull-results" model of fetching mined blocks from the remote processes. This may or may not be optimal, but it's a start.

Note also for @larskuhtz that I haven't yet attempted a unifying typeclass for "inner loops". Right now the inner loop implementations live in `Chainweb.Miner.Miners`.

### Testing this PR

1. Install `chainweb-node` and `chainweb-miner` by your usual means (`stack install`, etc.).
2. Run `chainweb-miner`as-is. You should see:
```
colin@yumi ~/c/h/chainweb> chainweb-miner
ClientArgs 
    { cores = 1
    , version = development
    , port = 8081
    } 
```
3. Ping the miner:
```
colin@yumi ~> curl localhost:8081/env
{"version":"development","cores":1,"port":8081}
```
4. Run `chainweb-node --chainweb-version=development --remote-miner=localhost:8081` and watch it work.
5. Stop `chainweb-node` and run a second miner: `chainweb-miner --port 8082`
6. Run `chainweb-node --chainweb-version=development --remote-miner=localhost:8081 --remote-miner=localhost:8082` to communicate with both clients at once!


src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -76,61 +74,45 @@ runMiner
...
+    testMiner :: IO (BlockHeader -> IO BlockHeader)
+    testMiner = do
+        gen <- MWC.createSystemRandom
+        pure $ localTest gen miners
+
+    powMiner :: IO (BlockHeader -> IO BlockHeader)
+    powMiner = case _configRemoteMiners conf of
+        [] -> pure $ localPOW v
+        rs -> do
+            m <- newManager defaultManagerSettings
```
fosskers: Why not `tlsManagerSettings` from `http-client-tls`? It seems precedent in Chainweb to use `defaultManagerSettings`, so I went with that here.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -76,61 +74,45 @@ runMiner
...
+            pure . remoteMining m $ map f rs
+
+    f :: HostAddress -> BaseUrl
+    f (HostAddress hn p) = BaseUrl Http hn' p' ""
```
fosskers: What `Scheme` to use? `Http` or `Https`?

src/Chainweb/Miner/Core.hs
```diff
@@ -67,22 +66,14 @@ mine _ h = BA.withByteArray initialTargetBytes $ \trgPtr -> do
 
             -- inner mining loop
             --
-            -- We do 100000 hashes before we update the creation time.
-            --
-            let go 100000 !n = do
-                    -- update the block creation time
-                    ct <- getCurrentTimeIntegral
-                    injectTime ct buf
-                    go 0 n
```
fosskers: @larskuhtz and I discussed it, and decided that resetting the time-stamp "mid-mine" isn't useful.

src/Chainweb/Miner/Miners.hs
```diff
@@ -0,0 +1,122 @@
...
+
+-- | Some remote process which is performing the low-level mining for us. May be
+-- on a different machine, may be on multiple machines, may be arbitrarily
+-- multithreaded.
+--
+remoteMining :: Manager -> [BaseUrl] -> BlockHeader -> IO BlockHeader
+remoteMining m urls bh = traverseConcurrently_ Par' submission urls >> polling
+  where
+    -- TODO Better error handling. Don't bail the entire thread if at least one
+    -- miner was successfully communicated with. Just log the rest.
```
fosskers: Need better error handling here.

src/Chainweb/Chainweb/MinerResources.hs
mightybyte: I'd with `Http`.  This seems like mostly an internal communication channel and `Https` is probably just extra ops burden for miners.

src/Chainweb/Miner/Core.hs
```diff
@@ -113,15 +104,10 @@ mine _ h = BA.withByteArray initialTargetBytes $ \trgPtr -> do
             hashInternalFinalize ctxPtr (castPtr pow)
     {-# INLINE hash #-}
 
-    -- | `injectTime` and `injectNonce` make low-level assumptions about the
-    -- byte layout of a hashed `BlockHeader`. If that layout changes, these
-    -- functions need to be updated. The assumption allows us to iterate on new
-    -- nonces quickly.
+    -- | `injectNonce` makes low-level assumptions about the byte layout of a
```
mightybyte: Maybe we can take this opportunity (or very soon hereafter) to address this ticket?

https://github.com/kadena-io/chainweb-node/issues/297

src/Chainweb/Miner/Miners.hs
fosskers: I will address this in a follow-up PR.

src/Chainweb/Miner/Core.hs
fosskers: We won't be doing this afterall.

PR #350, opened 2019-08-15
Move the implementation for efficiently querying ancestors to TreeDB

larskuhtz:


src/Chainweb/Logging/Config.hs
```diff
@@ -90,7 +90,7 @@ defaultLogConfig = LogConfig
     , _logConfigBackend = defaultBackendConfig
     , _logConfigTelemetryBackend = defaultEnableConfig defaultBackendConfig
     , _logConfigAmberdataBackend = EnableConfig False defaultAmberdataConfig
-      -- ^ Amberdata logging disabled by default
+        -- Amberdata logging disabled by default
```
fosskers: I noticed this elsewhere too. I'm surprised that CI didn't catch it. Perhaps something to do with specific Haddock versions on CI vs our machines.

src/Chainweb/Utils.hs
```diff
@@ -367,6 +374,42 @@ data DiffItem a
...
+
+traverseDiffItem
+    :: Applicative f
+    => (a -> f b)
+    -> (a -> f b)
+    -> DiffItem a
+    -> f (DiffItem b)
+traverseDiffItem f _ (LeftD a) = LeftD <$> f a
+traverseDiffItem _ g (RightD b) = RightD <$> g b
+traverseDiffItem f g (BothD a b) = BothD <$> f a <*> g b
```
fosskers: This could just be an instance of `Bitraverable`. The above functions could probably be reworked into `Bifoldable` too.

This also looks suspiciously like the `These` datatype.

larskuhtz: `Bifunctor`, `Bifoldable`, and `Bitraversable` are all parameterized over two types. But `DiffItem` has only a single type parameters. Both elements have the same type.

larskuhtz: We could consider using something like `type DiffItem a = These a a` (or `Or a a`), which would be a weaker type, but that would still be fine.

What I don't like about `these` is the large set of dependencies. Also that package was a bit unstable recently. `data-or` on the other hand depends only on base but isn't up to date feature wise. So, I think, that a bit of boilerplate here might be fine.

It would be nice if a type like `Or` or `These` existed in base along with the respective `Bi...` instances.

PR #351, opened 2019-08-15
[Transaction-Generator] Add REPL features

emmanueldenloye:
Luke! Use the REPL


tools/txg/TXGRepl.hs
```diff
@@ -0,0 +1,298 @@
...
+type ReceiverName = String
+
+type Balance = Decimal
+
+type Amount = Decimal
+
+data CoinCode
+    = CoinCreateAccount Account Guard
+    | CoinAccountBalance Account
+    | CoinTransfer SenderName ReceiverName Amount Keyset
```
mightybyte: We now have a new transfer function that doesn't require the `Keyset`.  Might be good to support that too.

emmanueldenloye: Ok, will do so

PR #352, opened 2019-08-15
DO NOT SUBMIT: Commit failing check: pact should refuse to validate duplicate txs

gregorycollins:


test/Chainweb/Test/Pact/PactReplay.hs
```diff
@@ -114,6 +116,23 @@ testMemPoolAccess  = MemPoolAccess
...
+    { mpaGetBlock = getTestBlock
+    , mpaSetLastHeader = \_ -> return ()
+    , mpaProcessFork = \_ -> return ()
+    }
+  where
+    ksData :: Text -> Value
+    ksData idx = object [("k" <> idx) .= object [ "keys" .= ([] :: [Text]), "pred" .= String ">=" ]]
+    getTestBlock _bHeight _bHash _bHeader = do
+        let nonce = "0"
+            moduleStr = defModule (T.pack nonce)
```
mightybyte: Hmmmmm, this module seems like a module that might be legal to submit twice.  How about testing this with a coin transfer and using the same nonce both times.  That would be the most convincing to me.

gregorycollins: The module would be legal to submit twice, but here the nonce is the same. All of the functions here (`goldenTestTransactions`, `ksData`) are pure so we are confident that this mempool will always produce the same block with exactly the same transaction hashes. You can verify this by inspecting the output from the failing test on master, the tx hashes are the same (and this is why the test is failing).

mightybyte: Ok, I see now.  Looks good!

PR #353, opened 2019-08-15
Single `Testnet*` Value

fosskers:
This PR follows up on a previous discussion to reduce the number of `Testnet*` values of the `ChainwebVersion` type. This PR removes `Testnet00` and `Testnet01`, leaving only `Testnet02`. In future, when we wish to release new testnet version, we are to increment the `02` *in place* (i.e. changing `Testnet02` directly to `Testnet03`). This has the effect of "cutting off" old running testnets.

Through this work I discovered some oversights and potential bugs, which are pointed out below.


node/ChainwebNode.hs
```diff
@@ -364,7 +364,7 @@ mainInfo :: ProgramInfo ChainwebNodeConfiguration
 mainInfo = programInfoValidate
     "Chainweb Node"
     pChainwebNodeConfiguration
-    (defaultChainwebNodeConfiguration Testnet01)
+    (defaultChainwebNodeConfiguration Testnet02)
```
fosskers: Oversight. This will now be caught by future changes.

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -121,12 +113,10 @@ genesisBlockPayload :: ChainwebVersion -> ChainId -> PayloadWithOutputs
 genesisBlockPayload Test{} _ = emptyPayload
 genesisBlockPayload TimedConsensus{} _ = emptyPayload
 genesisBlockPayload PowConsensus{} _ = emptyPayload
-genesisBlockPayload TimedCPM{} _ = TN0.payloadBlock
+genesisBlockPayload TimedCPM{} _ = TN2.payloadBlock
```
fosskers: Should this have its own  payload instead?

src/Chainweb/Pact/PactService.hs
```diff
@@ -70,7 +70,8 @@ import qualified Pact.Types.SPV as P
 import Chainweb.BlockHash
 import Chainweb.BlockHeader
 import Chainweb.BlockHeader.Genesis (genesisBlockHeader)
-import Chainweb.BlockHeader.Genesis.Testnet00Payload (payloadBlock)
```
fosskers: Bad bad bad! See below.

src/Chainweb/Pact/PactService.hs
```diff
@@ -186,35 +187,35 @@ initialPayloadState
...
   where
+    ghash :: BlockHash
     ghash = _blockHash genesisHeader
+
+    createCoinContract :: ReaderT (PactServiceEnv cas) (StateT PactServiceState IO) ()
     createCoinContract = do
         txs <- execValidateBlock mpa genesisHeader inputPayloadData
         bitraverse_ throwM pure $ validateHashes txs genesisHeader
 
-    PayloadWithOutputs{..} = payloadBlock
```
fosskers: Bug: Pact Service was being initialized with `Testnet00`'s payload, regardless of `ChainwebVersion`. We have no idea how the system worked despite this.

@emilypi @emmanueldenloye @marklnichols any thoughts?

src/Chainweb/Pact/PactService.hs
```diff
@@ -186,35 +187,35 @@ initialPayloadState
...
-    inputPayloadData = PayloadData (fmap fst _payloadWithOutputsTransactions)
-                       _payloadWithOutputsMiner
-                       _payloadWithOutputsPayloadHash
-                       _payloadWithOutputsTransactionsHash
-                       _payloadWithOutputsOutputsHash
+    inputPayloadData :: PayloadData
+    inputPayloadData = payloadWithOutputsToPayloadData pwo
```
fosskers: Oversight: use smart constructor.

src/Chainweb/Version.hs
```diff
@@ -220,15 +218,11 @@ chainwebVersionId v@TimedConsensus{} = toTestChainwebVersion v
 chainwebVersionId v@PowConsensus{} = toTestChainwebVersion v
 chainwebVersionId v@TimedCPM{} = toTestChainwebVersion v
 chainwebVersionId Development = 0x00000001
-chainwebVersionId Testnet00 = 0x00000002
-chainwebVersionId Testnet01 = 0x00000003
```
fosskers: What should be the policy concerning the incrementing of these codes?

test/Chainweb/Test/Orphans/Internal.hs
```diff
@@ -62,8 +62,8 @@ instance Arbitrary ChainwebVersion where
         , PowConsensus petersonChainGraph
         , TimedCPM singletonChainGraph
         , TimedCPM petersonChainGraph
-        , Testnet00
-        , Testnet01
+        , Development
+        , Testnet02
```
fosskers: Oversight: these were left out.

test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -17,34 +17,35 @@ module Chainweb.Test.Pact.PactExec
...
 import Chainweb.Test.Pact.Utils
 import Chainweb.Test.Utils
 import Chainweb.Version (ChainwebVersion(..), someChainId)
 
-import Pact.Types.Command
+import Pact.Types.Command (CommandResult(..), PactResult(..))
 
 testVersion :: ChainwebVersion
-testVersion = Testnet00
+testVersion = Development
```
fosskers: To avoid any possible conflict with testnet.

tools/ea/Ea.hs
```diff
@@ -93,7 +93,7 @@ defCoinContract :: FilePath
 defCoinContract = "pact/coin-contract/load-coin-contract.yaml"
 
 defGrants :: FilePath
-defGrants = "pact/genesis/testnet00/grants.yaml"
+defGrants = "pact/genesis/testnet02/grants.yaml"
```
fosskers: Oversight.

tools/ea/Ea.hs
```diff
@@ -126,7 +126,7 @@ genPayloadModule v txFiles =
 
         TIO.writeFile (T.unpack fileName) modl
   where
-    cid = someChainId Testnet00
+    cid = someChainId v
```
fosskers: Oversight.

tools/ea/Ea.hs
mightybyte: Again, I'd probably replace this with just `testnet` if we're only ever going to have one floating around at a time.

mightybyte:
Other than my above comments, looks like a positive direction to move in.  @larskuhtz ?

chainweb.cabal
```diff
@@ -100,8 +100,6 @@ library
         , Chainweb.BlockHeader
         , Chainweb.BlockHeader.Genesis
         , Chainweb.BlockHeader.Genesis.DevelopmentPayload
-        , Chainweb.BlockHeader.Genesis.Testnet00Payload
-        , Chainweb.BlockHeader.Genesis.Testnet01Payload
         , Chainweb.BlockHeader.Genesis.Testnet02Payload
```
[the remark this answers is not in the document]
fosskers: These module names are generated by `ea` and are a direct textual rendering from the `ChainwebVersion` value.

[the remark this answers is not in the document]
mightybyte: Hmmm, my concern here would be that now the process of generating a new set of genesis blocks starts to become an endeavor that requires expertise and that can easily be gotten wrong.  Not sure how to avoid that.

PR #354, opened 2019-08-16
Use batch inserts into CAS and RocksDbTables

larskuhtz:


src/Chainweb/BlockHeaderDB.hs
```diff
@@ -316,9 +316,12 @@ dbAddChecked db e = unlessM (casMember (_chainDbCas db) ek) dbAddCheckedInternal
...
     -- different tables)
-    add = do
-        casInsert (_chainDbCas db) (RankedBlockHeader e)
-        tableInsert (_chainDbRankTable db) (_blockHash e) (_blockHeight e)
+    add = updateBatch
+            [ RocksDbInsert (_chainDbCas db) (casKey rbh) rbh
+            , RocksDbInsert (_chainDbRankTable db) (_blockHash e) (_blockHeight e)
+            ]
+      where
+        rbh = RankedBlockHeader e
```
fosskers: `rbh` is only used in one place, so the `where` clause might not be necessary.

src/Chainweb/CutDB.hs
```diff
@@ -519,11 +520,11 @@ cutHashesToBlockHeaderMap
...
     trace logfun "Chainweb.CutDB.cutHashesToBlockHeaderMap" (_cutId hs) 1 $ do
-        hdrs <- emptyCas
-        traverse_ (casInsert hdrs) $ _cutHashesHeaders hs
-
         plds <- emptyCas
-        traverse_ (casInsert plds) $ _cutHashesPayloads hs
+        casInsertBatch plds $ V.fromList $ HM.elems $ _cutHashesPayloads hs
+
+        hdrs <- emptyCas
+        casInsertBatch hdrs $ V.fromList $ HM.elems $ _cutHashesHeaders hs
```
fosskers: How many items do we expect to insert at once? Could this ever be an expensive `Vector` conversion? GHC might be nice and fuse the inner list away anyway.

larskuhtz: Agreed, using vector in the class API was probably a mistake. Lazy `[]` would have been more flexible. But changing the `IsCas` class was beyond the scope of this PR.

In any case in practice this shouldn't make a big difference. Batches shouldn't be long in any case. Not sure, though if GHC is smart enough to scope the allocation for the vector.

src/Chainweb/BlockHeaderDB.hs
larskuhtz: Actually, it's used in two places.

fosskers: My mistake!

src/Chainweb/CutDB.hs
fosskers: Usually it is in the case of pulling keys lists from a `Map`.

PR #356, opened 2019-08-16
Adds a simpler coin transfer function that doesn't need a guard

mightybyte:


pact/genesis/testnet/grants.yaml
```diff
@@ -32,5 +34,5 @@ data:
       - "43f2adb1de192000cb3777bacc7f983b6614fd9c1715cd44cd484b6d3a0d34c8"
       - "2d70aa4f697c3a3b8dd6d97745ac074edcfd0eb65c37774cde25135483bea71e"
     pred: "keys-any"
-nonce: testnet00-grants
+nonce: testnet02-grants
```
fosskers: Possibly just `testnet-grants` now?

PR #359, opened 2019-08-19
Remote Miner Communication Resiliance 

fosskers:
This PR ensures that brittle/failing connections between `chainweb-node` and `chainweb-miner` processes will not cause problems.

### DIFF
--- a/src/Chainweb/Chainweb/MinerResources.hs
+++ b/src/Chainweb/Chainweb/MinerResources.hs
@@ -20,9 +20,10 @@ module Chainweb.Chainweb.MinerResources
   , runMiner
   ) where
 
+import Data.List.NonEmpty (NonEmpty)
 import qualified Data.List.NonEmpty as NEL
-import Data.Set.NonEmpty (NESet)
-import qualified Data.Set.NonEmpty as NES
+import Data.Set (Set)
+import qualified Data.Set as S
 import qualified Data.Text as T
 
 import Network.HTTP.Client (defaultManagerSettings, newManager)
@@ -110,8 +111,8 @@ runMiner v mr = do
             m <- newManager defaultManagerSettings
             pure $ remoteMining m rs
 
-    g :: [HostAddress] -> Maybe (NESet BaseUrl)
-    g = fmap (NES.fromList . NEL.map f) . NEL.nonEmpty
+    g :: Set HostAddress -> Maybe (NonEmpty BaseUrl)
+    g = fmap (NEL.map f) . NEL.nonEmpty . S.toList
 
     f :: HostAddress -> BaseUrl
     f (HostAddress hn p) = BaseUrl Http hn' p' ""
--- a/src/Chainweb/Chainweb/MinerResources.hs
+++ b/src/Chainweb/Chainweb/MinerResources.hs
@@ -21,6 +21,7 @@ module Chainweb.Chainweb.MinerResources
   ) where
 
 import qualified Data.List.NonEmpty as NEL
+import Data.Set.NonEmpty (NESet)
 import qualified Data.Set.NonEmpty as NES
 import qualified Data.Text as T
 
@@ -103,11 +104,14 @@ runMiner v mr = do
         pure $ localTest gen miners
 
     powMiner :: IO (BlockHeader -> IO BlockHeader)
-    powMiner = case NEL.nonEmpty $ _configRemoteMiners conf of
+    powMiner = case g $ _configRemoteMiners conf of
         Nothing -> pure $ localPOW v
         Just rs -> do
             m <- newManager defaultManagerSettings
-            pure . remoteMining m . NES.fromList $ NEL.map f rs
+            pure $ remoteMining m rs
+
+    g :: [HostAddress] -> Maybe (NESet BaseUrl)
+    g = fmap (NES.fromList . NEL.map f) . NEL.nonEmpty
 
     f :: HostAddress -> BaseUrl
     f (HostAddress hn p) = BaseUrl Http hn' p' ""
--- a/src/Chainweb/Chainweb/MinerResources.hs
+++ b/src/Chainweb/Chainweb/MinerResources.hs
@@ -20,6 +20,8 @@ module Chainweb.Chainweb.MinerResources
   , runMiner
   ) where
 
+import qualified Data.List.NonEmpty as NEL
+import qualified Data.Set.NonEmpty as NES
 import qualified Data.Text as T
 
 import Network.HTTP.Client (defaultManagerSettings, newManager)
@@ -101,11 +103,11 @@ runMiner v mr = do
         pure $ localTest gen miners
 
     powMiner :: IO (BlockHeader -> IO BlockHeader)
-    powMiner = case _configRemoteMiners conf of
-        [] -> pure $ localPOW v
-        rs -> do
+    powMiner = case NEL.nonEmpty $ _configRemoteMiners conf of
+        Nothing -> pure $ localPOW v
+        Just rs -> do
             m <- newManager defaultManagerSettings
-            pure . remoteMining m $ map f rs
+            pure . remoteMining m . NES.fromList $ NEL.map f rs
 
     f :: HostAddress -> BaseUrl
     f (HostAddress hn p) = BaseUrl Http hn' p' ""
--- a/src/Chainweb/Miner/Config.hs
+++ b/src/Chainweb/Miner/Config.hs
@@ -19,6 +19,9 @@ import Configuration.Utils
 
 import Control.Lens hiding ((.=))
 
+import Data.Set (Set)
+import qualified Data.Set as S
+
 import GHC.Generics (Generic)
 
 import Numeric.Natural (Natural)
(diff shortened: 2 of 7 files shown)


src/Chainweb/Miner/Miners.hs
```diff
@@ -96,13 +101,19 @@ submit :<|> poll = client (Proxy :: Proxy MiningAPI)
...
-    -- TODO Better error handling. Don't bail the entire thread if at least one
-    -- miner was successfully communicated with. Just log the rest.
-    submission :: BaseUrl -> IO ()
-    submission url = runClientM (submit bh) (ClientEnv m url Nothing) >>= either throwM pure
+    -- TODO Report /all/ miner calls that errored?
+    -- | Submit work to each given mining client. Will succeed so long as at
+    -- least one call returns back successful.
+    submission :: IO ()
+    submission = do
+        rs <- traverseConcurrently Par' f urls
```
fosskers: `Par'` will use all cores. Is that what we want?

src/Chainweb/Utils.hs
```diff
@@ -329,6 +334,13 @@ whenM :: Monad m => m Bool -> m () -> m ()
 whenM c a = c >>= flip when a
 {-# INLINE whenM #-}
 
+partitionEithersNEL :: NonEmpty (Either a b) -> These (NonEmpty a) (NonEmpty b)
+partitionEithersNEL (h :| es) = case bimap NEL.nonEmpty NEL.nonEmpty $ partitionEithers es of
+    (Nothing, Nothing) -> either (This . pure) (That . pure) h
+    (Just as, Nothing) -> This as
+    (Nothing, Just bs) -> That bs
+    (Just as, Just bs) -> These as bs
```
fosskers: Sadly missing elsewhere.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -101,11 +103,11 @@ runMiner v mr = do
         pure $ localTest gen miners
 
     powMiner :: IO (BlockHeader -> IO BlockHeader)
-    powMiner = case _configRemoteMiners conf of
-        [] -> pure $ localPOW v
-        rs -> do
+    powMiner = case NEL.nonEmpty $ _configRemoteMiners conf of
```
mightybyte: Why are we doing this?  It doesn't look like we gain anything here because we're still casing.  The fact that we have to add the partition function below seems to make this a net loss in my mind.

fosskers: It's convenient (in terms of type safety) to have the following signature on the other end of this:
```haskell
remoteMining :: Manager -> NESet BaseUrl -> BlockHeader -> IO BlockHeader
```
It means we won't be wasting calls on potentially duplicate endpoints. And since `powMiner` has the signature:
```haskell
    powMiner :: IO (BlockHeader -> IO BlockHeader)
```
these container conversions are only performed once.

fosskers: Give me a moment. I just improved this, and I'm going to improve it further.

fosskers: Give me a moment, I just improved this, and I'm going to improve it further.

src/Chainweb/Miner/Config.hs
```diff
@@ -39,7 +42,7 @@ makeLenses ''MinerCount
 data MinerConfig = MinerConfig
     { _configTestMiners :: !MinerCount
     , _configMinerInfo :: !MinerInfo
-    , _configRemoteMiners :: ![HostAddress]
+    , _configRemoteMiners :: !(Set HostAddress)
```
fosskers: To ensure uniqueness from the outset. There is never a reason to target the same miner twice.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -101,11 +105,14 @@ runMiner v mr = do
...
-        rs -> do
+    powMiner = case g $ _configRemoteMiners conf of
+        Nothing -> pure $ localPOW v
+        Just rs -> do
             m <- newManager defaultManagerSettings
-            pure . remoteMining m $ map f rs
+            pure $ remoteMining m rs
+
+    g :: Set HostAddress -> Maybe (NonEmpty BaseUrl)
+    g = fmap (NEL.map f) . NEL.nonEmpty . S.toList
```
larskuhtz: What is the benefit of this? I think I found the code before more readable. Is the new code safer?

src/Chainweb/Miner/Miners.hs
```diff
@@ -96,24 +99,39 @@ submit :<|> poll = client (Proxy :: Proxy MiningAPI)
 -- on a different machine, may be on multiple machines, may be arbitrarily
 -- multithreaded.
 --
-remoteMining :: Manager -> [BaseUrl] -> BlockHeader -> IO BlockHeader
-remoteMining m urls bh = traverseConcurrently_ Par' submission urls >> polling
+-- ASSUMPTION: The contents of the given @NonEmpty BaseUrl@ are unique.
```
larskuhtz: Why not use a non-empty set?

src/Chainweb/Miner/Miners.hs
```diff
@@ -96,24 +99,39 @@ submit :<|> poll = client (Proxy :: Proxy MiningAPI)
...
       where
+        -- TODO Don't bother scheduling retries for `url`s that fail?
         go :: Scheduler IO BlockHeader -> BaseUrl -> IO BlockHeader
-        go sch url = runClientM (poll cid bht) (ClientEnv m url Nothing) >>= \case
-            Left err -> throwM err
-            Right Nothing -> threadDelay 500000 >> go sch url
-            Right (Just new) -> terminateWith sch new
+        go sch url = do
+            -- This prevents scheduled retries from slamming the miners.
+            threadDelay 100000
```
larskuhtz: I think, this provides another argument to eventually move the logic in the module into an independent process. 

* for competitive production mining (with low level hardware) this interface will probably be to slow / reduce mining margins too much. For Testnet02 DA allows up to 6 blocks per chain per second. On 20 chains that 500ms per block. Polling every 100ms reduce the chances to win a block by up to 20%. This doesn't include the overhead of the reset call itself.

* for scenarios where competitive mining isn't a primary concern (if those are relevant at all), making an additional REST call every 100ms still adds additional networking overhead to consensus, possibly reducing the quality of consensus and causing forks.

A mining coordinator that would use the consensus REST API for communicating with consensus and uses a some custom protocol to communicate with the low-level hardware provides miner with more flexibility to find the best trade-offs for them.

src/Chainweb/Chainweb/MinerResources.hs
fosskers: Before this PR, the logic in the `remoteMiner` function in `Miners.hs` was incorrect. `NonEmpty` + `traverseConcurrently` + `These` lets me solve it elegantly.

src/Chainweb/Miner/Miners.hs
fosskers: It used to, before the final version of the PR. Unfortunately `NESet` doesn't have a `Traversable` instance, which is necessary for `traverseConcurrently` below.

src/Chainweb/Miner/Miners.hs
fosskers: Confirming the numbers...

fosskers: Because of the DA "allowance factor" of 3, the fastest (on average) we can expect to mine a block is in 10s (3x faster). The slowest would be 1.5 minutes (3x slower). Ignoring the cost of the call itself, and given that the 100ms wait occurs once **polling** begins (i.e. *after* the work has been submitted), the latest we could find out about the successful mining would be ~99ms after it occurred. Example:

- Millisecond `n`: we checked for a result and got `Nothing`.
- Millsecond `n + 1`: mining succeeds.
- Millisecond n + 100`: we check again, and get a `Just`.

So for the fastest case (10s), the 100ms wait represents a 1% inefficiency in the discovery time. For the slowest case (90s), it represents a 0.11% ineffiency. 

So while it would be strictly more efficient for the miner itself to report its results, the improvement will not be order-of-magnitude.

fosskers: Further, would the miner reporting the results itself back to `chainweb-node` be necessarily faster? Would the handling of that endpoint, plus feeding it back through whatever machinery is necessary to reassociate the new `BlockHeader` with its `Cut` and payload be _less_ than the 100ms here?

fosskers: 100ms of course is a made-up number - we can set that to whatever we wish.

larskuhtz: The mining coordinator would in any case has to the block to the cutdb. Depending on how it's implemented that would be via a Haskell function call or the reset API. So that adds latency, but since its push (and not poll), it should be in the order of microseconds or at most a single digit ms (for REST).

But that's at the level of the mining coordinator. At the level of core miner, I expect that, on the long run, people will use more low-latency protocols with less overhead. Running a web server on mining hardware that has to reply to an request every few (or tens of) seems a bit wasteful.

larskuhtz: On the other hand I can see how this Rest API could be useful for some intermediate miner, that bundles a number of hardware miners on a box.

PR #362, opened 2019-08-20
Add non-transactional Account Balance lookup

emilypi:
Discussed with @gregorycollins

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -23,6 +23,9 @@ module Chainweb.Pact.PactService
     , execValidateBlock
     , execTransactions
     , initPactService
+    , readCoinAccount
+    , readAccountBalance
+    , readAccountGuard
       -- * For Side-tooling
     , execNewGenesisBlock
     , initPactService'
@@ -47,6 +50,7 @@ import Data.Decimal
 import Data.Default (def)
 import Data.Either
 import Data.Foldable (toList)
+import qualified Data.Map as Map
 import Data.Maybe (isNothing)
 import Data.String.Conv (toS)
 import Data.Text (Text)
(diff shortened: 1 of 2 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -322,6 +329,50 @@ finalizeCheckpointer finalize = do
...
+readCoinAccount
+    :: PactDbEnv'
+      -- ^ pact db backend (sqlite)
+    -> Text
+      -- ^ account name
+    -> IO (Maybe (T2 Decimal (P.Guard (P.Term P.Name))))
+readCoinAccount (PactDbEnv' (P.PactDbEnv pdb pdbv)) a = rows >>= \case
+    Nothing -> return Nothfing
+    Just (P.ObjectMap o) -> case Map.toList o of
+      [(P.FieldKey "balance", b), (P.FieldKey "guard", g)] ->
```
gregorycollins: This is ok for now but it's too strict, i.e. if we decide to add a field to this map we'll have to change this code. Better to do two lookups?

src/Chainweb/Pact/PactService.hs
```diff
@@ -322,6 +329,50 @@ finalizeCheckpointer finalize = do
...
+-- associated with account name
+--
+readCoinAccount
+    :: PactDbEnv'
+      -- ^ pact db backend (sqlite)
+    -> Text
+      -- ^ account name
+    -> IO (Maybe (T2 Decimal (P.Guard (P.Term P.Name))))
+readCoinAccount (PactDbEnv' (P.PactDbEnv pdb pdbv)) a = rows >>= \case
+    Nothing -> return Nothfing
```
gregorycollins: `Nothfing` -- but the compiler will catch this for you :)

emilypi: fat fingered on save probably :)

src/Chainweb/Pact/PactService.hs
emilypi: if we add a field, we probably want this to fail. It involves changing the coin contract, which would be an entirely new contract with an entirely new table too, since `coin` is non-upgradeable.

emilypi:
Closing. I merged this directly into Greg's working branch

PR #363, opened 2019-08-20
Miner takes a Byte out of Life

fosskers:
This PR alters the `MiningAPI` type to expect pre-encoded versions of every argument. This way, our Mining Client (or any such 3rd-party client) needs to know *very* little about how our types are constructed. This should make it even easier to write new clients.

As evidence, the `Chainweb.Miner.Core` module (which powers both our in-process POW mining and our remote POW mining) now has but a single Chainweb import: the `Version` module.

### DIFF
--- a/miner/Miner.hs
+++ b/miner/Miner.hs
@@ -2,7 +2,9 @@
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
@@ -75,32 +77,35 @@
 
 module Main ( main ) where
 
-import BasePrelude hiding (app, option)
+import BasePrelude hiding (Handler, app, option)
 
 import Control.Concurrent.Async (async, race, wait)
 import Control.Concurrent.STM.TMVar
 import Control.Concurrent.STM.TVar (modifyTVar')
 import Control.Error.Util (note)
+import Control.Monad.Except (throwError)
 import Control.Scheduler (Comp(..), replicateWork, terminateWith, withScheduler)
 
 import Data.Aeson (ToJSON)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as M
 import qualified Data.Text as T
-import Data.Tuple.Strict (T2(..))
+import Data.Tuple.Strict (T2(..), T3(..))
 
 import qualified Network.Wai.Handler.Warp as W
 
 import Options.Applicative
 
 import Servant.API
-import Servant.Server (Application, Server, serve)
+import Servant.Server
+
+import qualified System.Random.MWC as MWC
 
 import Text.Pretty.Simple (pPrintNoColor)
 
 -- internal modules
 import Chainweb.BlockHeader
-import Chainweb.Miner.Core (mine, usePowHash)
+import Chainweb.Miner.Core
 import Chainweb.Miner.Miners (MiningAPI)
 import Chainweb.RestAPI.Orphans ()
 import Chainweb.Utils (int, toText)
@@ -112,10 +117,11 @@ import Chainweb.Version
 type API = "env" :> Get '[JSON] ClientArgs
     :<|> MiningAPI
 
+-- TODO Recheck the currying here.
 server :: Env -> Server API
 server e = pure (args e)
-    :<|> liftIO . submit e
-    :<|> (\cid h -> liftIO $ poll e cid h)
+    :<|> (\cid h bites -> liftIO $ submit e cid h bites)
+    :<|> poll e
 
 app :: Env -> Application
 app = serve (Proxy :: Proxy API) . server
@@ -123,7 +129,7 @@ app = serve (Proxy :: Proxy API) . server
 --------------------------------------------------------------------------------
 -- CLI
 
-type ResultMap = Map (T2 ChainId BlockHeight) BlockHeader
+type ResultMap = Map (T2 ChainId BlockHeight) HeaderBytes
 
 data ClientArgs = ClientArgs
     { cores :: Word16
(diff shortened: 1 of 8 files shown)


src/Chainweb/Miner/Core.hs
```diff
@@ -56,60 +73,57 @@ usePowHash Testnet02 f = f $ Proxy @SHA512t_256
...
 
     bufSize :: Int
-    !bufSize = B.length initialBytes
-
-    target :: HashTarget
-    !target = _blockTarget h
+    !bufSize = B.length hbytes
 
     powSize :: Int
-    !powSize = int $ hashDigestSize @a undefined
```
fosskers: `-Widentities` detects this as a spurious conversion, but only when supplied as `fromIntegral`. The over-sight goes undetected with just `int`.

miner/Miner.hs
```diff
@@ -180,26 +185,26 @@ main = do
...
-poll :: Env -> ChainId -> BlockHeight -> IO (Maybe BlockHeader)
-poll (results -> tm) cid h = M.lookup (T2 cid h) <$> readTVarIO tm
-
--- | Cease mining until another `submit` call is made.
---
--- halt :: IO ()
--- halt = undefined
+poll :: Env -> ChainId -> BlockHeight -> Handler HeaderBytes
+poll (results -> tm) cid h = M.lookup (T2 cid h) <$> liftIO (readTVarIO tm) >>= \case
+    Nothing -> throwError $ err404 { errBody = "No results for given ChainId/BlockHeight" }
```
fosskers: Poll failure is now communicated via a 404 instead of a 200 + `Maybe`.

miner/Miner.hs
```diff
@@ -212,36 +217,49 @@ mining e = do
...
-    go :: BlockHeader -> IO BlockHeader
-    go bh = fmap head . withScheduler comp $ \sch ->
-        replicateWork (int . cores $ args e) sch $
-            usePowHash (version $ args e) mine bh >>= terminateWith sch
+    go :: MWC.GenIO -> TargetBytes -> HeaderBytes -> IO HeaderBytes
+    go g tbytes hbytes = fmap head . withScheduler comp $ \sch ->
+        replicateWork (int . cores $ args e) sch $ do
+            -- TODO Be more clever about the Nonce that's picked to ensure that
+            -- there won't be any overlap?
+            n <- Nonce <$> MWC.uniform g
```
fosskers: The fix for a very dumb oversight. All the miners + their threads were starting work on the same nonce!

src/Chainweb/Miner/Core.hs
```diff
@@ -12,32 +14,61 @@
...
-import Chainweb.Difficulty (HashTarget, encodeHashTarget)
-import Chainweb.Utils (int, runGet)
+import Chainweb.BlockHeader (Nonce(..))
 import Chainweb.Version (ChainwebVersion(..))
 
 ---
 
+-- | Encoding of @HashTarget + BlockHeader@.
+newtype Bites = Bites { _bites :: B.ByteString }
+    deriving newtype (MimeRender OctetStream, MimeUnrender OctetStream)
```
fosskers: Perhaps a less stupid name?

emilypi: It's charming, but a little confusing a mnemonic 😄

miner/Miner.hs
```diff
@@ -75,35 +75,37 @@
...
 import Text.Pretty.Simple (pPrintNoColor)
 
 -- internal modules
-import Chainweb.BlockHeader
-import Chainweb.Miner.Core (mine, usePowHash)
+import Chainweb.BlockHeader (BlockHeight, Nonce(..))
+import Chainweb.Miner.Core
 import Chainweb.Miner.Miners (MiningAPI)
-import Chainweb.RestAPI.Orphans ()
-import Chainweb.Utils (int, toText)
```
fosskers: Much fewer concerns! Also in the `Core` module.

src/Chainweb/Miner/Core.hs
fosskers: Changed.

miner/Miner.hs
emilypi: Was that a specified change? That doesn't seem like a correct polling behavior. That's Server or Page Not Found, where a poll is simple "your request succeeded, but no corresponding data was found". It should return a 204 or 200 with empty data

miner/Miner.hs
```diff
@@ -180,26 +183,26 @@ main = do
 
 -- | Submit a new `BlockHeader` to mine (i.e. to determine a valid `Nonce`).
 --
-submit :: Env -> BlockHeader -> IO ()
-submit (work -> w) bh = atomically $
-    isEmptyTMVar w >>= bool (void $ swapTMVar w bh) (putTMVar w bh)
+submit :: Env -> ChainId -> BlockHeight -> WorkBytes -> IO ()
+submit (work -> w) cid bh bs = atomically $
```
emilypi: view patterns! Ackk!

miner/Miner.hs
mightybyte: :+1:

miner/Miner.hs
mightybyte: What kind of failure are we talking about?  If "poll failure" == nothing has been mined yet, then I think I agree with Emily.  That definitely shouldn't be a 404.  200 + `Maybe` would be correct.

fosskers: Let me introspect as to why I changed it in the first place (it original used `Maybe`).

miner/Miner.hs
fosskers: Dealwiddit :laughing:

miner/Miner.hs
fosskers: The change was because I altered the response type:
```haskell
    :<|> "poll" :> Capture "chainid" ChainId
                :> Capture "blockheight" BlockHeight
                :> Get '[OctetStream] HeaderBytes
```
It's not JSON anymore, so I can't return a `Maybe`.

fosskers: We've opted to return 200 in all cases, yielding an empty `ByteString` when lookup fails.

miner/Miner.hs
mightybyte: Oh, that is ick.

mightybyte:
Please don't force push, it obscures the history and is going to get squashed on merge anyway.

PR #365, opened 2019-08-22
run pact services synchronization on all chains concurrently

larskuhtz:


src/Chainweb/Chainweb.hs
```diff
@@ -455,7 +455,7 @@ withChainwebInternal conf logger peer rocksDb dbDir nodeid resetDb inner = do
 
     synchronizePactDb cs cutDb = do
         currentCut <- _cut cutDb
-        mapM_ syncOne $ mergeCutResources $ _cutMap currentCut
+        mapConcurrently_ syncOne $ mergeCutResources $ _cutMap currentCut
```
fosskers: This will use all available threads. Is that desired?

larskuhtz: For 20 chains during initialization, yes. It should be almost a no-op in most cases (we are investigating currently why it isn't currently). Also, if at all it takes longer it should be IO bound.

larskuhtz: But once we go to, say >100, chains, we should use a scheduler.

larskuhtz: @fosskers if you are familiar with the scheduler package and know how to fix this quickly, please do it.

PR #370, opened 2019-08-23
fix savepoint logic of checkpointer

larskuhtz:
There are two bugs in the checkpointer code:

1. `doDiscardBatch` and `doDicard` only roll back updates in a database transaction but do not delete the transaction from the transaction stack. Affected transactions remain on the stack without ever being committed.

2. `execNewBlock`, `execLocal`, and `execNewGenesisBlock` call `eestorecheckpointer`, which puts a new `Block` savepoint on the transaction stack. However these functions (or any subsequent function) dodn't make a call that would commit or release these savepoints. Again, this causes these transactions to remain open forever.

Effectively, the sqlite db is accumulating an unlimited stack of transactions. Only at node shutdown the transactions are rolled back. This has three consequences

1. The ever growing stack of uncommitted nested transactions is a huge memory leak.
2. It causes a performance degradation, because all updates are accumulated in a single big transactions (with many nested transactions). Sqlite is never able to do a definite commit and discard old transaction logs.
3. On shutdown the stack of all open transactions is canceled, which causes a rollback of all transactions (including those inner transactions that have been already committed). The only database updates that are persisted, are the blocks from the very first `rewind` call after the node started. As a consequence a newly started node has to replay all pact transactions since the time it was previously restarted.

It is likely that other observed issues are related to these bugs, like increasing levels of `iowait` with big databases and nodes running out of RAM during db maintenance operations and crashing. But that has yet to be confirmed.

This PR fixes these bugs

* by changing the `discard*` functions to release savepoints after rolling back the database updates in the savepoint and
* by calling `finalizeCheckpointer discard` for `execNewBlock`, `execLocal`, and `execNewGenesisBlock`.

Additionally, it simplifies the control flow a bit by making `rewindTo` return it's result instead of passing it to a provided continuation.

It also simplifies the error handling in `rewindTo` and eliminates the recursive call within an exception handler.

Finally, the PR adds some documentation, that clarifies some invariants about the savepoint state machine.


larskuhtz:
The PR is currently tested on devnet. It is deployed on `usCommonDevNodes` and `devBootstrapNodes`. `euCommonDevNodes` are currently in the process of being converted to the code of this PR. This take a long time, because, due to the bug, those nodes must replay all pact transactions since they were started last, which is close to genesis.

larskuhtz:
image

Memory consumption one devnet while nodes are successively migrated to the code of this PR.

At around 00:00 a pact transaction load of about 600 tx/min was added to the network.

src/Chainweb/Pact/PactService.hs
```diff
@@ -607,38 +620,49 @@ withParentBlockData phe action = action
...
-    -> (PactDbEnv' -> PactServiceM cas a)
-    -> PactServiceM cas a
-rewindTo mb act = do
+        -- ^ The block height @height@ to which to restore and the parent header
+        -- @parentHeader@.
+        --
+        -- It holds that @(_blockHeight parentHeader == pred height)@
+
+    -> PactServiceM cas PactDbEnv'
+rewindTo mb = do
```
larskuhtz: Rewind used to take continuation. That continuation was wrapped in the exception handler bracket in `playFork`. But only handling lookup failures for last block, the continuation isn't in that bracket any more and `rewindTo` can become a "normal" function that returns a result directly.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -71,7 +71,7 @@ initRelationalCheckpointer'
 initRelationalCheckpointer' bstate sqlenv loggr gasEnv = do
     let dbenv = BlockDbEnv sqlenv loggr
     db <- newMVar (BlockEnv dbenv bstate)
-    runBlockEnv db $ initSchema >> vacuumDb
+    runBlockEnv db $ initSchema >> vacuumDb -- TODO: remove?
```
emmanueldenloye: Do you think it might be worthwhile to consider putting on auto-vacuum?

src/Chainweb/Pact/PactService.hs
```diff
@@ -648,51 +672,51 @@ rewindTo mb act = do
 
     failNonGenesisOnEmptyDb = fail "impossible: playing non-genesis block to empty DB"
 
-    playFork cp bhdb payloadDb newH parentHash (lastBlockHeight, lastHash) =
-      flip catches exHandlers $ do
-          parentHeader <- liftIO $ lookupM bhdb parentHash
```
larskuhtz: If `lookupM bdhdb parentHash` throw an `TreeDbKeyNotFound` exception, the handler below would call call `playFork` again with the same value for `parentHash`, which would result in an infinite recursion.

The new code only recurses on lookup failures for `lastHash`, in `findValidParent`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -648,51 +672,51 @@ rewindTo mb act = do
...
+        V.mapM_ (fastForward payloadDb) newBlocks
+        -- play new block
+        restoreCheckpointer (Just (newH, parentHash))
       where
-        exHandlers = [Handler handleTreeDbFailure, Handler handlePayloadFailure]
-        handleEx :: forall e . Exception e => e -> PactServiceM cas a
-        handleEx e =
-              (liftIO $ getBlockParent cp (lastBlockHeight, lastHash)) >>= \case
-                Nothing -> throwM e
-                Just hash -> playFork cp bhdb payloadDb newH parentHash
```
larskuhtz: I usually try to avoid recursion (or any larger long-running operations) in exception handers, because, I think, those are executed under a mask. Also when the code in the exception handler itself throws and has non-trivial exception handling, it seems that the overall behavior can become quite complex to understand.

But maybe I am just to paranoid :-)

src/Chainweb/Pact/PactService.hs
```diff
@@ -538,9 +549,13 @@ execNewGenesisBlock
 execNewGenesisBlock miner newTrans = withDiscardedBatch $ do
     pdbenv <- restoreCheckpointer Nothing
     results <- execTransactions Nothing miner newTrans pdbenv
+    finalizeCheckpointer discard
     return $! toPayloadWithOutputs miner results
 
-
+-- |
+-- Precondition: restoreCheckpointer
```
larskuhtz:
```suggestion
```

`restoreCheckpointer` is not a precondition but called within the function in the call to `rewindTo`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -464,45 +479,37 @@ minerReward bh = do
 
 -- | Note: The BlockHeader param here is the PARENT HEADER of the new
 -- block-to-be
+--
+-- Precondition: restoreCheckpointer
```
larskuhtz:
```suggestion
```

`restoreCheckpointer` is not a precondition but provided below as a postcondition of the call to `rewindTo`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -464,45 +479,37 @@ minerReward bh = do
...
+    liftIO $ mpaProcessFork mpAccess parentHeader
+    liftIO $ mpaSetLastHeader mpAccess parentHeader
+    cp <- view (psCheckpointEnv . cpeCheckpointer)
+    let validate = validateChainwebTxsPreBlock pdbenv cp
+    newTrans <- liftIO $
+        mpaGetBlock mpAccess validate bHeight pHash parentHeader
+    -- locally run 'execTransactions' with updated blockheight data
+    results <- withParentBlockData parentHeader $
+        execTransactions (Just pHash) miner newTrans pdbenv
+    finalizeCheckpointer discard
```
larskuhtz: This call is new. Without it the function would leave an `Block` savepoint on the sqlite transaction stack.

src/Chainweb/Pact/PactService.hs
```diff
@@ -538,9 +549,13 @@ execNewGenesisBlock
 execNewGenesisBlock miner newTrans = withDiscardedBatch $ do
     pdbenv <- restoreCheckpointer Nothing
     results <- execTransactions Nothing miner newTrans pdbenv
+    finalizeCheckpointer discard
```
larskuhtz: This call is new. Without it the function would leave an Block savepoint on the sqlite transaction stack.

src/Chainweb/Pact/PactService.hs
```diff
@@ -648,51 +672,51 @@ rewindTo mb act = do
...
-        exHandlers = [Handler handleTreeDbFailure, Handler handlePayloadFailure]
-        handleEx :: forall e . Exception e => e -> PactServiceM cas a
-        handleEx e =
-              (liftIO $ getBlockParent cp (lastBlockHeight, lastHash)) >>= \case
-                Nothing -> throwM e
-                Just hash -> playFork cp bhdb payloadDb newH parentHash
-                                      (pred lastBlockHeight, hash)
-        handleTreeDbFailure e@(_ :: TreeDbException BlockHeaderDb) = handleEx e
-        handlePayloadFailure e@(_ :: PayloadNotFoundException) = handleEx e
+        findValidParent height hash = liftIO (lookup bhdb hash) >>= \case
```
emmanueldenloye: How could have the previous code bitten us in the butt?

larskuhtz: I guess it never threw an exception. If it did, the node probably would have run out of memory and crashed. Maybe it happened, and we didn't notice what was the reason.

src/Chainweb/Pact/PactService.hs
```diff
@@ -530,6 +537,10 @@ withDiscardedBatch act = bracket start end (const act)
 
 
 -- | only for use in generating genesis blocks in tools
+--
+-- Precondition: restoreCheckpointer
```
larskuhtz:
```suggestion
```

src/Chainweb/Pact/PactService.hs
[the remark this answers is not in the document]
larskuhtz: Yeah, something like:

```haskell
withRestoreCheckpointerDiscard target = bracket (restoreCheckpointer target) (const $ finalizeCheckpointer discard)
```

That would also be more exception safe.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -155,11 +155,20 @@ doSave dbenv hash = runBlockEnv dbenv $ do
...
 doDiscard :: Db -> IO ()
 doDiscard dbenv = runBlockEnv dbenv $ do
     clearPendingTxState
     rollbackSavepoint Block
 
+    -- @ROLLBACK TO n@ only rolls back updates up to @n@ but doesn't remove the
+    -- savepoint. In order to also pop the savepoint from the stack we commit it
+    -- (as empty transaction). <https://www.sqlite.org/lang_savepoint.html>
+    --
+    commitSavepoint Block
```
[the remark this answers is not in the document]
larskuhtz: My understanding of the code was that `rollBackSavepoint` and `commitSavepoint` closely resemble behavior of the respective sqlite commands and the `do*` commands at our own transaction DSL on top of that.

It may confuse people if we use sqlite terminology with different semantics. Not sure, though.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
larskuhtz: Not sure, I don't understand the details well enough. I think, we should do some more testing around that.

In past test runs `vacuumDb` was causing nodes to run out of memory and crash for large dbs. But maybe that was related to the savepoint nesting bug. So we should run more tests first.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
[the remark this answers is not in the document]
emmanueldenloye: Ok! that's a great argument for leaving it as is.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
emilypi: `VACUUM` makes a copy of the main DB to a temporary file and reloads - this probably copies unresolved snapshots as well, OOM'ing the node, since we were accumulating GB's of unresolved transactions. It's probably fine now that we've solved that problem if we can prove that the 2xdb size upper bound won't cause the node to run out of memory. `AUTOVACUUM` only removes stale pages, so it's probably not what we want on startup for a checkpointer.

src/Chainweb/Pact/PactService.hs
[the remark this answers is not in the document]
emilypi: Seconding that idea.

src/Chainweb/Pact/PactService.hs
```diff
@@ -648,51 +669,51 @@ rewindTo mb act = do
...
-          -- play fork blocks
-          V.mapM_ (fastForward payloadDb) newBlocks
-          -- play new block
-          restoreCheckpointer (Just (newH, parentHash)) >>= act
+    playFork cp bhdb payloadDb newH parentHash (lastBlockHeight, lastHash) = do
+        parentHeader <- liftIO $ lookupM bhdb parentHash
+        lastHeader <- findValidParent lastBlockHeight lastHash
+
+        (!_, _, newBlocks) <-
+            liftIO $ collectForkBlocks bhdb lastHeader parentHeader
```
emilypi: @gregorycollins one question we had from last night is this line here and what we're doing by forcing the db ref, and only retrieving the new blocks. Could either you or @emmanueldenloye annotate this with some docs?

src/Chainweb/Pact/PactService.hs
```diff
@@ -765,7 +785,7 @@ applyPactCmd isGenesis dbEnv cmdIn miner mcache v = do
 
     -- cvt from Command PayloadWithTexts to Command ((Payload PublicMeta ParsedCode)
     let !cmd = payloadObj <$> cmdIn
-    T2 result mcache' <- liftIO $ if isGenesis
+    T2 !result mcache' <- liftIO $ if isGenesis
```
emilypi: Is this necessary? `T2` is already strict in its arguments.

src/Chainweb/Pact/PactService.hs
emmanueldenloye: Would you rather it be done now or in follow up PR?

emmanueldenloye: Actually, I have taken a look. I feel pretty confident about removing the bang.

PR #371, opened 2019-08-23
Use TH to read miner rewards

mightybyte:


src/Chainweb/Pact/PactService.hs
```diff
@@ -426,15 +426,14 @@ readAccountGuard pdb account
...
     :: HasChainGraph v
-    => v -> IO (HashMap BlockHeight P.ParsedDecimal)
+    => v -> (HashMap BlockHeight P.ParsedDecimal)
 readRewards v = do
-    rs <- LBS.readFile "rewards/miner_rewards.csv"
-    case CSV.decode CSV.NoHeader rs of
-      Left e -> internalError
+    case CSV.decode CSV.NoHeader (toS rawMinerRewards) of
+      -- TODO Not sure what to do here
+      Left e -> error
```
emilypi: `error` is perfectly fine. We want this to be a 'fail fast and hard' scenario. We can also now `{-# INLINE readRewards #-}`

src/Chainweb/Miner/Rewards.hs
```diff
@@ -0,0 +1,9 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Chainweb.Miner.Rewards where
+
+import Data.ByteString (ByteString)
+import Data.FileEmbed
+
+rawMinerRewards :: ByteString
+rawMinerRewards = $(embedFile "rewards/miner_rewards.csv")
```
emilypi: We can just add this to `Chainweb.Pact.Miner`

mightybyte:
The thing I’m worried about with error is whether it’s strict enough to fail as early as we need it to.

src/Chainweb/Miner.hs
```diff
@@ -125,6 +141,37 @@ toMinerData = MinerData . encodeToByteString
...
+        . V.map formatRow
+        $ vs
+  where
+    formatRow :: (Word64, Double) -> (BlockHeight, ParsedDecimal)
+    formatRow (!a,!b) =
+      let
+        !n = v ^. chainGraph . to (int . size)
+        !m = fromRational $ toRational b
+      in (BlockHeight a, ParsedDecimal $ m / n)
+{-# INLINE readRewards #-}
```
larskuhtz: Is this used more than once during the lifetime of a node? Since this seems somewhat expensive, maybe we shouldn't inline it but memoize it instead?

If it's called only once, it's fine. Still, it maybe better to not inline it.

src/Chainweb/Miner.hs
```diff
@@ -125,6 +141,37 @@ toMinerData = MinerData . encodeToByteString
...
+
+-- | Read in the reward csv via TH for deployment purposes
+--
+rawMinerRewards :: ByteString
+rawMinerRewards = $(embedFile "rewards/miner_rewards.csv")
+{-# INLINE rawMinerRewards #-}
```
larskuhtz: I don't think this should be inlined. How big is the file? The body of this function contains the whole spread sheet, doesn't it?

So, instead, I'd rather add an explicit `{-# NOINLINE #-}` pragma, in order to make sure that we don't copy the data to each call site in the binaries.

src/Chainweb/Miner.hs
emilypi: It's only called once, so it should be okay

src/Chainweb/Miner.hs
emilypi: Good point

emilypi: I'll do that, thanks. I was being overzealous.

larskuhtz:
I wonder how the csv file should be included in the cabal file. First, I though it should be under `data-files`, but I think, those are usually runtime dependencies. So maybe it should be listed under `extra-source-files` instead? Not sure.

emilypi:
@larskuhtz it looks like it should be under `extra-source-files`

mightybyte:
I don't think it should be done with cabal infrastructure.  That still requires that certain things exist at runtime.  This should be embedded in the exe.

PR #373, opened 2019-08-23
Pact Test + Log Message  Cleanup

emilypi:


test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -318,7 +318,7 @@ testPactCtxSQLite v cid cdbv bhdb pdb sqlenv = do
     evalPactServiceM ctx (initialPayloadState v cid)
     return ctx
   where
-    loggers = pactTestLogger True
+    loggers = pactTestLogger False
```
fosskers: This is quite subtle. Can this documented, or even made into a flag?

PR #375, opened 2019-08-23
Add withCheckpointer bracket

larskuhtz:


src/Chainweb/Pact/PactService.hs
```diff
@@ -680,7 +710,11 @@ execValidateBlock
     -> PactServiceM cas PayloadWithOutputs
 execValidateBlock currHeader plData =
     -- TODO: are we actually validating the output hash here?
-    withBatch $ rewindTo mb >>= playOneBlock currHeader plData
+    withBatch $ do
+        rewindTo mb
+        withCheckpointer mb $ \pdbenv -> do
```
emmanueldenloye: Why didn't you use `withCheckpointerRewind` here?

larskuhtz: just missed it :-)

PR #381, opened 2019-08-27
Minist-Evoluo

fosskers:
This PR improves communication between the Coordination logic (found currently in `Chainweb.Miner.Coordinator`) and `chainweb-miner`. Specifically, it removes the need for a constant polling of results. Instead, the `chainweb-miner` knows how to send results back to `chainweb-node` itself. This requires a new endpoint, and some refactoring of the `Coordinator` logic.

The act of *starting* mining work and *consuming* mining results are now concurrent threads of activity, whereas previously this was one block of sequential logic. The *starting* work is represented by the `working` function. The *consuming* is represented by the `publishing` function.

For cases where we use `chainweb-miner`, the call graph looks as below. Green nodes represent shared data. Note: `type Prev = T2 PayloadWithOutputs PrevBlock`. This `TVar` is necessary, as data that used to naturally be in scope for the entire sequential flow no longer is.

remote-mining

For in-process POW mining and time-based test mining, we have the following. Here instead we run a function `listener` in a side-thread (likely launched in the same location as `working`) which informs `publishing` of the results.

local-mining


src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -66,56 +73,72 @@ withMinerResources
...
-        , _minerResConfig = conf
-        }
+    | otherwise = do
+        tms <- newTVarIO Nothing
+        inner . Just $ MinerResources
+            { _minerResLogger = logger
+            , _minerResNodeId = nid
+            , _minerResCutDb = cutDb
+            , _minerResConfig = conf
+            , _minerResState = tms
```
fosskers: Similar to `CutDb`, a new type `MiningState` is now shared between the `working` and `publishing` functions in the `Coordinator` module.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -167,6 +164,54 @@ filterAdjustments newBh as = case window $ _blockChainwebVersion newBh of
...
+-- originated, and publishes it to the `Cut` network.
+--
+publishing :: LogFunction -> TVar (Maybe MiningState) -> CutDb cas -> BlockHeader -> IO ()
+publishing lf tp cdb bh = do
+    -- Reassociate the new `BlockHeader` with the current `Cut`, if possible.
+    -- Otherwise, return silently.
+    --
+    c <- _cut cdb
+    readTVarIO tp >>= \case
+        Nothing -> pure ()  -- TODO Throw error?
```
fosskers: Should we `Warn` here?

src/Chainweb/Miner/RestAPI.hs
```diff
@@ -0,0 +1,65 @@
...
+
+---
+
+-- -----------------------------------------------------------------------------
+-- Mining Work Submission
+
+-- | Shared between the `Chainweb.Miner.Miners.remoteMining` function and the
+-- /chainweb-miner/ executable.
+--
+type MiningSubmissionApi = "submit" :> ReqBody '[OctetStream] WorkBytes :> Post '[JSON] ()
```
fosskers: Note: this is used to call a `chainweb-miner`, and hence doesn't follow the usual Chainweb URL structure. It is defined here for optimal code sharing.

src/Chainweb/Miner/RestAPI/Client.hs
```diff
@@ -0,0 +1,43 @@
...
+-- Mining Work Submission
+
+submitClient :: WorkBytes -> ClientM ()
+submitClient = client (Proxy @MiningSubmissionApi)
+
+-- -----------------------------------------------------------------------------
+-- Mining Results
+
+solvedClient :: ChainwebVersion -> HeaderBytes -> ClientM ()
+solvedClient (FromSing (SChainwebVersion :: Sing v)) = client (miningResultApi @v)
```
fosskers: Called by `chainweb-miner`.

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -0,0 +1,66 @@
...
+    lf :: LogFunction
+    lf = logFunction $ _minerResLogger mr
+
+    tms :: TVar (Maybe MiningState)
+    tms = _minerResState mr
+
+    cdb :: CutDb cas
+    cdb = _minerResCutDb mr
+
+-- TODO Use type-level `CutDbT`?
```
fosskers: ^

src/Chainweb/Miner/RestAPI.hs
```diff
@@ -0,0 +1,51 @@
...
+
+import Chainweb.Miner.Core (HeaderBytes)
+import Chainweb.RestAPI.Utils (ChainwebEndpoint(..), Reassoc, SomeApi(..))
+import Chainweb.Version
+
+import Data.Singletons
+
+---
+
+-- TODO Use `NetworkID` to be consistent with the other subnetworks?
```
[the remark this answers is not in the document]
larskuhtz: I don't think this matters at this point. This API shouldn't be on the public interface and also should be removed from the Chainweb-node API in future versions.

src/Chainweb/RestAPI/Orphans.hs
```diff
@@ -431,3 +432,6 @@ instance ToSchema ChainwebVersion where
 
 instance ToSchema MerkleLogHash where
   declareNamedSchema _ = pure $ NamedSchema (Just "MerkleLogHash") mempty
+
+instance ToSchema HeaderBytes where
+  declareNamedSchema _ = pure $ NamedSchema (Just "HeaderBytes") mempty
```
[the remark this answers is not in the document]
fosskers: To be done in follow-ups.

miner/Miner.hs
```diff
@@ -82,61 +82,71 @@ import BasePrelude hiding (Handler, app, option)
...
 app :: Env -> Application
 app = serve (Proxy :: Proxy API) . server
 
 --------------------------------------------------------------------------------
 -- CLI
 
-type ResultMap = Map (T2 ChainId BlockHeight) HeaderBytes
+-- | Newtype'd so that I can provide a custom `ToJSON` instance.
+--
+newtype URL = URL { _url :: BaseUrl }
```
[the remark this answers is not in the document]
fosskers: I'll get this in a follow-up.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -167,6 +164,54 @@ filterAdjustments newBh as = case window $ _blockChainwebVersion newBh of
...
+  where
+    -- | Even if a `Cut` is not the original `Cut` from which a `BlockHeader`
+    -- originated, they might still be compatible. If the highest header in the
+    -- Cut matches the parent of new header (for the chain that they share),
+    -- then this Cut is still compatible, and we haven't wasted mining effort.
+    --
+    compatibleCut :: Cut -> Bool
+    compatibleCut c = case HM.lookup (_blockChainId bh) $ _cutMap c of
+        Nothing -> False
+        -- TODO Is height a sufficient check, or should it go by parent hash?
```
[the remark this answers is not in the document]
larskuhtz: 4dcd935be9a77969d303a9d01b9190160cb8d7f7 seems to fix the issue.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -167,6 +164,54 @@ filterAdjustments newBh as = case window $ _blockChainwebVersion newBh of
...
+--
+publishing :: LogFunction -> TVar (Maybe MiningState) -> CutDb cas -> BlockHeader -> IO ()
+publishing lf tp cdb bh = do
+    -- Reassociate the new `BlockHeader` with the current `Cut`, if possible.
+    -- Otherwise, return silently.
+    --
+    c <- _cut cdb
+    readTVarIO tp >>= \case
+        Nothing -> pure ()  -- TODO Throw error?
+        Just (MiningState pl p) -> when (compatibleCut c && samePayload pl) $ do
```
larskuhtz: I assume that `samePayload` fails rarely enough, so that it doesn't pay of to cache the payload and cache in the rewards, even if the payload is not the most recent?

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -167,6 +164,54 @@ filterAdjustments newBh as = case window $ _blockChainwebVersion newBh of
...
+            -- Publish the new Cut into the CutDb (add to queue).
+            --
+            c' <- monotonicCutExtension c bh
```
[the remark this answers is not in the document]
larskuhtz: Tests with 4dcd935 indicate that about 15% of all mined cuts are mined on outdated cuts. 

There were between 4 and 5 cases per minute in the network at an overall block rate of 2.5 blocks per chain per minute for 10 chains.

That number seems a bit high, because it should depend only on local latencies and not on p2p network latencies.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -66,56 +73,72 @@ withMinerResources
...
-        ProofOfWork -> powMiner
+    loop :: IO () -> IO ()
+    loop = runForever lf "Chainweb.Miner.Coordinator.working"
+
+    -- | This thread will sit silently forever without burning resources when a
+    -- remote miner is being used, since the `TMVar` will never be written to in
+    -- that case.
+    --
+    listener :: TMVar BlockHeader -> IO ()
+    listener tmv = atomically (takeTMVar tmv) >>= publishing lf tms cdb >> listener tmv
```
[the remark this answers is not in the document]
larskuhtz: Cf.  a5aac999 for a fix.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -180,92 +225,12 @@ estimatedHashes (PrevBlock p) b = floor $ (d % t) * 1000000
...
-                    creationTime
-                    p
-
-            newHeader <- mine candidateHeader
-
-            -- create cut with new block
-            --
-            -- This is expected to succeed, since the cut invariants should
-            -- hold by construction
-            --
```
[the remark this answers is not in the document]
fosskers: I must have overlooked that during my copy-pasting, it wasn't an intentional removal.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -167,6 +164,54 @@ filterAdjustments newBh as = case window $ _blockChainwebVersion newBh of
             limit = bool (_blockHeight newBh - wh) 0 (_blockHeight newBh < wh)
         in HM.filter (\(T2 h _) -> h > limit) as
 
+-- | THREAD: Accepts a "solved" `BlockHeader` from some external source (likely
+-- a remote mining client), reassociates it with the `Cut` from which it
+-- originated, and publishes it to the `Cut` network.
+--
+publishing :: LogFunction -> TVar (Maybe MiningState) -> CutDb cas -> BlockHeader -> IO ()
+publishing lf tp cdb bh = do
```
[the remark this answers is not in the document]
fosskers: This will change yet again in a follow-up.

src/Chainweb/Chainweb/MinerResources.hs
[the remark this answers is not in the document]
fosskers: Ah, I had naively assumed that `publishing` would never throw.

src/Chainweb/Miner/Coordinator.hs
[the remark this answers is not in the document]
fosskers: And these are running in "in-process" mode - the remote miner isn't involved.

test/Chainweb/Test/Utils.hs
```diff
@@ -419,12 +422,13 @@ withChainServer
...
     => FromJSON t
     => PayloadCas cas
+    => Logger logger
     => ChainwebServerDbs t logger cas
     -> (ClientEnv -> IO a)
     -> IO a
 withChainServer dbs f = W.testWithApplication (pure app) work
   where
-    app = chainwebApplication (Test singletonChainGraph) dbs
+    app = chainwebApplication (Test singletonChainGraph) dbs undefined
```
larskuhtz: please use `error "undefined called at Chainweb.Test.Utils.withChainServer"` instead. Also consider adding a `HasCallStack` constraint.

test/Chainweb/Test/Utils.hs
```diff
@@ -544,19 +548,26 @@ withChainwebTestServer tls v appIO envIO test = withResource start stop $ \x ->
...
-    :: Show t
+    :: forall t cas
+    .  Show t
     => ToJSON t
     => FromJSON t
     => PayloadCas cas
     => Bool
     -> ChainwebVersion
-    -> IO (ChainwebServerDbs t logger cas)
+    -> IO (ChainwebServerDbs t GenericLogger cas)
```
larskuhtz: Why is this change needed?

test/Chainweb/Test/Utils.hs
fosskers: Oh crap, good catch, that was definitely just an oversight.

test/Chainweb/Test/Utils.hs
fosskers: To massage the types in the right way. I remember it getting very mad at me that it couldn't resolve a `logger` instance. Let me double-check.

fosskers: Yes, it was to get things compiling. I believe that in the old case there was no `Logger =>` constraint on that `logger` type variable, but now there is. Some other test utility function was forcing this to be `()`, which clearly doesn't have the instance. So I hardcode it to `GenericLogger` here, since that's what in scope.

PR #389, opened 2019-08-29
Tests for pacts chain-data built-in

larskuhtz:


larskuhtz:
@emilypi there are two failing tests: the `block-time` test, but also the `gas-fee` tests. 

`gas-fee` is listed as a field of `chain-data` in the [act documentation for `chain-data](https://pact-language.readthedocs.io/en/stable/pact-functions.html#chain-data). But it seems to not to be supported on chainweb.

test/Chainweb/Test/Pact/ChainData.hs
```diff
@@ -0,0 +1,311 @@
...
+
+-- | cf. <https://pact-language.readthedocs.io/en/stable/pact-functions.html#chain-data>
+--
+tests :: ScheduledTest
+tests = testGroupSch label
+    [ chainDataTest "block-time"
+    , chainDataTest "block-height"
+    , chainDataTest "gas-limit"
+    , chainDataTest "gas-price"
+    , chainDataTest "gas-fee"
```
larskuhtz:
```suggestion
```

PR #391, opened 2019-08-29
Add fix for `chain-data` block time

emilypi:
This makes sure we use a homogeneous time representation derived from the _parent header_. `chain-data` info now consists of the following: 

- Block time from parent block
- Previous Block hash from previous block
- Block height of the current block


src/Chainweb/Pact/PactService.hs
```diff
@@ -585,41 +585,28 @@ logError = logg "ERROR"
...
-    & locally (psPublicData . P.pdBlockHeight) (const bh)
-    & locally (psPublicData . P.pdPrevBlockHash) (const $ toText ph)
-  where
-    (BlockHeight !bh) = succ $ _blockHeight phe
-    (BlockHash !ph) = _blockHash phe
+    go t = t
+      { P._pdBlockHeight = succ bh
+      , P._pdBlockTime = bt
+      , P._pdPrevBlockHash = toText ph
+      }
```
emilypi: this is slightly more economical than using 3 calls to `locally`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -631,13 +618,23 @@ playOneBlock
...
     isGenesisBlock = isGenesisBlockHeader currHeader
-    pBlock = if isGenesisBlock then Nothing else Just bParent
+
+    go m txs =
+      if isGenesisBlock
+      then execTransactions Nothing m txs pdbenv
+      else do
+        bhDb <- asks _psBlockHeaderDb
+        ph <- liftIO $! lookupM bhDb (_blockParent currHeader)
+        withBlockData ph $! execTransactions (Just bParent) m txs pdbenv
```
emilypi: make sure we only call `withBlockData` on non-genesis blocks

PR #394, opened 2019-08-30
implement target and current block time validation

larskuhtz:
* [x] Add new `_blockEpochStart` field to `BlockHeader`,
* [x] use `_blockEpochStart` for target computation and DA,
* [x] improve some documentation in `BlockHeader` and `BlockHeader.Validation`,
* [x] improve some function name in Chainweb.BlockHeader.Validation`,
* [x] include parent header into `ValidationFailureException`,
* [x] implement pow target validation (including DA),
* [x] implement validation that block creation time is current, and
* [x] make list of 'ValidationFailureType's complete.


src/Chainweb/BlockHeader.hs
```diff
@@ -286,6 +297,76 @@ encodeBlockCreationTime (BlockCreationTime t) = encodeTime t
...
+
+decodeEpochStartTime :: MonadGet m => m EpochStartTime
+decodeEpochStartTime = EpochStartTime <$> decodeTime
+
+isEpochStartHeight :: HasChainwebVersion v => v -> BlockHeight -> Bool
+isEpochStartHeight v h = case window (_chainwebVersion v) of
+    Nothing -> False
+    Just (WindowWidth n)
+        | h == 0 -> True
+        | int h < n && int h `mod` 10 == (0 :: Int) -> True
```
fosskers: For those watching: this is how "Early DA" occurs.

src/Chainweb/BlockHeader.hs
```diff
@@ -286,6 +297,76 @@ encodeBlockCreationTime (BlockCreationTime t) = encodeTime t
...
+powTarget p (BlockCreationTime bt) = case window ver of
+    Nothing -> maxTarget
+    Just (WindowWidth w)
+        | isLastInEpoche p ->
+            adjust ver (windowSize w) (t .-. _blockEpochStart p) (_blockTarget p)
+        | otherwise -> _blockTarget p
+  where
+    t = EpochStartTime bt
+    ver = _blockChainwebVersion p
+    windowSize w = WindowWidth $ if int (_blockHeight p) <= w then 10 else w
```
fosskers: Is `10` here correct? Isn't it supposed to be the `w / 10`?

src/Chainweb/BlockHeader.hs
```diff
@@ -375,6 +458,12 @@ data BlockHeader :: Type where
             -- of the block reward and any transactional fees, but this is not
             -- enforced. This information is merely informational.
 
+        , _blockEpochStart :: {-# UNPACK #-} !EpochStartTime
+            -- ^ The start time of the current difficulty adjustment epoch.
+            -- Epochs devide the sequence of blocks in the chain into continues
```
fosskers:
```suggestion
            -- Epochs divide the sequence of blocks in the chain into continuous
```

src/Chainweb/TreeDB/Difficulty.hs
```diff
@@ -1,106 +0,0 @@
...
-          --
-          --   1. It is even.
-          --   2. It is large enough to be divisible by 10.
-          --
-          -- By splitting the early adjustments into 10 chunks, we guarantee
-          -- that the first adjustment comes quickly. Experimentally, it also
-          -- has the effect of forcing the network into a consensus almost
-          -- immediately - usually between 30s and a minute for 40 machines
-          -- spread across the earth.
-          | height < n -> n `div` 10
```
fosskers: Related to the above comment.

src/Chainweb/BlockHeader.hs
larskuhtz: I'll add a code comment. Thanks for pointing this out @fosskers

src/Chainweb/BlockHeader.hs
```diff
@@ -286,6 +297,76 @@ encodeBlockCreationTime (BlockCreationTime t) = encodeTime t
...
+isEpochStartHeight v h = case window (_chainwebVersion v) of
+    Nothing -> False
+    Just (WindowWidth n)
+        | h == 0 -> True
+        | int h < n && int h `mod` 10 == (0 :: Int) -> True
+        | int h `mod` n == 0 -> True
+        | otherwise -> False
+{-# INLINE isEpochStartHeight #-}
+
+isLastInEpoche :: BlockHeader -> Bool
```
gregorycollins: `s/Epoche/Epoch`

fosskers: Oh je.

larskuhtz: Ahhh. Is there a way to configure my keyboard (on hardware level) to prevent me from writing "epoche"?

src/Chainweb/BlockHeader.hs
```diff
@@ -318,19 +318,29 @@ encodeEpochStartTime (EpochStartTime t) = encodeTime t
...
-isEpochStartHeight v h = case window (_chainwebVersion v) of
+-- | During the first epoch after genesis there are 10 extra difficulty
+-- adjustments. This is to account for rapidly changing total hash power in this
+-- early stages of the network.
+--
+effectiveWindow :: BlockHeader -> Maybe WindowWidth
+effectiveWindow h = WindowWidth <$> case window ver of
+    Nothing -> Nothing
+    Just (WindowWidth w)
+        | int (_blockHeight h) <= w -> Just $ max 1 $ w `div` 10
```
fosskers: Yes, good catch, there's no such thing as a Window with width `0`.

src/Chainweb/BlockHeader.hs
```diff
@@ -318,19 +318,29 @@ encodeEpochStartTime (EpochStartTime t) = encodeTime t
...
+--
+isLastInEpoch :: BlockHeader -> Bool
+isLastInEpoch h = case effectiveWindow h of
     Nothing -> False
-    Just (WindowWidth n)
-        | h == 0 -> True
-        | int h < n && int h `mod` 10 == (0 :: Int) -> True
-        | int h `mod` n == 0 -> True
+    Just (WindowWidth w)
+        | int (_blockHeight h) `mod` w == 0 -> True
```
fosskers: The last `BlockHeader` in a particular epoch is the one *before* the "mod point". So for an effective window width of 12:
- 0 to 11 form the first epoch/window
- 12 - 23 form the second epoch

Therefore 11 and 23 should report `True` here, not `12` and `24` which is what would occur with `mod w == 0`.

fosskers: DA is expected to occur at the *first* block of the new epoch, so at height 12 and 24.

src/Chainweb/BlockHeader.hs
```diff
@@ -341,16 +351,15 @@ powTarget
...
         -- ^ POW target of new block
-powTarget p (BlockCreationTime bt) = case window ver of
+powTarget p (BlockCreationTime bt) = case effectiveWindow p of
     Nothing -> maxTarget
-    Just (WindowWidth w)
-        | isLastInEpoche p ->
-            adjust ver (windowSize w) (t .-. _blockEpochStart p) (_blockTarget p)
+    Just w
+        | isLastInEpoch p ->
+            adjust ver w (t .-. _blockEpochStart p) (_blockTarget p)
```
fosskers: Ah, it seems like `adjust` is occurring at the correct point anyway. So is the issue is just with the naming of the function `isLastInEpoch`. Really it should be `isFirstInEpoch`.

src/Chainweb/Version.hs
```diff
@@ -508,3 +488,87 @@ randomChainId :: HasChainwebVersion v => v -> IO ChainId
...
+blockRate :: ChainwebVersion -> Maybe BlockRate
+blockRate Test{} = Nothing
+blockRate TimedConsensus{} = Just $ BlockRate 4
+blockRate PowConsensus{} = Just $ BlockRate 10
+blockRate TimedCPM{} = Just $ BlockRate 4
+-- 120 blocks per hour, 2,880 per day, 20,160 per week, 1,048,320 per year.
+blockRate FastTimedCPM{} = Just $ BlockRate 1
+blockRate Development = Just $ BlockRate 30
+-- 120 blocks per hour, 2,880 per day, 20,160 per week, 1,048,320 per year.
+blockRate Testnet02 = Just $ BlockRate 30
```
fosskers: Why is it more useful for these functions to live in this module?

src/Chainweb/BlockHeader.hs
fosskers: See comment below.

src/Chainweb/Version.hs
larskuhtz: The idea is to eventually have all specifications of chainweb versions in a single place. That that when a new version is added or an old one removed, there is a well-defined place where all changes have to be made.

larskuhtz: Eventually, there should be data structure ala

```haskell
data ChainwebVersionSpecification = ChainwebVersionSpecification
    { _versionSpecName :: !T.Text
    , _versionSpecBlockRate :: !BlockRate
   , _versionSpecEpochSize :: !EpochSize
    , _versionSpecGraph :: !ChainwebGraph
    {- etc. -}
    }
```

And then there would be one constant value of this type defined for each version, e.g. in a module `Chainweb.Version.Specifications`.

That would also make it easier to define and run some basic consistency checks in a test suite.

fosskers: Ah I misread - I thought those had been moved to `BlockHeader`.

src/Chainweb/BlockHeader.hs
larskuhtz: Are you sure that this is correct? `p` is the parent header. The current header is the first in the epoch if the parent is the last in the previous epoch.

src/Chainweb/BlockHeader.hs
```diff
@@ -286,6 +297,85 @@ encodeBlockCreationTime (BlockCreationTime t) = encodeTime t
...
+    ver = _blockChainwebVersion h
+{-# INLINE effectiveWindow #-}
+
+-- | Return whether the given 'BlockHeader' is the last header in its epoch.
+--
+isLastInEpoch :: BlockHeader -> Bool
+isLastInEpoch h = case effectiveWindow h of
+    Nothing -> False
+    Just (WindowWidth w)
+        | int (_blockHeight h) `mod` w == 0 -> True
```
larskuhtz:
```suggestion
        | (int (_blockHeight h) + 1) `mod` w == 0 -> True
```

src/Chainweb/BlockHeader.hs
fosskers: Oh, in that case, I was right the first time, and DA needs to occur when **the parent was the last in the epoch** and **the parent's `BlockHeight` is 1 less than the "mod point" (11, 23, etc).**

PR #398, opened 2019-09-04
Mempool telemetry and bound

larskuhtz:
This PR provides to quick-and-dirty patches for the mempool that improve performance and debugability. 

* [x] add basic telemetry to Mempool
* [x] limit the size of the Mempool

Both should be improved in follow up PRs

### DIFF
--- a/src/Chainweb/Mempool/InMem.hs
+++ b/src/Chainweb/Mempool/InMem.hs
@@ -12,6 +12,7 @@ module Chainweb.Mempool.InMem
   (
    -- * Initialization functions
     withInMemoryMempool
+  , withInMemoryMempool_
 
     -- * Low-level create/destroy functions
   , makeInMemPool
@@ -20,9 +21,11 @@ module Chainweb.Mempool.InMem
 
 ------------------------------------------------------------------------------
 import Control.Applicative (pure, (<|>))
+import Control.Concurrent (threadDelay)
 import Control.Concurrent.MVar (MVar, newMVar, withMVar, withMVarMasked)
+import Control.Concurrent.Async
 import Control.DeepSeq
-import Control.Exception (bracket, mask_)
+import Control.Exception (bracket, mask_, throw)
 import Control.Monad (void, (<$!>))
 
 import Data.Aeson
@@ -38,12 +41,14 @@ import Pact.Types.Gas (GasPrice(..))
 
 import Prelude hiding (init, lookup, pred)
 
+import System.LogLevel
 import System.Random
 
 -- internal imports
 
 import Chainweb.BlockHash
 import Chainweb.BlockHeader
+import Chainweb.Logger
 import Chainweb.Mempool.InMemTypes
 import Chainweb.Mempool.Mempool
 import Chainweb.Utils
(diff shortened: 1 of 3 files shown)


src/Chainweb/Mempool/InMem.hs
```diff
@@ -118,6 +124,33 @@ withInMemoryMempool cfg f = do
...
+  where
+    monitor m = do
+        let lf = logFunction l
+        logFunctionText l Info $ "Initialized Mempool Monitor"
+        runForeverThrottled lf "Chainweb.Mempool.InMem.withInMemoryMempool_.monitor" 10 (10 * mega) $ do
+            stats <- getMempoolStats m
+            logFunctionText l Debug $ "got stats"
+            logFunctionJson l Info stats
+            logFunctionText l Debug $ "logged stats"
+            threadDelay 60000000 {- 1 minute -}
```
gregorycollins: We should randomize this interval slightly -- otherwise all of the threads are going to all wake up at the same time causing lumpy load spikes. (This is a problem throughout the codebase, but since we fork a monitor per chain it will be bad here)

src/Chainweb/Mempool/InMem.hs
```diff
@@ -150,18 +183,26 @@ markValidatedInMem lock txs = withMVarMasked lock $ \mdata -> do
...
         newHashes <- V.mapM (insOne mdata) txs
+        modifyIORef' (_inmemInserted mdata) (+ (length txs))
         modifyIORef' (_inmemRecentLog mdata) $
             recordRecentTransactions maxRecent newHashes
+        readIORef (_inmemInserted mdata)
+
+    unless (i < 5000) $ do
+        withMVarMasked lock $ \mdata -> do
+            modifyIORef' (_inmemPending mdata) $ \ps -> do
+                if (PSQ.size ps > 10000)
```
gregorycollins: Don't you have to update `_inmemInserted` here?

I think if you're going to do this, it might be best just to drop the new `_inmemInserted` field and just always do the check. If we are bounding the size of the set then calling `PSQ.size` every time is not that expensive.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -342,3 +383,9 @@ getRecentTxs maxNumRecent oldHw rlog
     txs = V.map ssnd $ V.takeWhile pred $ _rlRecent rlog
     pred (T2 x _) = x >= oldHw
 
+------------------------------------------------------------------------------
+getMempoolStats :: InMemoryMempool t -> IO MempoolStats
+getMempoolStats m = do
+    withMVar (_inmemDataLock m) $ \d -> MempoolStats
+        <$> (length <$> readIORef (_inmemPending d))
```
gregorycollins: Prefer `PSQ.size` to `length` here (since `length` probably goes via `toList` and will create garbage)

Use `<$!>`

src/Chainweb/Mempool/InMem.hs
larskuhtz: yes, `_inmemInserted` should be reset here.

src/Chainweb/Mempool/InMem.hs
larskuhtz: It seems that `PSQ.size` also uses the `Foldable` instance :-(. But using `size` seems better.

gregorycollins: Ahh! That's disappointing

PR #399, opened 2019-09-04
Switch to approximate thread delays to avoid "thundering herd" issues.

gregorycollins:
This is a routine I've used several times in the past -- instead of
"periodically sleep for one minute", you sleep for "one minute-ish". This has
the effect of spreading out the wake up times for periodic processes so that
they all don't wake up together.


src/Chainweb/Utils.hs
```diff
@@ -1119,3 +1126,22 @@ scurry k a b = k (T2 a b)
...
+approximately :: Int -> Prob.GenIO -> IO Int
+approximately k gen = max 0 <$!> sample
+  where
+    sample = (round . (/ 256.0) . head) <$!>
+             Prob.samples 1 (Prob.normal mean sdev) gen
+    mean   = fromIntegral $ k * 256
+    sdev   = mean / 6
+
+
+threadDelayRng :: MVar Prob.GenIO
```
larskuhtz: In the context of `threadDelay` one wouldn't expect a lot of races due to contention. So maybe an `IORef` with `atomicModifyIORef` would be more efficient here? An `MVar` is certainly the safer option, so it's probably fine.

gregorycollins: `GenIO` is not safe for concurrent modification, so you have to take the lock. I'm ok with these callers butting heads on this lock, they're going to sleep for a long time soon anyways!

PR #401, opened 2019-09-05
[PactService/Checkpointer] Amend logs

emmanueldenloye:


src/Chainweb/Pact/PactService.hs
```diff
@@ -341,10 +341,12 @@ restoreCheckpointer
...
         -- It holds that @(_blockHeight parentHeader == pred height)@
 
+    -> String
+        -- ^ Putative caller
     -> PactServiceM cas PactDbEnv'
-restoreCheckpointer maybeBB = do
+restoreCheckpointer maybeBB caller = do
     checkPointer <- view (psCheckpointEnv . cpeCheckpointer)
-    logInfo $ "restoring " <> sshow maybeBB
+    logInfo $ "restoring (with caller " <> caller <> ") " <> sshow maybeBB
```
larskuhtz: Do we have access to the underlying logger here? If so, we could set the `caller` as a field in the log scope, in which case it would appear structured in the log message and would be indexed as separate keyword field by ES.

emmanueldenloye: I had wanted to do this, but I looked at the actual logger being used and it is the logger from Pact and not the one from your package `yet-another-logger`. Perhaps, I could look into swapping changing the logger used in this module.

larskuhtz: Actually, the pact logger is using the is the log pipeline from `yet-another-logger`. (`yet-another-logger` is more of a framework than a single logger). But the pact logger doesn't expose the scope api. So, if the underlying logger is available, you could set the scope and still use the pact logger.

larskuhtz: One option would be to put the logger that is passed to `initPactService` into the `PactServiceEnv`. In that case it might even be possible to make `PactServiceM` an instance of `MonadLog`.

larskuhtz: Maybe we can to that as a followup PR?

larskuhtz: https://github.com/kadena-io/chainweb-node/pull/323/commits/469c7476d2efe47059bbf28b0f303c4d442074f6 does something similar (but the code would need some cleanup)

emmanueldenloye: Actually, I think it would be better to take on your suggestion in this PR and just unify on logging now. Do you think @slpopejoy would want to pipe in?

PR #404, opened 2019-09-06
increase maxBlockSize by factor of 10

larskuhtz:


fosskers:
Any potential downsides?

PR #407, opened 2019-09-06
Miner asks its own Questions

fosskers:
A `chainweb-node` now supports POW mining in two ways.

- In-process mining, single-core, ON by default, disabled by `--disable-mining`.
- Remote work requests / submission via REST endpoints, OFF by default, enabled by `--mining-coordination`.

`chainweb-miner` has been updated to take advantage of the new API structure, and in tests has been shown to be preempted efficiently when working alongside a `chainweb-node` which was also mining.

### TODO

- [x] Simplify `Coordinator` functions
- [x] Utilize new `Coordinator` functions, depending on mining scheme
- [x] Update `chainweb-miner` to call for work on its own (include `Miner` type to get rewards!)
- [x] Flag to enable mining Coordination, while keeping in-process mining off
- [x] Chain-specific work requests
- [x] Endpoint for Cut Updates
- [x] **Efficient preemption**


larskuhtz:
You may consider server sent event (SSE) instead of websockets. SSE are a simpler technology, that's sufficient here. They are part of HTML5.

fosskers:
It works :)

slpopejoy:
Best to name things without gender specificity, thanks.

miner/Miner.hs
```diff
@@ -185,49 +176,40 @@ pCores = option auto
...
-        CPU _ -> do
-            miner <- async $ mining (scheme env) env
-            pPrintNoColor as
-            W.run (port as) $ app env
-            wait miner
+    getWork env >>= \case
+        Nothing -> putStrLn "Failed to connect to the given Chainweb Node." >> exitFailure
+        Just wb -> case cmd as of
+            GPU _ -> putStrLn "GPU mining is not yet available."
+            CPU _ -> pPrintNoColor as >> mining (scheme env) env wb >> exitFailure
```
fosskers: The re-echo of the arguments may not be useful.

Prod-level logging hasn't yet been addressed at all, but will be done so in a follow-up.

miner/Miner.hs
```diff
@@ -244,29 +226,60 @@ scheme env = case cmd $ args env of
...
-submit (work -> w) bs = atomically $
-    isEmptyTMVar w >>= bool (void $ swapTMVar w bs) (putTMVar w bs)
+getWork :: Env -> IO (Maybe WorkBytes)
+getWork e = retrying policy (\_ -> pure . isNothing) $ const f
+  where
+    -- | If we wait longer than the average block time and still can't get
+    -- anything, then there's no point in continuing to wait.
+    --
+    policy :: RetryPolicy
+    policy = exponentialBackoff 500000 <> limitRetries 7
```
fosskers: 500ms, 1s, 2s, 4s, 8s, 16s, 32s... is seven attempts. I considered also making the `NodeUrl` flag instead accept a list, so that `getWork` can fallback to other nodes if an initial attempt fails.

miner/Miner.hs
```diff
@@ -244,29 +226,60 @@ scheme env = case cmd $ args env of
...
-    newWork = void . atomically . readTMVar $ work e
+    T3 cbytes tbytes hbytes = unWorkBytes wb
+
+    a :: ClientArgs
+    a = args e
+
+    -- TODO Rework to use Servant's streaming? Otherwise I can't use the
+    -- convenient client function here.
+    updateSignal :: IO ()
+    updateSignal = withEvents req (mgr e) (void . SP.head_ . SP.filter realEvent)
```
fosskers: The first Server-Sent-Event to appear here in the `Stream` will kill the race, thus prompting a new `getWork` call.

src/Chainweb/Chainweb.hs
```diff
@@ -205,6 +207,7 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     { _configChainwebVersion = v
     , _configNodeId = NodeId 0 -- FIXME
     , _configMiner = defaultEnableConfig defaultMinerConfig
+    , _configCoordinator = False
```
fosskers: `False` by default. Average people running full nodes will likely not want arbitrary `/mining/work/` calls to be allowed.

src/Chainweb/CutDB.hs
```diff
@@ -251,6 +253,26 @@ _cutStm = readTVar . _cutDbCut
...
+-- | As in `awaitNewCut`, but only updates when the specified `ChainId` has
+-- grown.
+--
+awaitNewCutByChainId :: CutDb cas -> ChainId -> Cut -> IO Cut
+awaitNewCutByChainId cdb cid c = atomically $ do
+    c' <- _cutStm cdb
+    let !b0 = HM.lookup cid $ _cutMap c
+        !b1 = HM.lookup cid $ _cutMap c'
+    when (b1 == b0) retry
+    return c'
```
fosskers: This function is the vehicle by which we do non-wasteful preempting for both local and remote mining. New `Cut`s come in all the time, but a particular mining process only wants to stop if **the chain they're currently working on** has changed. Otherwise, their work shouldn't be considered stale.

This will increase mining efficiency on our current network dramatically.

src/Chainweb/Miner/Miners.hs
```diff
@@ -87,41 +96,38 @@ localTest tmv gen miners bh =
...
+
 -- | A single-threaded in-process Proof-of-Work mining loop.
 --
-localPOW :: TMVar BlockHeader -> ChainwebVersion -> BlockHeader -> IO ()
-localPOW tmv v bh = do
-    HeaderBytes newBytes <- usePowHash v (\p -> mine p (_blockNonce bh) tbytes) hbytes
-    new <- runGet decodeBlockHeaderWithoutHash newBytes
-    atomically $ putTMVar tmv new
+localPOW :: LogFunction -> ChainwebVersion -> Miner -> CutDb cas -> IO ()
+localPOW lf v m cdb = runForever lf "Chainweb.Miner.Miners.localPOW" loop
```
fosskers: This could possibly be merged with `localTest`. They share some logic, but they're both so short by this point.

PR #409, opened 2019-09-09
Remove `_blockMiner` Field

fosskers:
This PR removes a field from the `BlockHeader` type which we've determined is no longer useful, as its information is stored in its associated `Payload`.

**Note:** some external documentation will likely need to be updated, since the byte structure of an encoded `BlockHeader` is now slightly smaller.


src/Chainweb/Logging/Amberdata.hs
```diff
@@ -89,7 +88,6 @@ data AmberdataBlock = AmberdataBlock
   , _amberdataTimestamp :: {-# UNPACK #-} !BlockCreationTime
   , _amberdataParentHash :: {-# UNPACK #-} !BlockHash
   , _amberdataNonce :: {-# UNPACK #-} !Nonce
-  , _amberdataMiner :: {-# UNPACK #-} !ChainNodeId
```
fosskers: @LindaOrtega will removal this screw up our AmberData dashboard?

src/Chainweb/MerkleUniverse.hs
```diff
@@ -69,7 +68,6 @@ instance MerkleUniverse ChainwebHashTag where
     type MerkleTagVal ChainwebHashTag 'PowHashTag = 0x0009
     type MerkleTagVal ChainwebHashTag 'BlockHashTag = 0x0010
     type MerkleTagVal ChainwebHashTag 'HashTargetTag = 0x0011
-    type MerkleTagVal ChainwebHashTag 'ChainNodeIdTag = 0x0012
```
fosskers: Will this tiny gap between 11 and 13 have a significant effect, @larskuhtz ?

PR #410, opened 2019-09-09
fix TreeDB ancestor query logic

larskuhtz:


src/Chainweb/TreeDB.hs
```diff
@@ -814,7 +819,7 @@ seekAncestor db h r
     -- have narrowed to a single block once we reach the target height.
     --
     fastRoute2 (i :: Int)
-        | 2 * l < hh = do
+        | r + 2 * off < hh = do
```
larskuhtz: This fixes the ancestor logic. Without this, if there was a more than one block header for at height r the function would use `fastRoute2`, but would start the search at block height 2 which would result in a `TreeDbAncestorMissing` failure.

This means the node during catch up would receive an connection failure and de-prioritize the failing node. If all node in the network had fork at height `r`, the catch-up node would be stuck, until some node got restarted and pruned its block header db.

PR #411, opened 2019-09-10
Add Guard Rotation Function to Coin Contract

emilypi:


pact/coin-contract/coin-sig.pact
```diff
@@ -22,20 +22,28 @@
 
   (defun transfer-and-create:string (sender:string receiver:string receiver-guard:guard amount:decimal)
     @doc "Transfer between accounts SENDER and RECEIVER on the same chain.    \
-    \This fails if the SENDER account does not exist. If the RECEIVER account \
-    \does not exist, it is created and associated with GUARD."
+    \This fails if both accounts do not exist. Create-on-transfer can be      \
```
mightybyte: Not sure I like this change.  The new wording `both accounts do` seems less clear to me than what was there before.  Also, the last sentence is weird because this *is* the `transfer-and-create` function.

pact/coin-contract/coin-sig.pact
```diff
@@ -22,20 +22,28 @@
...
+    @doc "Check an account's balance"
+    @model [ (property (not (= amount ""))) ]
+    )
 
   (defun account-info:object (account:string)
-    @doc "Get all of an account's info.  This includes the balance and the    \
-    \guard.")
+    @doc "Get all of an account's info. This includes the balance and the     \
+    \guard."
+    @model [ (property (not (= amount ""))) ])
```
mightybyte: Does this model make sense?  This function does not have an `amount` parameter.

pact/coin-contract/coin-sig.pact
```diff
@@ -22,20 +22,28 @@
...
 
     @model [ (property (> amount 0.0))
              (property (not (= sender receiver)))
            ]
     )
 
   (defun account-balance:decimal (account:string)
-    @doc "Check an account's balance")
+    @doc "Check an account's balance"
+    @model [ (property (not (= amount ""))) ]
```
mightybyte: No `amount` parameter here either.

pact/coin-contract/coin-sig.pact
```diff
@@ -22,20 +22,28 @@
...
   (defun account-info:object (account:string)
-    @doc "Get all of an account's info.  This includes the balance and the    \
-    \guard.")
+    @doc "Get all of an account's info. This includes the balance and the     \
+    \guard."
+    @model [ (property (not (= amount ""))) ])
+
+  (defun rotate-account-guard:string (account:string new-guard:guard)
+    @doc "Rotate guard associated with ACCOUNT to new guard NEW-GUARD"
+    @model [ (property (not (= amount ""))) ]
```
mightybyte: Ditto here

pact/coin-contract/coin-sig.pact
emilypi: should have been `account` - thanks.

pact/coin-contract/coin-sig.pact
emilypi: This is the wording from the new docs you wrote in your last PR to `coin.pact`. I'm just making sure the documentation is flush. Do you want me to amend the comments here and in `coin.pact`?

emilypi: Ah I added it to the wrong transfer function. My bad. Regarding the wording, this is what you wrote for the `transfer` function in your recent changes to `coin.pact`. Do you want me amend it?

pact/coin-contract/coin-sig.pact
emilypi: Thanks - it should be `account`

PR #412, opened 2019-09-11
Return correct gas in tx failure result

emilypi:
This PR returns the correct gas amount in the return `CommandResult` on tx failure.


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -132,10 +133,10 @@ applyCmd logger pactDbEnv miner gasModel pd spv cmd mcache = do
         case cmdResultE of
 
           Left e2 ->
-
-            jsonErrorResult payloadEnv requestKey e2 buyGasLogs (Gas 0) mcache'
+            let
+              (GasLimit (ParsedInteger !g)) = gasLimitOf cmd
+            in jsonErrorResult payloadEnv requestKey e2 buyGasLogs (Gas $ fromIntegral g) mcache'
```
fosskers: Is the `GasLimit` here the actual gas that was spent, or the limit that the user defined?

mightybyte: I think the question is more about whether it is limit vs spent, not whether the units are gas or coins.  Returning the limit gives no new info.  Seems like we should return the amount of gas consumed.

fosskers: Emily and I talked about this: in the case of failure, the limit _is_ what is spent. So returning the limit is correct.

PR #416, opened 2019-09-13
Add logging to Pact Endpoint handlers

emmanueldenloye:


larskuhtz:
I would propose that we log this to a telemetry backend instead of the normal application logs.

Logging to a telemetry backend has the advantage that JSON encoded data appears structured in the logs and is indexed as such by elastic search.

src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -125,45 +133,62 @@ somePactServer (SomePactServerData (db :: PactServerData_ v c logger cas))
...
-sendHandler mempool (SubmitBatch cmds) =
-    Handler $
+sendHandler
+    :: Logger logger
+    => logger
+    -> MempoolBackend ChainwebTransaction
+    -> SubmitBatch
+    -> Handler RequestKeys
+sendHandler logger mempool (SubmitBatch cmds) = Handler $ do
+    liftIO $ logg Info (sshow cmds)
```
larskuhtz: I would propose to log JSON log messages. Those can than be (optionally) handled by a telemetry backend. Which provides the advantage that ES can discover the structure and index them accordingly.

src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -83,12 +87,14 @@ newtype PactServerData_ (v :: ChainwebVersionT) (c :: ChainIdT) logger cas
...
        KnownChainIdSymbol c,
-       PayloadCas cas)
+       PayloadCas cas,
+       Logger logger)
     => SomePactServerData (PactServerData_ v c logger cas)
 
 
 somePactServerData
     :: PayloadCas cas
+    => Logger logger
```
larskuhtz: I am not sure if the handler is the best place for logging this data. What about adding the logger on the level of pact service. That would make sure that also internal calls would be logged. I don't think we have those currently, but it may be more future proof.

larskuhtz: hmm, I just noticed that `validateCommand` is called in the handlers. I don't think that's best practice (normally, handlers should not contain any busyness logic, but only handle HTTP endpoint related things.

Under these circumstances and given that we probably want to log before any business logic is applied to the incoming command, this module is the right place to do the logging.

src/Chainweb/Pact/RestAPI/Server.hs
larskuhtz: If the type of `cads` already has `ToJSON` and `NFData` instances, it could be passed directly to the to a JSON log function. Otherwise small wrapper data type may be used. Something like

```haskell
newtype PactCmdLog a = PactCmdLog a
    deriving (Show, ToJSON, NFData)
```

src/Chainweb/Utils.hs
```diff
@@ -188,45 +190,47 @@ import Control.Monad.IO.Class
...
+import qualified Data.ByteString as B
+import qualified Data.ByteString.Base64 as B64
+import qualified Data.ByteString.Base64.URL as B64U
+import qualified Data.ByteString.Lazy as BL
+import qualified Data.HashMap.Strict as HM
+import qualified Data.HashSet as HS
+import qualified Data.List.NonEmpty as NEL
+import qualified Data.Text as T
+import qualified Data.Text.Encoding as T
+import qualified Data.Text.Lazy as TL
```
larskuhtz: The imports in chainweb are usually sorted by module name (not by line). Could you please revert this reordering. It adds a lot of noise to the PR.

src/Chainweb/Utils.hs
```diff
@@ -1155,3 +1159,10 @@ threadDelayRng = unsafePerformIO (Prob.createSystemRandom >>= newMVar)
 approximateThreadDelay :: Int -> IO ()
 approximateThreadDelay d = withMVar threadDelayRng (approximately d)
                            >>= threadDelay
+
+data PactCmdLog
```
larskuhtz: Maybe move this to the logging site? Utils is more for things that are more generally useful.

src/Chainweb/Utils.hs
emmanueldenloye: It must have been a slip of the finger. I'll revert this.

src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -125,45 +138,69 @@ somePactServer (SomePactServerData (db :: PactServerData_ v c logger cas))
...
 somePactServers v =
     mconcat . fmap (somePactServer . uncurry (somePactServerData v))
 
-
-sendHandler :: MempoolBackend ChainwebTransaction -> SubmitBatch -> Handler RequestKeys
-sendHandler mempool (SubmitBatch cmds) =
-    Handler $
+data PactCmdLog
+  = PactCmdLogSend (NonEmpty (Command Text))
+  | PactCmdLogPoll (NonEmpty ByteString)
```
mightybyte: I think this needs to not be ByteString.  We're getting `Cannot decode byte '\xb4': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream.` in some places.

mightybyte: I pushed a commit fixing this.

mightybyte:
I ran this on devnet and it looks like the bytestring decoding problem is fixed now.

PR #417, opened 2019-09-13
Make pact /poll endpoint use the pactdb tx index.

gregorycollins:
Lars and Emily also need the new included internal pact execution service lookup method for SPV.

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -357,7 +357,6 @@ library
         , wai-middleware-throttle >= 0.3
         , warp >= 3.2
         , warp-tls >= 3.2.7
-        , witherable >= 0.2
         , x509 >=1.7
         , x509-system >=1.6
         , x509-validation >=1.6
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -218,6 +218,8 @@ initialPayloadState v@FastTimedCPM{} cid = initializeCoinContract v cid TN.paylo
 initialPayloadState v@Development cid = initializeCoinContract v cid DN.payloadBlock
 initialPayloadState v@Testnet02 cid = initializeCoinContract v cid PN.payloadBlock
 
+
+
 initializeCoinContract
     :: forall cas. PayloadCas cas
     => ChainwebVersion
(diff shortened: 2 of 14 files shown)


chainweb.cabal
```diff
@@ -357,7 +357,6 @@ library
         , wai-middleware-throttle >= 0.3
         , warp >= 3.2
         , warp-tls >= 3.2.7
-        , witherable >= 0.2
```
emilypi: 👍

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -123,6 +124,8 @@ responseGolden networkIO rksIO = golden "command-0-resp" $ do
     rks <- rksIO
     cwEnv <- _getClientEnv <$> networkIO
     (PollResponses theMap) <- testPoll testCmds cwEnv rks
+    putStrLn $ "keys: " ++ show rks
```
gregorycollins: I'll remove

src/Chainweb/Pact/PactService.hs
```diff
@@ -815,3 +826,18 @@ transactionsFromPayload plData = do
...
     toCWTransaction bs = codecDecode chainwebPayloadCodec bs
+
+execLookupPactTxs
+    :: PayloadCas cas
+    => T2 BlockHeight BlockHash
+    -> Vector P.PactHash
+    -> PactServiceM cas (Vector (Maybe (T2 BlockHeight BlockHash)))
+execLookupPactTxs (T2 leafHeight leafHash) txs =
+    withDiscardedBatch $
+    withCheckpointerRewind (Just (leafHeight + 1, leafHash))
```
emilypi: What's the justification for `leafHeight + 1`?  Do you mind commenting your thinking with these calls? (re: discardedBatch, rewind to specific height, where you think the bheight + bhash pair are coming from etc)

gregorycollins: This is just the way the api works -- it's documented in the outer interface but not in here. The functions here:

`withDiscardedBatch` sets a sqlite "Batch" savepoint, runs the action, then discards to "Batch", effectively making the change a no-op in terms of db mutation (another important reason to batch the requests) -- but maybe we don't care about moving the restore point here because we're moving it to the cut value and that's a good place for it to be. I'll remove the call

slpopejoy: @gregorycollins it might be worth making a version of `withCheckpointerRewind` that does the `succ` on the blockHeight to document this behavior; as it is we're manually doing it in three places -- here, `execLocal`, and `execNewBlock` which is creaky.

PR #419, opened 2019-09-13
New Payloads

fosskers:
The Coin Contract was recently updated.


mightybyte:
We might not need this since my PR has payload changes and will require a regen.

fosskers:
Same with my PR.

PR #422, opened 2019-09-15
NewMinedBlock: Mention the Miner

fosskers:
This makes it easy to see in our logging who was credited for mining each block.


src/Chainweb/Miner/Coordinator.hs
```diff
@@ -142,8 +145,9 @@ publish lf (MiningState ms) cdb bh = do
     c <- _cut cdb
     let !phash = _blockPayloadHash bh
     res <- runExceptT $ do
-        T2 p pl <- HM.lookup phash ms ?? "BlockHeader given with no associated Payload"
-        c' <- tryMonotonicCutExtension c bh !? "Newly mined block for outdate cut"
+        T3 m p pl <- HM.lookup phash ms ?? "BlockHeader given with no associated Payload"
+        let !miner = m ^. position @1 . _Unwrapped
```
mightybyte: Ditch `DataKinds` and `Data.Generics.Product.Positions` and just use `view _1`.

fosskers: I don't think those are the same thing. `position @1` grabs the first field of any product type with a `Generic` instance.

mightybyte: Oh, I saw the `T3` and incorrectly thought it was referring to the tuple.  In that case we should definitely not use this overcomplicated product stuff.  Add accessors to the `Miner` type.  That's dramatically more understandable.

mightybyte: Better yet, just use the `minerId` lens that is already there.

fosskers: Good catch, I somehow overlooked the `minerId` lens.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DataKinds #-}
```
mightybyte: Do we still need this?

PR #423, opened 2019-09-15
Reworked Miner Logging

fosskers:
This PR reworks the logging in `chainweb-miner`, giving it timestamps and proper configurable log levels. It also cleans up the Miner code in general and fixes a bug.

2019-09-15-170014_958x538_scrot

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -539,7 +539,6 @@ executable chainweb-miner
           chainweb
 
         -- external
-        , aeson >= 1.4
         , base >= 4.11 && < 5
         , connection >= 0.2
         , errors >= 2.3
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -324,7 +324,6 @@ library
         , resourcet >= 1.2
         , safeio >= 0.0
         , safe-exceptions >= 0.1
-        , scheduler >= 1.4
         , servant >= 0.14
         , servant-client >= 0.14
         , servant-client-core >= 0.14
@@ -550,13 +549,11 @@ executable chainweb-miner
         , retry >= 0.7
         , rio >= 0.1.9
         , scheduler >= 1.4.1
-        , servant >= 0.14
         , servant-client >= 0.14
         , streaming >= 0.2
         , streaming-events >= 1.0
         , strict-tuple >= 0.1
         , wai-extra >= 3.0
-        , warp >= 3.2
 
 executable cwtool
     import: warning-flags, debugging-flags
@@ -604,7 +601,6 @@ executable cwtool
         , base >= 4.11 && < 5
         , base-prelude >= 1.3
         , base16-bytestring >= 0.1
-        , bloomfilter >= 2.0
         , bytes >= 0.15
         , bytestring >= 0.10
         , chainweb-storage >= 0.1
@@ -637,7 +633,6 @@ executable cwtool
         , pretty-simple >= 2.1
         , primitive >= 0.6
         , random >= 1.1
-        , reflection >= 2.1
         , resourcet >= 1.2
         , servant >= 0.14.1
         , servant-client >= 0.14
@@ -646,7 +641,6 @@ executable cwtool
         , stm
         , streaming >= 0.2.2
         , streaming-commons >= 0.2
-        , strict-tuple >= 0.1
         , tasty >= 1.0
         , tasty-golden >= 2.3
         , tasty-hunit >= 0.9
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -547,7 +547,7 @@ executable chainweb-miner
         , mwc-random >= 0.13
         , optparse-applicative >= 0.14
         , retry >= 0.7
-        , rio >= 0.1.9
+        , rio >= 0.1.5
         , scheduler >= 1.4.1
         , servant-client >= 0.14
         , streaming >= 0.2
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -540,29 +540,22 @@ executable chainweb-miner
 
         -- external
         , aeson >= 1.4
-        , async >= 2.2
         , base >= 4.11 && < 5
-        , base-prelude >= 1.3
-        , base16-bytestring >= 0.1
-        , bytes >= 0.15
-        , bytestring >= 0.10
         , connection >= 0.2
-        , containers >= 0.5
         , errors >= 2.3
+        , generic-lens >= 1.2
         , http-client >= 0.5
         , http-client-tls >= 0.3
         , mwc-random >= 0.13
         , optparse-applicative >= 0.14
-        , pretty-simple >= 2.1
         , retry >= 0.7
+        , rio >= 0.1.9
         , scheduler >= 1.4.1
         , servant >= 0.14
         , servant-client >= 0.14
         , streaming >= 0.2
         , streaming-events >= 1.0
         , strict-tuple >= 0.1
-        , stm >= 2.4
-        , text >= 1.2
         , wai-extra >= 3.0
         , warp >= 3.2
 
--- a/miner/Miner.hs
+++ b/miner/Miner.hs
@@ -196,23 +196,25 @@ main = do
 
 run :: RIO Env ()
 run = do
+    logInfo "Starting Miner."
     env <- ask
     case cmd $ args env of
         GPU -> logError "GPU mining is not yet available."
-        CPU _ -> liftIO (getWork env) >>= \case
-            Nothing -> logError "Failed to connect to specified Node."
-            Just wb -> liftIO $ mining (scheme env) env wb
+        CPU _ -> getWork >>= traverse_ (mining (scheme env))
     exitFailure
 
-scheme :: Env -> (TargetBytes -> HeaderBytes -> IO HeaderBytes)
-scheme env = case cmd $ args env of
-    CPU e -> cpu (version $ args env) e (gen env)
-    GPU -> gpu
+scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
+scheme env = case cmd $ args env of CPU e -> cpu e; GPU -> gpu
 
 -- | Attempt to get new work while obeying a sane retry policy.
 --
-getWork :: Env -> IO (Maybe WorkBytes)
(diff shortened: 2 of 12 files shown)


chainweb.cabal
```diff
@@ -539,32 +538,22 @@ executable chainweb-miner
...
-        , containers >= 0.5
         , errors >= 2.3
+        , generic-lens >= 1.2
         , http-client >= 0.5
         , http-client-tls >= 0.3
         , mwc-random >= 0.13
         , optparse-applicative >= 0.14
-        , pretty-simple >= 2.1
         , retry >= 0.7
+        , rio >= 0.1.9
```
fosskers: RIO offered the simplest logging infrastructure that I knew how to get running quickly. Its logging has nice looking, reasonable defaults. It cleans up a lot of code below, too.

miner/Miner.hs
```diff
@@ -59,91 +56,67 @@
...
-    , mgr :: Manager
-    , args :: ClientArgs }
+    { gen :: !MWC.GenIO
+    , mgr :: !Manager
+    , log :: !LogFunc
+    , args :: !ClientArgs }
+    deriving stock (Generic)
+
+instance HasLogFunc Env where
+    logFuncL = field @"log"
```
fosskers: The only boilerplate that RIO demands for its logging.

miner/Miner.hs
```diff
@@ -208,107 +190,114 @@ main = do
...
-    CPU e -> cpu (version $ args env) e (gen env)
-    GPU _ -> gpu
+run :: RIO Env ()
+run = do
+    logInfo "Starting Miner."
+    env <- ask
+    case cmd $ args env of
+        GPU -> logError "GPU mining is not yet available."
+        CPU _ -> getWork >>= traverse_ (mining (scheme env))
+    exitFailure
```
fosskers: The miner is designed to never exit. If `mining` ever returns, we should consider it a failure, hence `exitFailure`.

miner/Miner.hs
```diff
@@ -208,107 +190,114 @@ main = do
     opts = info (pClientArgs <**> helper)
         (fullDesc <> progDesc "The Official Chainweb Mining Client")
 
-scheme :: Env -> (TargetBytes -> HeaderBytes -> IO HeaderBytes)
-scheme env = case cmd $ args env of
-    CPU e -> cpu (version $ args env) e (gen env)
-    GPU _ -> gpu
+run :: RIO Env ()
```
fosskers: Effectively `ReaderT` with built-in support for logging. Mostly a drop-in replacement.

miner/Miner.hs
```diff
@@ -208,107 +190,114 @@ main = do
...
     -- won the race instead, then the `go` call is automatically cancelled.
     --
-    miningSuccess :: HeaderBytes -> IO ()
+    miningSuccess :: HeaderBytes -> RIO Env ()
     miningSuccess h = do
-      bh <- bytesToBlockHeader h
-      putStrLn $ "Success!!!  " <> (show $ _blockNonce bh)
-      putStrLn $ "difficulty = " <> (T.unpack $ showTargetHex $ _blockTarget bh)
-      putStrLn $ "blockHash  = " <> (T.unpack $ hashToHex $ _blockHash bh)
-      putStrLn $ "powHash    = " <> (T.unpack $ powHashToHex $ _blockPow bh)
```
fosskers: @mightybyte we can reintroduce these if you like, possibly at the `Debug` level. The day-1 user probably just wants to know if they mined successfully or not, which the `logInfo` call below does.

miner/Miner.hs
```diff
@@ -208,107 +190,114 @@ main = do
...
-bytesToBlockHeader (HeaderBytes hbytes) = runGet decodeBlockHeaderWithoutHash hbytes
-
-cpu :: ChainwebVersion -> CPUEnv -> MWC.GenIO -> TargetBytes -> HeaderBytes -> IO HeaderBytes
-cpu v e g tbytes hbytes = fmap head . withScheduler comp $ \sch -> do
-    bh <- bytesToBlockHeader hbytes
-    let BlockHeight height = _blockHeight bh
-    printf "Mining block %d on chain %s with difficulty %s\n"
-      height
-      (T.unpack $ chainIdToText $ _blockChainId bh)
-      (T.unpack $ showTargetHex (_blockTarget bh))
```
fosskers: Doing all these here would slow down mining.

chainweb.cabal
```diff
@@ -539,32 +538,22 @@ executable chainweb-miner
...
-        , containers >= 0.5
         , errors >= 2.3
+        , generic-lens >= 1.2
         , http-client >= 0.5
         , http-client-tls >= 0.3
         , mwc-random >= 0.13
         , optparse-applicative >= 0.14
-        , pretty-simple >= 2.1
         , retry >= 0.7
+        , rio >= 0.1.5
```
fosskers: RIO offered the simplest logging infrastructure that I knew how to get running quickly. Its logging has nice looking, reasonable defaults. It cleans up a lot of code below, too.

miner/Miner.hs
```diff
@@ -208,107 +191,114 @@ main = do
...
-    CPU e -> cpu (version $ args env) e (gen env)
-    GPU _ -> gpu
+run :: RIO Env ()
+run = do
+    logInfo "Starting Miner."
+    env <- ask
+    case cmd $ args env of
+        GPU -> logError "GPU mining is not yet available."
+        CPU _ -> getWork >>= traverse_ (mining (scheme env))
+    liftIO exitFailure
```
fosskers: The miner is designed to never exit. If mining ever returns, we should consider it a failure, hence `exitFailure`.

miner/Miner.hs
```diff
@@ -188,15 +174,12 @@ pChainId = optional $ textOption
...
-        <$> MWC.createSystemRandom
-        <*> newManager (mkManagerSettings ss Nothing)
-        <*> execParser opts
-    getWork env >>= \case
-        Nothing -> putStrLn "Failed to connect to the given Chainweb Node." >> exitFailure
-        Just wb -> case cmd as of
-            GPU _ -> putStrLn "GPU mining is not yet available."
-            CPU _ -> pPrintNoColor as >> mining (scheme env) env wb >> exitFailure
+    cargs <- execParser opts
+    lopts <- setLogMinLevel (ll cargs) . setLogUseLoc False <$> logOptionsHandle stderr True
```
fosskers: "Be verbose, but set the minimum level to whatever came in from the CLI, and don't bother echoing source line locations."

chainweb.cabal
mightybyte: Yet another logging system?  Not a fan of that.  Also, rio seems like overkill here.  Let's stick with the logging system we've already got.  Either that or print statements are perfectly fine.

miner/Miner.hs
```diff
@@ -208,107 +191,114 @@ main = do
...
-scheme :: Env -> (TargetBytes -> HeaderBytes -> IO HeaderBytes)
-scheme env = case cmd $ args env of
-    CPU e -> cpu (version $ args env) e (gen env)
-    GPU _ -> gpu
+run :: RIO Env ()
+run = do
+    logInfo "Starting Miner."
+    env <- ask
+    case cmd $ args env of
+        GPU -> logError "GPU mining is not yet available."
```
mightybyte: If it's not working, we should take this out of the command-line help.  We shouldn't ship non-working features.

miner/Miner.hs
mightybyte: This is a few instructions once every time a new block gets mined.  This is so far down in the weeds it's not worth worrying about.  Also, the instructions for doing these conversions will be dwarfed by the IO itself.

miner/Miner.hs
mightybyte: I think day-1 users will probably want to know how to find the block that they mined.  This means at least showing them chain ID and block height.  Also probably block hash.  But then people expect to see zeroes in the block hash, so `powHash` makes sense as well.  Difficulty we could probably leave out.  That was more for my benefit as a sanity check.

miner/Miner.hs
```diff
@@ -208,107 +191,114 @@ main = do
...
-        -- there won't be any overlap?
-        n <- Nonce <$> MWC.uniform g
-        new <- usePowHash v (\p -> mine p n tbytes) hbytes
-        terminateWith sch new
+      e <- ask
+      let !v = version $ args e
+          !m = mgr e
+          !u = coordinator $ args e
+      cid <- liftIO chain
+      logInfo $ cid <> ": Mining Success!"
```
mightybyte: I think we need to tell people what that first number is.  They won't magically know.  Also, we should at least tell them the block height of the block they just mined.  That gives the minimum amount of information they need to go look the block up using the chainweb API or in the future a block explorer.

miner/Miner.hs
mightybyte: Please no type applications where a simple print statement would work just fine.

fosskers: Not quite the same thing. `logFuncL` is a lens, it's how the ReaderT stack gets its logging "for free".

miner/Miner.hs
fosskers: Very easy to add back as a follow-up.

chainweb.cabal
fosskers: I definitely thought the decision over. Here were my thoughts:

- Chainweb's logging is somewhat complicated, is tailored for Chainweb, is an amalgamation of 4 different module groups (`yet-another-logger` + `loglevel` + `Chainweb.Logger` + `Utils.Logging`), and Lars wants to change it anyway
- `yet-another-logger` states of itself "This Version is yet a preview", Lars tells me it's bit-rotten, it's also a `ReaderT` pattern, and I foresaw it interacting dubiously with the parallelism in the miner
- `rio` is maintained, it's tailored for CLI-based Haskell apps, it has sane defaults, very low boilerplate, easy to use, plays nice with the rest of the ecosystem, I already knew how to use it, and I had a single sitting to come up with something professional

So I went with it. It's certainly a leap above print statements, especially w.r.t. quality of output and configurable log levels. It was effectively a drop-in replacement, save the refactoring I did of `main` itself.

miner/Miner.hs
fosskers: I'll comment out (not delete) all the things that Greg will soon need, so that he can jump right in.

PR #424, opened 2019-09-16
Add configurable log message filter

larskuhtz:


mightybyte:
Looks like a good start.  @slpopejoy?

slpopejoy:
Aand it looks like @mightybyte wants to iterate!

PR #435, opened 2019-09-17
Corrections to Mining Guide

fosskers:
Spelling and so forth.

### DIFF
--- a/miner/README.org
+++ b/miner/README.org
@@ -31,7 +31,7 @@ process of mathematically "solving" a block. Unless solved, a block cannot be
 included in the chain. Without mining, a blockchain cannot progress and
 transactions will never finalize.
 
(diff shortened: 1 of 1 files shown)


miner/README.org
```diff
@@ -31,7 +31,7 @@ process of mathematically "solving" a block. Unless solved, a block cannot be
 included in the chain. Without mining, a blockchain cannot progress and
 transactions will never finalize.
 
-Economically, mining is the principle way to obtain currency on a Chainweb
+Economically, mining is the principal way to obtain currency on a Chainweb
```
mightybyte: This is incorrect.  In this case `principle` is the correct spelling.

mightybyte: Oops, my bad.  On re-checking this is correct.

fosskers: :gb:

PR #438, opened 2019-09-17
Pruning the Mining Coordination State

fosskers:
This prevents an eventual memory buildup.

### DIFF
--- a/src/Chainweb/Chainweb/MinerResources.hs
+++ b/src/Chainweb/Chainweb/MinerResources.hs
@@ -24,20 +24,29 @@ module Chainweb.Chainweb.MinerResources
   , withMiningCoordination
   ) where
 
-import Control.Concurrent.STM (TVar)
-import Control.Concurrent.STM.TVar (newTVarIO)
+import Data.Generics.Wrapped (_Unwrapped)
+import qualified Data.HashMap.Strict as HM
+import Data.Tuple.Strict (T3(..))
+
+import Control.Concurrent (threadDelay)
+import Control.Concurrent.Async (concurrently)
+import Control.Concurrent.STM (atomically)
+import Control.Concurrent.STM.TVar (TVar, modifyTVar', newTVarIO)
+import Control.Lens (over)
 
 import qualified System.Random.MWC as MWC
 
 -- internal modules
 
+import Chainweb.BlockHeader (_blockCreationTime, BlockCreationTime(..))
 import Chainweb.CutDB (CutDb)
 import Chainweb.Logger (Logger, logFunction)
 import Chainweb.Miner.Config (MinerConfig(..))
-import Chainweb.Miner.Coordinator (MiningState(..))
+import Chainweb.Miner.Coordinator (MiningState(..), PrevBlock(..))
 import Chainweb.Miner.Miners
 import Chainweb.Payload.PayloadStore
-import Chainweb.Utils (EnableConfig(..))
+import Chainweb.Time (Micros, Time(..), getCurrentTimeIntegral)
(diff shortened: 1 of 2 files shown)


src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -64,11 +73,21 @@ withMiningCoordination logger enabled cutDb inner
...
+        fmap snd . concurrently (prune t) $ inner . Just $ MiningCoordination
             { _coordLogger = logger
             , _coordCutDb = cutDb
-            , _coordState = t
-            }
+            , _coordState = t }
+  where
+    prune :: TVar MiningState -> IO a
+    prune t = do
+        let !d = 600000000  -- 10 minutes
```
fosskers: Is 10 minutes long enough?

larskuhtz: use `runForever`?

larskuhtz: Or something similar

larskuhtz: Or is it fine if the miner crashes if this throws an exception?

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -64,11 +73,21 @@ withMiningCoordination logger enabled cutDb inner
...
+        threadDelay d
+        ago <- over (_Unwrapped . _Unwrapped) (subtract (int d)) <$> getCurrentTimeIntegral
+        atomically . modifyTVar' t $ over _Unwrapped (HM.filter (f ago))
+        prune t
+
+    f :: Time Micros -> T3 a PrevBlock b -> Bool
+    f ago (T3 _ (PrevBlock p) _) = _blockCreationTime p > BlockCreationTime ago
```
larskuhtz: Would it make sense to use the block height as a criterium instead? For instance if the block is < current height - x? With x something like 3?

larskuhtz: An maybe an additional limit on total items. Say, reject new requests (e.g. via 503) if there are more than, say, 1000 item in the map?

src/Chainweb/Chainweb/MinerResources.hs
fosskers: No that's bad. This is in the Mining Coordination logic, which can't crash.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -64,11 +73,21 @@ withMiningCoordination logger enabled cutDb inner
...
             , _coordCutDb = cutDb
-            , _coordState = t
-            }
+            , _coordState = t }
+  where
+    prune :: TVar MiningState -> IO a
+    prune t = do
+        let !d = 600000000  -- 10 minutes
+        threadDelay d
+        ago <- over (_Unwrapped . _Unwrapped) (subtract (int d)) <$> getCurrentTimeIntegral
```
larskuhtz: Are you aware of Using functions from https://github.com/kadena-io/chainweb-node/blob/master/src/Numeric/AffineSpace.hs? Using it may make arithmetic over time values more readable. But it's at matter of style and preference.

src/Chainweb/Chainweb/MinerResources.hs
fosskers: The first isn't possible, but the second might be in a follow-up.

src/Chainweb/Chainweb/MinerResources.hs
fosskers: We might revisit this later.

PR #444, opened 2019-09-17
Coin minimum precision of 12

slpopejoy:
Enforces min precision of 12 decimal places for coin contract amounts.

Also rounds block rewards to 8 places.


pact/coin-contract/coin.pact
```diff
@@ -147,12 +165,24 @@
...
+      (property (> amount 0.0))
+      (property (not (= sender "")))
+      (property (not (= receiver "")))
+    ]
+
     (enforce (not (= sender receiver))
       "sender cannot be the receiver of a transfer")
 
+    (enforce (!= "" sender) "empty sender")
+    (enforce (!= "" receiver) "empty sender")
```
fosskers: This wasn't guarded against elsewhere?

pact/coin-contract/coin.pact
```diff
@@ -57,7 +72,7 @@
     \the transaction, and the price is the spot price of gas at that time.    \
     \The gas buy will be executed prior to executing SENDER's code."
 
-    @model [(property (> total 0.0))]
+    (enforce-unit total)
```
fosskers: Is the minimum defined here applied to all Pact numbers? Say if I compose `transfer` off some math calculation, like in `(transfer "foo" "bar" (/ numer denom))` and the result of the division is something funky and long. Would we expect the `transfer` to fail or to automatically truncate?

src/Chainweb/Miner/Pact.hs
```diff
@@ -165,7 +166,7 @@ readRewards v =
       let
         !n = v ^. chainGraph . to (int . size)
         !m = fromRational $ toRational b
-      in (BlockHeight a, ParsedDecimal $ m / n)
+      in (BlockHeight a, ParsedDecimal $ roundTo 8 (m / n))
```
fosskers: `roundTo` or `floorTo` ?

pact/coin-contract/coin.pact
emilypi: @fosskers No - since this costs gas, the sender of the tx is incentivized to not grief us by sending in empty values.

src/Chainweb/Miner/Pact.hs
emilypi: We agreed that `truncate` was what we wanted - we can use `roundTo' truncate 8 (m / n)` and get this behavior

pact/coin-contract/coin.pact
slpopejoy: I think we could extract a utility function to do the truncate to help smart contract math.

src/Chainweb/Miner/Pact.hs
slpopejoy: It's all pretty arbitrary for rewards ... also, we picked 8 places just to save space vs 12.

pact/coin-contract/coin.pact
slpopejoy: Also FV was busted and we did have some models for this, which once FV was fixed, failed ...

PR #447, opened 2019-09-18
MiningState Statistics

fosskers:
This PR helps us monitor Mining Coordination activity.

### DIFF
--- a/src/Chainweb/Chainweb/MinerResources.hs
+++ b/src/Chainweb/Chainweb/MinerResources.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
@@ -31,9 +32,10 @@ import Data.Tuple.Strict (T3(..))
 import Control.Concurrent (threadDelay)
 import Control.Concurrent.Async (concurrently)
 import Control.Concurrent.STM (atomically)
-import Control.Concurrent.STM.TVar (TVar, modifyTVar', newTVarIO)
+import Control.Concurrent.STM.TVar (TVar, modifyTVar', newTVarIO, readTVar)
 import Control.Lens (over)
 
+import System.LogLevel (LogLevel(..))
 import qualified System.Random.MWC as MWC
 
 -- internal modules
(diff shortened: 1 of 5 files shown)


src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -75,21 +82,38 @@ withMiningCoordination logger enabled cutDb inner
...
+            pure ms
+        count <- readIORef c
+        atomicWriteIORef c 0
+        logFunction logger Info . JsonLog $ MiningStats (M.size ms) count (avgTxs m)
+
+    f :: Time Micros -> T3 a PrevTime b -> Bool
+    f ago (T3 _ (PrevTime p) _) = p > BlockCreationTime ago
+
+    avgTxs :: MiningState -> Int
+    avgTxs (MiningState ms) = summed `div` M.size ms
```
fosskers: `div` should be close enough.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -75,21 +82,38 @@ withMiningCoordination logger enabled cutDb inner
...
-            , _coordLimit = 10000 }
+            , _coordLimit = 2500
+            , _coord503s = c }
   where
-    prune :: TVar MiningState -> IO ()
-    prune t = runForever (logFunction logger) "Chainweb.Chainweb.MinerResources.prune" $ do
-        let !d = 600000000  -- 10 minutes
+    prune :: TVar MiningState -> IORef Int -> IO ()
+    prune t c = runForever (logFunction logger) "Chainweb.Chainweb.MinerResources.prune" $ do
+        let !d = 300000000  -- 5 minutes
```
fosskers: Halved.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -75,21 +82,38 @@ withMiningCoordination logger enabled cutDb inner
...
     | otherwise = do
         t <- newTVarIO mempty
-        fmap snd . concurrently (prune t) $ inner . Just $ MiningCoordination
+        c <- newIORef 0
+        fmap snd . concurrently (prune t c) $ inner . Just $ MiningCoordination
             { _coordLogger = logger
             , _coordCutDb = cutDb
             , _coordState = t
-            , _coordLimit = 10000 }
+            , _coordLimit = 2500
```
fosskers: Quartered.

src/Chainweb/Chainweb/MinerResources.hs
larskuhtz: Would it be better to return the size from before prune? We may be interested in the `max` value.

fosskers: It is already the size from before prune :)

PR #452, opened 2019-09-18
Smooth some UX edges in the  SPV API 

emilypi:
Currently, we take a chainweb chain id as a parameter to the SPV api, but this creates some unsatisfying UX, where the user submits a Pact chain id as a parameter to a cross-chain transfer, but then has to remember to switch from pact to chainweb chain id's (raw word32's) for the SPV api _in the pact service_. This is bad UX. This PR makes it Pact Chain Ids the whole way so that the user only ever specifies pact types as parameters to the pact service.


src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -323,6 +324,7 @@ spvHandler l cutR cid chainR (SpvRequest rk tid) = do
         { errBody = "Index lookup for hash failed: " <> sshow e }
       Right !i -> return i
 
+    tid <- chainIdFromText ptid
```
larskuhtz: this function shouldn't be used directly, because it doesn't check the chain id for validity. I'll take care of this in an upcoming PR.

emilypi: Should I add a TODO to this @larskuhtz?

emilypi: Discussed, and it won't get in his way on this side of things.

PR #455, opened 2019-09-19
Fix #400 (`/poll` returns incorrect results) by fixing underlying payload bug.

### LINKED ISSUE #400: Polling/listening returns wrong tx result
There are instances where polling/listening with a request key returns the result of a different tx result with a different request key. Meanwhile, if you poll with the request key of the tx that was returned, then you get tx result you originally polled for. 

#396 seems to be the result of this problem, where polling tx with request key `SUsdcLI_CvKqngETqbiExnORQqkwalKhDHaiO1HTHn` returned tx result with request key `tIZQR3nXPV63oBezI6xB-asqzxy7oOjF-a2UwC__izM`. When you poll the other way, you can get the correct tx result which is shown below: 
```
{
    "tIZQR3nXPV63oBezI6xB-asqzxy7oOjF-a2UwC__izM": {
        "gas": 0,
        "result": {
            "status": "failure",
            "error": {
                "callStack": [],
                "type": "TxFailure",
                "message": "with-read: row not found: colin",
                "info": ""
            }
        },
        "reqKey": "SUsdcLI_CvKqngETqbiExnORQqkwalKhDHaiO1HTHn8",
        "logs": "RxDuyuxVGwjPnIr-gAUnzHhjkKEwhdVp_lCSfuEkeaI",
        "metaData": null,
        "continuation": null,
        "txId": null
    }
}
```
These behaviors have been witnessed multiple times, and it seems to happen with txs that happen within the same block. We have not recognized all of them. 
Below are pairs of request keys that behave this way:
```
["6ll8e0JPT1fUmVL8PjG6dPkSBHHbZzk2UL-Tv_WKhps","-EM7jDw6o4r9vZRbW7O5rbGjTn-nLkOtETEKxY7GFBg"]
["1Nx1jsfM4dN0-gtfQMLhnDoiiVy006yGpWwwmjnJXss","zBEDsrgmyxFcqfJta44-E5h-8Ul7y01ytGTqM7mZR3U"]
["g_3P5PAbJsbg-tc1FTdGSYuifQP-MLzV08JDLMqsF_Y","jFujgkYK1G-P-G0xJlvaSprP8-AKXd0K91k0KeWbm0g"]
["79pMMq58mQq1vw8WUPLmUSx2SK0C1y4rU3MzGYub3SI","rv7udbEWGRXoS9SyDXK6pdbPMqkxxMS3Sq5qDT867Hs"]
```

gregorycollins:
Callers were expecting `applyPactCmds` to return outputs in the same order as inputs like a `mapM`, and using `zip` and `zipWith` on the results -- unfortunately `applyPactCmds` was returning them in reverse order.

This was noticed by users of the `/poll` handler -- we used cmd hash as a key lookup into `PayloadWithOutputs` to get the cmd result output, but didn't check that the hashes matched. Since the outputs were zipped in reverse order, this means you'd often get results for some other transaction in the block.

Unfortunately, fixing this means breaking genesis. :hankey:


src/Chainweb/Payload.hs
```diff
@@ -774,18 +775,29 @@ instance IsCasValue PayloadWithOutputs where
...
-payloadWithOutputs d co outputs = PayloadWithOutputs
-    { _payloadWithOutputsTransactions = V.zip (_payloadDataTransactions d) outputs
-    , _payloadWithOutputsMiner = _payloadDataMiner d
-    , _payloadWithOutputsCoinbase = co
-    , _payloadWithOutputsPayloadHash = _payloadDataPayloadHash d
-    , _payloadWithOutputsTransactionsHash = _payloadDataTransactionsHash d
-    , _payloadWithOutputsOutputsHash = _payloadDataOutputsHash d
-    }
+payloadWithOutputs d co outputs =
+  if V.length (_payloadDataTransactions d) /= V.length outputs
```
gregorycollins: Lars do we still want this length check? The zip here wasn't an issue length-wise (just that the order was wrong)

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -175,8 +177,9 @@ spvRequests nio = testCaseSteps "spv client tests"$ \step -> do
       let pm = CM.PublicMeta (CM.ChainId "0") "sender00" 100000 0.01 100000 0
 
       tid <- mkChainId version (1::Int)
-      cmd <- liftIO $ mkExec txcode (txdata tid) pm ks (Just "0")
-      return $ SubmitBatch (pure cmd)
+      cmd1 <- liftIO $ mkExec txcode (txdata tid) pm ks (Just "0")
+      cmd2 <- liftIO $ mkExec txcode (txdata tid) pm ks (Just "1")
```
gregorycollins: N.B. putting multiple txs in the block is important for failing the test

src/Chainweb/Pact/PactService.hs
```diff
@@ -775,7 +775,7 @@ applyPactCmds
     -> Miner
     -> PactServiceM cas (Vector HashCommandResult)
 applyPactCmds isGenesis env cmds miner =
-    V.fromList . sfst <$> V.foldM f mempty cmds
+    V.fromList . reverse . sfst <$> V.foldM f mempty cmds
```
emilypi: lets just use `V.foldrM` - then we can do away with the `reverse` call

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -263,24 +259,52 @@ pollWithRetry
...
   sleep 3
   go maxPollRetries
     where
-      go retries = do
-          result <- runClientM (pollApiCmd cmds (Poll (_rkRequestKeys rks))) env
+      go !retries = do
+          let retry failure = do
+                if retries == 0
+                  then do
+                    putStrLn $ "poll failing after " ++ show maxSendRetries
```
emilypi: `putStrLn`'s here and below?

test/Chainweb/Test/Pact/RemotePactTest.hs
emilypi: Thanks

src/Chainweb/Payload.hs
slpopejoy: It's O(1) right? Doesn't hurt ...

src/Chainweb/Pact/PactService.hs
gregorycollins: There isn't one!

test/Chainweb/Test/Pact/RemotePactTest.hs
gregorycollins: There are some already in this location, I didn't change that

slpopejoy: shouldn't this be logging?

src/Chainweb/Payload.hs
larskuhtz: Yes, I think, it's cheap and good to have.

Add `HasCallStack` constraint?

Maybe prefix the message with "Internal code invariant violation: " so that it is clear that this should never happen and if it happens it is a code bug.

src/Chainweb/Payload.hs
```diff
@@ -774,18 +775,29 @@ instance IsCasValue PayloadWithOutputs where
     {-# INLINE casKey #-}
 
 payloadWithOutputs
```
larskuhtz: Add a haddock comment that makes the preconditions of the function explicit: Transaction and output vectors are of same length and have the tx and the matching outputs are at the same index.

test/Chainweb/Test/Pact/RemotePactTest.hs
larskuhtz: Tests shouldn't be chatty. (We have already too many of those.)

There are different ways to report test outputs such that it integrates with tasty's reporting.

The easiest for `IO` scenario tests is to use `testCaseSteps` which provides a "step" call back that logs messages as `tasty` progress reports messages.

larskuhtz: Oh, this is a golden test? Not sure if those provide a logging function.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -263,24 +259,52 @@ pollWithRetry
...
+                  ok <- checkResults (toList rkks) mp
+                  if ok
+                    then do
+                        putStrLn $ concat
+                            [ "poll succeeded after "
+                            , show (maxSendRetries - retries)
+                            , " retries" ]
+                        return $ PollResponses mp
+                    else do
+                        putStrLn "poll check failed, continuing"
```
larskuhtz: If these messages are needed for debugging, maybe put them behind some like:

```
#ifdef DEBUG_TESTS
debug = putStrLn
#else
debug = return ()
#endif
```

`DEBUG_TEST` could be define via the already existing [`debug` cabal flag](https://github.com/kadena-io/chainweb-node/blob/95ac85e95c5a8bd8ce17876bdc0ff44b7c562184/chainweb.cabal#L66).

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -263,24 +259,52 @@ pollWithRetry
...
-                      return result
-                  else do
-                      sleep 1
-                      go (retries - 1)
-              Right _ -> do
-                  putStrLn $ "poll succeeded after " ++ show (maxSendRetries - retries) ++ " retries"
-                  return result
+              Left e -> retry (show e)
+              Right (PollResponses mp) -> do
+                  ok <- checkResults (toList rkks) mp
```
slpopejoy: can we get this to be a list of errors or something instead of the opaque boolean?

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -263,24 +259,52 @@ pollWithRetry
...
+                                         , show rks
+                                         , "\nresult map was: "
+                                         , show mp
+                                         ]
+                        retry msg
+
+      checkCommandResult x cr =
+          _crReqKey cr == x && isOkPactResult (_crResult cr)
+
+      isOkPactResult (PactResult e) = isRight e
```
slpopejoy: Failures are valid results, this shouldn't result in a failed fetch

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -263,24 +259,52 @@ pollWithRetry
...
+                        putStrLn "poll check failed, continuing"
+                        let msg = concat [ "poll check failure. keys: "
+                                         , show rks
+                                         , "\nresult map was: "
+                                         , show mp
+                                         ]
+                        retry msg
+
+      checkCommandResult x cr =
+          _crReqKey cr == x && isOkPactResult (_crResult cr)
```
slpopejoy: the req key should be it's own check with informative error if it fails

test/Chainweb/Test/Pact/RemotePactTest.hs
gregorycollins: This is test code -- and below we do dump the inputs (the result map and the input key) for comparison

test/Chainweb/Test/Pact/RemotePactTest.hs
gregorycollins: Not in this case -- the transactions being input inside this test case are expected to succeed, and their failure would indeed be a reason to fail the test

test/Chainweb/Test/Pact/RemotePactTest.hs
gregorycollins: We dump the inputs if it fails so a debug session should be possible here. Again, this is test code, I'm loath to poke it much further. Emily wants to come in and rewrite this module, it's not a bad idea

PR #456, opened 2019-09-20
SPV UX Edge-Smoothing

emilypi:
This PR unnests the plaintext return value from the SPV Handler


src/Chainweb/RestAPI/Orphans.hs
```diff
@@ -334,6 +334,9 @@ instance ToSchema (TransactionProof SHA512t_256) where
 instance ToSchema (TransactionOutputProof SHA512t_256) where
     declareNamedSchema _ = return $ NamedSchema (Just "TransactionOutputProof") $ byteSchema
 
+instance ToSchema TransactionOutputProofB64 where
+    declareNamedSchema _ = return $ NamedSchema (Just "TransactionOutputProofB64") byteSchema
```
fosskers: Is this proof fixed-length? If so you can use the  `minLength` and `maxLength` lenses to specify that in the Swagger.

emilypi: Discussed, since this is cosmetic, we'll do this separately if at all

PR #459, opened 2019-09-20
Mining Guide Updates

fosskers:
Commits from Rebecca that update the mining guide.


miner/README.org
```diff
@@ -1,9 +1,9 @@
-#+TITLE: Mining for Chainweb
+#+TITLE: Mining for Kadena
```
fosskers: Perhaps "for the Kadena Public Blockchain"?

miner/README.org
```diff
@@ -25,24 +25,24 @@
...
+A blockchain is a series of /blocks/. Blocks contain /transactions/ that
 represent coin transfers or other Smart Contract interactions. Mining is the
 process of mathematically "solving" a block. Unless solved, a block cannot be
 included in the chain. Without mining, a blockchain cannot progress and
 transactions will never finalize.
 
-Economically, mining is the principal way to obtain currency on a Chainweb
-network. Since Chainweb uses Proof-of-Work, computational effort is rewarded
+Economically, mining is the principal way to obtain currency on a Kadena
+network. Since Kadena uses Proof-of-Work, computational effort is rewarded
```
fosskers: Is the Chainweb name going away? It's unclear in this line if "Kadena" refers to the network, the protocol, the company, etc.

miner/README.org
```diff
@@ -25,24 +25,24 @@
...
 with coins associated with each block you solve. The more computational power
 you have, the more likely you are to be successful at mining.
 
-Unlike blockchains that have a single chain, Chainweb has many. Everyone who
+Unlike blockchains that have a single chain, the Kadena public blockchain has many. Everyone who
```
fosskers: Seven syllables in "Kadena public blockchain" is a lot more than the two from "Chainweb".

miner/README.org
```diff
@@ -25,24 +25,24 @@
...
 chooses to mine can spread their effort across multiple chains, advancing each
 chain equally, thereby reducing competition and wasted effort.
 
-* Chainweb Mining via Official Kadena Software
+* Kadena Mining via Official Kadena Software
```
fosskers: We could just remove the name here altogether, and make it:

> Mining via Official Kadena Software

Where here, "Kadena" refers to the company.

miner/README.org
```diff
@@ -60,7 +60,7 @@ Pair before beginning.
 ** Obtaining a Key Pair
 
 The easiest way to obtain a Key Pair is via the [[https://pact.kadena.io/][Pact Web]] application. Keys can
-be generated by the /Wallet/ widget in the bottom-right. Once generated, take
+be generated by the /Wallet/ widget in the bottom right. Once generated, take
```
fosskers: My intuition says that the hyphen is correct, but the internet seems unable to agree. :thinking:

miner/README.org
```diff
@@ -92,7 +92,7 @@ Doing so, we'll see the following configuration of your "mining identity":
     predicate: keys-all
 #+end_src
 
-With values specific to you replaced above (and the rest of Chainweb configured
+With values specific to you replaced above (and the rest of the Kadena public blockchain configured
```
fosskers: In this case, the software is called `chainweb-node`, and that's what we mean here. The "Kadena public blockchain" is the entire network that exists "in the cloud" as a result of everyone running `chainweb-node`s.

miner/README.org
```diff
@@ -207,12 +207,12 @@ it bad luck.
...
 *** I specify ~--chain=...~ but am getting work for other chains. Why?
 
-It is fundamental to the design of Chainweb that chains cannot progress much
-further than their neighbour chains. It may be that by asking for ~--chain=9~,
-the Node couldn't find work to do! In this case, it falls back to picking a
-random chain. This balances the needs of the Miner, who may want a specific
-Chain to progress efficiently, with the needs of the network, which requires all
-chains to grow evenly.
+It is fundamental to the design of the Kadena public blockchain that chains
+cannot progress much further than their neighbor chains. It may be that by
```
fosskers: "Neighbour" is International English spelling. Do we have company consensus on which to use?

miner/README.org
mightybyte: Space looks better to my eye.

miner/README.org
mightybyte: Many of the changes from "Chainweb" to "the Kadena public blockchain" here make sense to me as a branding move.  But this one in particular seems weird.

miner/README.org
mightybyte: `neighbour` is definitely less common in the United States.

miner/README.org
fosskers: Then to get rid of the ambiguity completely we can say `chainweb-node` directly.

miner/README.org
fosskers: Right - do we want US spellings or international ones?

miner/README.org
rebecca-io: Yes! Much better.

miner/README.org
rebecca-io: You are correct. Moving forward, we will no longer be using Chainweb for user/client-facing documents. We can use chainweb, but only when speaking about chainwebs in general. 

For line 34, how would you position that? 'Kadena public blockchain network'?

miner/README.org
rebecca-io: This is okay, as we are wanting to make the transition from 'Chainweb' to 'Kadena public blockchain.'

miner/README.org
rebecca-io: Love it. Let's move forward with that change.

miner/README.org
rebecca-io: From my understanding, hyphens are needed when one or two more words (meant to modify) are placed before a noun. 

**Ex**: Keys can be generated by the Wallet widget in the bottom right vs. The keys are generated by the Wallet widget on the bottom-right corner.

miner/README.org
```diff
@@ -25,24 +25,25 @@
...
 process of mathematically "solving" a block. Unless solved, a block cannot be
 included in the chain. Without mining, a blockchain cannot progress and
 transactions will never finalize.
 
-Economically, mining is the principal way to obtain currency on a Chainweb
-network. Since Chainweb uses Proof-of-Work, computational effort is rewarded
-with coins associated with each block you solve. The more computational power
-you have, the more likely you are to be successful at mining.
+Economically, mining is the principal way to obtain currency on the Kadena
+network. Since we uses Proof-of-Work, computational effort is rewarded with
```
mightybyte: `we uses`

PR #461, opened 2019-09-23
individual clients for chainweb pact endpoints

larskuhtz:


src/Chainweb/Pact/RestAPI/Client.hs
```diff
@@ -16,16 +16,30 @@
 -- and its spv continuation proof endpoints.
 module Chainweb.Pact.RestAPI.Client
 ( pactSpvApiClient
```
emilypi: Do you mind adding `pactSpvApiClient_`?

larskuhtz:
```suggestion
( pactSpvApiClient_
, pactSpvApiClient
```

PR #462, opened 2019-09-23
block update monitor

larskuhtz:


src/Chainweb/CutDB.hs
```diff
@@ -524,13 +527,56 @@ cutStreamToHeaderStream db s = S.for (go Nothing s) $ \(p, n) ->
...
+    . MonadIO m
+    => CutDb cas
+    -> S.Stream (Of Cut) m r
+    -> S.Stream (Of (Either BlockHeader BlockHeader)) m r
+cutStreamToHeaderDiffStream db s = S.for (cutUpdates Nothing s) $ \(p, n) ->
+    S.foldrT
+        (\(cid, a, b) x -> void $ S.mergeOn toOrd x (branch cid a b))
+        (S.each $ zipCuts p n)
+  where
+    cutUpdates :: Maybe Cut -> S.Stream (Of Cut) m r -> S.Stream (Of (Cut, Cut)) m r
```
fosskers: `Stream (Of (a,b))` will cause a leak. Consider making the tuple a `T2`.

node/ChainwebNode.hs
```diff
@@ -195,6 +198,52 @@ runCutMonitor logger db = L.withLoggerLabel ("component", "cut-monitor") logger
             $ S.map (cutToCutHashes Nothing)
             $ cutStream db
 
+data BlockUpdate = BlockUpdate
+    { _blockUpdateBlockHeader :: !BlockHeader
```
larskuhtz: This should be `ObjectEncoded BlockHeader`.

node/ChainwebNode.hs
```diff
@@ -195,6 +198,52 @@ runCutMonitor logger db = L.withLoggerLabel ("component", "cut-monitor") logger
...
+
+    toUpdate :: Either BlockHeader BlockHeader -> IO BlockUpdate
+    toUpdate (Right bh) = BlockUpdate
+        <$> pure bh -- _blockUpdateBlockHeader
+        <*> pure False -- _blockUpdateOrphaned
+        <*> txCount bh -- _blockUpdateTxCount
+    toUpdate (Left bh) = BlockUpdate
+        <$> pure bh -- _blockUpdateBlockHeader
+        <*> pure True -- _blockUpdateOrphaned
+        <*> ((1 -) <$> txCount bh) -- _blockUpdateTxCount
```
larskuhtz:
```suggestion
        <*> ((0 -) <$> txCount bh) -- _blockUpdateTxCount
```

src/Chainweb/CutDB.hs
larskuhtz: I would the streaming steps in this function get fused into to a single loop body. So the lazy tuples shouldn't be an issue here.

I updated the PR to use `T2`, to be on the safe side.

PR #463, opened 2019-09-23
Add a badlist to mempool

gregorycollins:
Currently we don't remember when transactions fail pre-block validation. Since pre-block validation only happens when we call `mempoolGetBlock`, and nodes pick a chain to mine on at random, there is always a window where:

  - we call `mempoolGetBlock`, the transaction gets dropped
  - we initiate a connection with a peer that still has the bad tx
  - the bad tx gets regossiped back to us. Since we don't keep a badlist, we accept it, and happily re-gossip it to other peers.

Lars thinks that a precise mapping of bad transactions (keeping the hashes until they TTL-expire) might be memory-intensive, but we should implement this fix for now and follow up with (perhaps) a probabilistic check later.


src/Chainweb/Mempool/InMem.hs
```diff
@@ -260,16 +267,28 @@ getBlockInMem :: forall t .
 getBlockInMem cfg lock txValidate bheight phash size0 = do
     withMVar lock $ \mdata -> do
         !psq0 <- readIORef $ _inmemPending mdata
+        -- TODO: temporary: delete this in preference for Emmanuel's TTL branch
```
fosskers: Can this be dealt with yet?

src/Chainweb/Mempool/InMem.hs
```diff
@@ -260,16 +267,28 @@ getBlockInMem :: forall t .
...
+        -- which passes block creation time as a parameter
+        now <- getCurrentTimeIntegral
+        !badmap <- pruneBadMap now <$!> readIORef (_inmemBadMap mdata)
         let !psq = HashMap.map decodeTx psq0
-        (!psq', out) <- go psq size0 []
-        let ins !h !t = HashMap.insert (hasher t) t h
-        -- put the pending txs back into the map.
+        (T3 !psq' !badmap' out) <- go psq badmap size0 []
```
fosskers: They probably don't need to be banged if already in a `T3`.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -340,17 +363,18 @@ getBlockInMem cfg lock txValidate bheight phash size0 = do
...
-    go !psq !remainingGas !soFar = do
+       -> IO (T3 (HashMap TransactionHash t) BadMap (Vector t))
+    go !psq !badmap !remainingGas !soFar = do
         nb <- nextBatch psq remainingGas
         if null nb
-          then return (psq, V.concat soFar)
+          then return $! T3 psq badmap (V.concat soFar)
           else do
-            (good, psq') <- validateBatch psq $! V.fromList nb
+            (T3 good psq' badmap') <- validateBatch psq badmap $! V.fromList nb
```
fosskers: Don't need parens here.

src/Chainweb/Mempool/InMem.hs
gregorycollins: no

gregorycollins: ...I expect this one to get in first at this rate

larskuhtz:
This has been tested on devnet over the past > 12h.

mightybyte:
I was going to try to merge this tonight but wasn't quite sure how to resolve the conflict.

src/Chainweb/Mempool/InMem.hs
gregorycollins: ... the time check is being done elsewhere and we are only using 'now' to clean out badmap -- i think this is ok

PR #464, opened 2019-09-24
Header Stream

fosskers:
This PR adds an endpoint that sends a stream of `BlockHeader`s as Server-Sent Events for consumption by our block explorer.

Can be tested with:
```
chainweb-node --chainweb-version=development --port=8082 --log-level=warn --header-stream

curl -k -v https://localhost:8082/chainweb/0.0/development/header/updates
```

### DIFF
--- a/src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
+++ b/src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
@@ -32,6 +32,7 @@ module Chainweb.BlockHeaderDB.RestAPI.Server
 ) where
 
 import Control.Applicative
+import Control.Concurrent.Async (concurrently)
 import Control.Concurrent.Chan (newChan, writeChan)
 import Control.Lens hiding (children, (.=))
 import Control.Monad
@@ -313,8 +314,9 @@ headerStreamServer cdb = headerStreamHandler cdb
 headerStreamHandler :: forall cas. PayloadCas cas => CutDb cas -> Tagged Handler Application
 headerStreamHandler db = Tagged $ \req respond -> do
     chan <- newChan
-    void . SP.mapM_ (g >=> writeChan chan . f) . SP.concat $ blockDiffStream db
-    eventSourceAppChan chan req respond
+    snd <$> concurrently
+        (void . SP.mapM_ (g >=> writeChan chan . f) . SP.concat $ blockDiffStream db)
+        (eventSourceAppChan chan req respond)
   where
     cas :: PayloadDb cas
     cas = _webBlockPayloadStoreCas $ view cutDbPayloadStore db
--- a/src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
+++ b/src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
 
 -- |
 -- Module: Chainweb.BlockHeaderDB.RestAPI.Server
@@ -55,19 +56,23 @@ import qualified Streaming.Prelude as SP
 
 -- internal modules
 
-import Chainweb.BlockHeader (BlockHeader, ObjectEncoded(..))
+import Chainweb.BlockHeader (BlockHeader(..), ObjectEncoded(..))
 import Chainweb.BlockHeader.Validation
 import Chainweb.BlockHeaderDB
 import Chainweb.BlockHeaderDB.RestAPI
 import Chainweb.ChainId
-import Chainweb.CutDB (CutDb, blockDiffStream)
+import Chainweb.CutDB (CutDb, blockDiffStream, cutDbPayloadStore)
+import Chainweb.Payload (PayloadWithOutputs(..))
+import Chainweb.Payload.PayloadStore.Types (PayloadCas, PayloadDb)
 import Chainweb.RestAPI.Orphans ()
 import Chainweb.RestAPI.Utils
+import Chainweb.Sync.WebBlockHeaderStore (_webBlockPayloadStoreCas)
 import Chainweb.TreeDB
 import Chainweb.Utils
 import Chainweb.Utils.Paging hiding (properties)
 import Chainweb.Version
 
+import Data.CAS (casLookupM)
 import Data.Singletons
 
 -- -------------------------------------------------------------------------- --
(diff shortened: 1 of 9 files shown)


src/Chainweb/CutDB/RestAPI/Server.hs
```diff
@@ -67,14 +74,27 @@ cutGetHandler db (Just (MaxRank (Max mar))) = liftIO $ do
 cutPutHandler :: CutDb cas -> CutHashes -> Handler NoContent
 cutPutHandler db c = NoContent <$ liftIO (addCutHashes db c)
 
+headerStreamHandler :: CutDb cas -> Tagged Handler Application
+headerStreamHandler db = Tagged $ \req respond -> do
+    chan <- newChan
+    void . SP.mapM_ (writeChan chan . f) . SP.concat $ blockDiffStream db
```
fosskers: `concat` to throw away all the `Left` values (orphaned headers).

src/Chainweb/CutDB/RestAPI/Server.hs
```diff
@@ -67,14 +74,27 @@ cutGetHandler db (Just (MaxRank (Max mar))) = liftIO $ do
...
+    eventSourceAppChan chan req respond
+  where
+    f :: BlockHeader -> ServerEvent
+    f bh = ServerEvent (Just $ fromByteString "BlockHeader") Nothing
+        [ fromLazyByteString . encode . toJSON $ ObjectEncoded bh ]
```
fosskers: Should probably return more than just the header. Needs stats too.

src/Chainweb/CutDB/RestAPI.hs
```diff
@@ -89,3 +94,14 @@ cutApi = Proxy
 someCutApi :: ChainwebVersion -> SomeApi
 someCutApi (FromSing (SChainwebVersion :: Sing v)) = SomeApi $ cutApi @v
 
+-- -------------------------------------------------------------------------- --
+-- BlockHeader Event Stream
+
+type HeaderStreamApi_ = "header" :> "updates" :> Raw
```
fosskers: Probably the wrong endpoint path.

larskuhtz:
I am not sure if this API should be part of the "standard" chainweb-node. I am concerned that we keep adding features and functionality to the consensus nodes eventually overloading them.

HTTP long polling is potentially expensive, since it allows a client to keep a TLS connection along with a Haskell thread open with verify little resource overhead on the client side.

I wonder providing this kind of information should be the role of dedicated chainweb monitoring nodes.

Or, if we really want just a single chainweb-node application that servers everybody's needs (in addition to providing consensus with best possible performance), which should at least make serving certain non-critical APIs optional (via configuration settings)?

larskuhtz:
I'd like use to rather focus on performance and soundness of the core features of a chainweb-node then adding new non-essential APIs and features.

fosskers:
Lars and I have discussed this and I'll be making certain improvements to ensure this endpoint isn't abused.

src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
```diff
@@ -276,3 +295,35 @@ someBlockHeaderDbServer (SomeBlockHeaderDb (db :: BlockHeaderDb_ v c))
...
+headerStreamServer cdb = headerStreamHandler cdb
+
+headerStreamHandler :: forall cas. PayloadCas cas => CutDb cas -> Tagged Handler Application
+headerStreamHandler db = Tagged $ \req respond -> do
+    chan <- newChan
+    void . SP.mapM_ (g >=> writeChan chan . f) . SP.concat $ blockDiffStream db
+    eventSourceAppChan chan req respond
+  where
+    cas :: PayloadDb cas
+    cas = _webBlockPayloadStoreCas $ view cutDbPayloadStore db
```
fosskers: Causes a payload DB hit per `BlockHeader`!

src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
```diff
@@ -276,3 +295,35 @@ someBlockHeaderDbServer (SomeBlockHeaderDb (db :: BlockHeaderDb_ v c))
...
+
+    g :: BlockHeader -> IO HeaderUpdate
+    g bh = do
+        x <- casLookupM cas $ _blockPayloadHash bh
+        pure $ HeaderUpdate (ObjectEncoded bh) (length $ _payloadWithOutputsTransactions x)
```
fosskers: `Foldable.length` on `Vector` is safe, it's overridden to `V.length` which is `O(1)`.

src/Chainweb/Chainweb.hs
```diff
@@ -265,6 +272,9 @@ pChainwebConfiguration = id
     <*< configCoordinator .:: boolOption_
         % long "mining-coordination"
         <> help "whether to enable external requests for mining work"
+    <*< configHeaderStream .:: boolOption_
+        % long "header-stream"
+        <> help "whether to enable an endpoint for streaming block updates"
```
fosskers: Off by default. Specifying this flag turns it on.

src/Chainweb/RestAPI.hs
```diff
@@ -210,6 +213,8 @@ someChainwebServer v dbs mr =
         <> someP2pServers v (_chainwebServerPeerDbs dbs)
         <> PactAPI.somePactServers v (_chainwebServerPactDbs dbs)
         <> maybe mempty (Mining.someMiningServer v) mr
+        <> maybe mempty (someHeaderStreamServer v)
+               (bool Nothing (Just ()) hs >> _chainwebServerCutDb dbs)
```
fosskers: I wish there were better ergonomics between `Bool` and `Maybe`.

src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
```diff
@@ -276,3 +295,35 @@ someBlockHeaderDbServer (SomeBlockHeaderDb (db :: BlockHeaderDb_ v c))
...
+    :: forall cas (v :: ChainwebVersionT)
+    .  PayloadCas cas
+    => CutDb cas
+    -> Server (HeaderStreamApi v)
+headerStreamServer cdb = headerStreamHandler cdb
+
+headerStreamHandler :: forall cas. PayloadCas cas => CutDb cas -> Tagged Handler Application
+headerStreamHandler db = Tagged $ \req respond -> do
+    chan <- newChan
+    void . SP.mapM_ (g >=> writeChan chan . f) . SP.concat $ blockDiffStream db
```
emilypi: - What does `Tagged` do here? 

- `void` is unnecessary 

- Why split `g` and `f`? Kleisli fish are nice and all but you've defined the functions locally below, and there doesn't seem to be any multiple usages.

src/Chainweb/RestAPI.hs
emilypi: This is definitely a code smell. `Maybe ()` is isomorphic to `Bool`, so this effectively `bool False True <some bool>`

fosskers: `_chainwebServerCutDb` gives a `Maybe`, but you're right that I can just put that in place of `Just ()`.

src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
fosskers: The endpoint that this handler corresponds to ends in `Raw`. This is the strategy that we've discovered for doing Server-Sent Events with Servant.

`void` was indeed necessary.

`f` and `g` are split for cleanliness.

src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
mightybyte: If this proves to be too much overhead, we can consider removing it in the future.

PR #470, opened 2019-09-25
[Checkpointer] Update bug fix

emmanueldenloye:


test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -395,6 +394,41 @@ checkpointerTest name initdata =
...
+          hash06 <- BlockHash <$> merkleLogHash "0000000000000000000000000000006a"
+
+          blockEnv07 <- _cpRestore _cpeCheckpointer (Just (BlockHeight 6, hash15))
+
+          void $ runExec blockEnv07 Nothing "(m6.insertTbl 'b 2)"
+          void $ runExec blockEnv07 Nothing "(m6.updateTbl 'b 3)"
+          void $ runExec blockEnv07 Nothing "(m6.updateTbl 'b 4)"
+
+          runExec blockEnv06 Nothing "(m6.readTbl)" >>= \EvalResult{..} -> Right _erOutput @?= traverse toPactValue [tIntList [1,4]]
+
```
slpopejoy: so, the bug needs to verify the final table value in a subsequent transaction. If you reproduce the bug, it will be 2 or 3, not 4.

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -395,6 +398,62 @@ checkpointerTest name initdata =
...
+          hash08 <- BlockHash <$> merkleLogHash "0000000000000000000000000000008a"
+
+          blockEnv09 <- _cpRestore _cpeCheckpointer (Just (BlockHeight 8, hash07))
+
+          -- FOR DEBUGGING/INSPECTING VALUES AT SPECIFIC KEYS
+          -- void $ runExec blockEnv09 Nothing "(let ((written (at 'col (read m6.tbl 'a [\"col\"])))) (enforce (= written 1) \"key a\"))"
+          -- void $ runExec blockEnv09 Nothing "(let ((written (at 'col (read m6.tbl 'b [\"col\"])))) (enforce (= written 4) \"key b\"))"
+          -- FOR DEBUGGING/INSPECTING VALUES AT SPECIFIC KEYS
+
+          runExec blockEnv09 Nothing "(m6.readTbl)" >>= \EvalResult{..} -> Right _erOutput @?= traverse toPactValue [tIntList [1,4]]
```
mightybyte: It probably won't be obvious to someone coming along later that the 4 here correspond with the 4 on line 438.  In a future PR it might be worth figuring out how to make the relationship more obvious.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -207,7 +207,7 @@ recordPendingUpdate (Utf8 key) (Utf8 tn) txid v = modifyPendingData modf
     deltaKey = SQLiteDeltaKey tn key
 
     modf (a, m, l, p) = (a, upd m, l, p)
-    upd = HashMap.insertWith DL.append deltaKey (DL.singleton delta)
+    upd = HashMap.insertWith const deltaKey (DL.singleton delta)
```
gregorycollins: Just add a comment here that the "const" means we prefer the latest update for the roekey for the current transaction.

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -395,6 +398,62 @@ checkpointerTest name initdata =
 
           _cpDiscard _cpeCheckpointer
 
+          next "2nd mini-regression test for debugging updates (part 1) empty block"
+
```
gregorycollins: kill all the blank lines here

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
emmanueldenloye: done

PR #476, opened 2019-09-27
fix tx lookup for last block

larskuhtz:


src/Chainweb/Pact/PactService.hs
```diff
@@ -866,21 +851,38 @@ execLookupPactTxs restorePoint txs
...
+findLatestValidBlock :: PactServiceM cas (Maybe BlockHeader)
+findLatestValidBlock = getCheckpointer >>= liftIO . _cpGetLatestBlock >>= \case
+    Nothing -> return Nothing
+    Just (height, hash) -> go height hash
+  where
+    go height hash = do
+        bhdb <- view psBlockHeaderDb
+        liftIO (lookup bhdb hash) >>= \case
+            Nothing -> do
+                logInfo $ "Latest block isn't valid."
```
gregorycollins: If we expect this case to be common we should probably reduce priority from `Info`

PR #477, opened 2019-09-27
Perform the same checks at validateBlock time as at newBlock

gregorycollins:


src/Chainweb/Pact/PactService.hs
```diff
@@ -445,15 +441,32 @@ validateChainwebTxsPreBlock dbEnv cp blockOriginationTime bh hash txs = do
...
+    -> Checkpointer
+    -> BlockCreationTime
+    -> BlockHeight
+    -> BlockHash
+    -> Vector ChainwebTransaction
+    -> IO (Vector Bool)
+validateChainwebTxsPreBlock dbEnv cp blockOriginationTime bh hash txs = do
+    lb <- _cpGetLatestBlock cp
+    when (Just (pred bh, hash) /= lb) $
+        fail "internal error: restore point is wrong, refusing to validate."
```
emmanueldenloye: Instead of this being a general `fail`, why don't we make this a `PactInternalError`?

gregorycollins: done

PR #478, opened 2019-09-27
Pact api: don't rewind checkpointer for /local.

gregorycollins:


test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -379,12 +399,14 @@ generatePactApi cwVersion chainid =
 
 apiCmds :: ChainwebVersion -> ChainId -> PactTestApiCmds
 apiCmds cwVersion theChainId =
-    let sendCmd :<|> pollCmd :<|> _ :<|> _ = generatePactApi cwVersion theChainId
-    in PactTestApiCmds sendCmd pollCmd
+    let sendCmd :<|> pollCmd :<|> _ :<|> localCmd = generatePactApi cwVersion theChainId
+    in PactTestApiCmds sendCmd pollCmd localCmd
```
larskuhtz: `Chainweb.Pact.RestAPI.Client` now exports client functions for the pact API. We may consider using those here.

PR #481, opened 2019-09-27
Add failure diagnostics for `/send` call.

gregorycollins:
- split mempool insert out into two functions -- the check, and the insert
  - make mempool insert check chattier about the reason it failed
  - make `/send` call the new insert check first and send the failure reason back as a 400
  - add a test that this works.


test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -158,6 +163,31 @@ responseGolden networkIO rksIO = golden "command-0-resp" $ do
...
+    expectSendFailure $ flip runClientM cenv $ do
+        sendApiCmd (testCmdsChainId sid) batch
+
+  where
+    expectSendFailure act = do
+        m <- (wrap `catch` h)
+        maybe (return ()) (\msg -> assertBool msg False) m
+      where
+        wrap = do
+            let ef out = Just ("expected exception on bad tx, got: "
```
gregorycollins: Note we are handling both exceptions and `Left foo` (what it actually returns) here -- I could change that

PR #486, opened 2019-09-30
[Validations] Adding back greppable properties

emmanueldenloye:
I'm hoping that this PR can be used by others to add back anything that they may have missed while implementing the validations from the Chainweb validations document. If not, then this should be a pretty small PR that may not even have to go through CI (I just added comments).


fosskers:
Should the same be done for the coin contract?

PR #487, opened 2019-09-30
Cumulative Gas Limits

fosskers:
This PR implements the validations regarding cumulative gas counts in a `SubmitBatch` that contains multiple TXs from the same `sender`.


src/Chainweb/Pact/PactService.hs
```diff
@@ -418,40 +422,80 @@ finalizeCheckpointer finalize = do
...
     :: PactDbEnv'
     -> Checkpointer
     -> BlockCreationTime
     -> BlockHeight
     -> Vector ChainwebTransaction
     -> IO (Vector Bool)
-validateChainwebTxs dbEnv cp blockOriginationTime bh txs =
-    V.mapM checkOne txs
+validateChainwebTxs dbEnv cp blockOriginationTime bh txs
+    | bh == 0 = pure $ V.replicate (V.length txs) False
```
gregorycollins: Previous code had `True` here -- does this work? It might prevent genesis txs from validating. Change `$` to `$!`

src/Chainweb/Pact/PactService.hs
```diff
@@ -418,40 +422,80 @@ finalizeCheckpointer finalize = do
...
+    | otherwise = do
+        bs <- balances txs
```
gregorycollins: If a sender sends in _k_ txs, this will perform _k_ balance checks for that sender, and your `gasTotals` runs two passes over the transactions. I was thinking more something like this:

  - do away with the `GasTotals` map (shouldn't be necessary)
  - implement `checkDebit :: Balances -> Text -> Decimal -> IO (Balances, Bool)` that does a forwarding lookup (i.e. check the map first, then check the db) and fails if the account has insufficient balance -- this ensures one DB lookup per sender. 
  - `validate` becomes `validate :: IORef Balances -> ChainwebTransaction -> IO Bool` -- you could use a fold instead if you want but it's hard to efficiently generate the vector in pure code that way.

Alternatively collect `HashSet Text` for sender first and then do the lookup that way -- the latter has the advantage of being future-proof in the event that we get smart and batch the lookup.

Btw now that my `/send` diagnostic branch is in, the `Bool` probably must be `Mempool.InsertError`

src/Chainweb/Pact/PactService.hs
```diff
@@ -418,40 +422,80 @@ finalizeCheckpointer finalize = do
...
+    --
+    balances :: Vector ChainwebTransaction -> IO Balances
+    balances = foldlM balLookup mempty
+
+    balLookup :: Balances -> ChainwebTransaction -> IO Balances
+    balLookup acc tx
+        | HM.member sender acc = pure acc
+        | otherwise = do
+            let pactHash = view P.cmdHash tx
+            mb <- _cpLookupProcessedTx cp pactHash
```
gregorycollins: The dupe check shouldn't be moved down into the balance lookup like this, it's really unexpected! Also the semantics are now wrong, if the dupe check fails we need to fail the validation.

src/Chainweb/Pact/PactService.hs
```diff
@@ -418,41 +421,50 @@ finalizeCheckpointer finalize = do
...
+        sender = P._pmSender pm
+        P.GasLimit (P.ParsedInteger limit) = P._pmGasLimit pm
+        P.GasPrice (P.ParsedDecimal price) = P._pmGasPrice pm
+        limitInCoin = price * fromIntegral limit
 
-    -- skip validation for genesis -- gas accounts / etc don't exist yet
-    checkOne tx = if bh == 0 then return True else checkOneReal tx
-    checkOneReal tx = do
+    checkOne :: P.Command PayloadWithText -> IO Bool
+    checkOne tx = do
```
gregorycollins: P.S. also I think this used to be/was supposed to be

`checkOne = if bh == 0 then const (return True) else checkOneReal`

which would only test bh once

src/Chainweb/Pact/PactService.hs
fosskers: You're right, I got that backwards.

src/Chainweb/Pact/PactService.hs
fosskers: > this will perform k balance checks for that sender

Previously it was doing an `IO`-based balance check per TX. This approach here front-loads that effort. Yes it does mean multiple passes over the `Vector` (which is the conversation of a `SubmitBatch`, which we wouldn't expect to be larger than a few thousand TXs at max), but amortized I think the reduction in `IO` calls is an improvement.

I'll think about the other points.

src/Chainweb/Pact/PactService.hs
fosskers: It effectively does fail it in future steps. Having an entry in the `Balances` map means "we can attempt moving this TX forward."

src/Chainweb/Pact/PactService.hs
fosskers: Once per Transaction, yes. Now it's once per call to `validateChainwebTxs`.

fosskers:
There's a logic mistake I have to fix before this can be merged.

src/Chainweb/Pact/PactService.hs
```diff
@@ -439,42 +439,54 @@ validateChainwebTxs
...
+            Nothing -> pure False
+            Just bs' -> do
+                let !valid = all ($ tx) validations
+                when valid $ writeIORef bsr bs'
+                pure valid
       where
-        gasCheck :: Decimal -> Decimal -> b -> Bool
-        gasCheck b g _ = b >= g
-
+        validations = [checkTimes]
```
gregorycollins: good because I think I'm adding more in here

src/Chainweb/Pact/PactService.hs
```diff
@@ -439,42 +439,54 @@ validateChainwebTxs
...
-        f gt tx
-            | not (HM.member sender bs) = gt
-            | otherwise = HM.insertWith (+) sender limitInCoin gt
-          where
-            pm = P._pMeta . payloadObj $ P._cmdPayload tx
-            sender = P._pmSender pm
-            P.GasLimit (P.ParsedInteger limit) = P._pmGasLimit pm
-            P.GasPrice (P.ParsedDecimal price) = P._pmGasPrice pm
-            limitInCoin = price * fromIntegral limit
+    -- gasTotals :: Balances -> Vector ChainwebTransaction -> GasTotals
```
gregorycollins: clean this up and we are good to go

PR #488, opened 2019-10-01
add x-chainweb-time header

larskuhtz:
This adds a custom HTTP response header to all API endpoints that contains the POSIX timestamp of server.

Clients can check that and decide whether they (or the server) is out of sync. Clients that can't reliably sync there clock can use this header to successfully submit requests.


larskuhtz:
Another way to implement this would have been to add the header to the `ChainwebEndpointApi` type. That would cause the header to show up in servant generated documentation, but it would also require to change all handler functions in the code.

src/Chainweb/RestAPI.hs
```diff
@@ -224,15 +228,27 @@ chainwebApplication
...
     { corsRequestHeaders = simpleHeaders
     }
 
+chainwebTime :: Middleware
+chainwebTime app req resp = app req resp'
+  where
+    resp' res = do
+        timestamp <- realToFrac <$> getPOSIXTime :: IO Double
+        resp $ mapResponseHeaders
+            ((:) ("X-Chainweb-Time", B8.pack (printf "%.0f" timestamp)))
```
gregorycollins: a) avoid `Double` for timestamps -- we already have a dep on the clock package so here I would prefer http://hackage.haskell.org/package/clock-0.8/docs/src/System.Clock.html#getTime with CLOCK_REALTIME_COARSE -- easy to convert the resulting `TimeSpec` to `Ratio Int64` for `printf`, or do the job by hand with `div` on the `nsec` field of the timespec

b) the header should be called "X-Server-Timestamp" (and the bike shed should be blue)

larskuhtz: It seems that `CLOCK_REALTIME_COARSE` is available only on linux (or at least it isn't obvious from the package code if it is also available on other platforms).

There's also no code for easy detection or graceful fallback available. So, I'll stick with `Realtime`.

I don't think that the `RealTime` clock is (much) faster than `getPOSIXTime` from `time` and we already use `time` in the same code path. But happy to use clock, instead.

src/Servant/Client_.hs
```diff
@@ -0,0 +1,118 @@
...
+    ( Functor, Applicative, Monad, MonadIO
+    , Generic
+    , MonadError ClientError, MonadThrow , MonadCatch
+    )
+
+instance MonadReader ClientEnv ClientM_ where
+    ask = ClientM_ $ lift ask
+    local f = ClientM_ . mapReaderT (local f) . unClientM_
+    {-# INLINE ask #-}
+    {-# INLINE local #-}
```
emilypi: isn't this the derived instance for `ClientM_`?

src/Servant/Client_.hs
```diff
@@ -0,0 +1,118 @@
...
+-- | Locally overwrite 'Request's made the inner computation.
+--
+modifyRequest :: (Request -> IO Request) -> ClientM_ a -> ClientM_ a
+modifyRequest f = ClientM_ . local (set modReq f) . unClientM_
+{-# INLINE modifyResponse #-}
+
+-- | Locally overwrite 'Response's received the inner computation.
+--
+modifyResponse :: (Response -> IO Response) -> ClientM_ a -> ClientM_ a
+modifyResponse f = ClientM_ . local (set modRes f) . unClientM_
```
emilypi: same as above - isn't this just `local (set modRes f)` for `ClientM_`?

test/Chainweb/Test/RestAPI.hs
```diff
@@ -186,6 +229,103 @@ simpleClientSession envIO cid =
                 (Expected h)
                 (Actual r)
 
+        -- brancheHeaders
```
emilypi: typo

test/Chainweb/Test/RestAPI.hs
```diff
@@ -186,6 +229,103 @@ simpleClientSession envIO cid =
...
+
+        forM_ (newHeaders `zip` drop 2 newHeaders) $ \(h0, h1) -> do
+            void $ liftIO $ step "branchHeadersClient: get two block headers with lower and upper bound"
+            bhs5 <- branchHeadersClient version cid Nothing Nothing Nothing Nothing
+                (BranchBounds (HS.singleton (LowerBound $ key h0)) (HS.singleton (UpperBound $ key h1)))
+            assertExpectation "branchHeadersClient returned wrong number of entries"
+                (Expected 2)
+                (Actual $ _pageLimit bhs5)
+
+        -- brancheHeaders
```
emilypi: (see above)

test/Chainweb/Test/RestAPI.hs
larskuhtz:
```suggestion
        -- branchHeaders
```

test/Chainweb/Test/RestAPI.hs
larskuhtz:
```suggestion
        -- branchHeaders
```

src/Servant/Client_.hs
larskuhtz: `ClientM_` implements `MonadReader ClientEnv`, i.e. it exposes the environment of the underlying `ClientM` monad. 

The fact that `ClientEnv_` itself is a `ReaderT Mods` is a hidden implementation detail and not exported.

src/Servant/Client_.hs
larskuhtz: `ClientM_` implements `MonadReader ClientEnv`. In order to access the `Mods` environment, we have to unwrap it to get the internal `ReaderT Mods`.

PR #489, opened 2019-10-01
Bump Pact Version 

emilypi:
This bump gets us the following

- Network Id's 
- The caplist work 
- Charset natives + native consts
- Removed `ByteString` orphans

This breaks the following: 

- Literally everything because so much of our work was relying on those `ByteString` instances!

I've fixed up everything to be basically the same encoding, just done explicitly. If you want it done differently, lets discuss a best solution for each of the affected parts of the code.


src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -361,12 +360,12 @@ spvHandler l cutR cid chainR (SpvRequest rk (Pact.ChainId ptid)) = do
     bdb = _chainResBlockHeaderDb chainR
     pdb = view CutDB.cutDbPayloadCas cdb
     b64 = TransactionOutputProofB64
-      . LBS.toStrict
+      . decodeUtf8
+      . BSL8.toStrict
       . Base64.encode
```
larskuhtz: Everywhere else in chain web we use unpadded base64. Shouldn't we be consistent throughout the chainweb API?

larskuhtz: In `Chainweb.Utils` we already have `encodeB64UrlText` (or `encodeB64UrlTextNoPadding`, if we want to be consistent).

emilypi: In Pact we're using unpadded Base64 so yes, definitely we should standardize on this.

src/Chainweb/Mempool/RestAPI/Server.hs
```diff
@@ -32,15 +32,20 @@ import Chainweb.Utils
...
+
+    go h = case decode h of
+      Left e -> throwM . DecodeException $ T.pack e
+      Right t -> return t
+
     begin = do
-        txs <- mapM decode txsBS
-        let txV = V.fromList txs
-        liftIO $ mempoolInsert mempool CheckedInsert txV
+      txs <- mapM go txsT
```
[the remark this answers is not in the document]
emilypi: updated, thanks

src/Chainweb/Mempool/RestAPI/Client.hs
```diff
@@ -131,9 +139,14 @@ lookupClient
 lookupClient txcfg v c txs = do
     SomeChainwebVersionT (_ :: Proxy v) <- return $ someChainwebVersionVal v
     SomeChainIdT (_ :: Proxy c) <- return $ someChainIdVal c
-    let decode = either (throw . DecodeException . T.pack) return . codecDecode (txCodec txcfg)
     cs <- lookupClient_ @v @c txs
-    mapM (traverse decode) cs
+    mapM (traverse go) cs
+  where
+    go h = case decode h of
+      Left e -> throw . DecodeException $ T.pack e
```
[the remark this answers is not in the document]
emilypi: In the original code it was `throw`, but we're in something `~ IO` so it should be the same. I'll use `throwM`

src/Chainweb/Pact/RestAPI/Server.hs
larskuhtz: But this code is using padded Base64.

emilypi: Switched to yours 😄

PR #490, opened 2019-10-02
Add mempool/pact-send chainId validation check.

gregorycollins:


src/Chainweb/Pact/Utils.hs
```diff
@@ -17,22 +17,29 @@ module Chainweb.Pact.Utils
...
+import Pact.Types.Command
 
 import Chainweb.BlockHeader
+import Chainweb.ChainId
 import Chainweb.Pact.Backend.Types
 import Chainweb.Time
 import Chainweb.Transaction
 
+fromPactChainId :: MonadThrow m => P.ChainId -> m ChainId
+fromPactChainId (P.ChainId t) = chainIdFromText t
```
emilypi: @larskuhtz You good with this? iirc you had a version of chainweb chainid to textual pact chain id that was what you wanted in a different PR

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -180,16 +182,23 @@ localTest iot nio = do
...
+        SubmitBatch batch2 <- testBatch' (return $ Time $ TimeSpan 0) 2 mv
+        let batch = SubmitBatch $ batch1 <> batch2
+        sid <- mkChainId version (0 :: Int)
+        expectSendFailure $ flip runClientM cenv $ do
+            sendApiCmd (testCmdsChainId sid) batch
+
+        step "check sending mismatched chain id"
+        batch3 <- testBatch'' "40" iot 20000 mv
+        expectSendFailure $ flip runClientM cenv $ do
+            sendApiCmd (testCmdsChainId sid) batch3
```
emilypi: 👍

src/Chainweb/Pact/Utils.hs
larskuhtz: It's actually an "unsafe" construction of `ChainId` but it's fine for now. 

This function should actually be called `uncheckedChainIdFromText` and `chainIdFromText` should be

```haskell
chainIdFromText :: MonadThrow m => HasChainwebVersion v => v -> T.Text -> m ChainId
```

I have a pending branch that locks down construction of chainIds, that I hope to find some time to finish soon.

larskuhtz: So 👍 for now, since there is still no safe/checked version of `chainIdFromText`.

PR #492, opened 2019-10-02
Reject Payloads by Version Incompat

fosskers:
This PR prevents Payloads that report the wrong `ChainwebVersion` to be rejected before even reaching the mempool.

### DIFF
--- a/src/Chainweb/Pact/RestAPI/Server.hs
+++ b/src/Chainweb/Pact/RestAPI/Server.hs
@@ -136,14 +136,15 @@ pactServer (cut, chain) =
     pactApiHandlers :<|> pactSpvHandler
   where
     cid = FromSing (SChainId :: Sing c)
+    v = FromSing (SChainwebVersion :: Sing v)
     mempool = _chainResMempool chain
     logger = _chainResLogger chain
 
     pactApiHandlers
-      = sendHandler logger mempool
+      = sendHandler logger v mempool
       :<|> pollHandler logger cut cid chain
       :<|> listenHandler logger cut cid chain
-      :<|> localHandler logger cut cid chain
+      :<|> localHandler logger v cut cid chain
 
     pactSpvHandler = spvHandler logger cut cid chain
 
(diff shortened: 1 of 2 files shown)


src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -425,12 +428,16 @@ internalPoll cutR cid chain cut requestKeys0 = do
...
+        | length (_cmdSigs cmd) > 100 -> Left "More than 100 signatures given."
+        | payloadVer cmd /= Just v -> Left "Incompatible ChainwebVersion given."
+        | otherwise -> Right (mkPayloadWithText <$> cmd)
     ProcFail err -> Left err
   where
     cmdBS :: Command ByteString
     cmdBS = encodeUtf8 <$> cmdText
+
+    payloadVer :: Command (Payload m c) -> Maybe ChainwebVersion
+    payloadVer = (>>= fromText . Pact._networkId) . _pNetworkId . _cmdPayload
```
emilypi: i have a util in `TransactionExec` that does the first part of this - see `networkIdOf`

src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -425,12 +428,16 @@ internalPoll cutR cid chain cut requestKeys0 = do
...
 
-validateCommand :: Command Text -> Either String ChainwebTransaction
-validateCommand cmdText = case verifyCommand cmdBS of
+validateCommand :: ChainwebVersion -> Command Text -> Either String ChainwebTransaction
+validateCommand v cmdText = case verifyCommand cmdBS of
     ProcSucc cmd
-        | length (_cmdSigs cmd) <= 100 -> Right (mkPayloadWithText <$> cmd)
-        | otherwise -> Left "More than 100 signatures given."
+        | length (_cmdSigs cmd) > 100 -> Left "More than 100 signatures given."
+        | payloadVer cmd /= Just v -> Left "Incompatible ChainwebVersion given."
```
gregorycollins: FYI -- I put the `ChainId` check into `Chainweb.Chainweb`. They should probably agree, i.e. either you move this check there or you move that check here?

fosskers: I will resolve the locations of these checks in a follow-up, and ensure that they're occurring everywhere they're supposed to be.

PR #494, opened 2019-10-03
avoid usage of Chan in eventsource handler

larskuhtz:
We observe that nodes, that have the block streaming endpoint enabled on devnet, occasionally become unresponsive with max CPU load.

The current handler uses an unbounded buffer between the `blockDiffStream` and the event source handler request thread.

This PR changes the control flow to be demand driven. The 'Chan' and separate thread is replaced by an `IORef` that holds the current state of the `blockDiffStream`.

Since the `blockDiffStream` is backed by and STM `TVar` for obtaining new cuts, the `IO ServerEvent` call back blocks until a new cut becomes available. If, for some reason sending data becomes slow, the stream would just skip a few cuts instead of buffering an unbounded amount of blocks. On the next call `IO ServerEvent` the `blockDiffStream` would compute the intermediate blocks. 

This can, however, result in higher latencies in the deliver of blocks and temporary increase of server side resources, since the blocks for the skipped cuts must be fetched from the database. If this becomes in issue, we could close the connection if the event source falls too much behind (the difference between two cuts becomes too big).


src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
```diff
@@ -313,11 +313,14 @@ headerStreamServer cdb = headerStreamHandler cdb
 
 headerStreamHandler :: forall cas. PayloadCas cas => CutDb cas -> Tagged Handler Application
 headerStreamHandler db = Tagged $ \req respond -> do
-    chan <- newChan
-    snd <$> concurrently
-        (SP.mapM_ (g >=> writeChan chan . f) . SP.concat $ blockDiffStream db)
-        (eventSourceAppChan chan req respond)
+    streamRef <- newIORef $ SP.map f $ SP.mapM g $ SP.concat $ blockDiffStream db
```
fosskers: Why not compose `f` and `g` within the `mapM`?

larskuhtz: Wouldn't that be the same? May understanding is that `SP.map f . SP.mapM g` get's fused into `SP.mapM (fmap f . g)`. But I find the former easier to read.

PR #496, opened 2019-10-04
Consolidate one-off Transaction Validations

fosskers:
This PR moves one-off Transaction validations from near the `send/` handler to within the official Mempool "preinsertion" check logic. In the process, this PR reworks that area of the code to:

- short-circuit at the first sight of validation failure
- avoid repeated iterations over `Vector`s
- avoid repeatedly hashing a `ChainwebTransaction` to get its `TransactionHash`

Overall, we should see an improvement in TX validation performance.


src/Chainweb/Mempool/InMem.hs
```diff
@@ -201,60 +206,106 @@ markValidatedInMem :: MVar (InMemoryMempoolData t)
...
-        if chk tx
-          then runPreChecks chks tx
-          else Just reason
+    notInBadMap :: Either InsertError ()
+    notInBadMap = maybe (Right ()) (const $ Left InsertErrorBadlisted) $ HM.lookup h badmap
+
+-- | Validation: Similar to `insertCheckInMem`, but does not short circuit.
+-- Instead, bad transactions are filtered out and the successful ones are kept.
+--
+insertCheckInMem'
```
fosskers: This variation is used below in `insertInMem`, who doesn't want to fail fast, and instead wants to insert all the TXs that passed anyway, even if there were some failures in the group.

src/Chainweb/Mempool/InMemTypes.hs
```diff
@@ -47,8 +47,10 @@ data InMemConfig t = InMemConfig {
     _inmemTxCfg :: {-# UNPACK #-} !(TransactionConfig t)
   , _inmemTxBlockSizeLimit :: !GasLimit
   , _inmemMaxRecentItems :: {-# UNPACK #-} !Int
-    -- Here Nothing means 'OK to insert'
-  , _inmemPreInsertCheck :: !(V.Vector t -> IO (V.Vector (Maybe InsertError)))
+  , _inmemPreInsertPureChecks :: t -> Either InsertError t
+  , _inmemPreInsertBatchChecks
+        :: V.Vector (TransactionHash, t)
+        -> IO (V.Vector (Either (TransactionHash, InsertError) (TransactionHash, t)))
```
fosskers: One-off and batch checks are now split. This increases the chance of being able to short-circuit.

test/Chainweb/Test/Mempool/InMem.hs
```diff
@@ -20,16 +20,17 @@ tests = testGroup "Chainweb.Mempool.InMem"
             $ Chainweb.Test.Mempool.tests
             $ MempoolWithFunc wf
   where
+    wf :: (InsertCheck -> MempoolBackend MockTx -> IO a) -> IO a
     wf f = do
-        mv <- newMVar (\v -> return $! V.map (const Nothing) v)
-        let cfg = InMemConfig txcfg mockBlockGasLimit 2048 (checkMv mv)
+        mv <- newMVar (pure . V.map Right)
+        let cfg = InMemConfig txcfg mockBlockGasLimit 2048 Right (checkMv mv)
```
fosskers: Here and elsewhere I'm concerned that the one-off validations are no longer checked, and are reported as passing regardless of their contents.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -201,60 +206,106 @@ markValidatedInMem :: MVar (InMemoryMempoolData t)
...
+    -- We hash the tx here and pass it around around to avoid needing to repeat
+    -- the hashing effort.
+    let withHashes :: Either (TransactionHash, InsertError) (Vector (TransactionHash, t))
+        withHashes = for txs $ \tx ->
+          let !h = hasher tx
+          in bimap (h,) (h,) $ validateOne cfg badmap now tx h
+
+    case withHashes of
+        Left _ -> pure $ void withHashes
+        Right r -> void . sequenceA <$> _inmemPreInsertBatchChecks cfg r
```
fosskers: As mentioned below, the one-off checks and batch checks are now separated. Within `withHashes`, we fail utterly at the first sign of *any* TX not passing validation. If all pass, then we move on to the batch checks, which unfortunately have a semi-inefficient `sequenceA` (at least w.r.t. short-circuiting).

stack.yaml
```diff
@@ -6,6 +6,7 @@ extra-deps:
   # --- Missing from Stackage --- #
   - QuickCheck-GenT-0.2.0
   - bloomfilter-2.0.1.0
+  - compactable-0.1.2.3
```
emilypi: since this is just for `fforMaybe`, I say we just vendor that code in an monomorphize it at its call site rather than adding the whole library as a dep

src/Chainweb/Mempool/InMem.hs
```diff
@@ -201,60 +206,106 @@ markValidatedInMem :: MVar (InMemoryMempoolData t)
...
+-- the first detection of any validation failure on any Transaction.
+--
 insertCheckInMem
-    :: InMemConfig t    -- ^ in-memory config
+    :: forall t
+    .  InMemConfig t    -- ^ in-memory config
     -> MVar (InMemoryMempoolData t)  -- ^ in-memory state
     -> Vector t  -- ^ new transactions
-    -> IO (Vector (TransactionHash, Maybe InsertError))
+    -> IO (Either (TransactionHash, InsertError) ())
```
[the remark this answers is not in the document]
fosskers: I moved away from "failure in the positive" on purpose. I've been down that road a few times, and judged it not worth it. Yes while `Either e ()` is technically isomorphic to `Maybe e`, they have different Monad semantics, and it's easier for people to understand the intent of the `Either`.

stack.yaml
fosskers: With proper compilation with `-split-sections` (or even a post-compile `strip` call) this doesn't become so much of an issue - all the dead code is removed.

src/Chainweb/Chainweb.hs
```diff
@@ -367,51 +370,63 @@ withChainweb c logger rocksDb dbDir resetDb inner =
...
-                in Just (combine heads, tails)
+        let !pay = payloadObj . P._cmdPayload $ tx
+            pcid = P._pmChainId $ P._pMeta pay
+            sigs = length (P._cmdSigs tx)
+            ver  = P._pNetworkId pay >>= fromText @ChainwebVersion . P._networkId
+        tcid <- note (MP.InsertErrorOther "Unparsable ChainId") $ fromPactChainId pcid
+        if | tcid /= cid   -> Left MP.InsertErrorMetadataMismatch
+           | sigs > 100    -> Left $ MP.InsertErrorOther "Too many signatures"
+           | ver /= Just v -> Left MP.InsertErrorMetadataMismatch
+           | otherwise     -> Right tx
```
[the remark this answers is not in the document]
fosskers: Well, it has the following `Show` instance:
```
    show InsertErrorMetadataMismatch =
        "Transaction metadata (chain id, chainweb version) conflicts with this \
        \endpoint"
```
which may be sufficient.

src/Chainweb/Chainweb.hs
```diff
@@ -367,51 +370,63 @@ withChainweb c logger rocksDb dbDir resetDb inner =
...
+        let hashes = V.map (toPactHash . fst) txs
         pex <- readMVar mv
-        out1 <- V.map toDupeResult <$>
-                (_pactLookup pex (Left cid) hashes >>= either throwM return)
-        out2 <- V.mapM checkMetadata txs
-        -- N.B. this could just be "zipWith" but I wanted to generalize it to
-        -- future-proof adding new checks here
-        return $! zipAllWith [out1, out2] (getFirst . mconcat . map First)
+        rs <- _pactLookup pex (Left cid) hashes >>= either throwM pure
+        pure $ V.zipWith (\r (h,t) -> maybe (Right (h,t)) (Left . (h,)) $ toDupeResult r) rs txs
```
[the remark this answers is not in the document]
fosskers: Done.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -201,60 +206,106 @@ markValidatedInMem :: MVar (InMemoryMempoolData t)
...
+-- Instead, bad transactions are filtered out and the successful ones are kept.
+--
+insertCheckInMem'
+    :: forall t
+    .  InMemConfig t    -- ^ in-memory config
+    -> MVar (InMemoryMempoolData t)  -- ^ in-memory state
+    -> Vector t  -- ^ new transactions
+    -> IO (Vector (TransactionHash, t))
+insertCheckInMem' cfg lock txs = do
+    now <- getCurrentTimeIntegral
```
gregorycollins: It's sort of a shame that this is two functions that do roughly the same thing now. Can you refactor with a policy argument (short circuiting, or not?) so they can share implementation?

src/Chainweb/Mempool/InMem.hs
```diff
@@ -456,7 +507,7 @@ getPendingInMem cfg nonce lock since callback = do
 
 ------------------------------------------------------------------------------
 clearInMem :: MVar (InMemoryMempoolData t) -> IO ()
-clearInMem lock = do
+clearInMem lock =
```
gregorycollins: I don't know why lint tools complain about these "superfluous" `do` statements. OK, you've cluttered up this PR by "cleaning it up", but now if someone else has to come in and extend this function the `do` needs to come back in.

To me, from a turd-polishing perspective the PR noise triggers my OCD as much or even more than the harmless lingering `do` does.

stack.yaml
gregorycollins: Agree with Emily -- pull in the definition of the combinator you need vs. making us link with yet another abstruse utilities library

test/Chainweb/Test/Mempool.hs
```diff
@@ -1,9 +1,12 @@
--- | Tests and test infrastructure common to all mempool backends.
+{-# LANGUAGE TupleSections #-}
```
gregorycollins: Sort the `LANGUAGE` list (now here's my OCD talking)

test/Chainweb/Test/Mempool.hs
```diff
@@ -15,6 +18,7 @@ module Chainweb.Test.Mempool
   , lookupIsMissing
   ) where
 
+import Data.Bifunctor (bimap)
```
gregorycollins: Sort import list

test/Chainweb/Test/Mempool.hs
```diff
@@ -102,15 +106,22 @@ arbitraryGasPrice :: Gen GasPrice
...
       emptyMeta = TransactionMetadata zero Time.maxTime
       zero = Time.Time (Time.TimeSpan (Time.Micros 0))
 
-type InsertCheck = MVar (Vector MockTx -> IO (Vector (Maybe InsertError)))
+type BatchCheck =
+    Vector (TransactionHash, MockTx)
+    -> IO (V.Vector (Either (TransactionHash, InsertError) (TransactionHash, MockTx)))
+
+-- TODO Why is this inside an MVar?
+type InsertCheck = MVar BatchCheck
```
gregorycollins: Remove TODO and add this explanation. We have an MVar so tests can override/modify the instance's insert check for tests. To make quickcheck testing not take _forever_ for remote we run the server and client mempools in a resource pool -- the test pulls an already running instance from the pool so we need a way to twiddle its check function for the tests.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -416,21 +468,20 @@ getPendingInMem cfg nonce lock since callback = do
...
+        let keys = HM.keys psq
         (dl, sz) <- foldlM go initState keys
         void $ sendChunk dl sz
 
-    sendSome psq rlog (rNonce, oHw) = do
-        if rNonce /= nonce
-          then sendAll psq
-          else sendSince psq rlog oHw
+    sendSome psq rlog (rNonce, oHw)
+        | rNonce /= nonce = sendAll psq
```
gregorycollins: This is more unhelpful churn (re replacing the `if` with guards). Both sides are semantically identical, one doesn't read any better than the other, and use of `otherwise` causes false negatives in test coverage reports

test/Chainweb/Test/Mempool/InMem.hs
gregorycollins: a) we test the low-level check code by populating the `MVar` with a check
b) some specific checks on chainweb transactions (`Mempool` is intentionally abstract in the transaction type for mock/testability) are performed in the remote pact tests -- but we need to make sure we cover this handler thoroughly

test/Chainweb/Test/Mempool/InMem.hs
```diff
@@ -20,16 +20,17 @@ tests = testGroup "Chainweb.Mempool.InMem"
...
         InMem.withInMemoryMempool cfg $ f mv
 
+    checkMv :: MVar (t -> IO b) -> t -> IO b
     checkMv mv xs = do
         f <- readMVar mv
         f xs
 
-    txcfg = TransactionConfig mockCodec hasher hashmeta mockGasPrice mockGasLimit
-                              mockMeta
+    txcfg = TransactionConfig mockCodec hasher hashmeta mockGasPrice mockGasLimit mockMeta
```
gregorycollins: House style is to wrap at 80 col -- this line is 89 now

src/Chainweb/Chainweb.hs
```diff
@@ -367,51 +372,67 @@ withChainweb c logger rocksDb dbDir resetDb inner =
...
     -> MVar PactExecutionService
-    -> Mempool.InMemConfig ChainwebTransaction
-validatingMempoolConfig cid mv = Mempool.InMemConfig
-    { Mempool._inmemTxCfg = txcfg
-    , Mempool._inmemTxBlockSizeLimit = blockGasLimit
-    , Mempool._inmemMaxRecentItems = maxRecentLog
-    , Mempool._inmemPreInsertCheck = preInsertCheck
+    -> MP.InMemConfig ChainwebTransaction
+validatingMempoolConfig cid v mv = MP.InMemConfig
+    { MP._inmemTxCfg = txcfg
```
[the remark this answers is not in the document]
fosskers: Within the context of Chainweb it should be obvious what `MP` means, and we use the one-or-two-letter convention basically everywhere (e.g. `T` for Text, `BL` for lazy bytestrings, `HM`, for HashMap, etc.)

I can switch it back if you prefer.

src/Chainweb/Mempool/InMem.hs
fosskers: Yes, although the return types are quite different.

test/Chainweb/Test/Mempool.hs
fosskers: I'm surprised `stylish-haskell` didn't catch that :thinking:

PR #500, opened 2019-10-06
Use bounded queue for pact service

larskuhtz:
* [x] Replace `TQueue` with `TBQueue`.
* [ ] Add telemetry to monitor statistics of the size of the pact queue

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
@@ -8,6 +9,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeSynonymInstances #-}
 
 -- |
 -- Module: Chainweb.Pact.PactService
@@ -35,7 +37,9 @@ module Chainweb.Pact.PactService
     , minerReward
     ) where
 ------------------------------------------------------------------------------
-import Control.Concurrent
+import Control.Concurrent.MVar
+import Control.Concurrent.Async
+import Control.Exception (SomeAsyncException)
 import Control.Lens
 import Control.Monad
 import Control.Monad.Catch
@@ -248,7 +252,8 @@ initializeCoinContract v cid pwo = do
 
 -- | Loop forever, serving Pact execution requests and reponses from the queues
 serviceRequests
-    :: PayloadCas cas
+    :: forall cas
+    . PayloadCas cas
     => MemPoolAccess
     -> PactQueue
     -> PactServiceM cas ()
@@ -273,7 +278,7 @@ serviceRequests memPoolAccess reqQ = do
             ValidateBlockMsg ValidateBlockReq {..} -> do
                 tryOne' "execValidateBlock"
                         _valResultVar
-                        (return . flip validateHashes _valBlockHeader)
+                        (flip validateHashes _valBlockHeader)
                         (execValidateBlock _valBlockHeader _valPayloadData)
                 go
             LookupPactTxsMsg (LookupPactTxsReq restorePoint txHashes resultVar) -> do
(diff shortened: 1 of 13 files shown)


src/Chainweb/Pact/Service/PactInProcApi.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
```
fosskers: Necessary?

test/Chainweb/Test/Pact/ChainData.hs
```diff
@@ -157,14 +156,14 @@ withPact
...
-    -> (IO (TQueue RequestMsg) -> TestTree)
+    -> (IO PactQueue -> TestTree)
     -> TestTree
 withPact iopdb iobhdb mempool iodir f =
     withResource startPact stopPact $ f . fmap snd
   where
     startPact = do
         mv <- newEmptyMVar
-        reqQ <- atomically newTQueue
+        reqQ <- atomically (newTBQueue 2000)
```
fosskers: Shouldn't this use the `pactQueueSize` value defined above?

test/Chainweb/Test/Pact/PactInProcApi.hs
```diff
@@ -80,13 +79,13 @@ tests = ScheduledTest label $ withRocksResource $ \rocksIO ->
...
     -> MemPoolAccess
-    -> (IO (TQueue RequestMsg) -> TestTree)
+    -> (IO PactQueue -> TestTree)
     -> TestTree
 withPact rocksIO mempool f = withResource startPact stopPact $ f . fmap snd
   where
     startPact = do
         mv <- newEmptyMVar
-        reqQ <- atomically newTQueue
+        reqQ <- atomically (newTBQueue 2000)
```
fosskers: Here too.

test/Chainweb/Test/Pact/PactReplay.hs
```diff
@@ -286,14 +285,14 @@ withPact
...
-    -> (IO (TQueue RequestMsg) -> TestTree)
+    -> (IO PactQueue -> TestTree)
     -> TestTree
 withPact logLevel iopdb iobhdb mempool iodir f =
     withResource startPact stopPact $ f . fmap snd
   where
     startPact = do
         mv <- newEmptyMVar
-        reqQ <- atomically newTQueue
+        reqQ <- atomically (newTBQueue 2000)
```
fosskers: Here too.

test/Chainweb/Test/Pact/TTL.hs
```diff
@@ -117,14 +116,14 @@ withPact
...
-    -> (IO (TQueue RequestMsg) -> TestTree)
+    -> (IO PactQueue -> TestTree)
     -> TestTree
 withPact iopdb iobhdb mempool iodir f =
     withResource startPact stopPact $ f . fmap snd
   where
     startPact = do
         mv <- newEmptyMVar
-        reqQ <- atomically newTQueue
+        reqQ <- atomically (newTBQueue 2000)
```
fosskers: Here too.

fosskers:
Otherwise looks good. If we attempt a `addRequest :: PactQueue -> RequestMsg -> IO ()` when the Queue is full, will it just block on the IO action?

src/Chainweb/Pact/PactService.hs
```diff
@@ -284,20 +288,67 @@ serviceRequests memPoolAccess reqQ = do
...
+    tryOne which mvar m = tryOne' which mvar Right m
+
+    tryOne'
+        :: String
+        -> MVar (Either PactException b)
+        -> (a -> Either PactException b)
+        -> PactServiceM cas a
+        -> PactServiceM cas ()
+    tryOne' which mvar post m =
+        (evalPact (post <$> m) >>= (liftIO . putMVar mvar))
```
emilypi: `evalPact` the name clashes with some Pact constructs that we use. Can we use a different name for this (e.g. `eval_`)?

src/Chainweb/Pact/Service/PactInProcApi.hs
```diff
@@ -1,10 +1,12 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
```
emilypi: unused

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -547,24 +547,22 @@ withPact
...
   where
     startPact = do
         mv <- newEmptyMVar
-        reqQ <- atomically newTQueue
+        reqQ <- atomically $ newTBQueue pactQueueSize
         pdb <- iopdb
         bhdb <- iobhdb
         dir <- iodir
-        a <- withLink $ initPactService version cid logger reqQ mempool mv
+        a <- async $ initPactService version cid logger reqQ mempool mv
```
emilypi: This isn't the only place `withLink` was used - we can remove it from `Chainweb.Utils`, and there are 2 more tests using it (`PactReplayTest` and `PactInProcApiTest` iirc)

src/Chainweb/Pact/Service/PactInProcApi.hs
```diff
@@ -89,22 +90,29 @@ withPactService'
...
-        reqQ <- atomically (newTQueue :: STM (TQueue RequestMsg))
-        a <- withLink $
-             PS.initPactService ver cid logger reqQ memPoolAccess cdbv bhDb
-                                pdb dbDir nodeid resetDb
-        evaluate =<< rst (action reqQ) `finally` closeQueue reqQ `finally` wait a
+withPactService' ver cid logger memPoolAccess cdbv bhDb pdb dbDir nodeid resetDb action = do
+    reqQ <- atomically $ newTBQueue pactQueueSize
+    race (server reqQ) (client reqQ) >>= \case
+        Left () -> error "pact service terminated unexpectedly"
+        Right a -> return a
```
fosskers: Do we ever expect this to exit?

src/Chainweb/Pact/Service/PactInProcApi.hs
```diff
@@ -89,22 +90,29 @@ withPactService'
...
+  where
+    client reqQ = action reqQ
+    server reqQ = runForever logg "pact-service" $
```
fosskers: Is `withPactService'` expected to run forever too?

src/Chainweb/Pact/Service/PactInProcApi.hs
larskuhtz: Only the client (right side). I think we have test cases that just terminate and expect the server to shut down.

src/Chainweb/Pact/Service/PactInProcApi.hs
larskuhtz: No, only the server component is expect to stay alive as long as the client is running.

In practice, in a production chainweb node, the client above will also run forever, but that's not a concern of this function.

PR #503, opened 2019-10-07
[mempool] BlockGasLimit is now configurable

fosskers:


src/Chainweb/Chainweb.hs
```diff
@@ -230,6 +232,7 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configThrottleRate = 1000
     , _configMempoolP2p = defaultEnableConfig defaultMempoolP2pConfig
     , _configPruneChainDatabase = True
+    , _configBlockGasLimit = 100000
```
fosskers: Is this default still appropriate, given that gas values went up for many primitives?

gregorycollins: We need to set it so a full block takes less than a half second to process, that's the target I think.

fosskers:
Shall we address that in a follow-up, once we have more data?

gregorycollins:
> Shall we address that in a follow-up, once we have more experiment data?

Probably -- it would probably behoove us, now that most of the validations are in, to spend some time again on block validation performance before we adjust it

PR #507, opened 2019-10-07
LTS 13

fosskers:
This has Chainweb match Pact.

### DIFF
--- a/src/Servant/Client_.hs
+++ b/src/Servant/Client_.hs
@@ -83,6 +83,9 @@ instance RunClient ClientM_ where
 #if MIN_VERSION_servant_server(0,16,0)
     throwClientError = throwError
     {-# INLINE throwClientError #-}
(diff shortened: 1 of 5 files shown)


src/Servant/Client_.hs
```diff
@@ -83,6 +83,9 @@ instance RunClient ClientM_ where
 #if MIN_VERSION_servant_server(0,16,0)
     throwClientError = throwError
     {-# INLINE throwClientError #-}
+#elif MIN_VERSION_servant_server(0,15,0)
+    throwServantError = throwError
+    {-# INLINE throwServantError #-}
```
larskuhtz: Is this correct that `throwClientError` and `throwServantError` were introduced in different versions of servant? Or is one bound here wrong?

fosskers: The name changed to `throwClientError` in `0.16`, but the streaming function below that was removed in `0.15`. Annoying :P

PR #509, opened 2019-10-08
[chainweb-node Main] add systemd watchdog support.

gregorycollins:


src/Chainweb/Utils.hs
```diff
@@ -877,16 +878,19 @@ runForeverThrottled logfun name burst rate a = mask $ \umask -> do
...
-leadingZeros b = int (B.length x) * 8 + case B.uncons y of
-    Just (h, _) -> int $ countLeadingZeros h
-    Nothing -> 0
-  where
-    (x, y) = B.span (== 0x00) b
+-- Count leading zero bits of a bytestring
+
+-- | Count leading zero bits of a bytestring
+--
+leadingZeros :: Integral int => B.ByteString -> int
```
gregorycollins: I should probably move this hunk into its own PR -- I noticed it while I was reading this. Will delete

chainweb.cabal
```diff
@@ -367,6 +372,9 @@ library
 
     if flag(ed25519)
         cpp-options: -DWITH_ED25519=1
+    if flag(use_systemd)
```
gregorycollins: I forgot to add an `&& os(linux)` here

stack.yaml
```diff
@@ -14,6 +14,7 @@ extra-deps:
   - paths-0.2.0.0
   - streaming-events-1.0.0
   - strict-tuple-0.1.2
+  - systemd-1.2.0
```
fosskers: @emilypi ran a `stack build` on her Mac with this - and everything compiled fine.

default.nix
```diff
@@ -50,6 +51,17 @@ in
...
 
+        systemd =
+          if systemdFeatures then
+            callHackageDirect {
+              pkg = "systemd";
+              # we'd like to use a newer one but it requires network >= 3
+              ver = "1.2.0";
+              sha256 = "1mwrrki3zsc4ncr7psjv9iqkzh7f25c2ch4lf2784fh6q46i997j";
+            }
+          else null;
```
fosskers: `null` is okay?

src/Chainweb/Utils/Watchdog.hs
```diff
@@ -0,0 +1,41 @@
...
+        either return impossible
+  where
+    impossible _ = fail "impossible: watchdog returned first"
+
+    watchdogTimer :: Int -> IO ()
+    watchdogTimer usec0 = forever $ do
+        void $ tryAllSynchronous $ SD.notifyWatchdog
+        threadDelay usec
+      where
+        usec = if usec0 <= 0 then defaultWatchdogUsec else usec0
```
fosskers: Any potential issues here the `usec` is *just* too close to what the external `systemd` is expecting?

default.nix
gregorycollins: The osx nix build wouldn't have passed otherwise

src/Chainweb/Utils/Watchdog.hs
gregorycollins: In general no -- typical values for watchdog in seconds are 30s - 2m

PR #512, opened 2019-10-09
Patch GHC stm bug

mightybyte:
Using diff from here:

https://gitlab.haskell.org/ghc/ghc/commit/7fc418df856d9b58034eeec48915646e67a7a562

Raw diff link:

https://gitlab.haskell.org/ghc/ghc/commit/7fc418df856d9b58034eeec48915646e67a7a562.diff

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -101,11 +101,11 @@ in
           sha256 = "1wcvs705b377zm0l33mhpzy1kyhwclxqkd5gzhk7dsd6ymnrlmpm";
         };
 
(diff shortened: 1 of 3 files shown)


larskuhtz:
Are we sure that there are no other relevant related changes in the RTS between the GHC versions? I would be extremely careful in making custom changes to `rts/RaiseAsync.c`.

mightybyte:
No longer necessary because 8.6 has the patch.

PR #516, opened 2019-10-10
LTS 14 + Latest Pact

fosskers:
This also makes our Nix config work with the latest Pact.

**Note:** This PR requires these two PRs to be merged first:

- https://github.com/kadena-io/pact/pull/636
- https://github.com/kadena-io/pact/pull/685

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ result-*
 .ghc.environment.*
 .ghci
 *~
-\#*
 .stylish-haskell.yaml
 public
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -32,7 +32,6 @@ result-*
 .ghc.environment.*
 .ghci
 *~
-\#*
```
fosskers: This breaks `nix-gitignore`.

PR #517, opened 2019-10-10
Fix a validation dupe check bug

gregorycollins:
If two txs came in with the same sender, and the first was NOT a dupe but the second one was, we were dupe-checking the first one but not the second.


fosskers:
Let me double-check this before merging.

larskuhtz:
I deployed this to devnet and added some load. So far there are no duplicate tx.

src/Chainweb/Pact/PactService.hs
```diff
@@ -489,17 +489,25 @@ validateChainwebTxs
...
-    validate :: IORef Balances -> ChainwebTransaction -> IO Bool
-    validate bsr tx = do
-        bs <- readIORef bsr
-        case HM.lookup sender bs >>= debitGas bs tx of
-            Nothing -> pure False
-            Just bs' -> do
-                let !valid = all ($ tx) validations
-                when valid $ writeIORef bsr bs'
-                pure valid
+    validate :: IORef Balances -> (ChainwebTransaction, Bool) -> IO Bool
```
larskuhtz: Does returning just `Bool` provide sufficient information in case of an error? It seems that there are different things validated within this function (dup-checks and balance checks).

gregorycollins: Colin and I talked about this. Return type should probably be `Either Mempool.InsertError ()` but this is ok for now

PR #518, opened 2019-10-10
Upgrade to GHC 8.6

mightybyte:


chainweb.cabal
```diff
@@ -348,7 +349,7 @@ library
         , swagger2 >= 2.2
         , temporary >= 1.3
         , text >= 1.2
-        , these >= 0.7 && < 1
+        , these >= 0.7 && < 1.1
```
fosskers: The `semialign` dep effectively forces `these >= 1`.

overrides.nix
```diff
@@ -0,0 +1,112 @@
...
+         }) {};
+
+    ourOverrides = {
+      pact = dontCheck ( addBuildDepend (self.callCabal2nix "pact" pactSrc {}) pkgs.z3);
+
+      aeson = callHackageDirect {
+        pkg = "aeson";
+        ver = "1.4.3.0";
+        sha256 = "13lim8vv78m9lhn7qfjswg7ax825gn0v75gcb80hckxawgk8zxc1";
+      };
```
fosskers: This should be safe to unpin by now.

chainweb.cabal
mightybyte: Oh, right.  I'll make that change.

overrides.nix
mightybyte: It's not safe to unpin.  This nixpkgs comes with aeson-1.4.2.  :(

fosskers: Ah :(

PR #519, opened 2019-10-11
Tx Gen test

emmanueldenloye:


chainweb.cabal
```diff
@@ -572,114 +614,6 @@ executable chainweb-miner
         , strict-tuple >= 0.1
         , wai-extra >= 3.0
 
-executable cwtool
```
mightybyte: We shouldn't delete this section completely.  We still need the cwtool executable.  It's just the modules in `other-modules` that we want to re-use that need to be moved into the `library` section.

emmanueldenloye: Ok, that makes sense. I'll make that change.

test/Chainweb/Test/Pact/TxGen.hs
```diff
@@ -0,0 +1,616 @@
...
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# OPTIONS_GHC -fno-warn-unused-matches #-}
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+-- |
+-- Module: Chainweb.Test.TxGen
```
mightybyte: I don't see this module listed in the `.cabal` file.  Also, this comment doesn't match the module name.

emmanueldenloye: I must have left it in by accident. It is a module that I was going to create but later realized was more trouble than it was worth. I'll delete the line.

chainweb.cabal
```diff
@@ -265,6 +267,33 @@ library
         , Utils.Logging.Config
         , Utils.Logging.Trace
 
+        -- tools
+        , Tools.Chain2Gexf
+        , Tools.CwTool
+        , Tools.Ea
+        , Tools.Gexf
```
larskuhtz: This adds new dependencies, which shouldn't go into the main library.

chainweb.cabal
```diff
@@ -265,6 +267,33 @@ library
...
+        , Tools.TXG.Simulate.Utils
+        , Tools.TXG.Types
+        , Tools.Tests.MultiNode
+        , Tools.Tests.Network.X509.SelfSigned.Test
+        , Tools.Tests.Orphans.External
+        , Tools.Tests.Orphans.Internal
+        , Tools.Tests.P2P.Peer.BootstrapConfig
+        , Tools.Tests.SlowTests
+        , Tools.Tests.Utils
+        , Tools.TxStream
```
larskuhtz: Is this really needed?

I think, this generally goes into the wrong direction. The chainweb library is already far too big. It takes a long time to build and becomes increasingly harder to maintain.

chainweb.cabal
```diff
@@ -350,6 +387,10 @@ library
         , strict-tuple >= 0.1
         , string-conv >= 0.1
         , swagger2 >= 2.2
+        , tasty >= 1.0
+        , tasty-golden >= 2.3
+        , tasty-hunit >= 0.9
+        , tasty-quickcheck >= 0.9
```
larskuhtz: I don't think that we want this dependencies in the main application

PR #523, opened 2019-10-11
Rename two types

fosskers:
This PR does a simple find-and-replace on `ChainwebTransaction` and `TransactionHash` to combat some line length issues.

### DIFF
--- a/src/Chainweb/Chainweb.hs
+++ b/src/Chainweb/Chainweb.hs
@@ -77,7 +77,7 @@ module Chainweb.Chainweb
 , chainwebConfig
 
 -- ** Mempool integration
-, ChainwebTransaction
+, ChainwebTX
 , Mempool.chainwebTransactionConfig
 , validatingMempoolConfig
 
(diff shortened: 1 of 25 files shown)


gregorycollins:
I'm not sure. Longer names are more descriptive -- is brevity the soul of wit?? Let's get some more opinions before we go forward with this

fosskers:
> Longer names are more descriptive

I agree, so long as symbol names don't approach Java lengths. And within Pact and Chainweb, the shortening of "Transaction" to "TX" specifically is fairly widespread.

emilypi:
I know this is not a technical argument, and that I'm painting exclusively what looks to be a bikeshed in front of me right now, but I know for a fact I'm going to hate working with `TXHash`. It's very aesthetically displeasing. Same with `ChainwebTx`. I'm perfectly fine with the way those in particular are named. 

There are long and less-than-descriptive names floating around chainweb though, like `HashCommandResult` and `ChainwebTransaction` which I always end up having to look up that are higher priority in my mind.

PR #525, opened 2019-10-11
Add `get-all-keys` function to coin contract

emilypi:
@mightybyte 

This gets you what you need. What you can do, which is far more cost-effective than a table scan or adding columns, is to simple map `(account-info)` over keys. I've amended account info to return the account name as well.


emilypi:
Decided against. Closing.

PR #526, opened 2019-10-11
Charge gas for transaction size

### LINKED ISSUE #504: Gas charge for transaction size
Charge x units of gas (i.e. 1/100th) per bytes for transaction size.
If this exceeds gas limit, fail. Otherwise, initialize the gas function with this as the started gas amount.

Motivation: 
For database native functions (mainly write), Pact will traverse pact values and calculating their size in bytes. This could be a potential attack vector. To mitigate it, we should charge some gas for the transaction size.

LindaOrtega:
Closes #504 

Charges a small amount (1/100th) of gas per byte of a transaction.
Excludes the size of a continuation proof, if one is present. 

If gas supply is less then the gas charged for transaction size, errors out with a Gas Pact Error.
Otherwise, this transaction size gas is added to the gas accumulated throughout the Pact execution.

Potential bug:
Not discounting the initial Gas charge from the Gas available to Pact.
My solution: https://github.com/kadena-io/chainweb-node/pull/526/files#diff-4da418a8f6368420870915674d7f8e85R146


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -116,78 +118,94 @@ applyCmd
...
+              logDebugRequestKey logger requestKey "success for requestKey"
 
-            let !redeemGasEnv = set ceGasEnv freeGasEnv payloadEnv
-                cmdLogs = fromMaybe [] $ _crLogs cmdResult
-            redeemResultE <- catchesPactError $!
-              redeemGas redeemGasEnv cmd cmdResult pactId cmdLogs mcache''
+              let !redeemGasEnv = set ceGasEnv freeGasEnv payloadEnv
+                  cmdLogs = fromMaybe [] $ _crLogs cmdResult
+              redeemResultE <- catchesPactError $!
+                               redeemGas redeemGasEnv cmdIn cmdResult pactId cmdLogs mcache''
```
emilypi: `cmd` + indentation

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -116,78 +118,94 @@ applyCmd
...
 
-        -- initialize refstate with cached module definitions
-        let st0 = set (evalRefs . rsLoadedModules) mcache' def
+          -- initialize refstate with cached module definitions
+          let st0 = set (evalRefs . rsLoadedModules) mcache' def
 
-        cmdResultE <- catchesPactError $!
-          runPayload payloadEnv st0 cmd buyGasLogs managedNamespacePolicy
+          cmdResultE <- catchesPactError $!
+                        runPayload payloadEnv st0 cmd buyGasLogs managedNamespacePolicy
```
emilypi: indentation?

src/Chainweb/Transaction.hs
```diff
@@ -121,6 +122,8 @@ creationTimeOf :: forall c . Command (Payload PublicMeta c) -> TxCreationTime
 creationTimeOf = _pmCreationTime . _pMeta . _cmdPayload
 {-# INLINE creationTimeOf #-}
 
+payloadWithText :: Command PayloadWithText -> PayloadWithText
+payloadWithText = _cmdPayload
```
emilypi: i don't think this passes the Fairbarn threshold - it's just an accessor

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -467,16 +502,19 @@ buyGas env cmd (Miner mid mks) supply mcache = do
...
     -> [TxLog Value]   -- ^ previous txlogs
     -> ModuleCache
     -> IO (T2 (CommandResult [TxLog Value]) ModuleCache)
-redeemGas env cmd cmdResult gid prevLogs mcache = do
-    let fee       = gasFeeOf (_crGas cmdResult) (gasPriceOf cmd)
-        rk        = cmdToRequestKey cmd
-        initState = set (evalRefs . rsLoadedModules) mcache
+redeemGas env cmdIn cmdResult gid prevLogs mcache = do
+    let cmd        = payloadObj <$> cmdIn
+        initialGas = initialGasOf (payloadWithText cmdIn)
```
slpopejoy: this function is only called from within `applyCmd`, it should remain largely unchanged and definitely should not be recomputing `initialGasOf`, pass it in from `applyCmd`

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -198,13 +216,14 @@ applyGenesisCmd
       -- ^ Contains block height, time, prev hash + metadata
     -> SPVSupport
       -- ^ SPV support (validates cont proofs)
-    -> Command (Payload PublicMeta ParsedCode)
+    -> Command PayloadWithText
       -- ^ command with payload to execute
     -> IO (T2 (CommandResult [TxLog Value]) ModuleCache)
-applyGenesisCmd logger dbEnv pd spv cmd = do
+applyGenesisCmd logger dbEnv pd spv cmdIn = do
```
slpopejoy: this should be unchanged as we don't charge gas on genesis

src/Chainweb/Pact/TransactionExec.hs
LindaOrtega: See latest commit

src/Chainweb/Pact/TransactionExec.hs
LindaOrtega: See latest commit

src/Chainweb/Transaction.hs
LindaOrtega: see latest commit

src/Chainweb/Pact/TransactionExec.hs
LindaOrtega: see latest commit

src/Chainweb/Pact/TransactionExec.hs
LindaOrtega: see latest commit

PR #530, opened 2019-10-14
Show miner hash rate stats

mightybyte:
This PR updates `chainweb-miner` to report Hash Rates upon successful mining of a block.
```
2019-10-15 16:24:07.824275: [info] Starting Miner.
2019-10-15 16:24:26.653637: [info] Chain 0: Mined block at Height 142. (1.22 MH/s)
2019-10-15 16:25:03.724461: [info] Chain 3: Mined block at Height 143. (1.14 MH/s)
2019-10-15 16:27:55.014416: [info] Chain 3: Mined block at Height 151. (0.93 MH/s)
```


src/Chainweb/Miner/Core.hs
```diff
@@ -107,26 +108,29 @@ mine
...
+                      hash ctx buf pow
+
+                      -- check whether the nonce meets the target
+                      fastCheckTarget trgPtr (castPtr pow) >>= \case
+                          True -> writeIORef nonces (nv - o)
+                          False -> go (succ n)
+
+              -- Start inner mining loop
+              go orig
+      T2 new <$> readIORef nonces
```
fosskers: Note for the future: this code has been heavily profiled and is not to be altered without confirming that no performance regression has been introduced.

miner/Miner.hs
```diff
@@ -249,25 +255,26 @@ getWork = do
...
-        m = mgr e
+        m = envMgr e
         u = coordinator a
 
 -- | A supervisor thread that listens for new work and manages mining threads.
 --
 mining :: (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes) -> WorkBytes -> RIO Env ()
 mining go wb = do
-    race updateSignal (go tbytes hbytes) >>= traverse_ miningSuccess
+    !now <- liftIO getPOSIXTime
```
larskuhtz: Since the code uses the timestamp only to compute a time difference and doesn't care about absolute time, using the monotonic clock form `clock` would be more efficient.

src/Chainweb/Miner/Core.hs
mightybyte: Dropping the stats printing from this loop makes the stats night unto worthless because in the real world mining blocks is quite a rare event.  We need to figure out a way to report results at regular intervals.

PR #531, opened 2019-10-15
Gas Indeterminacy and payer draft

slpopejoy:
Now that interfaces can express capabilities, we can present a `gas-payer` API that modules can implement to approve an on-behalf-of coin contract account to pay gas.

EDIT this is now covering lots of gas issues, so should merge for stability once confirmed that gas sizing is causing issues.


slpopejoy:
- probably needs a DB reset, as post-genesis gas changes

- mac CI passed

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -155,64 +159,61 @@ applyCmd logger pactDbEnv miner gasModel pd spv cmdIn mcache
...
-        Left e1 ->
-          jsonErrorResult buyGasEnv requestKey e1 [] (Gas 0) mcache
-          "tx failure for requestKey when buying gas"
-        Right (Left e) ->
-          -- this call needs to fail hard if Left. It means the continuation did not process
-          -- correctly, and we should fail the transaction
-          internalError e
-        Right (Right (T3 pactId buyGasLogs mcache')) -> do
-          logDebugRequestKey logger requestKey "successful gas buy for request key"
+        Left e1 -> fatal $ "tx failure for requestKey when buying gas: " <> tShow e1
```
emilypi: debugging this without logging has been a frustrating pain. Do you you mind logging this error as well?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -483,25 +488,71 @@ buyGas
...
+      (m,qn,as) <- MaybeT findPayerCap
+      pMod <- MaybeT $ lookupModule qn m
+      MaybeT $ return $ validateMod pMod
+      capRef <- MaybeT $ resolveRef qn (QName qn)
+      return $ runCap (getInfo qn) capRef as
+
+    findPayerCap :: Eval e (Maybe (ModuleName,QualifiedName,[PactValue]))
+    findPayerCap = preview $ eeMsgSigs . folded . folded . to sigPayerCap . _Just
+
+    sigPayerCap (UserCapability q@(QualifiedName m n _) as) | n == "GAS_PAYER" = Just (m,q,as)
```
emilypi: Shouldn't we also check the module name here to avoid collision? This needs to be `gas-payer-v1-blah.GAS_PAYER` doesn't it?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -521,9 +572,9 @@ redeemGas env cmd initialGas cmdResult gid prevLogs mcache = do
     let totalGas   = initialGas + _crGas cmdResult
         fee        = gasFeeOf totalGas (gasPriceOf cmd)
         rk         = cmdToRequestKey cmd
-        initState  = set (evalRefs . rsLoadedModules) mcache
+        initState  = initStateInterpreter $ setModuleCache mcache
           $ initCapabilities [magic_FUND_TX]
-
+    -- print ("\n\nREDEEM GAS:",rk,initialGas,_crGas cmdResult,fee, "\n\n")
```
emilypi: there's a few print statements here we can remove unless you need them for debugging a next PR

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: We validate the module next. It doesn't really matter what it is as long as it implements the interface, which we check.

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: Yeah they're helpful for gas debugging while we're digging

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: I'll add logging to `fatal`

PR #533, opened 2019-10-15
up-pin pact, coinbase module cache, 0 gas genesis

slpopejoy:
Coinbase module cache is fed to transactions, should make for small speed up

Latest (final?) pact

Genesis writes 0 gas in output to make it not sensitive to gas charging


test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -271,15 +271,15 @@ allocationTest iot nio = testCaseSteps "genesis allocation tests" $ \step -> do
 
     case r of
       Left e -> assertFailure $ "test failure: " <> show e
-      Right cr -> assertEqual "expect /local allocation balance" (resultOf cr) accountInfo
+      Right cr -> assertEqual "expect /local allocation balance" accountInfo (resultOf cr)
```
emilypi: 👍 thanks

PR #534, opened 2019-10-15
Chainweb Validations 37 (and possibly 39)

emmanueldenloye:
- prop_tx_gas_rounding: Gheck that gas prices in the input are correctly rounded.
- prop_redeem_gas_refund_minimum_unit: Check that gas refunds are rounded to the nearest hop.


src/Chainweb/Mempool/InMem.hs
```diff
@@ -270,6 +273,19 @@ validateOne cfg badmap (Time (TimeSpan now)) t h =
       where
         TransactionMetadata (Time (TimeSpan ct)) (Time (TimeSpan et)) = txMetadata txcfg t
 
+    -- prop_tx_gas_rounding
+    gasPriceRoundingChcek :: Either InsertError ()
+    gasPriceRoundingChcek = ebool_ (InsertErrorOther msg) (f (txGasPrice txcfg t))
+        where
+          -- Should a hop be specified as a constant in Chainweb.Utils?
+          f (GasPrice (ParsedDecimal d)) = decimalPlaces d <= 12
```
fosskers: It feels to me like this could be caught in the JSON instance for `GasPrice`, and possibly also in a smart constructor, making it impossible to create a `GasPrice` with the wrong rounding.

emilypi: The minimum units is a chainweb-specific gas policy. If we did this at the Pact level, then this would affect any chain that interoperated with us, which is too strong of a constraint to place on our partners.

PR #536, opened 2019-10-15
SPV bugfix: validate target chain id

emilypi:
This PR addresses the following: 

- Fixes a bug where target chain id of a cont proof is not checked against the service chain of execution (i.e. i could ask for a proof `0 -> 9` and as long as a cross-chain-transfer was `0 -> <whatever>` it would pass inspection)

- Adds a test for said bug

- Change the graph used in SPV tests to make use of `triangle graph`. This drops our test times for spv to _30 seconds total_.


test/Chainweb/Test/Pact/SPV.hs
```diff
@@ -91,12 +92,13 @@ import Data.LogMessage
...
-    , testCase "wrong chain execution fails" wrongchain
-    , testCase "invalid proof formats fail" badproof
+    , testCase "wrong chain execution fails" wrongChain
+    , testCase "invalid proof formats fail" invalidProof
+    , testCase "wrong target chain in proofs fail" wrongChainProof
     ]
 
 v :: ChainwebVersion
-v = FastTimedCPM petersonChainGraph
+v = FastTimedCPM triangleChainGraph
```
larskuhtz: Nice

PR #541, opened 2019-10-16
Emergency Difficulty Adjustment

fosskers:
This PR introduces an extra condition for triggering DA. When block rates have slowed to an extreme degree we detect this and perform DA as if the next epoch had been reached.
```
| NHP Lost (%) | Blocks-per-Minute | Block appears in (sec) | Full Epoch Time (hours) |
|--------------+-------------------+------------------------+-------------------------|
|            0 |                 2 |                     30 |                     1.0 |
|           50 |                 1 |                     60 |                     2.0 |
|           65 |               0.7 |                     86 |                     2.9 |
|           80 |               0.4 |                    150 |                     5.0 |
|           95 |               0.1 |                    600 |                    20.0 |
|           99 |              0.02 |                   3000 |                   100.0 |
|--------------+-------------------+------------------------+-------------------------|
```


src/Chainweb/BlockHeader.hs
```diff
@@ -345,7 +345,25 @@ isLastInEpoch h = case effectiveWindow h of
...
 {-# INLINE isLastInEpoch #-}
 
--- | Compute the POW target for a new BlockHeader
+-- | If it is discovered that the last DA occured significantly in the past, we
+-- assume that a large amount of hash power has suddenly dropped out of the
+-- network. Thus we must perform Emergency Difficulty Adjustment to avoid
+-- stalling the chain.
+--
+slowEpoch :: BlockHeader -> BlockCreationTime -> Bool
+slowEpoch p (BlockCreationTime ct) = actual > (expected * 5)
```
fosskers: This compares against the `BlockCreationTime` of the current `BlockHeader`. The parent is only consulted for the `EpochStartTime`.

While we imagined two scenarios (slow / stalled), I believe this check guards against both of those.

slpopejoy:
What does the table mean?

fosskers:
It's a demonstration of how long we'd currently have to wait for DA if certain amounts of hash power were lost. This PR defends against sudden losses.

src/Chainweb/Version.hs
```diff
@@ -521,20 +519,20 @@ newtype BlockRate = BlockRate Seconds
 -- number of seconds we expect to pass while a miner mines on various chains,
 -- eventually succeeding on one.
 --
-blockRate :: ChainwebVersion -> Maybe BlockRate
-blockRate Test{} = Nothing
-blockRate TimedConsensus{} = Just $ BlockRate 4
-blockRate PowConsensus{} = Just $ BlockRate 10
-blockRate TimedCPM{} = Just $ BlockRate 4
+blockRate :: ChainwebVersion -> BlockRate
+blockRate Test{} = BlockRate 0
```
larskuhtz: What will happen if adjustment is called for this? Before there would be a meaningful runtime error:

```haskell
error $ "No BlockRate available for given ChainwebVersion: " <> show v
```

With this change will the code just crash with division by 0 or similar, will it provide a meaningful error, or will it just work?

fosskers: Unknown, but the `Test` value is rarely used.

larskuhtz: Maybe just replace it by something like `error "ChainwebVersion 'Test' doesn't most not use the BlockRate setting".

That would help identify the error if a tests hits it. Otherwise, we my just get `Division by zero` but have no idea where it's coming from.

PR #544, opened 2019-10-17
Time-based Disabling of Transactions/Mining 

fosskers:
This PR disables Transactions from being submitted before a certain END_DATE. It also prevents a Node from participating meaningfully in the network after that END_DATE.

### TODO

- [x] `chainweb-node` will not start if past the END_DATE
- [x] Mempool rejects all TXs if `START_DATE < now`
- [x] Halt mining (and coordination) after END_DATE


src/Chainweb/Miner/Coordinator.hs
```diff
@@ -162,12 +162,23 @@ chainChoice c choice = case choice of
...
+-- a future version of Chainweb. It prevents this Node from generating any new
+-- Cuts after a specified date.
+--
+publish :: LogFunction -> MiningState -> CutDb cas -> BlockHeader -> IO ()
+publish lf ms cdb bh = do
+    now <- getCurrentTimeIntegral
+    case txSilenceDates $ _chainwebVersion bh of
+        Just (_, end) | now > end -> pure ()
+        _ -> publish' lf ms cdb bh
+
```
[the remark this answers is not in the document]
larskuhtz: On the other hand, if we just added a watch-dog process in `node/ChainwebNode.hs` that kill the whole node, we wouldn't need any kill switch here or elsewhere.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -244,20 +249,30 @@ insertCheckInMem cfg lock txs
...
     txcfg = _inmemTxCfg cfg
 
+    -- | KILLSWITCH: This is to be removed in a future version of Chainweb. This
+    -- prevents any transaction from entering the mempool.
+    --
+    transactionsEnabled :: Either InsertError ()
+    transactionsEnabled = case txSilenceDates v of
+        Just (start, _) | start < nw -> Left InsertErrorTransactionsDisabled
+        _ -> pure ()
+
```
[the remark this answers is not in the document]
larskuhtz: I think this isn't enough. Someone could write a hacked node that mines blocks with tx. We should reject those, too. 

Since `txSilentDates` is already a property of the `ChainwebVersion` we can add a payload validation rule ala

```Haskell
prop_silentDates hdr payload = case txSilenceDates v of
    Just (start, _) | start < blockCreationTime hdr -> length (playloadTransactions payload) == 0
    _ -> True
```

larskuhtz:
Disabling the mempool REST API can by changing `mempoolsToServer` in `src/Chainweb/Chainweb.hs` line 651 to

```Haskell
    mempoolsToServe :: [(ChainId, Mempool.MempoolBackend ChainwebTransaction)]
    mempoolsToServe = case
        Just (_, end) | now > end -> []
        _ -> prop _chainResMempool
```

node/ChainwebNode.hs
```diff
@@ -449,7 +450,14 @@ mainInfo = programInfoValidate
...
+-- past a specified date.
+--
 main :: IO ()
-main = withWatchdog $ runWithPkgInfoConfiguration mainInfo pkgInfo $ \conf -> do
+main = withWatchdog . runWithPkgInfoConfiguration mainInfo pkgInfo $ \conf -> do
     let v = _configChainwebVersion $ _nodeConfigChainweb conf
-    withNodeLogger (_nodeConfigLog conf) v $ node conf
+    now <- getCurrentTimeIntegral
+    case txSilenceDates v of
+        Just (_, end) | now > end -> pure ()
```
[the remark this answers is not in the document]
fosskers: I argue that running nodes should be left alive, since:

1. They will no longer be able to produce new Cuts anyway
2. They're still useful to keep around as p2p/cut sync nodes
3. If all nodes in the network suddenly died, it could affect overall stability in a bad way.

src/Chainweb/Chainweb.hs
```diff
@@ -616,9 +617,10 @@ runChainweb
     -> IO ()
 runChainweb cw = do
     logg Info "start chainweb node"
+    now <- getCurrentTimeIntegral
```
larskuhtz: What about moving this right before `mempoolsToServe` in line 652? That would avoid having to pass the parameter around (something that we'll have to undo later).

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -60,18 +62,29 @@ import Data.Singletons
 
 ---
 
+-- | KILLSWITCH: The logic here involing `txSilenceDates` is to be removed in a
+-- future version of Chainweb. This logic errors on remote requests for new
+-- mining work, such that this Node can no longer meaningfully participate in
+-- mining.
+--
```
larskuhtz: What about using a similar approach as with the mempool, of just completely disabling the miner API? What that be easier to implement?

src/Chainweb/Pact/PactService.hs
```diff
@@ -856,16 +856,26 @@ execValidateBlock
...
-        return $! Save currHeader result
+execValidateBlock currHeader plData = do
+    -- KILLSWITCH: The logic here involving `txSilenceDates` is to be removed in
+    -- a future version of Chainweb. This rejects any non-genesis block which is
+    -- found to contain transactions.
+    --
+    now <- liftIO getCurrentTimeIntegral
+    case txSilenceDates (_blockChainwebVersion currHeader) of
+        Just (start, _) | now > start && not isGenesisBlock && not payloadIsEmpty ->
+            throwM $ BlockValidationFailure "Transactions are disabled."
```
larskuhtz: Should this error message contain some more details? For instance, that this is temporarily, it isn't a bug but a feature, and a date as of when transactions are being enabled.

src/Chainweb/Pact/PactService.hs
```diff
@@ -856,16 +856,26 @@ execValidateBlock
...
-execValidateBlock currHeader plData =
-    -- TODO: are we actually validating the output hash here?
-    withBatch $ withCheckpointerRewind mb "execValidateBlock" $ \pdbenv -> do
-        !result <- playOneBlock currHeader plData pdbenv
-        return $! Save currHeader result
+execValidateBlock currHeader plData = do
+    -- KILLSWITCH: The logic here involving `txSilenceDates` is to be removed in
+    -- a future version of Chainweb. This rejects any non-genesis block which is
+    -- found to contain transactions.
+    --
```
larskuhtz: I think, this is wrong. This logic has to stay. I must not be that a transaction that was invalid yesterday becomes valid tomorrow.

Otherwise, a selfish miner with a lot of hash power could mine a branch that has tx before the cut-off date and reveal the branch to the network after this code got remove.

src/Chainweb/Pact/PactService.hs
```diff
@@ -856,16 +856,26 @@ execValidateBlock
     => BlockHeader
     -> PayloadData
     -> PactServiceM cas PayloadWithOutputs
-execValidateBlock currHeader plData =
-    -- TODO: are we actually validating the output hash here?
-    withBatch $ withCheckpointerRewind mb "execValidateBlock" $ \pdbenv -> do
-        !result <- playOneBlock currHeader plData pdbenv
-        return $! Save currHeader result
+execValidateBlock currHeader plData = do
```
larskuhtz: @emilypi is this the correct place for payload Validations? Do we have a canonical place for doing all payload validation checks?

src/Chainweb/Pact/PactService.hs
```diff
@@ -856,16 +856,26 @@ execValidateBlock
...
+    -- KILLSWITCH: The logic here involving `txSilenceDates` is to be removed in
+    -- a future version of Chainweb. This rejects any non-genesis block which is
+    -- found to contain transactions.
+    --
+    now <- liftIO getCurrentTimeIntegral
```
larskuhtz: This should be the block creation time and not the current time. Once a block got included into the chain, payload validation must always yield the same result.

src/Chainweb/Pact/PactService.hs
emilypi: Payload validation is done here and `execNewBlock` which is a place we may want to have redundant checks

src/Chainweb/Pact/PactService.hs
emilypi: Agree

node/ChainwebNode.hs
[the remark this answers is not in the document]
emilypi: I agree with @fosskers here, though, these useless nodes might make consensus needlessly complex

src/Chainweb/Chainweb.hs
fosskers: To avoid forcing `mempoolsToServe` into `IO`, which would have to be undone in future anyway.

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -60,18 +62,29 @@ import Data.Singletons
 
 ---
 
+-- | KILLSWITCH: The logic here involing `txSilenceDates` is to be removed in a
```
emilypi: spelling

src/Chainweb/Pact/PactService.hs
```diff
@@ -905,14 +906,10 @@ execValidateBlock
...
-    -- a future version of Chainweb. This rejects any non-genesis block which is
-    -- found to contain transactions.
-    --
-    now <- liftIO getCurrentTimeIntegral
+    let !now = _bct $ _blockCreationTime currHeader
     case txSilenceDates (_blockChainwebVersion currHeader) of
-        Just (start, _) | now > start && not isGenesisBlock && not payloadIsEmpty ->
-            throwM $ BlockValidationFailure "Transactions are disabled."
+        Just end | end > now && not isGenesisBlock && not payloadIsEmpty ->
+            throwM $ BlockValidationFailure "Transactions are disabled until December 5."
```
larskuhtz: The code would be easier to understand and maintain if the logic for the silence date validation would be moved into a dedicated validation function:

```haskell
validateSilenceDates :: MonadThrow m => BlockHeader -> PayloadData -> m ()
validateSilenceDates bh plData = case txSilenceDates (_chainwebVersion bh) of
    Just end | end > blockTime && not isGenesisBlock && not payloadIsEmpty ->
        throwM $ BlockValidationFailure "Transactions are disabled until December 5. BlockHeader: " <> encodeToText (ObjectEncoded bh)
    Nothing -> return ()
  where
    blockTime = _bct (_blockCreationTime bh)
    payloadIsEmpty = V.null $ _payloadDataTransactions plData
```

The error message should also contain the block header encoded as JSON.

This function would then just be:

```Haskell
execValidateBlock currHeader plData = do
    validateSilenceDates currHeader plData
    -- TODO: are we actually validating the output hash here?
    withBatch $ withCheckpointerRewind mb "execValidateBlock" $ \pdbenv -> do
    !result <- playOneBlock currHeader plData pdbenv
    return $! Save currHeader result
  where
    -- ...
```

src/Chainweb/Pact/PactService.hs
larskuhtz: @fosskers would we have to add `validateSilenceDates` (see below) to `execNewBlock`, too?

src/Chainweb/Mempool/InMem.hs
```diff
@@ -96,9 +97,10 @@ newInMemMempoolData =
 
 ------------------------------------------------------------------------------
 toMempoolBackend
-    :: InMemoryMempool t
+    :: ChainwebVersion
+    -> InMemoryMempool t
     -> IO (MempoolBackend t)
-toMempoolBackend mempool = do
+toMempoolBackend v mempool = do
```
larskuhtz: I wonder if it would be better to just add the chainweb version to `InMemConfig` and make it an instance of `HasChainwebVersion`.

PR #546, opened 2019-10-18
Coin contract UX fixes

slpopejoy:
FUND_TX -> GAS, teleport -> crosschain, don't revalidate sender on redeem


slpopejoy:
Obviated by Emily's upcoming PR

PR #551, opened 2019-10-19
TXG coin renames, transfer caps

slpopejoy:


larskuhtz:
This Draft was merged as part of PR #552

PR #556, opened 2019-10-21
Attempts to buy gas during transaction validation

LindaOrtega:
TODO: 
~~Add test that would cause buyGas to fail~~


src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +478,134 @@ finalizeCheckpointer finalize = do
...
+    -- | Attempt to run buy gas for transactions
     --
-    debitGas :: Balances -> ChainwebTransaction -> Decimal -> Maybe Balances
-    debitGas bs tx bal
-        | newBal < 0 = Nothing
-        | otherwise = Just $ HM.adjust (const newBal) sender bs
+    debitGas
+      :: Vector (ChainwebTransaction, Uniqueness)
+      -> IO (Vector (ChainwebTransaction, Uniqueness, Bool))
+    debitGas uniqueTxs = fst <$> runPactServiceM psState psEnv runGas
```
emmanueldenloye: Maybe make the `<$>` a `<$!>`?

src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +478,134 @@ finalizeCheckpointer finalize = do
...
+    -> Vector (ChainwebTransaction, Uniqueness)
+    -> PactServiceM cas (Vector (ChainwebTransaction, Uniqueness, Bool))
+attemptBuyGas cp miner nonGenesisParentHash txs = withDiscardedBatch $ do
+  mbLatestBlock <- liftIO $ _cpGetLatestBlock cp
+  (bh, bhash) <- case mbLatestBlock of
+                     Nothing -> throwM NoBlockValidatedYet
+                     (Just !p) -> return p
+  let target = Just (succ bh, bhash)
+  withCheckpointer target "validateChainwebTxs/attemptBuyGas" $ \(PactDbEnv' dbEnv) -> do
+      (T2 _ mcache) <- runCoinbase nonGenesisParentHash dbEnv miner
```
slpopejoy: Is this to catch miner problems?

src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +478,134 @@ finalizeCheckpointer finalize = do
...
+      psEnv <- ask
+      res <- mapM (runBuyGas psEnv dbEnv mcache) txs
```
slpopejoy: We need to fold the mcache through the various txs here, as loading capabilities could cause other modules to be loaded. See `applyPactCmds` for how this is done.

src/Chainweb/Pact/PactService.hs
LindaOrtega: `buyGas` takes a ModuleCache as an input. And after tracing back to see where it got it from, I came across these:
- https://github.com/kadena-io/chainweb-node/blob/master/src/Chainweb/Pact/TransactionExec.hs#L164
-> https://github.com/kadena-io/chainweb-node/blob/masg/src/Chainweb/Pact/PactService.hs#L938
-> https://github.com/kadena-io/chainweb-node/blob/masg/src/Chainweb/Pact/PactService.hs#L878
(it seems that the `ModuleCache` that eventually gets run with `runGas` get created in the latter via: `T2 coinOut mc <- runCoinbase nonGenesisParentHash pactdbenv miner`. So I tried to replicate this.)

src/Chainweb/Pact/PactService.hs
LindaOrtega: So store the ModuleCache that's returned from `runGas` (https://github.com/kadena-io/chainweb-node/blob/master/src/Chainweb/Pact/TransactionExec.hs#L476) in each transaction and pass it to the other transactions?

src/Chainweb/Pact/PactService.hs
LindaOrtega: call to `runCoinbase` omitted in current commit.

src/Chainweb/Pact/PactService.hs
LindaOrtega: See latest commit

src/Chainweb/Pact/PactService.hs
LindaOrtega: See latest commit

LindaOrtega:
@gregorycollins I added a remote pact test.
@slpopejoy re: 
>more feedback than just Bool
I was going to implement that in a follow-up PR.

src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +480,135 @@ finalizeCheckpointer finalize = do
...
+    :: PayloadCas cas
+    => Checkpointer
+    -> Miner
+    -> Vector (ChainwebTransaction, Uniqueness)
+    -> PactServiceM cas (Vector (ChainwebTransaction, Uniqueness, Bool))
+attemptBuyGas cp miner txs = withDiscardedBatch $ do
+  mbLatestBlock <- liftIO $ _cpGetLatestBlock cp
+  (bh, bhash) <- case mbLatestBlock of
+                     Nothing -> throwM NoBlockValidatedYet
+                     (Just !p) -> return p
```
emilypi: Unnecessary parens

src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +480,135 @@ finalizeCheckpointer finalize = do
...
+          gasPrice = gasPriceOf cmd
+          gasLimit = fromIntegral $ gasLimitOf cmd
+          supply = gasFeeOf gasLimit gasPrice
+          buyGasEnv = createGasEnv envM db cmd 
+
+      buyGasResultE <- liftIO $! P.catchesPactError $!
+        buyGas buyGasEnv cmd miner supply mcache
+
+      case buyGasResultE of
+        Left _ -> return $ T2 (tx, u, False) mcache
```
emilypi: If the transaction failed and it wasn't the result of the the buy-gas command, we probably want to raise this differently. @slpopejoy suggestions?

src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +480,135 @@ finalizeCheckpointer finalize = do
...
+
+    runBuyGas
+        :: PayloadCas cas
+        => PactServiceEnv cas
+        -> P.PactDbEnv a
+        -> ModuleCache
+        -> (ChainwebTransaction, Uniqueness)
+        -> PactServiceM cas (T2 (ChainwebTransaction, Uniqueness, Bool) ModuleCache)
+    runBuyGas _ _ mcache (tx, u@Duplicate) = return $ T2 (tx, u, False) mcache
+    runBuyGas envM db mcache (tx, u@Unique) = do
```
emilypi: `Duplicate` and `Unique` are singletons, so we can match directly on them instead of binding them to a variable. GHC will inline this appopriately.

src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +480,135 @@ finalizeCheckpointer finalize = do
...
     -> BlockHeight
     -> Vector ChainwebTransaction
     -> IO (Vector Bool)
-validateChainwebTxs dbEnv cp blockOriginationTime bh txs
+validateChainwebTxs psEnv psState cp miner blockOriginationTime bh txs
     | bh == 0 = pure $! V.replicate (V.length txs) True
     | V.null txs = pure V.empty
     | otherwise = do
           let f t = let p = view P.cmdHash t
                     in (bool Unique Duplicate . isJust) <$> _cpLookupProcessedTx cp p
```
emilypi: Suggesting

```haskell
let f t = (bool Unique Duplicate . isJust) <$!> _cpLookupProcessedTx cp (P._cmdHash t)
```

Instead to avoid a thunk and strict map to execute immediately

src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +480,135 @@ finalizeCheckpointer finalize = do
...
-            Nothing -> pure False
-            Just bs' -> do
-                let !valid = all ($ tx) validations
-                when valid $ writeIORef bsr bs'
-                pure valid
+    validate :: (ChainwebTransaction, Uniqueness, Bool) -> IO Bool
+    validate (_, Duplicate, _) = pure False
+    validate (tx, Unique, canBuyGas) = do
+      let !valid = all ($ tx) validations
+      pure valid
```
emilypi: this can be inlined - there's no reason to assign a variable here

src/Chainweb/Pact/PactService.hs
```diff
@@ -477,97 +480,135 @@ finalizeCheckpointer finalize = do
...
+validateChainwebTxs psEnv psState cp miner blockOriginationTime bh txs
     | bh == 0 = pure $! V.replicate (V.length txs) True
     | V.null txs = pure V.empty
     | otherwise = do
           let f t = let p = view P.cmdHash t
                     in (bool Unique Duplicate . isJust) <$> _cpLookupProcessedTx cp p
-          dupecheckOks <- V.mapM f txs
+          -- TODO miner attach: list of txs being validated could contain duplicates themselves
+          dupecheckOks <- liftIO $ V.mapM f txs
           let txs' = V.zip txs dupecheckOks
```
emilypi: can we length-check this zip or are we gauranteed this as an invariant. @emmanueldenloye ?

src/Chainweb/Pact/PactService.hs
slpopejoy: Any failure is worth throwing out. We need more feedback in any case

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -312,6 +315,41 @@ txTooBigGasTest iot nio = testCaseSteps "transaction size gas tests" $ \step ->
...
+          (PollResponses resp) <- liftIO $ polling sid cenv rks expectation
+
+          return (HashMap.lookup (NEL.head $ _rkRequestKeys rks) resp)
+
+    -- batch with incorrect sender
+    batch0 <- mkBadGasTxBatch "(+ 1 2)" "some-unknown-sender" sender00KeyPair Nothing
+    res0 <- catches (Right <$> run batch0 ExpectPactResult)
+      [ Handler (\(e :: PactTestFailure) -> return $ Left e) ]
+
+    void $ liftIO $ step "tx signed with incorrect sender should fail buy gas validation"
```
gregorycollins: Awesome :+1:

src/Chainweb/Pact/PactService.hs
LindaOrtega: I added a TODO for a future follow up

PR #558, opened 2019-10-21
TX generator test (version 2)

emmanueldenloye:


chainweb.cabal
```diff
@@ -499,6 +499,91 @@ test-suite chainweb-tests
...
+test-suite chainweb-tools-tests
+    import: warning-flags, debugging-flags
+    default-language: Haskell2010
+    ghc-options:
+        -threaded
+        -with-rtsopts=-N2
+    type: exitcode-stdio-1.0
+    hs-source-dirs: test tools/txg
+    main-is: ChainwebToolsTests.hs
+    other-modules:
```
fosskers: If empty, this can be removed.

larskuhtz:
@mightybyte I think a separat test-suite only increases build time for a complete CI build including tools tests. All other builds would profit. In particular, I don't see a need to build and test txgen locally, when working on unrelated parts of the library, which I do most of the time.

Actually, I am not even sure if we have to test tools (currently only txgen) on a per CI build basis, or if a nightly check would be sufficient. Our unit test suite is already pretty big.

I think it makes a lot of sense, to keep production code and production tests free of all non-production code.

larskuhtz:
@mightybyte in the end, I don't care that much about test code, as long as it doesn't clutter or is mixed with production code.

test/Chainweb/Test/Tools/TXGen.hs
```diff
@@ -0,0 +1,154 @@
...
+v :: ChainwebVersion
+v = FastTimedCPM petersonChainGraph
+
+tests :: RocksDb -> ScheduledTest
+tests rdb = testGroupSch "Chainweb.Test.Tools.TxGen"
+    [ withNodes rdb nNodes $ \net ->
+        withMVarResource 0 $ \iomvar ->
+        withTime $ \iot ->
+        testGroup "txgen tests" [
+            withRequestKeys iot iomvar net $ responseGolden net
```
larskuhtz: What is the purpose of this golden test? Golden test tent to be fragile and are best used to catch and get manual confirmation to changes for things like API data or data that is hashed and put on the chain. Golden tests are less suited for testing functionality.

PR #560, opened 2019-10-23
Use JSON encoding for BlockValidationFailure log messages

larskuhtz:


src/Chainweb/Pact/PactService.hs
```diff
@@ -390,11 +390,12 @@ validateHashes pwo bHeader =
...
-            "Hash from Pact execution: " ++ show newHash ++
-            " does not match the previously stored hash: " ++ show prevHash ++
-            ". Payload with outputs: " ++ show pwo
-
+        else Left $ BlockValidationFailure $ A.object
+            [ "message" A..= ("Payload hash from Pact execution does not match previously stored hash" :: T.Text)
+            , "actual" A..= newHash
+            , "expected" A..= prevHash
+            , "payloadWithOutputs" A..= pwo
+            ]
```
LindaOrtega: @emmanueldenloye woudl it be helpful to also output the `PayloadWithOutputs` associated with `prevHash` not just the `PayloadWithOutputs` associated with `newHash`?

larskuhtz: Generally, we I don't think we have those available at this point. We don't keep the outputs for non-validated blocks around.

PR #563, opened 2019-10-23
Miner Rewards 2.0

emilypi:
This is our final rewards schedule for Mainnet release

TODO: 

- [x] Scaffolding + new csv
- [x] test showing us that miner rewards are changing per block height


src/Chainweb/Pact/PactService.hs
```diff
@@ -616,18 +615,37 @@ readAccountGuard pdb account
...
-    m <- view $ psMinerRewards . at (roundBy bh 500000)
-    case m of
+minerReward bh = use psMinerRewards >>= go
+  where
+    go (MinerRewards rewards q) = do
+      when (V.null q) err
+      let (!h, !t) = (V.unsafeHead q, V.unsafeTail t)
+      if bh <= h
+      then lookup_ h rewards
+      else modifyRewards t rewards
```
emilypi: The rewards map is now part of pact service state, maintaining a list of sorted keys. This list acts as a kind of queue, where each block height is compared against head and looked up at the appropriate upper bound in the map. There is no closed form for this. If the blockheight exceeds `head`'s value, then `head` is popped off the list and the state is updated with `tail`.

rewards/miner_rewards.csv
```diff
@@ -1,435 +1,1436 @@
...
+125005200,1
+125092800,1
+125180400,1
+125268000,1
+125355600,1
+125443200,1
+125530800,1
+125538055,1
+125538056,0.023999333
+125538057,0
```
fosskers: Can you elaborate on these final lines? You're also missing a newline char at EOF.

src/Chainweb/Miner/Pact.hs
```diff
@@ -148,26 +151,42 @@ fromMinerData :: MonadThrow m => MinerData -> m Miner
 fromMinerData = decodeStrictOrThrow' . _minerData
 {-# INLINABLE fromMinerData #-}
 
+data MinerRewards = MinerRewards
+    { _minerRewards :: !(HashMap BlockHeight ParsedDecimal)
+      -- ^ The map of blockheight thresholds to miner rewards
+    , _minerRewardHeights :: !(Vector BlockHeight)
+      -- ^ A (sorted) vector of blockheights (head is most recent)
+    } deriving (Eq, Ord, Show, Generic)
```
fosskers: The state of the `Vector` isn't lost if the Node shuts down?

src/Chainweb/Miner/Pact.hs
```diff
@@ -148,26 +151,42 @@ fromMinerData :: MonadThrow m => MinerData -> m Miner
...
-      Right vs -> HM.fromList
-        . V.toList
-        . V.map formatRow
-        $ vs
+      Right vs ->
+        let !rs = HM.fromList . V.toList . V.map formatRow $ vs
+        in MinerRewards rs (V.fromList . sort $! HM.keys rs)
   where
-    formatRow :: (Word64, Double) -> (BlockHeight, ParsedDecimal)
+    formatRow :: (Integer, Double) -> (BlockHeight, ParsedDecimal)
```
fosskers: Why `Integer`?

src/Chainweb/Miner/Pact.hs
emilypi: It would be - that's a good point. I wonder if we should make the lookup less `head`-based and a comparison instead.

rewards/miner_rewards.csv
emilypi: at this point, we've shut off the blockchain and the universe has imploded. There is no plan after this 120-year point.

src/Chainweb/Miner/Pact.hs
emilypi: It's a harmless intermediate type only used for inference. It acts the same, and I changed it for some reason at some point that I forget.

PR #566, opened 2019-10-23
Avoid div-by-zero

fosskers:
@mightybyte ran into this.

### DIFF
--- a/miner/Miner.hs
+++ b/miner/Miner.hs
@@ -333,7 +333,7 @@ mining go wb = do
       let !v = version $ envArgs e
           !m = envMgr e
           !u = coordinator $ envArgs e
-          !r = (fromIntegral hashes :: Double) / fromIntegral secs / 1000000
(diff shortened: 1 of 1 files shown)


miner/Miner.hs
```diff
@@ -374,7 +374,7 @@ gpu (GPUEnv mpath margs) (TargetBytes target) (HeaderBytes blockbytes) = do
           throwString err
       Right (MiningResult nonceBytes numNonces hps _) -> do
           let newBytes = nonceBytes <> B.drop 8 blockbytes
-              secs = numNonces `div` hps
+              secs = numNonces `div` max 1 hps
```
mightybyte: Is this right?  Dividing `numNonces` by `hps` doesn't seem like the right thing.

fosskers: That's how you get the seconds back out.

mightybyte: Oh, disregard.  This is `secs`.

PR #567, opened 2019-10-23
[Tool] Generate sqlite database checksums

emmanueldenloye:


tools/cwtool/CwTool.hs
```diff
@@ -81,6 +82,10 @@ topLevelCommands =
       "test-miner"
       "Test an external miner"
       TestMiner.main
+  , CommandSpec
+      "db-checksum"
+      "Generate a checksum of all the checkpointer database tables in a block (or range of blocks)"
```
gregorycollins: We need to decide what we want. It's either going to be

  - "Give me all the db row state between blocks a and b"
  - "Give me all the db row state up to block b"

Originally I meant the latter.

tools/db-checksum/CheckpointerDBChecksum.hs
```diff
@@ -0,0 +1,131 @@
...
+    getEndingTxId :: BlockHeight -> Database -> IO Int64
+    getEndingTxId bh db = do
+        r <- qry db stmt [SInt $ fromIntegral bh] [RInt]
+        case r of
+          [[SInt txid]] -> return txid
+          _ -> error "Cannot find ending txid."
+      where
+        stmt = "SELECT endingtxid FROM BlockHistory WHERE blockheight = ?;"
+
+    tistmt = "SELECT * FROM TransactionIndex ORDER BY blockheight DESC WHERE blockheight <= ?;"
```
gregorycollins: In order to make a deterministic hash you need a deterministic ordering -- need an `ORDER BY` clause that mentions every field. Same for the other queries.

tools/db-checksum/CheckpointerDBChecksum.hs
```diff
@@ -0,0 +1,131 @@
...
+          qry db (usertablestmt t) [SInt txid] [RText, RInt, RBlob]
+
+    getEndingTxId :: BlockHeight -> Database -> IO Int64
+    getEndingTxId bh db = do
+        r <- qry db stmt [SInt $ fromIntegral bh] [RInt]
+        case r of
+          [[SInt txid]] -> return txid
+          _ -> error "Cannot find ending txid."
+      where
+        stmt = "SELECT endingtxid FROM BlockHistory WHERE blockheight = ?;"
```
gregorycollins: add `ORDER BY`

tools/db-checksum/CheckpointerDBChecksum.hs
```diff
@@ -0,0 +1,131 @@
...
+
+callDb :: Text -> (Database -> ReaderT SQLiteEnv IO a) -> ReaderT SQLiteEnv IO a
+callDb callerName action = do
+    c <- asks _sConn
+    tryAny (action c) >>= \case
+      Left err -> internalError $ "callDb (" <> callerName <> "): " <> (pack $ show err)
+      Right r -> return r
+
+getTableNames :: ReaderT SQLiteEnv IO [ByteString]
+getTableNames = callDb "getTableNames" $ \db -> do
```
gregorycollins: Table names should come from our internal creation manifest. One cross check we can do here is to _also_ query sqlite for the tables, and complain if we find any that aren't supposed to be there

tools/db-checksum/CheckpointerDBChecksum.hs
```diff
@@ -0,0 +1,131 @@
...
+  where
+    stmt = "SELECT name FROM sqlite_master WHERE type='table';"
+    toByteString (SText (Utf8 bytestring)) = bytestring
+    toByteString _ = error "impossible case"
+    errMsg = "Somehow there are no tables in this connection. This should be an impossible case."
+
+
+data Args = Args
+   {  aFilePath    :: FilePath
+   , aBlockHeight :: BlockHeight
```
gregorycollins: You can just do `BlockHeight` here since the block hash is inferrable based on db state

tools/db-checksum/CheckpointerDBChecksum.hs
emmanueldenloye: Understood.

tools/db-checksum/CheckpointerDBChecksum.hs
emmanueldenloye: Why would we need `ORDER BY` here? There can only be one result by design.

tools/db-checksum/db-checksum-config.yaml
```diff
@@ -0,0 +1,8 @@
+tablesOutputLocation: dboutputhashes
+getAllTables: true
+endBlockHeight: 5
+sqliteFile: ../pact-v1-chain-8.sqlite
```
emmanueldenloye: This is sqlite file I have on disk. I'll change it some reasonable default value before the close of this PR.

PR #570, opened 2019-10-24
Fail early with meaningful error when tx decoding fails

larskuhtz:


src/Chainweb/Pact/PactService.hs
```diff
@@ -949,16 +949,17 @@ toHashCommandResult = over (P.crLogs . _Just) $ P.pactHash . encodeToByteString
...
-    let !transSeq = _payloadDataTransactions plData
-    let !transList = toList transSeq
-    let !bytes = _transactionBytes <$!> transList
-    let !eithers = toCWTransaction <$!> bytes
-    -- Note: if any transactions fail to convert, the final validation hash
-    -- will fail to match the one computed during newBlock
-    let theRights = rights eithers
-    return $! V.fromList theRights
+    unless (null theLefts) $ do
+        throwM $ TransactionDecodeFailure $ "Failed to decode pact transactions: "
```
mightybyte: I'm not sure that throwing is the right thing here.  It might be, but we should run it by @gregorycollins and @slpopejoy

larskuhtz: We reach this code only after either the mempool parsed the tx or we got the block from another node. So, if parsing fails, it's almost certainly due to the latter, in which case we reject the block. Throwing an exception during pact validation is a way to reject the block.

(Currently, we don't blacklist the sender, so we may try to validate the same block again latter, but that problem is a broader issue and isn't specific to this failure case)

src/Chainweb/Pact/PactService.hs
```diff
@@ -949,16 +949,17 @@ toHashCommandResult = over (P.crLogs . _Just) $ P.pactHash . encodeToByteString
...
+            <> (T.intercalate ". " $ T.pack <$> theLefts)
+    return $! V.fromList (rights eithers)
   where
     toCWTransaction bs = codecDecode chainwebPayloadCodec bs
+    !transSeq = _payloadDataTransactions plData
+    !transList = toList transSeq
+    !bytes = _transactionBytes <$!> transList
+    !eithers = toCWTransaction <$!> bytes
+    theLefts = lefts eithers
```
emilypi: request: can we just inline some of this stuff? Like, 

```
!bytes = _transactionBytes <$!> 
    toList (_payloadDataTransactions plData)
```

There's a lot of indirection here that I've been meaning to factor out

larskuhtz: I leave this for a follow up PR, since it would be done best for all places in this module in a single PR.

PR #574, opened 2019-10-24
Mainnet

fosskers:
This PR defines a new `Mainnet01` value for `ChainwebVersion`. Critically, this is the only version that defines a TX Blackout Date.


src/Chainweb/Version.hs
```diff
@@ -554,15 +563,18 @@ window FastTimedCPM{} = Nothing
...
+txSilenceEndDate :: ChainwebVersion -> Maybe (Time Micros)
+txSilenceEndDate Test{} = Nothing
+txSilenceEndDate TimedConsensus{} = Nothing
+txSilenceEndDate PowConsensus{} = Nothing
+txSilenceEndDate TimedCPM{} = Nothing
+txSilenceEndDate FastTimedCPM{} = Nothing
+txSilenceEndDate Development = Nothing
+txSilenceEndDate Testnet02 = Nothing
+-- Thursday, 2019 December 5, 12:00 AM
+txSilenceEndDate Mainnet01 = Just . Time $ TimeSpan 1575504000000000
```
fosskers: @slpopejoy @buckie @mightybyte This will disable mining coordination and new cut submission the instant this time is passed. If stopped, a node can't be restarted after this date either. Are we concerned about sudden drop-offs of Nodes/Miners?

larskuhtz: Is there a value to have many stalled nodes in the network? Those nodes would still validate and forward data, but what's the value for the network?

Less nodes mean a smaller diameter and lower latencies. More nodes mean more protection against partitions and attacks.

Assuming that we, and others, trust our bootstrap nodes, the benefits of a larger network don't seem that important.

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -90,6 +90,8 @@ genesisTime FastTimedCPM{} = BlockCreationTime epoch
 genesisTime Development = BlockCreationTime . Time $ TimeSpan 1563388117613832
```
slpopejoy: We need these times to be human-readable in code

src/Chainweb/Version.hs
slpopejoy: We have no choice but to shut them down, if they don't switch we'll just have to deal with it.

slpopejoy: We need the code for these time values to be human-readable, not comments

larskuhtz: This code is executed on each block validation. Parsing time values is expensive. Let's make sure that the implementation is efficient (e.g. by caching the result).

larskuhtz: Also, if we use something human readable, we should use a standardized format, that is unambiguous with respect to the time system. In block creation times (and this time value is compared with block creation times) we use SI seconds since POSIX epoch. 

So, it should be clear how a the human readable format is converted to that standard.

src/Chainweb/BlockHeader/Genesis.hs
fosskers: Hence the comment above it. In particular, this is a field that goes into a `BlockHeader`, which is a number-based `BlockCreationTime`. These are often used for comparisons.

slpopejoy: Obviously, but you can create these here using a utility function.

src/Chainweb/Version.hs
slpopejoy: See comment above.

larskuhtz: If it has to be human readable, I propose that we use ISO 8601 format `yyyy-mm-ddTHH:MM:SS+00:00`. @fosskers is using `NOINLINE` enough to ensure that it's not parsed each time it is used? Otherwise, with template Haskell it could be reduced to a numeric constant at compile time and inlined (`Time` and `TimeSpan` are both newtypes).

When comparing POSIX seconds with current times, the system clock is used which is usually based on UTC. That means, that we ignore leap seconds and do conversions assuming that each day has 86400 seconds. (I would prefer to convert based on TAI seconds, but most systems don't provide a clock for that. Also UTC is much more common.) Therefore, @slpopejoy has a point here, that providing the constant in UTC format is more explicit about the time standard we are using.

src/Chainweb/BlockHeader/Genesis.hs
slpopejoy: We need a human-verifiable value for these constants as they constitute serious business risk if they are wrong. Either
1) Offer a solution to generate them performantly or
2) Write a unit test that proves they are correct.

mightybyte: Also notice that the comment for `Testnet02` has gone stale because it is the same value as `Development` but the comment says something different.  Furthermore, none of these comments are in UTC and they don't specify the timezone that they are in.  So yeah...see the new PR #581 I just opened to be merged into this one.

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -117,6 +119,9 @@ genesisBlockPayload Development cid = case chainIdInt @Int cid of
 genesisBlockPayload Testnet02 cid = case chainIdInt @Int cid of
     0 -> PN0.payloadBlock
     _ -> PNN.payloadBlock
+genesisBlockPayload Mainnet01 cid = case chainIdInt @Int cid of
+    0 -> PN0.payloadBlock
+    _ -> PNN.payloadBlock
```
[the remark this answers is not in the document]
mightybyte: I don't see any `Ea` changes here.  Is that an oversight?

src/Chainweb/Version.hs
```diff
@@ -552,12 +562,14 @@ window FastTimedCPM{} = Nothing
...
+
+txSilenceEndDate :: ChainwebVersion -> Maybe (Time Micros)
+txSilenceEndDate Test{} = Nothing
+txSilenceEndDate TimedConsensus{} = Nothing
+txSilenceEndDate PowConsensus{} = Nothing
+txSilenceEndDate TimedCPM{} = Nothing
+txSilenceEndDate FastTimedCPM{} = Nothing
+txSilenceEndDate Development = Nothing
+txSilenceEndDate Testnet02 = Nothing
+txSilenceEndDate Mainnet01 = Just [timeMicrosQQ| 2019-12-05T00:00:00.0 |]
```
slpopejoy: 👍

PR #579, opened 2019-10-25
use url in elastic search config to support https

larskuhtz:


fosskers:
Parsing the config file doesn't fail when the URL is malformed, instead it fails at runtime. This is different from the previous behaviour.

PR #581, opened 2019-10-26
Add quasiquoter for time constants

mightybyte:


mightybyte:
Note that the target of this PR is @fosskers other branch for which there is an open PR to master.

mightybyte:
Yes, I confirmed the outputs.  This second commit `77fad51` generates exactly the same values that were there before.

larskuhtz:
I think, we should still write a small test case to protect against regressions in the time library. But we can do that later.

mightybyte:
Agreed.

PR #583, opened 2019-10-26
[WIP] Debugging `attemptBuyGas`

LindaOrtega:
(Research PR)
Couple of observations:

- `beginTx: rolling back`: occurs when there are 1+ transactions in a block and at least one of those is a failing one (any pact failure not just failing the buy gas validation). This PR changed the `invalidBuyGasTest` to send in 6 failing transactions (from the debugging messages I'm printing out, all 6 are included in one block). The error doesn't appear if its just one failing transaction in the block. I also suspect that it won't appear unless there's another transaction that runs after the failing transaction in the block.

- `validation error due to hash mismatch`: Only shows up in the test when (1) at least one transaction in the block fails to buy gas, (2) when `runCoinbase` is included in the buy gas validation function, and (3) I can't replicate the error when there's at least one transaction in the block that passes buy gas validation. This PR adds `runCoinbase` to `attemptBuyGas` to replicate this validation error.

- `validation error due to hash mismatch`: In my tests it fails because the outputs hash differs for the block in new block vs. validate block. After adding more debugging statements, its seems that the coinbase outputs hash is what's different. Specifically, the txId differs by one (txId = 6 in new block vs txId = 5 in validate block).


slpopejoy:
Superseded by #595

PR #586, opened 2019-10-27
Gas Payer Final that builds

slpopejoy:


mightybyte:
Closing because we got #585 fixed and merged.

PR #594, opened 2019-10-28
Miner Configuration

larskuhtz:
* [x] Turn miner off by default
* [x] Set default `MinerInfo` to empty account (invalid)
* [x] Add config validation rule that reminds user to configure `MinerInfo`
* [x] Enable mining in tests and provide a valid configuration

Misc:

* [x] Don't change type of asynchronous exceptions when catching and re-throwing them


larskuhtz:
Here is the new behavior of miner config. Default is, that in-process mining is disabled:

```sh
lars@paul:~/Code/chainweb/chainweb-master (lars/miner-config) [GHC-8.8.1]$ ./dist-newstyle/build/x86_64-osx/ghc-8.8.1/chainweb-0.2/build/chainweb-node/chainweb-node --print-config
databaseDirectory: null
resetChainDatabases: false
chainweb:
  throttleRate: 1000
  miner:
    enabled: false
    configuration:
      minerInfo:
        account: ''
        predicate: keys-all
        public-keys: []
      minimumBlockHeight: 1200
  includeOrigin: true
...
```

If mining is enabled, it is validated that the miner info is defined:

```sh
lars@paul:~/Code/chainweb/chainweb-master (lars/miner-config) [GHC-8.8.1]$ ./dist-newstyle/build/x86_64-osx/ghc-8.8.1/chainweb-0.2/build/chainweb-node/chainweb-node --print-config --enable-mining
chainweb-node: Mining is enabled but not miner id is configured
  error, called at src/Configuration/Utils.hs:680:39 in cnfgrtn-tls-0.4.1-b60165ea:Configuration.Utils
```

larskuhtz:
I wonder, if we should change the default miner configuration only for mainnet and leave it as is for the other networks?

src/Chainweb/Miner/Config.hs
```diff
@@ -45,24 +66,49 @@ makeLenses ''MinerConfig
...
+        <> internal
+    <*< configMinimumBlockHeight .:: fmap (int @Int) % option auto
+        % long "minimum-mining-height"
+        <> short 'm'
+        <> help "minimum block height for mining"
+
+validateMinerConfig :: ConfigValidation MinerConfig l
+validateMinerConfig c =
+    when (mid == "")
+        $ throwError "Mining is enabled but not miner id is configured"
```
larskuhtz:
```suggestion
        $ throwError "Mining is enabled but no miner id is configured"
```

src/Chainweb/Miner/Config.hs
```diff
@@ -45,24 +66,49 @@ makeLenses ''MinerConfig
 defaultMinerConfig :: MinerConfig
 defaultMinerConfig = MinerConfig
     { _configTestMiners = MinerCount 10
-    , _configMinerInfo = noMiner
+        -- hidden configuration option that is only used in testing
+    , _configMinerInfo = invalidMiner
+    , _configMinimumBlockHeight = 1200
     }
+  where
+    invalidMiner = Miner
```
slpopejoy: Shouldn't this be the same as `noMiner` (or `noMiner` should be renamed to `invalidMiner`) which is used elsewhere?

larskuhtz: `noMiner` the currently used for mining by test-nodes on devnet network and testnet. It's probably a poorly chosen name. `invalidMiner` here is just a configuration "template". Configuration validation detects it and reminds the user to fill in a value.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -442,7 +442,7 @@ doCommit = use bsMode >>= \mm -> case mm of
     Just m -> do
         txrs <- if m == Transactional
           then do
-              modify' (over bsTxId succ)
+              bsTxId %= succ
```
mightybyte: Can we revert this change?  It's not obvious how this impacts strictness.  And it's unrelated to the rest of the PR.

src/Chainweb/Pact/Backend/Utils.hs
```diff
@@ -82,6 +82,7 @@ withSavepoint name action = mask $ \resetMask -> do
         liftIO $ evaluate r
     throwErr s = internalError $ "withSavepoint (" <> asString name <> "): " <> pack s
     handlers = [ Handler $ \(e :: PactException) -> throwErr (show e)
+               , Handler $ \(e :: AsyncException) -> throwM e
```
mightybyte: @slpopejoy @gregorycollins should sign off on this

slpopejoy: Looks good to me

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
larskuhtz:
```suggestion
              modify' (over bsTxId succ)
```

PR #595, opened 2019-10-28
Buy gas sequencing fix

slpopejoy:


slpopejoy:
Reopened to merge #596 onto master

PR #596, opened 2019-10-28
untangle buyGas check DRAFT

slpopejoy:


src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -440,15 +441,22 @@ doCreateUserTable tn@(TableName ttxt) mn = do
...
 {-# INLINE doCreateUserTable #-}
 
+logit :: Show a => a -> BlockHandler SQLiteEnv ()
+logit a = asks _logger >>= \l -> liftIO $ logLog l "ERROR" ("logit: " ++ show a)
+
 doRollback :: BlockHandler SQLiteEnv ()
-doRollback = bsPendingTx .= Nothing
+doRollback = do
+  logit ("TX","doRollback")
+  resetTemp
```
gregorycollins: I think this will break the checkpointer. If you've had successful txs played in the block, but you play a subsequent tx, and roll it back, won't the `resetTemp` clear out all of the pending transactions for the block? We _do_ need to set `bsMode .= Nothing` here though.

src/Chainweb/Pact/PactService.hs
```diff
@@ -517,24 +517,18 @@ type BuyGasValidation = T3 ChainwebTransaction Uniqueness AbleToBuyGas
...
     -> Vector (T2 ChainwebTransaction Uniqueness)
     -> PactServiceM cas (Vector BuyGasValidation)
-attemptBuyGas cp miner nonGenesisParentHash txs = withDiscardedBatch $ do
-    mbLatestBlock <- liftIO $ _cpGetLatestBlock cp
-    (bh, bhash) <- case mbLatestBlock of
-        Nothing -> throwM NoBlockValidatedYet
-        Just !p -> return p
-    let target = Just (succ bh, bhash)
-    withCheckpointer target "attemptBuyGas" $ \(PactDbEnv' dbEnv) -> do
+attemptBuyGas miner nonGenesisParentHash (PactDbEnv' dbEnv) txs = do
```
gregorycollins: :+1: this was a good catch that we were unintentionally recursing on the savepoints here. _WHY_ sqlite lets you do this instead of throwing an error is beyond me

src/Chainweb/Pact/PactService.hs
```diff
@@ -750,6 +746,9 @@ execNewBlock mpAccess parentHeader miner creationTime = withDiscardedBatch $ do
     pHeight = _blockHeight parentHeader
     pHash = _blockHash parentHeader
     bHeight = succ pHeight
+    updateMP = liftIO $ do
```
gregorycollins: rename as `updateMempool`

src/Chainweb/Pact/PactService.hs
```diff
@@ -720,20 +707,29 @@ execNewBlock
     -> Miner
     -> BlockCreationTime
     -> PactServiceM cas PayloadWithOutputs
-execNewBlock mpAccess parentHeader miner creationTime = withDiscardedBatch $ do
-    withCheckpointerRewind (Just (bHeight, pHash)) "execNewBlock" $ \pdbenv -> do
+execNewBlock mpAccess parentHeader miner creationTime =
+  do
+    updateMP
```
gregorycollins: You moved it outside but I'm not sure there's a reason to, neither of the functions called should touch the DB

src/Chainweb/Pact/PactService.hs
```diff
@@ -720,20 +707,29 @@ execNewBlock
...
+    newTrans <- withDiscardedBatch $ withCheckpointerRewind (Just (bHeight, pHash)) "preBlock" $ \pdbenv -> do
```
gregorycollins: Rewrite to avoid `withDiscardedBatch` / `withCheckpointerRewind` twice as discussed in chat -- you can just `getCheckpointer`, run your stuff, then `_cpDiscard cp >> _cpRestore cp (Just (bheight, pHash))` should reset your state to play the next block

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
slpopejoy: ok. I confess to not understanding how the txlog works in the pending stuff ...

src/Chainweb/Pact/PactService.hs
slpopejoy: This was just to make it clear that this doesn't "roll back" or something. When I see stuff bracketed I assume the scope of the update is somehow contained, so putting it outside makes it easier for me at least to understand.

slpopejoy:
Closed to merge off master.

PR #597, opened 2019-10-28
Make chainweb-miner a bit smarter

emmanueldenloye:
This PR adds these features:

* Generate keypair for miners
* Report the estimated time that it took to successfully mine a block
* Allow the miner to specify multiple nodes to mine from (in a round robin fashion)
* Following on the previous feature, the user now doesn't have to specify the Chainweb version. This is automatically retrieved for each url, when the script is started. We might want to consider different behavior should any of the version retrievals fail. Right now, the entire script fails if any one of the urls does not have an accompanying Chainweb version.


miner/Miner.hs
```diff
@@ -125,21 +138,25 @@ data Env = Env
     { envGen :: !MWC.GenIO
     , envMgr :: !Manager
     , envLog :: !LogFunc
+    , envCmd :: !Command
     , envArgs :: !ClientArgs
     , envHashes :: IORef Word64
-    , envSecs :: IORef Word64 }
+    , envSecs :: IORef Word64
+    , envSuccessInitTime :: IORef Int
+    , envUrls :: MVar (NESeq (T2 BaseUrl ChainwebVersion)) }
```
fosskers: Make this an `IORef` to unify it with the other fields and reduce the number of imports necessary.

PR #598, opened 2019-10-28
Primed Coordinator

fosskers:
This PR allows a Node to set a list of "blessed" Miner identities. Each such miner will have Payloads preformed for it, such that new work requests will return (almost) immediately without incurring the usual costs of going through `_pactNewBlock`.


src/Chainweb/Chainweb.hs
```diff
@@ -202,6 +204,7 @@ data ChainwebConfiguration = ChainwebConfiguration
     { _configChainwebVersion :: !ChainwebVersion
     , _configNodeId :: !NodeId
     , _configMiner :: !(EnableConfig MinerConfig)
+    , _configBlessedMiners :: ![Miner]
```
fosskers: Any better name?

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -68,27 +77,85 @@ data MiningCoordination logger cas = MiningCoordination
     , _coordCutDb :: !(CutDb cas)
     , _coordState :: !(TVar MiningState)
     , _coordLimit :: !Int
-    , _coord503s :: IORef Int }
+    , _coord503s :: !(IORef Int)
+    , _coordMiners :: !(TVar CachedPayloads)
```
fosskers: A newtype'd `Map` defined in `Coordinator.hs`.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -68,27 +77,85 @@ data MiningCoordination logger cas = MiningCoordination
...
+        creationTime <- BlockCreationTime <$> getCurrentTimeIntegral
+        payload <- _pactNewBlock pact m parent creationTime
+        pure $ T2 payload creationTime
+
+    cachePayloads :: TVar CachedPayloads -> Cut -> IO ()
+    cachePayloads tcp old = do
+        new <- atomically $ _cutStm cdb
+        let oldChains = HM.keysSet $ _cutMap old
+            newChains = HM.keysSet $ _cutMap new
+            difChains = HS.toList  $ HS.difference newChains oldChains
```
fosskers: The `HashSet`s are only size 10, so this is a cheap `difference` call.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -68,27 +77,85 @@ data MiningCoordination logger cas = MiningCoordination
...
+            parents   = map (\cid -> new ^?! ixg cid) difChains
+        CachedPayloads cached <- readTVarIO tcp
+        newCached <- M.traverseWithKey (g parents) cached
+        atomically . writeTVar tcp $ CachedPayloads newCached
```
fosskers: No other thread *writes* to this `TVar`, so it's safe to have its read/write separated. There is still a risk that the "best" Cut updates while all this is happening, though.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -68,27 +77,85 @@ data MiningCoordination logger cas = MiningCoordination
...
+      where
+        g :: [BlockHeader]
+          -> Miner
+          -> M.Map ChainId (T2 PayloadWithOutputs BlockCreationTime)
+          -> IO (M.Map ChainId (T2 PayloadWithOutputs BlockCreationTime))
+        g ps miner pays = do
+            news <- M.fromList <$> traverse (\p -> (_chainId p,) <$> getPayload miner p) ps
+            -- `union` is left-biased, and so prefers items from the first Map
+            -- when duplicate keys are encountered.
+            pure $ M.union news pays
```
fosskers: Cheap bulk update of the `Map`.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -134,13 +140,20 @@ newWork choice miner pact c = do
...
         Just adjParents -> do
-            -- Fetch a Pact Transaction payload. This is an expensive call
-            -- that shouldn't be repeated.
+            -- Fetch creation time and payload.
             --
-            creationTime <- getCurrentTimeIntegral
-            payload <- _pactNewBlock pact miner p (BlockCreationTime creationTime)
+            CachedPayloads cached <- readTVarIO tcp
+            T2 payload creationTime <- case M.lookup miner cached >>= M.lookup cid of
+                Just pt -> pure pt
```
fosskers: This is where the speed comes from.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -68,27 +77,85 @@ data MiningCoordination logger cas = MiningCoordination
...
+    fromCut m cut = M.fromList <$> traverse (\(cid, bh) -> (cid,) <$> getPayload m bh) cut
+
+    getPayload :: Miner -> BlockHeader -> IO (T2 PayloadWithOutputs BlockCreationTime)
+    getPayload m parent = do
+        creationTime <- BlockCreationTime <$> getCurrentTimeIntegral
+        payload <- _pactNewBlock pact m parent creationTime
+        pure $ T2 payload creationTime
+
+    cachePayloads :: TVar CachedPayloads -> Cut -> IO ()
+    cachePayloads tcp old = do
```
fosskers: @larskuhtz I'd say this should be in `runForever`, but the `Cut` has to update on each loop. What `Cut` would serve as the default when an exception is thrown?

fosskers:
Found a big mistake. Correcting now...

PR #601, opened 2019-10-29
[mempool] don't reintroduce old transactions

gregorycollins:
When resolving a fork, there's no reason to reintroduce transactions to the mempool that would already be expired if they were played now.

### DIFF
--- a/src/Chainweb/Mempool/Consensus.hs
+++ b/src/Chainweb/Mempool/Consensus.hs
@@ -35,6 +35,8 @@ import qualified Data.Vector as V
 
 import GHC.Generics
 
+import Pact.Types.ChainMeta (TTLSeconds(..), TxCreationTime(..))
+
 import System.LogLevel
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


mightybyte:
This is a pretty significant change.  @gregorycollins you need to add a test for this or there's no way we can put this in today.

gregorycollins:
OK, I'm adding a test for the filtering behavior, and just reusing the existing TTL check code which already has a test case

src/Chainweb/Mempool/Consensus.hs
```diff
@@ -51,8 +54,8 @@ import Data.CAS
 import Data.LogMessage (JsonLog(..), LogFunction)
 
 ------------------------------------------------------------------------------
-data MempoolConsensus t = MempoolConsensus
-    { mpcMempool :: !(MempoolBackend t)
+data MempoolConsensus = MempoolConsensus
+    { mpcMempool :: !(MempoolBackend ChainwebTransaction)
```
larskuhtz: 👍

PR #604, opened 2019-10-29
[mempool] run a thread to force mempool pruning in case mining is off.

gregorycollins:


src/Chainweb/Miner/Miners.hs
```diff
@@ -93,6 +95,20 @@ localTest lf v m cdb gen miners = runForever lf "Chainweb.Miner.Miners.localTest
...
+-- regularly to prune mempool.
+mempoolNoopMiner :: LogFunction -> ChainwebVersion -> Miner -> CutDb cas -> IO ()
+mempoolNoopMiner lf v m cdb = runForever lf "Chainweb.Miner.Miners.mempoolNoopMiner" loop
+  where
+    chains = toList $ chainIds v
+    pact = _webPactExecutionService . _webBlockPayloadStorePact $
+           view cutDbPayloadStore cdb
+    loop = do
+        c <- _cut cdb
+        mapM_ (\cid -> newWork (Suggestion cid) m pact c) chains
```
slpopejoy: Does `IO` guarantee execution of the `mapM_`?

gregorycollins: Yes

PR #612, opened 2019-10-30
Randomize peerdb list

larskuhtz:


src/P2P/Node/PeerDB.hs
```diff
@@ -175,8 +180,21 @@ instance Arbitrary PeerEntry where
 -- -------------------------------------------------------------------------- --
 -- Peer Entry Set
 
+pdNonce :: Int
+pdNonce = unsafePerformIO randomIO
+{-# NOINLINE  pdNonce #-}
```
fosskers: Known technique, should be fine.

src/P2P/Node/PeerDB.hs
```diff
@@ -175,8 +180,21 @@ instance Arbitrary PeerEntry where
 -- -------------------------------------------------------------------------- --
 -- Peer Entry Set
 
+pdNonce :: Int
+pdNonce = unsafePerformIO randomIO
```
larskuhtz: Consider generating this in `newEmptyPeerDb`

emilypi: This is safe because all `randomIO` needs is an incrementing state token to be sufficiently random. For ergonomics' sake, I agree with @larskuhtz here. Less to think about/deal with and it does roughly the same thing.

src/P2P/Node/PeerDB.hs
```diff
@@ -98,6 +101,8 @@ import Prelude hiding (null)
 
 import System.IO.SafeWrite
 import System.IO.Temp
+import System.IO.Unsafe
+import System.Random
```
gregorycollins: In general: prefer `mwc-random` always, although I understand this is a one-off

src/P2P/Node/PeerDB.hs
lehins: lol. Was looking at `randomIO` use cases and accidentally stumbled upon this. You realize that it is possible to set this global "pure" `pdNonce` binding to some desired value ahead of time. 
```haskell
λ> x = mkStdGen 217
λ> :set -XTypeApplications
λ> (myNonce, _) = random @Int x 
λ> setStdGen x
λ> pdNonce == myNonce
True
```
Considering that it is a nonce it seems like a security concern. I am not familiar with the rest of the app of course, so it might be fine.

larskuhtz: What we need here is that each node uses a different order when sorting peer entries.

I assume (not sure that I actually checked) that the standard generator used by `randomIO` is seeded with system entropy. If that's the case, I think, this code is fine. I don't see a security concern. Randomness here is used for performance rather than security. Also, if a remote attacker could guess the nonce, I don't see how it could be exploited for a DOS attack.

lehins: Current time is used for initialization of `theStdGen` instead of system entropy.

> Randomness here is used for performance rather than security.

In that case this code should be fine.

PR #615, opened 2019-10-30
Bump cabal version to 1.0

mightybyte:


chainweb.cabal
```diff
@@ -1,7 +1,7 @@
 cabal-version: 2.2
 
 name:         chainweb
-version:      0.2
+version:      1.0
```
fosskers: Perhaps `1.0.0` to follow semver?

mightybyte: This is Haskell.  We're using the PVP, not SemVer.

PR #617, opened 2019-10-30
Additional Tx Gen Repl functions

marklnichols:


tools/txg/TXG/Repl.hs
```diff
@@ -70,6 +96,10 @@ import TXG.Simulate.Contracts.CoinContract
 import TXG.Simulate.Contracts.Common
 import TXG.Simulate.Utils
 
+sender00Public, sender00Secret :: Text
```
mightybyte: Can we get this from somewhere else?  I recall there being something that read these directly from the YAML file.

marklnichols: I know we do in SBFT for tests and the demo, probably the same somewhere in Chainweb

PR #626, opened 2019-10-31
Guard against GPU False Positives

fosskers:


miner/Miner.hs
```diff
@@ -412,6 +414,15 @@ gpu (GPUEnv mpath margs) (TargetBytes target) (HeaderBytes blockbytes) = do
...
               secs = numNonces `div` max 1 hps
-          modifyIORef' (envHashes e) (+ numNonces)
-          modifyIORef' (envSecs e) (+ secs)
-          return $! HeaderBytes newBytes
+
+          -- FIXME Consider removing this check if during later benchmarking it
+          -- proves to be an issue.
+          bh <- runGet decodeBlockHeaderWithoutHash newBytes
+
+          if | not (prop_block_pow bh) -> do
```
mightybyte: Why do we need `MultiWayIf` for this?

PR #628, opened 2019-10-31
Fix Testnet Genesis

emilypi:


pact/genesis/devnet/development/ns.yaml
```diff
@@ -0,0 +1,23 @@
+codeFile: ../../../namespaces/ns.pact
```
slpopejoy: Let's swap this ns.yaml (and other devnet ones) with what's in testnet, and not duplicate the devnet ones.

Thus
- devnet/development and devnet/fasttimedCPM share the "short" ns.yaml currently in testnet
- testnet/ns.yaml uses this file

pact/genesis/testnet/grantsN.yaml
```diff
@@ -23,7 +23,7 @@ code: |-
   (coin.coinbase "tony" (read-keyset "tony") 500.0)
   (coin.coinbase "viv" (read-keyset "viv") 500.0)
   (coin.coinbase "will" (read-keyset "will") 500.0)
-
+  (coin.coinbase "Kadena" (keyset-ref-guard "Kadena") 10000000.0)
```
mightybyte: Change the nonce below

pact/genesis/devnet/development/ns.yaml
emilypi: Fixed

PR #629, opened 2019-10-31
Round-robin Chain Selection

fosskers:


src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -80,14 +82,16 @@ withMiningCoordination
 withMiningCoordination logger enabled cutDb inner
     | not enabled = inner Nothing
     | otherwise = do
-        t <- newTVarIO mempty
-        c <- newIORef 0
-        fmap snd . concurrently (prune t c) $ inner . Just $ MiningCoordination
+        ms <- newTVarIO mempty
+        ec <- newIORef 0
+        cs <- newTVarIO . cycle . HS.toList $ chainIds cutDb
```
fosskers: `cycle` is supposed to be constant time / space.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -157,17 +160,23 @@ newWork logFun choice miner pact c = do
...
-      new <- randomChainId c
+      new <- nextChainId tchains
       bool (pure new) (loop cid) $ new == cid
 
+nextChainId :: TVar [ChainId] -> IO ChainId
+nextChainId tchains = atomically $ do
+    cids <- readTVar tchains
+    writeTVar tchains $ tail cids
+    pure $ head cids
+
```
fosskers: The main contribution of the PR.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -157,17 +160,23 @@ newWork logFun choice miner pact c = do
...
+chainChoice :: TVar [ChainId] -> ChainChoice -> IO ChainId
+chainChoice tchains choice = case choice of
+    Anything -> nextChainId tchains
     Suggestion cid -> pure cid
     TriedLast cid -> loop cid
   where
     loop :: ChainId -> IO ChainId
     loop cid = do
-      new <- randomChainId c
+      new <- nextChainId tchains
```
larskuhtz: I am a bit uncomfortable with this change. This means all chains are visited in the same order by all nodes. Sure, different nodes should mine at different chains, but, if for instance many chains become blocked, all nodes would synchronize and start in the same order.

Usually, uniform random choice is a powerful algorithmic tool that enforces many desired properties at very low cost (assuming availability of a reliable random source). In particular it is efficient in mitigating many attacks. 

At the very least I propose that each node picks a random permutation of the chains at startup.

larskuhtz: Since this is at the core of Chainweb mining strategy: @buckie what are your thoughts on this?

fosskers: In practice this list will progress _very_ quickly, and also at different rates on each node that is doing coordination. The nodes won't in any way be synchronized.

gregorycollins: Yes we have to shuffle the list on startup. :+1:

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -157,17 +160,23 @@ newWork logFun choice miner pact c = do
...
       bool (pure new) (loop cid) $ new == cid
 
+nextChainId :: TVar [ChainId] -> IO ChainId
+nextChainId tchains = atomically $ do
+    cids <- readTVar tchains
+    writeTVar tchains $ tail cids
```
gregorycollins: force tail w/ `$!` here

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -68,7 +70,8 @@ data MiningCoordination logger cas = MiningCoordination
     , _coordCutDb :: !(CutDb cas)
     , _coordState :: !(TVar MiningState)
     , _coordLimit :: !Int
-    , _coord503s :: IORef Int }
+    , _coord503s :: IORef Int
+    , _coordChains :: TVar [ChainId] }
```
emilypi: Why `TVar`? Do you mind adding haddocks for this structure?

src/Chainweb/Chainweb/MinerResources.hs
mightybyte: This makes me nervous.  What are we trying to accomplish here?

src/Chainweb/Chainweb/MinerResources.hs
fosskers: `TVar` because it is shared between many server threads calling `newWork`.

src/Chainweb/Chainweb/MinerResources.hs
mightybyte: This seems like a recipe for disaster of someone inadvertently forces this.  We need a more obviously foolproof solution.

src/Chainweb/Miner/Coordinator.hs
larskuhtz: Actually, I don't think that's enough. In the beginning there will be relatively few coordinator nodes serving many miners.

larskuhtz: Also this approach only works for < 10 miners. Beyond that it's not better than random sampling, which is safer.

src/Chainweb/Chainweb/MinerResources.hs
larskuhtz: The goal here is to avoid mining twice on the same parent. Instead of making chain selection completely deterministic, we should just do a minimal solution.

I propose:

```suggestion
    , _coordChains :: !(TVar (Vector (ChainId, BlockHash))) }
```

Each time we mine a block, we update the Vector with the block hash. When we look for new work we consider a chain as blocked if the block has for the chain in the cut is equal to the respective hash in `_coordChains`. So, we just extend the notion of "blocked" chain which already exists in the code.

The selection among the non-blocked chains remains the same `radomChainId`.

That has the advantage of less congestion: we update the `TVar` only when we mine a block (which is rare) and read only when we assign work. It also works with more than 10 miners. 

A deadlock can only happen when all nodes in the network get stuck at the same time with validation failure of block that they mind themselves, which is unlikely (and would destroy the network in any case).

larskuhtz:
I am not in favor of this PR for three reasons:

1. I don't see the benefit. Under normal conditions, mining of local orphans should be very rare and hardly an issue. The behavior that we observed with GPU mining is clearly a bug and we should rather investigate the bug than hiding it.

2. I makes chain selection deterministic. For instance, in the case of a single miner chains mined always in the same order. That could for instance increase the likelyhood that chains are skipped more often than others because they are blocked.

3. The approach changes a core algorithmic aspect of mining (random uniform choice of the chain), but only works for a narrow scenario of < 10 miners.

src/Chainweb/Chainweb/MinerResources.hs
fosskers: > Instead of making chain selection completely deterministic

It isn't truly deterministic, since chains would often be skipped due to blockage / cuts coming in from other nodes.

> Each time we mine a block, we update the Vector with the block hash. 

Mine a block, or accept a new cut? Mined blocks are not definitive.

fosskers:
Neither (2) nor (3) mentioned above are strictly true.

PR #633, opened 2019-11-01
More Miner Logging

fosskers:
This PR also introduces a delay between attempts to connect to the update stream.


miner/Miner.hs
```diff
@@ -354,7 +357,8 @@ mining go wb = do
             , port = baseUrlPort u
             , secure = True
             , method = "GET"
-            , requestBody = RequestBodyBS cbs }
+            , requestBody = RequestBodyBS cbs
+            , responseTimeout = responseTimeoutNone }
```
fosskers: This prevents the miner from giving up too early when the block rate naturally extends beyond 30s.

Is `None` too much?

PR #636, opened 2019-11-01
[DO NOT MERGE] generate 2048 bit RSA keys for self-signed certificates

larskuhtz:


larskuhtz:
Testing on devnet indicate that this change doesn't fix the issue of TLS Handshake failures, but instead makes the issue worse.

Beside that this change also causes a considerable delay in increase in node startup time.

PR #638, opened 2019-11-01
[Miner] Add account balance lookup

emmanueldenloye:


miner/README.org
```diff
@@ -175,6 +175,66 @@ You may only be interested in warning or error messages. If so, use the
...
+*** Specifying Mulitple Nodes
+
+You can specify multiple nodes to mine from the command line.
+*However*, what this actually means needs to be clarified. Consider
+this example:
+
+#+BEGIN_SRC bash
+  ./chainweb-miner cpu --node=us-e3.chainweb.com:443
+  --node=us-e2.chainweb.com:443 --miner-account example
+  --miner-key example --miner-pred keys-all --chain 0
```
LindaOrtega:
```
./chainweb-miner cpu --node=us-e3.chainweb.com:443
  --node=us-e2.chainweb.com:443
```
I'd rather have this omit the rest of the arguments and fail to prevent users from just copy-ing and pasting with an incorrect key

miner/README.org
```diff
@@ -175,6 +175,66 @@ You may only be interested in warning or error messages. If so, use the
...
+block you mine is from us-e3.chainweb.com while the second is from
+us-e2.chainweb.com. Instead, the first node found on the command line
+is the first node the chainweb-miner attempts to communicate with to
+find blocks to mine. If for some reason, the chainweb-miner is unable
+to establish a connection with the node, then it will attempt to
+establish a connection with the specified next node. This process will
+continue until there are no more nodes left to consider. Upon this
+event, the chainweb-miner process will halt and exit. For example,
+given the above example, if the chainweb-miner cannot communicate with
+either us-e3.chainweb and us-e2.chainweb.com, the chainweb-miner
```
LindaOrtega:
```suggestion
either us-e3.chainweb and us-e2.chainweb.com, its
```

miner/README.org
```diff
@@ -175,6 +175,66 @@ You may only be interested in warning or error messages. If so, use the
...
+The balance on chain 3 is 53.004029.
+The balance on chain 4 is 57.613075.
+The balance on chain 5 is 76.049259.
+The balance on chain 6 is 53.004029.
+The balance on chain 7 is 50.699506.
+The balance on chain 8 is 64.526644.
+The balance on chain 9 is 48.394983.
+Your total is 580.739796000000
+#+END_EXAMPLE
+
```
LindaOrtega:
```suggestion
We recommend querying your balance from the node you have been mining to as well as a couple other nodes (i.e. the bootstrap nodes).
```

miner/Miner.hs
```diff
@@ -153,6 +170,7 @@ pCommand = hsubparser
     (  command "cpu" (info cpuOpts (progDesc "Perform multicore CPU mining"))
     <> command "gpu" (info gpuOpts (progDesc "Perform GPU mining"))
     <> command "keys" (info keysOpts (progDesc "Generate public/private key pair"))
+    <> command "balance-check" (info balanceCheckOpts (progDesc "Get balances on all chains"))
```
emilypi: Bikeshedding suggestion: `balances` instead of `balance-check`

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
 scheme env = case envCmd env of
     CPU e _ -> cpu e
     GPU e _ -> gpu e
-    Keys -> error "Impossible: You shouldn't reach this case."
+    _ -> error "Impossible: You shouldn't reach this case."
```
emilypi: swallowed up `Keys`?

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
...
+    :: ChainwebVersion
+    -> ChainId
+    -> P.Command Text
+    -> ClientM (P.CommandResult Hash)
+genLocal version cid =
+  case someChainwebVersionVal version of
+    SomeChainwebVersionT (_ :: Proxy cv) ->
+      case someChainIdVal cid of
+        SomeChainIdT (_ :: Proxy ccid) ->
+          client (Proxy :: Proxy (PactLocalApi cv ccid))
```
emilypi: This code exists. See https://github.com/kadena-io/chainweb-node/blob/master/src/Chainweb/Pact/RestAPI/Client.hs#L79

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
...
+
+getBalances :: BaseUrl -> Miner -> IO ()
+getBalances url (Miner (MinerId mi) _) = do
```
emilypi: If you're just after an identifier here, there's no reason to ask for the full `Miner` type. You really just need the account name. All balances and account names are public data, so this is fine.

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
...
+        printf $ "Your total is " <> show (roundTo 12 tot) <> " kda.\n"
+  where
+    ss :: TLSSettings
+    ss = TLSSettingsSimple True True True
+    cenv m = ClientEnv m url Nothing
+    go = do
+      NodeInfo v _ cs _ <- client (Proxy @NodeInfoApi)
+      forM (sort cs) $ \cidtext -> do
+          c <- chainIdFromText cidtext
+          ks <- liftIO $ testSomeKeyPairs
```
emilypi: Test code shouldn't leak into this. Lets create some blessed dummy data to handle local balance checks.

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
...
+          let _code = printf "(coin.get-balance \"%s\")" (T.unpack mi)
+              meta = def { _pmChainId = P.ChainId cidtext }
```
emilypi: doesn't actually make a difference in this case - we don't validate metadata in local calls - we just care about what chain you posted to.

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
...
+-- | note this is "sender00"'s key
+someED25519Pair :: (PublicKeyBS, PrivateKeyBS, Text, PPKScheme)
+someED25519Pair =
+    ( PubBS $ decodeKey
+        "368820f80c324bbc7c2b0610688a7da43e39f91d118732671cd9c7500ff43cca"
+    , PrivBS $ decodeKey
+        "251a920c403ae8c8f65f59142316af3c82b631fba46ddea92ee8c95035bd2898"
+    , "368820f80c324bbc7c2b0610688a7da43e39f91d118732671cd9c7500ff43cca"
+    , ED25519
+    )
```
emilypi: Pls don't cargo cult the test data

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
...
+              return (tot + bal)
+        tot <- foldM f (0 :: Decimal) bbs
+        printf $ "Your total is " <> show (roundTo 12 tot) <> " kda.\n"
+  where
+    ss :: TLSSettings
+    ss = TLSSettingsSimple True True True
+    cenv m = ClientEnv m url Nothing
+    go = do
+      NodeInfo v _ cs _ <- client (Proxy @NodeInfoApi)
+      forM (sort cs) $ \cidtext -> do
```
emilypi: `forM` in this case is `forM :: [Text] -> (Text -> IO (Decimal, Text)) -> IO [(Decimal, Text)]`, which should make it apparent what I'm about to say - it'd be much easier if you handled this as a map 😄.

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
...
+          c <- chainIdFromText cidtext
+          ks <- liftIO $ testSomeKeyPairs
+          let _code = printf "(coin.get-balance \"%s\")" (T.unpack mi)
+              meta = def { _pmChainId = P.ChainId cidtext }
+          cmd <- liftIO $ mkExec _code Null meta (NEL.toList ks) (Just $ P.NetworkId $ toText v) Nothing
+          (P.PactResult result) <- P._crResult <$> genLocal v c cmd
+          return $ case result of
+            Right (PLiteral (LDecimal bal)) ->
+              (bal, "The balance on chain " <> cidtext <> " is " <> (T.pack $ show bal) <> " kda.\n")
+            _ -> (0, "Something went wrong when looking for the balance on chain " <> cidtext <> "\n")
```
emilypi: capital `KDA` for currency codes  or better, use `₭`!

miner/Miner.hs
fosskers: Or singular `balance` even.

miner/Miner.hs
```diff
@@ -215,6 +236,7 @@ main :: IO ()
 main = do
     execParser opts >>= \case
         Keys -> genKeys
+        BalanceCheck url _miner -> getBalances url _miner
```
fosskers: Let's avoid starting names with `_` outside of record field names - it can interact poorly with warnings and hole fits.

miner/README.org
```diff
@@ -175,6 +175,66 @@ You may only be interested in warning or error messages. If so, use the
...
+#+BEGIN_SRC bash
+  ./chainweb-miner cpu --node=us-e3.chainweb.com:443
+  --node=us-e2.chainweb.com:443 --miner-account example
+  --miner-key example --miner-pred keys-all --chain 0
+#+END_SRC
+
+Before discussing the aforementioned necessary clarification, note
+that specifying multiple nodes is rather straightforward: just add
+another --node= clause (as found in the example)! Now, when you
```
fosskers:
```suggestion
another ~--node=~ clause (as found in the example)! Now, when you
```

miner/Miner.hs
```diff
@@ -263,7 +285,72 @@ scheme :: Env -> (TargetBytes -> HeaderBytes -> RIO Env HeaderBytes)
...
+getBalances :: BaseUrl -> Miner -> IO ()
+getBalances url (Miner (MinerId mi) _) = do
+    balanceStmts <- newManager (mkManagerSettings ss Nothing) >>= runClientM go . cenv
+    case balanceStmts of
+      Left e -> throwM e
+      Right bbs -> do
+        let f :: Decimal -> (Decimal, Text) -> IO Decimal
+            f tot (bal, bs) = do
+              printf (T.unpack bs)
+              return (tot + bal)
```
fosskers: Please move this `f` to the `where` clause.

miner/README.org
```diff
@@ -175,6 +175,67 @@ You may only be interested in warning or error messages. If so, use the
   2019-09-16 16:58:23.646547: [error] Failed to fetch work! Is the Node down?
 #+end_example
 
+*** Specifying Mulitple Nodes
```
fosskers: Spelling.

miner/Miner.hs
mightybyte: Use `KDA`, not the Loas Kip symbol.

miner/Miner.hs
mightybyte: Agreed

fosskers:
:wave: please reopen this PR here: https://github.com/kadena-io/chainweb-miner

PR #642, opened 2019-11-04
guard the peer db from unreachable nodes

larskuhtz:


src/P2P/Node.hs
```diff
@@ -298,39 +296,87 @@ setInactive :: P2pNode -> STM ()
...
 
 -- -------------------------------------------------------------------------- --
--- Sync Peers
+-- New Peer Validation Manager
+
+-- | Global Manager for checking reachability of new Peers
+--
+newPeerManager :: IORef HTTP.Manager
+newPeerManager = unsafePerformIO
+    $ unsafeManager 2000000 {- 1 second -} >>= newIORef
```
fosskers: 1 second?

src/P2P/Node.hs
```diff
@@ -298,39 +296,87 @@ setInactive :: P2pNode -> STM ()
...
+        Testnet02 -> isReservedHostAddress (_peerAddr pinf)
+        Development -> isReservedHostAddress (_peerAddr pinf)
         _ -> False
 
-    canConnect :: IO Bool
-    canConnect = runClientM (peerGetClient v nid (Just 0) Nothing) env >>= \case
-        Left _ -> pure False
-        Right _ -> pure True
+    canConnect = do
+        mgr <- getNewPeerManager
```
fosskers: This is just calling the `Manager` that was already created once, not generating a new `Manager` each time? The latter would be quite expensive.

src/P2P/Node.hs
```diff
@@ -350,7 +396,10 @@ syncFromPeer node info = runClientM sync env >>= \case
             logg node Warn $ "failed to sync peers from " <> showInfo info <> ": " <> sshow e
             return False
     Right p -> do
-        goods <- wither (removeTrivialPeers node) $ _pageItems p
+        goods <- fmap catMaybes
+            $ mapConcurrently (guardPeerDbOfNode node)
+            $ filter (\i -> me /= _peerId i)
+            $ _pageItems p
```
fosskers: We expect this page to be small, yeah? So the `mapConcurrently` won't trash the runtime?

PR #643, opened 2019-11-04
CutDB: rate limiting and bad-sender penalties

gregorycollins:
If senders send us cuts we can't quickly verify (e.g. they are behind NAT) it can gum up our queue. We will apply a 20-second cool-down on cut processing to any peer who sends us a bad cut.


src/Chainweb/CutDB.hs
```diff
@@ -132,13 +139,22 @@ import Data.CAS.RocksDB
...
 import Utils.Logging.Trace
 
 -- -------------------------------------------------------------------------- --
 -- Cut DB Configuration
 
+data LimitConfig = LimitConfig
+    { maxBucketTokens :: Int
+    , initialBucketTokens :: Int
+    , bucketRefillTokensPerSecond :: Int
+    , badSenderPenaltySecs :: Int
```
larskuhtz: Make strict and unpack

src/Chainweb/CutDB.hs
```diff
@@ -161,9 +179,13 @@ defaultCutDbConfig v ft = CutDbConfig
     , _cutDbConfigTelemetryLevel = Warn
     , _cutDbConfigUseOrigin = True
     , _cutDbFetchTimeout = ft
+    , _cutDbConfigRateLimitCachePolicy = Limiter.TokenLimitCachePolicy exptime
```
larskuhtz: could you put a doc string?

src/Chainweb/CutDB.hs
```diff
@@ -161,9 +179,13 @@ defaultCutDbConfig v ft = CutDbConfig
...
+    , _cutDbConfigRateLimitPolicy = rlPolicy
     }
   where
     g = _chainGraph v
+    exptime = Clock.TimeSpec { Clock.sec = 10 * 60, Clock.nsec = 0 }
+    rlPolicy = LimitConfig 60 60 30 20
```
larskuhtz: Could you use the field names for readability

src/Chainweb/Chainweb.hs
```diff
@@ -270,7 +275,9 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configMempoolP2p = defaultEnableConfig defaultMempoolP2pConfig
     , _configPruneChainDatabase = True
     , _configBlockGasLimit = 100000
-    , _configCutFetchTimeout = 3000000
+    , _configCutFetchTimeout = 10000000
```
slpopejoy: Attention

fosskers: Yes, is this still needed now that the catchup step size is reduced?

src/Chainweb/Chainweb.hs
```diff
@@ -271,6 +273,7 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configPruneChainDatabase = True
     , _configBlockGasLimit = 100000
     , _configCutFetchTimeout = 10000000
+    , _configCutLimitConfig = LimitConfig 60 60 30 240
```
larskuhtz: What does this mean? Documentation?

fosskers:
```haskell
data LimitConfig = LimitConfig
    { maxBucketTokens :: {-# UNPACK #-} !Int
    , initialBucketTokens :: {-# UNPACK #-} !Int
    , bucketRefillTokensPerSecond :: {-# UNPACK #-} !Int
    , badSenderPenaltySecs :: {-# UNPACK #-} !Int
    } deriving (Eq, Ord, Show)
```

larskuhtz: We should use the the fields when constructing a `LimitConfig` value.

Also are these values exposed in the configuration file? If that's the case:

1. isn't this to complex for a normal user to handle?

2. should we hide this from the user? (not sure if hidden config options are good, because we would have to document them somewhere carefully, to not forget about them)

3. Does it have a proper syntax?

fosskers: > Also are these values exposed in the configuration file?

Yes, at least in this PR.

> isn't this to complex for a normal user to handle?

Likely, yes. We wanted it fully configurable at the time for rapid testing.

> should we hide this from the user?

Probably. @gregorycollins ?

> Does it have a proper syntax?

I gave it a JSON instance.

fosskers:
The branch should be kept around, since the technique itself useful. It could be simplified, however.

PR #645, opened 2019-11-04
Add chainweb miner as a nixos service

Fresheyeball:
I wanted to run the miner on my NixOS machine. Naturally that wasn't going to fly with `chainweb-miner --flags &`. It needs to be a systemd service. 

I added `service.nix` to the root of the project to serve as the provider for chain web components as systemd services. Currently I supported a naive implementation of `chainweb-miner` only. You can use this service like so.

```nix
let 
  chainweb-miner = builtins.fetchTarball { url = "https://github.com/Fresheyeball/chainweb-node/archive/master.tar.gz"; } + "/service.nix";
in {
  imports = [ chainweb-miner ];
  services.chainweb-miner = {
    enable = true;
    account = "<my account>";
    nodeHostname = "<my node>";
    key = "<my key>";
  };
}
```

---

Additionally you where using the hercules-ci fork of gitignore at root. This introduces unclean nix hash calculation in 2 ways.

1. The hash depends on the directory name in which the project is cloned
2. The hash depends on local user development environment including their home directory

This means your `nix-build` is not deterministic between computers. I moved you to my fork of gitignore in which both of these problems are resolved. Additionally you where no ignoring markdown or nix files, I addressed this as well.

### DIFF
--- a/project.nix
+++ b/project.nix
@@ -27,7 +27,7 @@ in {
     overrides = import ./overrides.nix pactSrc pkgs hackGet;
 
     packages = {
-      chainweb = gitignoreSource [ ".gitlab-ci.yml" "*.md" "**/*.md" ] ./.;
+      chainweb = gitignoreSource [ ".gitlab-ci.yml" "*.md" "**/*.md" "*.nix" "**/.nix" ] ./.;
     };
 
     shellToolOverrides = ghc: super: {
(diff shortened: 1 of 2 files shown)


Fresheyeball:
@mightybyte @emilypi This code is not merge-able until `nix-build` works. Right now the nix expression for `chainweb-miner` is broken, so importing it is not going to fly.

fosskers:
This could be reintroduced into the new repo: https://github.com/kadena-io/chainweb-miner

PR #651, opened 2019-11-04
fix link to linux binaries

LindaOrtega:


LindaOrtega:
https://github.com/kadena-io/chainweb-node/pull/650 fixes the issue

PR #653, opened 2019-11-04
Bounded Peer Check Concurrency

fosskers:


src/P2P/Node.hs
```diff
@@ -396,8 +397,9 @@ syncFromPeer node info = runClientM sync env >>= \case
             logg node Warn $ "failed to sync peers from " <> showInfo info <> ": " <> sshow e
             return False
     Right p -> do
+        caps <- getNumCapabilities
         goods <- fmap catMaybes
-            $ mapConcurrently (guardPeerDbOfNode node)
+            $ traverseConcurrently (ParN $ int caps * 3) (guardPeerDbOfNode node)
```
fosskers: 3x the number of caps should be more than safe.

PR #658, opened 2019-11-05
[PactService] chain-data on /local

emmanueldenloye:


gregorycollins:
Document what it does and why you did it -- it's short but I don't know what's going on here.

PR #662, opened 2019-11-07
Improve CPU hashing speed by ~60%

jwiegley:
Pinging @mightybyte @emilypi

This patch does three separate things to achieve the speedup:

1. Updates BLAKE C code to the latest version.

2. Compiles cryponite with SSE enabled and `-msse4.1`.

3. Makes the `c_blake2s_update` function an `unsafe` FFI import. For general purpose applications it might not be a good idea to block the Haskell RTS when updating the hash, but if hashing is the *only* thing that you're doing, there's no reason to allow anything else to happen until the call is done. And since this function is being called vastly more than any other, the 10x speed improvement from the unsafe call is worth any contention issues.

### DIFF
--- a/blake2s.patch
+++ b/blake2s.patch
@@ -1,7 +1,18 @@
+--- i/Crypto/Hash/Blake2s.hs
++++ w/Crypto/Hash/Blake2s.hs
+@@ -71,7 +71,7 @@ instance HashAlgorithm Blake2s_256 where
+ foreign import ccall unsafe "cryptonite_blake2s_init"
+     c_blake2s_init :: Ptr (Context a) -> Word32 -> IO ()
+ 
+-foreign import ccall "cryptonite_blake2s_update"
++foreign import ccall unsafe "cryptonite_blake2s_update"
+     c_blake2s_update :: Ptr (Context a) -> Ptr Word8 -> Word32 -> IO ()
+ 
+ foreign import ccall unsafe "cryptonite_blake2s_finalize"
 --- i/cbits/blake2/sse/blake2s-load-sse2.h
 +++ w/cbits/blake2/sse/blake2s-load-sse2.h
 @@ -17,44 +17,44 @@
-
+ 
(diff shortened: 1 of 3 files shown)


jwiegley:
Hashing numbers from a machine that ordinarily averages 15.7-16:

```
2019-11-06 22:13:20.300171: [info] Chain 0: Mined block at Height 24514. (25.31 MH/s - 77s since last)
2019-11-06 22:14:52.080238: [info] Chain 4: Mined block at Height 24518. (25.29 MH/s - 92s since last)
2019-11-06 22:21:28.096838: [info] Chain 5: Mined block at Height 24534. (26.02 MH/s - 397s since last)
2019-11-06 22:21:34.709427: [info] Chain 1: Mined block at Height 24534. (25.59 MH/s - 7s since last)
2019-11-06 22:22:15.140355: [info] Chain 1: Mined block at Height 24536. (25.70 MH/s - 41s since last)
2019-11-06 22:22:36.307337: [info] Chain 1: Mined block at Height 24537. (25.74 MH/s - 22s since last)
```

jwiegley:
I think I also know why hashing performance wasn’t scaling as wells might be expected:

If hashing is 100% of what you’re doing, then doubling the core count should double the hashing power. But if hashing is only 80% of what you’re doing, and you increase by 10 cores, effectively two whole cores worth won’t be doing any hashing at all.

I think enough time was being spent in the FFI bridging, that only part of each core was being used for actual hashing, and this inefficiency was also being multiplied by the additional cores.

jwiegley:
@fosskers Since I'm not a Stack user, could someone help with that?

emilypi:
@jwiegley sure thing. I'll fix up the PR to work for cabal and stack

jwiegley:
As @mightybyte mentioned to me today, these changes affect core crytography routines that are bit too sensitive to change for `chainweb-node` just yet. As a change to `-miner` it's fine, but it will need some vetting to be ready for inclusion.

I'll also try to push these up to cryptonite.

fosskers:
:wave: Please reopen this PR here: https://github.com/kadena-io/chainweb-miner

PR #668, opened 2019-11-08
Reduce catchup step size for now

gregorycollins:


fosskers:
@larskuhtz and I concluded that this shouldn't be necessary any more.

PR #670, opened 2019-11-08
Allow cross-compiling: nix-build --argstr system x86_64-linux

jwiegley:
This is helpful if you're on a Mac, and you have access to a Linux builder.


mightybyte:
This got into master already as a part of another ticket.  Thanks for the suggestion!

PR #671, opened 2019-11-08
Remove slow epoch check

fosskers:


larskuhtz:
### IMPORTANT

*This is technically a fork of the chain.* It must be merged only after confirming that there has been no emergency DA in the history of the current chain.

This can be verified after #674 got merged: Install this PR on a node, start a node with `initialCutHeightLimit: 0` and let it catchup up with the network. That will be reasonably fast because there are no tx and all data is in the db. (Disabling db pruning at during that catchup, ensures that the node can be restarted during the process without losing the data. Using `private` makes sure the node will go straight to the longest chain.)

There must be no emergency DA until this PR is rolled out to 51% of the network.

If we can't guarantee that there is no emergency DA in the history until this has been rolled out to 51% of the network, we would have to guard this chain with a minimal block height.

larskuhtz:
I had a node trying to catchup with this PR. It got stuck at around cut height 64,000 with

```
Nov 18 23:33:59 ip-172-31-2-98 chainweb-node[2449]: 2019-11-18T23:33:59.365Z [Error] [chainwebVersion=mainnet01|cluster=lars|peerId=JlCg2p|port=443|host=34.201.31.147|component=cut] CutDB failed: Validation failure.
Nov 18 23:33:59 ip-172-31-2-98 chainweb-node[2449]:  Parent: Nothing.
Nov 18 23:33:59 ip-172-31-2-98 chainweb-node[2449]:  Header: BlockHeader {_blockNonce = Nonce 13908635616494432660, _blockCreationTime = BlockCreationTime {_bct = Time (TimeSpan (Micros 1572593790043972))}, _blockParent = "Aqy9NKHUitTXFh-wD8EV6AFwmNPzqhu
Nov 18 23:33:59 ip-172-31-2-98 chainweb-node[2449]: The given target for the block is incorrect for its history
Nov 18 23:33:59 ip-172-31-2-98 chainweb-node[2449]: . Restarting ...
```

After switching back to master, the node continued. So, it looks like there was at least on emergency DA.

src/Chainweb/BlockHeader.hs
```diff
@@ -373,10 +355,10 @@ powTarget
         -- ^ block creation time of new block
     -> HashTarget
         -- ^ POW target of new block
-powTarget p bct@(BlockCreationTime bt) = case effectiveWindow p of
+powTarget p (BlockCreationTime bt) = case effectiveWindow p of
     Nothing -> maxTarget
     Just w
-        | isLastInEpoch p || slowEpoch p bct ->
+        | isLastInEpoch p ->
```
larskuhtz: It seems that there has been at least one `slowEpoch` event in the history of `mainnet01`. So we'll have to keep this rule for all block height smaller than some height that is larger than the height of the last `slowEpoch` event.

I propose something like

```haskell
| slowEpochGuard p | isLastInEpoch p || slowEpoch p bct ->
```

With `slowEpochGuard p = _blockHeight p < 60,000`. (The constant should be adapted to a few days in the future from the release of the version that has this change.)

fosskers: There may have been more than one. Once we add the new logic, let's do the catch up again and see if it gets caught anywhere else.

src/Chainweb/BlockHeader.hs
```diff
@@ -33,8 +33,17 @@
 --
 module Chainweb.BlockHeader
 (
+-- * Newtype wrappers for function parameters
+  ParentHeader(..)
+, ParentCreationTime(..)
```
fosskers: There are likely more places such a type would be useful. At one point I had a `PrevBlock` type in `Coordination.hs`. Your type here should take precedence.

fosskers:
The change looks solid. "Soft approval" until it has been fully tested on a fresh node.

src/Chainweb/BlockHeader.hs
larskuhtz: Yeah, I didn't want to clutter this PR. But the intention is to use this in other places, too.

larskuhtz:
I ran the following tests:

```
id                      height   versions and revision
----------------------  ------- ------------------------------------------
i-0ef70b1e877a99814:    268000  chainweb-1.0.4 revision e944ae27-HEAD
i-09052d636efd91bb1:    63000   chainweb-1.0.4 revision ccebb63c-HEAD
```

The first node (`i-0ef70b1e877a99814`) is running the current HEAD version of this PR. While catching up with mainnet it went past the existing emergency DAs in the history. This confirms that the PR is compatible with `mainnet01`.

The second node (`i-09052d636efd91bb1`) is running a modified versions of this PR with `slowEpochGuard (ParentHeader p) = _blockHeight p < 6000`, which deactivates slow epochs before the first known emergency in the history of `mainnet01`. It got stuck at cut height 63,000 with the following failure:

```
Nov 20 00:06:21 ip-172-31-18-139 chainweb-node[3074]: 2019-11-20T00:06:21.221Z [Error] [chainwebVersion=mainnet01|cluster=lars|peerId=Ltbtr9|port=443|host=34.201.104.122|component=cut] CutDB failed: Validation failure.
Nov 20 00:06:21 ip-172-31-18-139 chainweb-node[3074]:  Parent: Nothing.
Nov 20 00:06:21 ip-172-31-18-139 chainweb-node[3074]:  Header: BlockHeader {_blockNonce = Nonce 2174796561123830821, _blockCreationTime = BlockCreationTime {_bct = Time (TimeSpan (Micros 1572593879503577))}, _blockParent = "FKyn6uN0VXqk_6d-qcwDxsbOSZNC5C
Nov 20 00:06:21 ip-172-31-18-139 chainweb-node[3074]: The given target for the block is incorrect for its history
Nov 20 00:06:21 ip-172-31-18-139 chainweb-node[3074]: . Restarting ...
```

This confirms the effectiveness of this PR in preventing future emergency DAs.

PR #674, opened 2019-11-08
Initial cut height limit and private flag

larskuhtz:
* `private: true`: prevents the addition of any new peers to the PeerDb. The only peers that a private node connects to are the configured known peers. With this PR other nodes can still push cuts to a private node. PR #681 prevents unknown peers from pushing cuts.

* `initialCutHeightLimit`: When switching between forks, long rewinds are very expensive. Another issue is that a node that switches from fork `A` to `B` will go to `B` at a height larger than the current height an `A`. If the fork point is deep down it will attempt to do this in a single step, which might cause timeouts during rewind or pact replay (which doesn't cache intermediate states).

    This PR allows the user to provide an estimate of the fork point or any lower bound (0 is always a safe, but inefficient choice), to cause he node to start cut processing at a block height blow the fork point.

    Another benefit of this option is, that it allows a node to re-validate the history of the block chain from some point onwards. This verifies the consistency of the database and the chain.

### DIFF
--- a/src/Chainweb/Chainweb.hs
+++ b/src/Chainweb/Chainweb.hs
@@ -270,6 +270,7 @@ data ChainwebConfiguration = ChainwebConfiguration
     , _configPruneChainDatabase :: !Bool
     , _configBlockGasLimit :: !Mempool.GasLimit
     , _configCutFetchTimeout :: !Int
+    , _configInitialCutHeightLimit :: !(Maybe BlockHeight)
     } deriving (Show, Eq, Generic)
 
 makeLenses ''ChainwebConfiguration
@@ -302,6 +303,7 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configPruneChainDatabase = True
     , _configBlockGasLimit = 100000
     , _configCutFetchTimeout = 3000000
+    , _configInitialCutHeightLimit = Nothing
     }
 
 instance ToJSON ChainwebConfiguration where
@@ -320,6 +322,7 @@ instance ToJSON ChainwebConfiguration where
         , "pruneChainDatabase" .= _configPruneChainDatabase o
         , "blockGasLimit" .= _configBlockGasLimit o
         , "cutFetchTimeout" .= _configCutFetchTimeout o
+        , "initialCutHeightLimit" .= _configInitialCutHeightLimit o
         ]
 
 instance FromJSON (ChainwebConfiguration -> ChainwebConfiguration) where
@@ -338,6 +341,7 @@ instance FromJSON (ChainwebConfiguration -> ChainwebConfiguration) where
         <*< configPruneChainDatabase ..: "pruneChainDatabase" % o
         <*< configBlockGasLimit ..: "blockGasLimit" % o
         <*< configCutFetchTimeout ..: "cutFetchTimeout" % o
+        <*< configInitialCutHeightLimit ..: "initialCutHeightLimit" % o
 
 pChainwebConfiguration :: MParser ChainwebConfiguration
 pChainwebConfiguration = id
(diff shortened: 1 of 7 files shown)


fosskers:
This should go in after #677 .

src/Chainweb/Chainweb.hs
```diff
@@ -376,6 +380,9 @@ pChainwebConfiguration = id
     <*< configCutFetchTimeout .:: option auto
         % long "cut-fetch-timeout"
         <> help "After this many microseconds, give up trying to sync a particular Cut"
+    <*< configInitialCutHeightLimit .:: fmap (Just . BlockHeight) % option auto
```
emilypi: should this even be configurable? This sounds like best practice for catchup. We should poll our bootstraps for their most recent highest cut to determine this number on startup and defer to 0 otherwise.

src/Chainweb/Cut.hs
```diff
@@ -246,6 +247,18 @@ limitCut wdb h c
         -- this is safe because it's guaranteed that the requested rank is
         -- smaller then the block height of the argument
 
+limitCutHeaders
+    :: HasCallStack
+    => WebBlockHeaderDb
+    -> BlockHeight
+        -- upper bound for the cut height. This is not a tight bound.
```
emilypi:
```suggestion
        -- ^ upper bound for the cut height. This is not a tight bound.
```

src/Chainweb/CutDB.hs
```diff
@@ -475,12 +481,17 @@ processCuts conf logFun headerStore payloadStore cutHashesStore queue cutVar = q
             <> ", got: " <> sshow (_cutHashesHeight x)
         return r
 
+    -- This could be problematic if there is a very lighweight fork that is far
```
emilypi:
```suggestion
    -- This could be problematic if there is a very lightweight fork that is far
```

src/Chainweb/Chainweb.hs
larskuhtz: For this to be effective, we need an estimate of the height of the fork point. The better the estimate, the better the performance. 0 is always a valid estimate, but it's inefficient, because pact replay takes a long time -- in particular once we have transactions in the system.

To me it seems that those extreme scenarios with very deep forks are best resolved by manually switching to an "repair" mode. This options plus the `--private` flag would be such a "repair" mode.

src/P2P/Node/PeerDB.hs
```diff
@@ -308,23 +309,23 @@ insertPeerEntryList l m = foldl' (flip addPeerEntry) m l
 -- -------------------------------------------------------------------------- --
 -- Peer Database
 
-data PeerDb = PeerDb (MVar ()) (TVar PeerSet)
+data PeerDb = PeerDb !Bool !(MVar ()) !(TVar PeerSet)
```
larskuhtz: This needs documentation. Should probably be a record.

src/P2P/Node/Configuration.hs
```diff
@@ -81,6 +81,13 @@ data P2pConfiguration = P2pConfiguration
         -- ^ the path where the peer database is persisted
 
     , _p2pConfigIgnoreBootstrapNodes :: !Bool
+        -- ^ ignore builtin bootstrap nodes.
+
+    , _p2pConfigPrivate :: !Bool
+        -- ^ make this node private, so that it only communicates with the
+        -- initially configured know peers. Use this option with care, because
```
fosskers:
```suggestion
        -- initially configured known peers. Use this option with care, because
```

src/P2P/Node/Configuration.hs
```diff
@@ -161,6 +172,9 @@ pP2pConfiguration networkId = id
     <*< p2pConfigIgnoreBootstrapNodes .:: enableDisableFlag
         % prefixLong net "ignore-bootstrap-nodes"
         <> help ("when enabled the hard-coded bootstrap nodes for network are ignored")
+    <*< p2pConfigPrivate .:: enableDisableFlag
+        % prefixLong net "private"
+        <> help ("when enabled this node private and communicates only with the initially configured known peers")
```
fosskers:
```suggestion
        <> help ("when enabled this node becomes private and communicates only with the initially configured known peers")
```

PR #675, opened 2019-11-08
Only allow small rewinds on new block

slpopejoy:
This limits rewinds on new block to a difference of 8 blockheight from checkpointer head. It also notes that the SPV use of checkpointer rewind needs to go away soon.

### DIFF
--- a/src/Chainweb/Pact/RestAPI/Server.hs
+++ b/src/Chainweb/Pact/RestAPI/Server.hs
@@ -338,6 +338,11 @@ spvHandler l cutR cid chainR (SpvRequest rk (Pact.ChainId ptid)) = do
     cut <- liftIO $! CutDB._cut cdb
     bh <- liftIO $! lookupCutM cid cut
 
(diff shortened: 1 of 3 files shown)


src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -338,6 +338,11 @@ spvHandler l cutR cid chainR (SpvRequest rk (Pact.ChainId ptid)) = do
     cut <- liftIO $! CutDB._cut cdb
     bh <- liftIO $! lookupCutM cid cut
 
+    -- TODO get rid of (Right bh) below, rewinding checkpointer for
+    -- this could cause much problems. Conversely, we could set the
+    -- max rewind limit for this case in PactService. But in any case
+    -- it would seem this is the wrong way to check for transaction
+    -- existence at a particular depth.
```
emilypi: 👍

larskuhtz:
The CI build has two failing pact test on the linux build. Is it safe to ignore those?

PR #676, opened 2019-11-08
Allow Insecure TLS Connections

emilypi:
From what we've seen, TLS connections and the current way we're doing fingerprint validation is broken. This is a quick fix to turn _on_ accepting invalid certificates so nodes can talk to each other.

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -66,6 +66,12 @@ flag debug
     default: False
     manual:True
 
+flag ghc-flags
+    description:
+        Generate .ghc.flags files during compilation
+    default: False
+    manual: True
+
 common debugging-flags
     if flag(debug)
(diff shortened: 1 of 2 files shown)


chainweb.cabal
```diff
@@ -394,6 +400,9 @@ library
     if flag(use_systemd) && os(linux)
         build-depends: systemd >= 1.2
         cpp-options: -DWITH_SYSTEMD=1
+    if flag(ghc-flags)
+        ghc-options: -fplugin GhcFlags.Plugin
+        build-depends: ghcflags
```
emilypi: `ghc-flags` added so i can get some better tooling metrics.

slpopejoy:
Superseded by #678

PR #677, opened 2019-11-08
Authenticate Peers and lock down cut db

larskuhtz:


src/Chainweb/RestAPI.hs
```diff
@@ -250,6 +252,30 @@ chainwebTime app req resp = app req resp'
...
+chainwebValidatePeer pdb app req resp = do
+    sn <- peerDbSnapshot pdb
+    let pinfo = case hostaddr of
+            Nothing -> Nothing
+            Just x -> _peerEntryInfo <$> getOne (getEQ x sn)
+    case pinfo of
+        Nothing -> app req resp
+        Just x -> do
+            let req' = req
+                    { requestHeaders
```
fosskers: Why not just avoid binding to `pinfo` and just pattern matching once?

PR #681, opened 2019-11-09
require cut origin to be known

larskuhtz:


larskuhtz:
This is running for a while now on most bootstraps. Its an important piece for making `private: true` effective in preventing bad nodes from publishing cuts.

Preventing attackers from injecting bad cuts is an important step to protect the chain from attacks. The implementation in this PR doesn't provide full protection against a skilled attacker.  It is, however, an important step into that direction.

PR #684, opened 2019-11-10
Move the Miner

fosskers:
This PR removes the Mining Client code, which has been moved to https://github.com/kadena-io/chainweb-miner.

#638 and #662 should take heed.


emmanueldenloye:
I've moved my PR (#638) to the chainweb-miner repository.

PR #686, opened 2019-11-10
Private Mining

fosskers:
The mining configuration section has been consolidated and now looks like this:
```yaml
  mining:
    coordination:
      enabled: false
      mode: private
      miners: []
    nodeMining:
      miner:
        account: ''
        predicate: keys-all
        public-keys: []
      enabled: false
```
`mode` can be `public` or `private`. If `private`, then associated `miners` list must contain at least 1 Account Name, the same name you run your `chainweb-miner` with.

**PSA:** Please regenerate your Node configs by doing the following:
```
chainweb-node --config-file=config.yaml --print-config > new-config.yaml
mv new-config.yaml config.yaml
```

### DIFF
--- a/src/Chainweb/Miner/RestAPI/Server.hs
+++ b/src/Chainweb/Miner/RestAPI/Server.hs
@@ -26,6 +26,7 @@ import Control.Monad.STM (atomically)
 
 import Data.Binary.Builder (fromByteString)
 import Data.Generics.Wrapped (_Unwrapped)
+import qualified Data.HashSet as HS
 import Data.IORef (IORef, atomicModifyIORef', newIORef, readIORef, writeIORef)
 import qualified Data.Map.Strict as M
 import Data.Proxy (Proxy(..))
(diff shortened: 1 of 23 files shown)


src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -84,6 +86,10 @@ workHandler mr v mcid m = do
             when (M.size ms > _coordLimit mr) $ do
                 liftIO $ atomicModifyIORef' (_coord503s mr) (\c -> (c + 1, ()))
                 throwError err503 { errBody = "Too many work requests" }
+            let !conf = _coordConf mr
+            when (_coordinationMode conf == Private
+                  && not (HS.member (view minerId m) (_coordinationMiners conf))) $
+                throwError err403 { errBody = "Unauthorized Miner" }
```
fosskers: 403 if an unauthorized minor requests work.

PR #705, opened 2019-11-14
[TXG] Beautify REST API calls

emmanueldenloye:
This is a small PR created so that the transaction generator code
reads better than it did before.

### DIFF
--- a/tools/txg/TXG.hs
+++ b/tools/txg/TXG.hs
@@ -78,7 +78,7 @@ import Pact.Types.Exp (Literal(..))
 import Chainweb.ChainId
 import Chainweb.Graph
 import Chainweb.HostAddress
-import Chainweb.Pact.RestAPI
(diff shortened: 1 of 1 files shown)


tools/txg/TXG.hs
```diff
@@ -550,28 +550,28 @@ defaultContractLoaders :: ChainwebVersion -> NonEmpty ContractLoader
...
+
+-- poll :: ChainwebVersion -> ChainId -> Poll -> ClientM PollResponses
+-- poll version chainid = go
+--   where
+--     _ :<|> go :<|> _ :<|> _ = api version chainid
+
+-- listen :: ChainwebVersion -> ChainId -> ListenerRequest -> ClientM ListenResponse
+-- listen version chainid = go
+--   where
+--     _ :<|> _ :<|> go :<|> _ = api version chainid
```
emilypi: You can just delete this :)

PR #710, opened 2019-11-15
Don't Rewind on SPV Request Lookup

emilypi:


src/Chainweb/Pact/Types.hs
```diff
@@ -105,4 +105,16 @@ newtype GasId = GasId PactId deriving (Eq, Show)
...
+    | NoRewind !a
+    | Genesis {-# UNPACK #-} !cid
+    deriving (Functor, Applicative, Foldable, Traversable, Monad)
+
+instance HasChainId a => HasChainId (Rewind a) where
+    _chainId = \case
+      DoRewind !a -> _chainId a
+      NoRewind !a -> _chainId a
+      Genesis !cid -> cid
+
```
emmanueldenloye: Why do you need Functor, Applicative, Foldable, Traversable and Monad instances?

emilypi: Cautionary - i don't know where I'll be going so I want to know i have everything I need and I'll pare it down as I work the PR. It's a WIP draft!

src/Chainweb/Chainweb.hs
```diff
@@ -495,7 +496,7 @@ validatingMempoolConfig cid v gl mv = Mempool.InMemConfig
     preInsertBatch txs = do
         let hashes = V.map (toPactHash . sfst) txs
         pex <- readMVar mv
-        rs <- _pactLookup pex (Left cid) hashes >>= either throwM pure
+        rs <- _pactLookup pex (Genesis cid) hashes >>= either throwM pure
```
emilypi: I'm not sure if "Genesis" is the correct name for this case when we're looking up by chain id. Paint my bikeshed please

src/Chainweb/Pact/PactService.hs
```diff
@@ -1077,26 +1077,29 @@ transactionsFromPayload plData = do
...
+      case restorePoint of
+        NoRewind _ -> liftIO $!
+          V.mapM (_cpLookupProcessedTx cp) txs
+        DoRewind a -> doRewind cp a
+        _ -> do
+          mrp <- fmap (\(!bh) -> T2 (_blockHeight bh) (_blockHash bh))
+            <$!> findLatestValidBlock
+          case mrp of
+            Nothing -> return mempty -- nothing denotes genesis
+            Just a -> doRewind cp a
```
emilypi: If `NoRewind`, lookup all txs without rewinding. This is only used when doing SPV lookups. When `DoRewind` or `Genesis`, do the rewind.

src/Chainweb/Pact/Types.hs
```diff
@@ -105,4 +110,16 @@ newtype GasId = GasId PactId deriving (Eq, Show)
 
 type TransactionM p a = ReaderT (CommandEnv p) IO a
 
+data Rewind a
+    = DoRewind !a
```
gregorycollins: There should only be two cases here:

```
DoRewind (T2 BlockHeight BlockHash) | NoRewind ChainId
```

src/Chainweb/Pact/PactService.hs
gregorycollins: PreInsert shouldn't rewind, that's a regression -- everything that was `Left` before should be `NoRewind`.

src/Chainweb/Pact/Types.hs
gregorycollins: Does `Rewind` need to be polymorphic?

emilypi: Made it monomorphic.

PR #719, opened 2019-11-20
Configurable block gas limit (finally)

slpopejoy:
Actually implements existing config. 

Need help from @gregorycollins on failing test


test/Chainweb/Test/Mempool.hs
```diff
@@ -266,7 +266,7 @@ propTrivial txs _ mempool = runExceptT $ do
     lookup = mempoolLookup mempool . V.fromList . map hash
 
     getBlock = mempoolGetBlock mempool noopMempoolPreBlockCheck 0 nullBlockHash
-                               (mempoolBlockGasLimit mempool)
+
```
slpopejoy: This is the failing test. I tried setting `mockBlockGasLimit` to 1000000 but it didn't fix it

PR #722, opened 2019-11-20
Seed Initial Module Cache in Pact Service at Genesis

emilypi:
This seeds every module cache with the modules/interfaces loaded at genesis, keeping it in Pact Service state in perpetuity.

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -53,6 +53,8 @@ import Data.Bool (bool)
 import qualified Data.ByteString.Short as SB
 import Data.Decimal
 import Data.Default (def)
+import Data.DList (DList(..))
+import qualified Data.DList as DL
 import Data.Either
 import Data.Foldable (foldl', toList)
 import qualified Data.HashMap.Strict as HM
@@ -198,7 +200,7 @@ initPactService' ver cid chainwebLogger spv bhDb pdb dbDir nodeid
           gasModel = tableGasModelNoSize defaultGasConfig -- TODO get sizing working
       let !pse = PactServiceEnv Nothing checkpointEnv spv def pdb
                                 bhDb gasModel rs
-      evalStateT (runReaderT act pse) (PactServiceState Nothing)
+      evalStateT (runReaderT act pse) (PactServiceState Nothing mempty)
   where
     loggers = pactLoggers chainwebLogger
     logger = P.newLogger loggers $ P.LogName ("PactService" <> show cid)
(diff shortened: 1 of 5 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -988,24 +990,26 @@ execTransactions
     -> PactDbEnv'
     -> PactServiceM cas Transactions
 execTransactions nonGenesisParentHash miner ctxs enfCBFail (PactDbEnv' pactdbenv) = do
-    T2 coinOut mc <- runCoinbase nonGenesisParentHash pactdbenv miner enfCBFail
+    mc <- use psInitCache
+    coinOut <- runCoinbase nonGenesisParentHash pactdbenv miner enfCBFail mc
```
emilypi: coinbase will now always have `coin` and all of its dependencies in scope

src/Chainweb/Pact/PactService.hs
```diff
@@ -988,24 +990,26 @@ execTransactions
...
     txOuts <- applyPactCmds isGenesis pactdbenv ctxs miner mc
     return $! Transactions (paired txOuts) coinOut
   where
     !isGenesis = isNothing nonGenesisParentHash
     cmdBSToTx = toTransactionBytes
       . fmap (T.decodeUtf8 . SB.fromShort . payloadBytes)
-    paired = V.zipWith (curry $ first cmdBSToTx) ctxs
+    paired outs = V.zipWith (curry $ first cmdBSToTx) ctxs outs
```
emilypi: legibility

src/Chainweb/Pact/PactService.hs
```diff
@@ -1039,24 +1043,28 @@ applyPactCmd
     -> ChainwebTransaction
     -> Miner
     -> ModuleCache
-    -> ([HashCommandResult] -> [HashCommandResult])  -- ^ difference list
-    -> PactServiceM cas (T2 ([HashCommandResult] -> [HashCommandResult]) ModuleCache)
+    -> DList HashCommandResult
+    -> PactServiceM cas (T2 (DList HashCommandResult) ModuleCache)
```
emilypi: I missed this before, but we were rolling our own DList when the package was already in scope.

src/Chainweb/Pact/PactService.hs
```diff
@@ -1039,24 +1043,28 @@ applyPactCmd
...
         pactHash  = view P.cmdHash cmdIn
 
-    T2 !result mcache' <- liftIO $ if isGenesis
+    T2 result mcache' <- liftIO $ if isGenesis
         then applyGenesisCmd logger dbEnv pd spv (payloadObj <$> cmdIn)
         else applyCmd logger dbEnv miner (_psGasModel psEnv) pd spv cmdIn mcache
 
+    -- on genesis, seed initial service state with loaded modules
+    when isGenesis $
+      psInitCache .= mcache'
```
emilypi: When we're at genesis, seed the cache.

fosskers: Will this still work for nodes that are way past genesis height? Does this trigger upon startup?

emilypi: This is something I'd want to talk with stuart about. I don't know. Tbh i think this would only work in general for services starting at 0. I'm not honestly sure how to get older pact services synced up otherwise.

PR #723, opened 2019-11-20
Fix insufficient gas UX on `/send`

### LINKED ISSUE #712: Show validation failure message at `/poll` or `/listen` endpoint
When transactions are rejected, validation failure messages are returned as text at `/send` request. There are two types of behavior that these validation-failed txs show: 

1) The `/send` request returns text at once. (metadata conflicts behave this way)
2) The `/send` request returns request key at first, but if the request key is sent to `/poll` or `/listen`, the request times out or doesn't return anything. (Insufficient gas failure behave this way)

In the second case, the only way to see the validation failure message is to make request with the the exact send command to `/send` endpoint one more time. If the validation has failed, then it will return the validation failure message as text. If transaction didn't fail yet, the response will be the same request key it returned at first.

It would make sense to respond to `/poll` or `/listen` for the validation-failed txs to respond with validation failure messages instead of `/send` endpoint. In the case of batch transactions, the validation failure message will only return the first command in the array of commands.

gregorycollins:
Fixes #712. We need to do a gas check on pre-insert to do this.


src/Chainweb/Mempool/Mempool.hs
```diff
@@ -206,12 +207,12 @@ instance Show InsertError
     show InsertErrorInvalidTime = "Transaction time is invalid or TTL is expired"
     show InsertErrorOversized = "Transaction gas limit exceeds block gas limit"
     show InsertErrorBadlisted =
-        "Transaction is badlisted because it previously failed to validate \
-        \(e.g. insufficient gas)"
+        "Transaction is badlisted because it previously failed to validate."
     show InsertErrorMetadataMismatch =
         "Transaction metadata (chain id, chainweb version) conflicts with this \
         \endpoint"
     show InsertErrorTransactionsDisabled = "Transactions are disabled until December 5"
```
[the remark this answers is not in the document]
gregorycollins: The year?

src/Chainweb/Mempool/Mempool.hs
```diff
@@ -206,12 +207,12 @@ instance Show InsertError
...
+    show InsertErrorNoGas = "Sender account has insufficient gas."
```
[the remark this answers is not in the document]
gregorycollins: The error is already associated to the transaction hash in the pre-insert check..

src/Chainweb/Pact/PactService.hs
```diff
@@ -581,9 +586,9 @@ validateChainwebTxs
     -> BlockHeight
     -> Vector ChainwebTransaction
     -> RunGas
-    -> IO (Vector Bool)
+    -> IO (Vector (Either Mempool.InsertError ()))
```
emilypi: Was there a reason this function is in `IO` as opposed to `PactServiceM` here? I grepped and it's only used in this file as part of the service flow.

src/Chainweb/Mempool/Mempool.hs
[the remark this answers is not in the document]
fosskers:
```suggestion
     show InsertErrorTransactionsDisabled = "Transactions are disabled until 2019 Dec 5"
```

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -159,9 +161,6 @@ tests rdb = testGroupSch "Chainweb.Test.Pact.RemotePactTest"
               , after AllSucceed "genesis allocations" $
                 testGroup "caplistTests"
                 [ caplistTest iot net ]
-              , after AllSucceed "caplist tests" $
-                testGroup "gasBuyingErrors"
-                [ invalidBuyGasTest iot net ]
```
fosskers: Okay to pull out?

src/Chainweb/Pact/PactService.hs
gregorycollins: No, we can fix it later.

test/Chainweb/Test/Pact/RemotePactTest.hs
gregorycollins: Yes, it's moved to the `/send` test now

PR #728, opened 2019-11-21
benches with diff tx counts

slpopejoy:


slpopejoy:
> How come this benchmark code is part of the `chainweb` library and not in the benchmark suite?

Mainly for dev speed and avoiding target change hassles.

PR #733, opened 2019-11-21
Complete Config

fosskers:


complete-config.yaml
```diff
@@ -0,0 +1,145 @@
...
+      - key: component
+        value: local-handler
+        level: info
+    default: error
+
+# --- Miscellaneous Settings --- #
+
+# Location of your database files. Leaving this as `null` tells Chainweb to use
+# its default locations.
+databaseDirectory: null
```
mightybyte: Should we take this out or suggest that people not use this until the DB location bug is fixed?

fosskers: The fix only applies to the default location, I think. @larskuhtz ?

mightybyte: The default location puts the DBs in the right place.  I believe the bug happens when people override it.

fosskers: I'll hide it for now.

PR #735, opened 2019-11-22
Close update streams after about 4 minutes

larskuhtz:
Server events are long running HTTP connections. Without a server side timeout the server depends on the mercy of the client to promptly close outdated connections. In addition it seems that with warp/wai server event connections that are closed by the client are not promptly garbage collected and lead to a large number of connections in CLOSE_WAIT state.

There is no way for a client to close an server event stream on the application level. The client has to close the underlying TCP connection of a stream, which means that it can't be reused.

When server event stream is closed by the server the HTTP request is terminated on the application level and the TCP connection can be reused.


larskuhtz:
Effect onto Memory and CPU on mainnet

image

larskuhtz:
Effect on number of open file descriptors

image

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -135,7 +137,9 @@ updatesHandler :: CutDb cas -> ChainBytes -> Tagged Handler Application
 updatesHandler cdb (ChainBytes cbytes) = Tagged $ \req respond -> do
     cid <- runGet decodeChainId cbytes
     cv  <- _cut cdb >>= newIORef
-    eventSourceAppIO (go cid cv) req respond
+    x <- randomRIO @Double (0.9, 1.1)
+    timer <- registerDelay (round $ 1000000 * 240 * x)
+    eventSourceAppIO (go timer cid cv) req respond
```
fosskers: Can you put comments here to explain the units?

larskuhtz: Good point. Also, I wonder if the use of the STM timer should be avoided. It's probably more efficient to just get the current time after each event. An even simpler solution would terminate the stream based on the number of events instead of time.

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -131,21 +138,54 @@ solvedHandler mr (HeaderBytes hbytes) = do
...
+
+    count = _coordUpdateStreamCount mr
+
+    withLimit respond inner = bracket
+        (atomicModifyIORef' count $ \x -> (x - 1, x - 1))
+        (const $ atomicModifyIORef' count $ \x -> (x + 1, ()))
+        (\x -> if x <= 0 then ret503 respond else inner)
+
+    ret503 respond = do
+        atomicModifyIORef' (_coord503s mr) (\c -> (c + 1, ()))
```
larskuhtz:
```suggestion
```

incrementing the count is already taken care of in the finalizer of the `withLimit` bracket.

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -131,21 +138,54 @@ solvedHandler mr (HeaderBytes hbytes) = do
...
+        respond $ responseLBS status503 [] "Too many work requests"
```
larskuhtz:
```suggestion
        respond $ responseLBS status503 [] "No more update streams available currently. Retry later."
```

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -86,14 +88,17 @@ withMiningCoordination logger conf cutDb inner
         t <- newTVarIO mempty
         c503 <- newIORef 0
         c403 <- newIORef 0
+        l <- newIORef 0
```
larskuhtz:
```suggestion
        l <- newIORef (_coordinationUpdateStreamLimit conf)
```

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -86,14 +88,17 @@ withMiningCoordination logger conf cutDb inner
         t <- newTVarIO mempty
         c503 <- newIORef 0
         c403 <- newIORef 0
+        l <- newIORef (_coordinationUpdateStreamLimit conf)
```
fosskers: I thought you changed this back to 0?

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -131,21 +138,53 @@ solvedHandler mr (HeaderBytes hbytes) = do
...
+    => MiningCoordination l cas
+    -> ChainBytes
+    -> Tagged Handler Application
+updatesHandler mr (ChainBytes cbytes) = Tagged $ \req respond -> withLimit respond $ do
     cid <- runGet decodeChainId cbytes
-    cv  <- _cut cdb >>= newIORef
-    eventSourceAppIO (go cid cv) req respond
+    cv  <- _cut (_coordCutDb mr) >>= newIORef
+    jitter <- randomRIO @Double (0.9, 1.1)
+    timer <- registerDelay (round $ jitter * realToFrac timeout * 1000000)
```
fosskers: Can you add comments to explain these constants?

src/Chainweb/Chainweb/MinerResources.hs
larskuhtz: When this is zero all requests will be throttled. The initial value is the number of available streams. Requests borrow a unit from this counter and put it back when down.

PR #741, opened 2019-11-25
Fix typo

mightybyte:


mightybyte:
This requires a reset of the testnet DBs.

mightybyte:
Yep, I checked very carefully and also did a diff between `grants0` and `grantsN` since `grants0` has seemed to be fine.  Looks good.

PR #742, opened 2019-11-25
[mempool] Resurrect (with tweaks) PR #604

gregorycollins:
Mempool is currently not being pruned when mining is turned off. This PR spawns a reaper thread to prune old txs out of the mempool periodically if mining is disabled, and tweaks the pruning to work better.

  - move pruning logic to its own toplevel function so that we don't have to call `mempoolGetBlock` just for its side-effects
  - consider all txs for pruning, not just those that are considered for block inclusion
  - cache tx expiry times in our pending map so we don't have to decode transaction bytes to prune
  - add a no-op miner thread that prunes mempools periodically when mining is off.


gregorycollins:
N.B. this PR is based atop PR #732 , that one needs to go in first

src/Chainweb/Mempool/InMem.hs
```diff
@@ -611,3 +624,36 @@ getMempoolStats m = do
...
+------------------------------------------------------------------------------
+pruneInternal
+    :: forall t . NFData t
+    => InMemoryMempoolData t
+    -> Time Micros
+    -> IO ()
+pruneInternal mdata now = do
+    let pref = _inmemPending mdata
+    !pending <- readIORef pref
+    !pending' <- evaluate $ force $ HashMap.filter flt pending
```
larskuhtz: How expensive is this and how often it is down? Would it make sense to use a better data structure for this? It would probably be overkill if this isn't called too often.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -611,3 +624,36 @@ getMempoolStats m = do
...
+    :: forall t . NFData t
+    => MVar (InMemoryMempoolData t)
+    -> IO ()
+pruneInMem lock = do
+    now <- getCurrentTimeIntegral
+    withMVar lock $ \mdata -> pruneInternal mdata now
+
+
+------------------------------------------------------------------------------
+pruneInternal
```
larskuhtz: It seems like the complexity is linear in the number of pending tx. Is that correct? Could you add and Haddock comment, that states the complexity -- to make sure nobody accidentally uses it in a performance critical path.

src/Chainweb/Miner/Miners.hs
```diff
@@ -96,6 +102,22 @@ localTest lf v m cdb gen miners = runForever lf "Chainweb.Miner.Miners.localTest
...
 
+-- | A miner that grabs new blocks from mempool and discards them. Mempool
+-- pruning happens during new-block time, so we need to ask for a new block
+-- regularly to prune mempool.
+mempoolNoopMiner
+    :: LogFunction
+    -> HashMap ChainId (ChainResources logger)
+    -> IO ()
+mempoolNoopMiner lf chainRes =
+    runForever lf "Chainweb.Miner.Miners.mempoolNoopMiner" loop
```
larskuhtz: Only the `_chainResMempool` is required. Passing the whole structure my keep references to more resources alive than actually needed.

Generally, the `*Resources` types are supposed to only be used during initialization. The idea is that on initialization each component picks from the resources only what it actually needs.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -164,10 +168,16 @@ runMiner
...
     -> IO ()
-runMiner v mr = case window v of
-    Nothing -> testMiner
-    Just _ -> powMiner
+runMiner v mr =
+    if enabled
+        then case window v of
+                 Nothing -> testMiner
+                 Just _ -> powMiner
+        else mempoolNoopMiner lf (_minerChainResources mr)
```
larskuhtz: `mempoolNoopMiner` only uses `_chainResMempool`. 

```suggestion
        else mempoolNoopMiner lf (_chainResMempool <$> _minerChainResources mr)
```

PR #743, opened 2019-11-25
1.1 Transition, Part 2: Tweaked Guards

fosskers:
#### Currently (`1.0.x` Nodes)

As of now:

- Upon startup, no `MempoolBackend`s are offered to be served.
- Upon startup, no Mempool sync threads are ran.
- A special mempool validation rule immediately fails any transaction attempting
  to pass through validation.
- Anything sent to `/send` is rejected.

As of Dec 5 00:00:00 UTC:

- Nodes can no longer be restarted.
- Any request to `/mining/work` is rejected.
- Anything sent to `/mining/solved` is rejected.

#### `1.1.x` Nodes

As of Dec 2 (`1.1.0` release):

1. Upon startup, mempool threads are live, but any calls to `/mempool/insert`
  will fail.
2. A new permanent mempool pre-insert check fails any transaction created before
  the Dec 6 activation date.
3. Anything sent to `/send` is rejected.
4. Mining continues as usual (extra shutoff logic removed).
5. Gas limits are set such that only ~10 `transfer` calls could make it into a
   single block.
6. Submission of new modules is blocked at the Pact level.

As of Dec 6 ??:??:?? UTC, all without user intervention:

7. `/send` will start accepting transactions.
8. `/mempool/insert` will start accepting traffic from other mempools.

**Notes:**

- (5) and (6) are not part of this PR.

**Questions:**

- Are the guards in (1) and (8) necessary, given (2)?


larskuhtz:
I think the overall order of events is (ignoring version 1.0.5):

**Release 1.1:**

0. Released at `d0`
1. "hard” reject tx before `d1` , i.e. start successfully validating tx from `d1` onward,
2. “soft” reject tx before `d2` , i.e. start processing tx from `d2` onward,
3. “hard” reject user code,
4. “soft” reject user code,
5. stop working completely (shut-off) at date `d4`,

**Release: 1.2:**

6. Release at `d3`,
7. "hard" reject user code before `d5`,
8. "soft" reject user code before `d6`,

with `d0 < d1 < d2 < d3 < d4 < d5 < d6`. 

**Dates**

* `d0` is Dec 2.
* `d1` is already hard coded as [`txSilenceEndDate Mainnet01 = Just [timeMicrosQQ| 2019-12-05T00:00:00.0 |]`](https://github.com/kadena-io/chainweb-node/blob/1a2fd839745436df478b4e6427b5083aca9c334f/src/Chainweb/Version.hs#L575).
* `d2` should be `2019-12-05T00:00:01`. 
* `d3` should be several days (2 weeks?) before `2020-01-15`.
* `d4` should be `2020-01-14T23:00:00` (maybe even `2020-01-14T00:00:00`?)
* `d5` is `2020-01-15T00:00:00`
* `d6` should be `2020-01-15T00:00:01`

**Notes:**

(1) is already implemented in chainweb-node-1.0.5 in [`validateSilenceDates`](https://github.com/kadena-io/chainweb-node/blob/1a2fd839745436df478b4e6427b5083aca9c334f/src/Chainweb/Pact/PactService.hs#L973).

For (3) we have to add another validation rule that rejects “pact user code”. This rule will be guarded in (7) and made *permanent*.

(4) is what most code changes in the PR are about. (4) and (8) ensure that invalid tx fail early and don't get validation errors. This points are not sufficient to protect against modified nodes that create blocks with tx the violate the rules. 

(5) will be single central kill switch in the node that shuts the node off and prevent restarts from `d4` onward.

src/Chainweb/Version.hs
```diff
@@ -572,4 +572,4 @@ txSilenceEndDate TimedCPM{} = Nothing
 txSilenceEndDate FastTimedCPM{} = Nothing
 txSilenceEndDate Development = Nothing
 txSilenceEndDate Testnet03 = Nothing
-txSilenceEndDate Mainnet01 = Just [timeMicrosQQ| 2019-12-05T00:00:00.0 |]
+txSilenceEndDate Mainnet01 = Just [timeMicrosQQ| 2019-12-06T00:00:00.0 |]
```
larskuhtz: We can't change this. This would be a hard fork.

larskuhtz: This is a property of the `ChainwebVersion`. Any change of the `chainwebVersion` or its properties is a fork.

fosskers: I will add a second one.

src/Chainweb/Chainweb.hs
```diff
@@ -884,7 +875,7 @@ runChainweb cw = do
             FastTimedCPM{} -> enabled c
             Development -> enabled c
             Testnet03 -> enabled c
-            Mainnet01 -> disabled
+            Mainnet01 -> enabled c
```
fosskers: Mempools are now completely on, but TXs are still blocked (for the mempool) in three places:

1. `/mempool/insert`
2. The pre-insert check
3. Block Validation

Do you foresee any issues with thus @gregorycollins ?

src/Chainweb/Mempool/InMem.hs
```diff
@@ -264,12 +264,12 @@ validateOne cfg v badmap now t h =
...
-    -- | KILLSWITCH: This is to be removed in a future version of Chainweb. This
-    -- prevents any transaction from entering the mempool.
+    -- | KILLSWITCH: This can be removed once the date itself has passed. Until
+    -- then, this prevents any transaction from entering the mempool.
     --
     transactionsEnabled :: Either InsertError ()
-    transactionsEnabled = case txSilenceEndDate v of
-        Just _ -> Left InsertErrorTransactionsDisabled
+    transactionsEnabled = case txActivationDate v of
+        Just start | now < start -> Left InsertErrorTransactionsDisabled
```
fosskers: Actually consider the date, instead of rejecting everything outright.

src/Chainweb/Mempool/RestAPI/Server.hs
```diff
@@ -28,24 +27,42 @@ import Chainweb.Mempool.Mempool
...
+    -- this will prevent mempools from inserting data into each other.
+    --
+    begin :: Handler ()
     begin = do
-        txs <- mapM go txsT
-        let txV = V.fromList txs
-        liftIO $ mempoolInsert mempool CheckedInsert txV
+        now <- liftIO getCurrentTimeIntegral
+        case txActivationDate v of
+            Just start | start < now -> pure ()
```
fosskers: @gregorycollins should this throw instead? Do the mempools have certain p2p assumptions (demotion, etc.) when it comes to success/failure when calling this endpoint?

src/Chainweb/Mempool/RestAPI/Server.hs
```diff
@@ -105,31 +122,23 @@ handleErrs = flip catchAllSynchronous $ \e ->
...
     :<|> getPendingHandler mempool
-
-mempoolApp
-    :: forall v c t
-    . KnownChainwebVersionSymbol v
-    => KnownChainIdSymbol c
-    => Show t
-    => Mempool_ v c t
-    -> Application
-mempoolApp mempool = serve (Proxy @(MempoolApi v c)) (mempoolServer mempool)
```
fosskers: Not used anywhere.

src/Chainweb/Pact/PactService.hs
```diff
@@ -971,7 +971,7 @@ execValidateBlock currHeader plData = do
     isGenesisBlock = isGenesisBlockHeader currHeader
 
 validateSilenceDates :: MonadThrow m => BlockHeader -> PayloadData -> m ()
-validateSilenceDates bh plData = case txSilenceEndDate (_blockChainwebVersion bh) of
+validateSilenceDates bh plData = case txActivationDate (_blockChainwebVersion bh) of
```
fosskers: This is a change in a core validation rule. Merging this change is a hard fork.

src/Chainweb/Version.hs
```diff
@@ -564,12 +564,15 @@ window Development = Just $ WindowWidth 120
...
 
-txSilenceEndDate :: ChainwebVersion -> Maybe (Time Micros)
-txSilenceEndDate Test{} = Nothing
-txSilenceEndDate TimedConsensus{} = Nothing
-txSilenceEndDate PowConsensus{} = Nothing
-txSilenceEndDate TimedCPM{} = Nothing
-txSilenceEndDate FastTimedCPM{} = Nothing
-txSilenceEndDate Development = Nothing
-txSilenceEndDate Testnet03 = Nothing
-txSilenceEndDate Mainnet01 = Just [timeMicrosQQ| 2019-12-05T00:00:00.0 |]
```
fosskers: No longer used.

src/Chainweb/Version.hs
larskuhtz: The second one, that remarks the end of the soft reject of tx should be larger than `txSilenceEndDate`. In terms of above timeline `txSilenceEndDate == d1` and date for ending the soft reject is `d2`.

src/Chainweb/Pact/PactService.hs
larskuhtz: Don't change this rule! It's eternal.

larskuhtz:
```suggestion
validateSilenceDates bh plData = case txSilenceEndDate (_blockChainwebVersion bh) of
```

We must not change this existing validation rule.

src/Chainweb/Version.hs
larskuhtz: This must be reverted. We can't change history. These rules existed to validate the history of the block chains. They should stay forever.

src/Chainweb/Pact/PactService.hs
fosskers: We must, or else transactions from Dec 5 to Dec 6 will be allowed.

src/Chainweb/Version.hs
larskuhtz: I propose that we keep this (date `d1`) to not change existing validation rules) and define a new constant for date `d2`:

```Haskell
softTxSilenceEndDate Mainnet01 = Just [timeMicrosQQ| 2019-12-07T00:00:00.0 |]
```

Which will be used to implement step (2) from the timeline.

src/Chainweb/Chainweb.hs
fosskers: A node of mine has been running with just these mempool changes for several hours without issue.

src/Chainweb/Pact/PactService.hs
```diff
@@ -978,8 +978,8 @@ execValidateBlock currHeader plData = do
     bParent = _blockParent currHeader
     isGenesisBlock = isGenesisBlockHeader currHeader
 
-validateSilenceDates :: MonadThrow m => BlockHeader -> PayloadData -> m ()
-validateSilenceDates bh plData = case txSilenceEndDate (_blockChainwebVersion bh) of
+validateTxEnabled :: MonadThrow m => BlockHeader -> PayloadData -> m ()
+validateTxEnabled bh plData = case txEnabledDate (_blockChainwebVersion bh) of
```
fosskers: These are just renamings. The original logic is preserved.

src/Chainweb/Version.hs
```diff
@@ -565,26 +565,29 @@ window Development = Just $ WindowWidth 120
...
+--
+txEnabledDate :: ChainwebVersion -> Maybe (Time Micros)
+txEnabledDate Test{} = Nothing
+txEnabledDate TimedConsensus{} = Nothing
+txEnabledDate PowConsensus{} = Nothing
+txEnabledDate TimedCPM{} = Nothing
+txEnabledDate FastTimedCPM{} = Nothing
+txEnabledDate Development = Nothing
+txEnabledDate Testnet03 = Nothing
+txEnabledDate Mainnet01 = Just [timeMicrosQQ| 2019-12-05T00:00:00.0 |]
```
fosskers: Same date, just renamed.

src/Chainweb/Version.hs
```diff
@@ -565,26 +565,29 @@ window Development = Just $ WindowWidth 120
...
-txActivationDate Mainnet01 = Just [timeMicrosQQ| 2019-12-06T00:00:00.0 |]
+transferActivationDate :: ChainwebVersion -> Maybe (Time Micros)
+transferActivationDate Test{} = Nothing
+transferActivationDate TimedConsensus{} = Nothing
+transferActivationDate PowConsensus{} = Nothing
+transferActivationDate TimedCPM{} = Nothing
+transferActivationDate FastTimedCPM{} = Nothing
+transferActivationDate Development = Nothing
+transferActivationDate Testnet03 = Nothing
+transferActivationDate Mainnet01 = Just [timeMicrosQQ| 2019-12-06T00:00:00.0 |]
```
fosskers: It's our intention that this code will be removed almost immediately after the specified date. Note that this specific date here gives a 24 hour gap between the "shut off" of the old nodes.

node/ChainwebNode.hs
```diff
@@ -452,16 +450,7 @@ mainInfo = programInfoValidate
...
 main :: IO ()
 main = withWatchdog . runWithPkgInfoConfiguration mainInfo pkgInfo $ \conf -> do
     let v = _configChainwebVersion $ _nodeConfigChainweb conf
-    now <- getCurrentTimeIntegral
-    case txSilenceEndDate v of
-        Just end | now > end -> do
-            putStrLn "Transactions are now possible - please update your Chainweb binary."
-            exitFailure
-        _ -> withNodeLogger (_nodeConfigLog conf) v $ node conf
+    withNodeLogger (_nodeConfigLog conf) v $ node conf
```
larskuhtz: We need a new global kill switch here, that shuts off the node completely (not only preventing restarts).

I propose that we implement something like `withKillSwitch :: Maybe Date -> IO a -> IO a` and use it at the top level in `main`. For 1.1. it must be set to `2020-01-15T00:00:00Z`, according to the overall timeline.

Should that be done in this PR or in a follow up PR?

src/Chainweb/Mempool/InMem.hs
```diff
@@ -264,12 +264,12 @@ validateOne cfg v badmap now t h =
     txcfg :: TransactionConfig t
     txcfg = _inmemTxCfg cfg
 
-    -- | KILLSWITCH: This is to be removed in a future version of Chainweb. This
-    -- prevents any transaction from entering the mempool.
+    -- | KILLSWITCH: This can be removed once the date itself has passed. Until
```
larskuhtz: Let's put the concrete date into the comment for easier use of `grep`.

```suggestion
    -- | KILLSWITCH 2019-12-05T16:00:00Z: This can be removed once the date itself has passed. Until
```

src/Chainweb/Mempool/RestAPI/Server.hs
```diff
@@ -50,14 +50,14 @@ insertHandler v mempool txsT = handleErrs (NoContent <$ begin)
         Left e -> throwM . DecodeException $ T.pack e
         Right t -> return t
 
-    -- | KILLSWITCH: The logic involving `txActivationDate` can be removed once
+    -- | KILLSWITCH: The logic involving `transferActivationDate` can be removed once
```
larskuhtz:
```suggestion
    -- | KILLSWITCH 2019-12-05T16:00:00Z: The logic involving `transferActivationDate` can be removed once
```

src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -173,9 +174,9 @@ data PactCmdLog
   deriving (Show, Generic, ToJSON, NFData)
 
 
--- | KILLSWITCH: The logic here involving `txSilenceEndDate` is to be removed in a
--- future version of Chainweb. This prevents any "real" Pact transactions from
--- being submitted to the system.
+-- | KILLSWITCH: The logic here involving `transferActivationDate` can be removed once
```
larskuhtz:
```suggestion
-- | KILLSWITCH 2019-12-05T16:00:00Z: The logic here involving `transferActivationDate` can be removed once
```

src/Chainweb/Version.hs
larskuhtz: The latest version of the timeline states `2019-12-05T16:00:00Z`. Assuming that the quasi quoter parsers UTC it should be:

```suggestion
transferActivationDate Mainnet01 = Just [timeMicrosQQ| 2019-12-05T16:00:00.0 |]
```

node/ChainwebNode.hs
fosskers: Follow-up (Part 3).

src/Chainweb/Mempool/InMem.hs
fosskers: That date isn't quite right.

src/Chainweb/Mempool/RestAPI/Server.hs
fosskers: Same here.

src/Chainweb/Pact/RestAPI/Server.hs
fosskers: Same here.

PR #744, opened 2019-11-25
Bump pact, update gas model, add gas to local

LindaOrtega:
- Bumps pact to have the latest version of the gas model
- Adds gas to /local (i.e. local will fail if gas limit not big enough)
- Adds `(env-gaslog)` unit tests to coin contract repl

TODO:
- ~Investigate possible error in tests (appeared and then didn't when run again)~ The error seemed to be tied to "AsyncCancelled" exception and it seemed to be occurring even before this PR.
- ~Finish adding gas log unit tests to coin contract repl~ DONE


test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -558,7 +560,7 @@ allocationTest iot nio = testCaseSteps "genesis allocation tests" $ \step -> do
       $ ObjectMap
       $ M.fromList
         [ (FieldKey "account", PLiteral $ LString "allocation02")
-        , (FieldKey "balance", PLiteral $ LDecimal 1099918.43) -- 1k + 1mm - gas
+        , (FieldKey "balance", PLiteral $ LDecimal 1099994.9) -- 1k + 1mm - gas
```
emilypi: Wow these got way cheaper!

src/Chainweb/Pact/PactService.hs
```diff
@@ -1122,9 +1122,11 @@ getCheckpointer :: PactServiceM cas Checkpointer
 getCheckpointer = view (psCheckpointEnv . cpeCheckpointer)
 
 -- | temporary gas model without sizing
-tableGasModelNoSize :: GasCostConfig -> GasModel
-tableGasModelNoSize gasConfig =
+_tableGasModelNoSize :: GasCostConfig -> GasModel
+_tableGasModelNoSize gasConfig =
```
emilypi: Since we've switched to the actual gas table, this should be removed, no?

slpopejoy: Yes this should be removed entirely

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -586,20 +581,41 @@ initStateInterpreter :: EvalState -> Interpreter p
...
   where
-    feePerByte :: Decimal = 0.01
-
     contProofSize =
       case _pPayload (payloadObj cmd) of
         Continuation (ContMsg _ _ _ _ (Just (ContProof p))) -> B.length p
         _ -> 0
     txSize = SB.length (payloadBytes cmd) - contProofSize
-    gasFee = round $ fromIntegral txSize * feePerByte
+    gasFee = expLengthPenalty txSize
```
slpopejoy: We probably still need the factor, charging straight byte size is most likely too big.

src/Chainweb/Pact/PactService.hs
LindaOrtega: deleted in latest commit

src/Chainweb/Pact/TransactionExec.hs
LindaOrtega: see latest commit

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -586,22 +581,56 @@ initStateInterpreter :: EvalState -> Interpreter p
...
+
+    costPerByte = fromIntegral txSize * feePerByte
+    sizePenalty = txSizeAccelerationFee costPerByte
+    gasFee = ceiling (costPerByte + sizePenalty)
 {-# INLINE initialGasOf #-}
 
+txSizeAccelerationFee :: Decimal -> Decimal
+txSizeAccelerationFee costPerByte = total
+  where
+    total = (costPerByte / bytePenalty) ^ power
```
slpopejoy: Where did the factor go? I thought it was (s * F) + ((s / P) ^ 7)

test/Chainweb/Test/Pact/SPV.hs
```diff
@@ -302,7 +302,7 @@ txGenerator1 time pidv sid tid = do
 
                 let pcid = Pact.ChainId $ chainIdToText sid
 
-                cmd <- mkTestExecTransactions "sender00" pcid ks "1" 10 0.01 100000 (toTxCreationTime time) txs
+                cmd <- mkTestExecTransactions "sender00" pcid ks "1" 24 0.01 100000 (toTxCreationTime time) txs
```
slpopejoy: Instead of 100000 let's try to size these accurately

src/Chainweb/Pact/TransactionExec.hs
LindaOrtega: I thought the formula was 
```
(s * F) + (((s * F) / 512) ^ 7)
```
and 
```
costPerByte = s * F
```

test/Chainweb/Test/Pact/SPV.hs
LindaOrtega: I think 100000  is TTLSeconds, 24 is the gas limit

slpopejoy: ok

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: ok got it

PR #745, opened 2019-11-26
[PactService] Truly provide sensible values for (chain-data) on local calls

emmanueldenloye:
I added an  extra test to `ChainData.hs` to cover the local case.


src/Chainweb/Pact/PactService.hs
```diff
@@ -796,9 +796,9 @@ execLocal cmd = withDiscardedBatch $ do
     -- necessary information for this call to return sensible values.
     parentHeader <- liftIO $! lookupM bhDb bhash
     withCheckpointer target "execLocal" $ \(PactDbEnv' pdbenv) -> do
-        PactServiceEnv{..} <- ask
         mc <- use psInitCache
-        r <- withBlockData parentHeader $
+        r <- withBlockData parentHeader $ do
+             PactServiceEnv{..} <- ask
```
slpopejoy: This seems like an unnecessary change -- if it is let's remove it to keep this PR focused on tests

slpopejoy: Is this the fix?

emmanueldenloye: Yes, this is the fix.

PR #751, opened 2019-11-26
Display Unauthorized Miner

fosskers:
A simple error message change that's helpful for debugging.

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -87,6 +87,8 @@ common warning-flags
         -Wincomplete-record-updates
         -Wincomplete-uni-patterns
         -Widentities
+        -funclutter-valid-hole-fits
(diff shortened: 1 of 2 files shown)


chainweb.cabal
```diff
@@ -95,7 +97,7 @@ common warning-flags
 custom-setup
     setup-depends:
           Cabal
-        , base >= 4.11 && < 5
+        , base >= 4.12 && < 5
```
mightybyte: Why are we bumping base here?

fosskers: It establishes GHC 8.6 officially as our lower bound. It also makes sure `-funclutter-valid-hole-fits` works for anybody who compiles this (it doesn't exist for <= 8.4)

PR #755, opened 2019-11-27
1.1 Transition, Part 1: Soft Mempool Reactivation

fosskers:
This PR reactivates the mempool threads, while leaving the `/mempool/insert` call "blocked". All calls to this endpoint will silently succeed with a no-op, so as not to upset mempool peer lists.

**Note:** All other Transaction blocks are still in place. To reiterate their locations:

1. `/send` rejects everything.
2. `/mempool/insert` no-ops on anything.
3. Mempool pre-insert rejects everything.
4. Block Validation rejects any payload with non-zero TX count.


src/Chainweb/Mempool/RestAPI/Server.hs
```diff
@@ -28,24 +27,38 @@ import Chainweb.Mempool.Mempool
...
-        txs <- mapM go txsT
-        let txV = V.fromList txs
-        liftIO $ mempoolInsert mempool CheckedInsert txV
+        now <- liftIO getCurrentTimeIntegral
+        case txActivationDate v of
+            Just start | start < now -> pure ()
+            _ -> do
+                txs <- mapM go txsT
+                let txV = V.fromList txs
+                liftIO $ mempoolInsert mempool CheckedInsert txV
```
larskuhtz: Isn't there already a guard in place in `mempooliksert`? If not, I think, that would be a better place for it.

larskuhtz: This needs a comment, that allows to identify the code via `grep` and remove it in the first release after Dec 6.

src/Chainweb/Version.hs
```diff
@@ -40,6 +40,7 @@ module Chainweb.Version
 , window
 -- ** Date-based Transaction Disabling
 , txSilenceEndDate
+, txActivationDate
```
larskuhtz: I am not sure if `Chainweb.Version` is the right place for this constant, since it's not a property of the version. Normally, things in this file stay forever.

It's probably fine to define it here for convenience. But it needs a comment (that can be identified with `grep`) that explain that it can be removed after Dec 6.

src/Chainweb/Mempool/RestAPI/Server.hs
fosskers: There is, yes, but having it here cuts out some overhead.

src/Chainweb/Version.hs
fosskers: I *will* be a property of the Version very soon. You'll notice that in the other PR, this is the chief value used to turn everything else on.

src/Chainweb/Mempool/RestAPI/Server.hs
larskuhtz: If a check in `mempooliksert` would cause an error, this handler would have to catch it. In that case the current solution in this PR seems better.

PR #757, opened 2019-11-28
[DO NOT MERGE] Use Contextual SPV Verification

emilypi:
Must go in after https://github.com/kadena-io/chainweb-node/pull/756

### DIFF
--- a/src/Chainweb/Pact/SPV.hs
+++ b/src/Chainweb/Pact/SPV.hs
@@ -114,7 +114,8 @@ verifySPV cid cdbv typ proof = readMVar cdbv >>= go typ proof
             --  3. Extract tx outputs as a pact object and return the
             --  object.
 
-            TransactionOutput p <- verifyTransactionOutputProof cdb u
+            bh <- runTransactionOutputProof u
+            TransactionOutput p <- verifyTransactionOutputProofAt cdb u bh
 
             q <- case decodeStrict' p :: Maybe HashCommandResult of
               Nothing -> internalError "unable to decode spv transaction output"
@@ -161,7 +162,8 @@ verifyCont cid cdbv (ContProof cp) = do
           --  3. Extract continuation 'PactExec' from decoded result
           --  and return the cont exec object
 
-          TransactionOutput p <- verifyTransactionOutputProof cdb u
+          bh <- runTransactionOutputProof u
+          TransactionOutput p <- verifyTransactionOutputProofAt cdb u bh
 
           q <- case decodeStrict' p :: Maybe HashCommandResult of
             Nothing -> internalError "unable to decode spv transaction output"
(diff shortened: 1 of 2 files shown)


src/Chainweb/Pact/SPV.hs
```diff
@@ -114,7 +114,8 @@ verifySPV cid cdbv typ proof = readMVar cdbv >>= go typ proof
             --  3. Extract tx outputs as a pact object and return the
             --  object.
 
-            TransactionOutput p <- verifyTransactionOutputProof cdb u
+            bh <- runTransactionOutputProof u
+            TransactionOutput p <- verifyTransactionOutputProofAt cdb u bh
```
larskuhtz: This wouldn't work. The third argument to `verifyTransactionOutputProofAt` is the checkpointer context of the current pact validation, i.e. the parent header of the block that is currently validated by pact.

### Some Background

When using a proof in the  validation block one needs to confirm that the subject of the proof (the "burn" tx in case of cross chain transfers) is contained in the history of the currently validated block. This is witnessed in two steps:

1. the self-contained proof confirms that the subject is contained in the history of the target of the proof. The target is a block on the target chain (which is chain where the "create" of a cross chain transfer is validated). The target block is determined when the proof is created.

2. a chain oracle is used to confirm that the target block is an ancestor of the currently validated block (the "create" of the cross chain transfer)

`runTransactionOutputProof` only performs the first step. Since the proof is self contained this is a pure function and doesn't depend on any context.

`verifyTransactionOutputProofAt` performs both steps. It first runs `runTransactionOutputProof` and then calls the chain oracle to confirm that the result of `runTransactionOutputProof` is in the current history. In order to perform the second step it needs to know for what block it should perform the membership query.

`verifyTransactionOutputProof` also performs both steps. It uses the current consensus state as the context for the chain oracle query. But that block can be different form the currently validated block. So, it can't be used during pact validation.

src/Chainweb/Pact/SPV.hs
```diff
@@ -161,7 +162,8 @@ verifyCont cid cdbv (ContProof cp) = do
           --  3. Extract continuation 'PactExec' from decoded result
           --  and return the cont exec object
 
-          TransactionOutput p <- verifyTransactionOutputProof cdb u
+          bh <- runTransactionOutputProof u
+          TransactionOutput p <- verifyTransactionOutputProofAt cdb u bh
```
larskuhtz: This, too, is trivially true, but not correct.

src/Chainweb/Pact/SPV.hs
larskuhtz: This code is trivially true. It confirms that the target of the proof is an ancestor of the target of the proof. What needs to be confirmed is that the target of the proof is an ancestor of the currently validated block.

emilypi: Ah so you're providing an upper bound with this block hash and I don't end up running the proof explicitly. Gotcha that will take more effort to track the parent hash.

larskuhtz:
```
 
         proof
         subject
            v
 Chain 0:   * --> * --> * --> * --> * --> * --> * --> *
                   \
                    \
                     \
 Chain 1:   * --> * --> * --> * --> * --> * --> * --> *
                              ^           ^           ^
                            proof       validated    current
                            target      block       consensus

Verification:

1. Proof verification: proof subejct is contained in proof target
2. chain oracle:  proof target is contained in context block


`verifyTransactionOutputProof == verifyTransactionOutputProof currentConsensus`

```

emilypi:
Closing due to supersedence

PR #758, opened 2019-11-28
[PactService] Do compile check in PreBlock

emmanueldenloye:
This PR has two primary goals:

1. It provides a check of successful compilation of pact code in the "preblock" stage of 
`execNewBlock` in `PactService.hs`.
2. It disallows module installations within pact code until  `txEnabledDate`. However, if another date is preferred, let's hash out what that should be with due haste.


slpopejoy:
Re-doing this with a sensible refactoring of validation logic

PR #759, opened 2019-11-28
1.1 Transition, Part 3: 1.2 Killswitch 

larskuhtz:


node/ChainwebNode.hs
```diff
@@ -443,6 +482,11 @@ pkgInfoScopes =
 -- -------------------------------------------------------------------------- --
 -- main
 
+-- KILLSWITCH 2020-01-15T00:00:00Z for version 1.1
+--
+killSwitchDate :: Maybe String
+killSwitchDate = Just "2020-01-15T00:00:00Z"
```
fosskers: Should this not be a function of `ChainwebVersion`?

larskuhtz: I don't think so. It's a property of the cabal version.

Strictly speaking the chainwebVersion is `mainnet01` and this doesn't change when cabal version 1.1 goes offline.

Chainweb Version `mainnet01` is immutable, so everything we add to it has to stay. But a kill switch supposed to be removed while the chainweb version stays the same. 

So, `mainnet01` has a property that cabal version 1.1 isn't aware of yet. So cabal version 1.1 must go away before the respective property of `mainnet01` becomes relevant.

Thinking of the chainweb version as something immutable, and thinking of this change as property of the chainweb version that always existed (and always will exist) but wasn't yet represented in the code, makes it easier to reasoning about the correctness of the validation logic.

node/ChainwebNode.hs
```diff
@@ -427,6 +429,43 @@ mkTelemetryLogger
...
+-- Kill Switch
+
+withKillSwitch
+    :: (LogLevel -> T.Text -> IO ())
+    -> Maybe UTCTime
+    -> IO a
+    -> IO a
+withKillSwitch _ Nothing inner = inner
+withKillSwitch lf (Just t) inner = race timer inner >>= \case
+    Left () -> error "kill switch thread termianted unexpectedly"
```
fosskers:
```suggestion
    Left () -> error "Kill switch thread terminated unexpectedly"
```

fosskers:
Currently does not work as intended.

node/ChainwebNode.hs
```diff
@@ -427,6 +429,43 @@ mkTelemetryLogger
...
+    Right a -> return a
+  where
+    timer = do
+        runForever lf "KillSwitch" $ do
+            now <- getCurrentTime
+            when (now >= t) $ do
+                lf Error killMessage
+                error $ T.unpack killMessage
```
larskuhtz: `error` isn't an asynchronous exception, so it won't terminate the `runForever` loop.

fosskers:
Observed to work.

PR #760, opened 2019-11-28
Use validation context for pact SPV verification instead of global consensus

larskuhtz:
* [x] add chain oracle queries to TreeDB
* [x] add proof verification functions that use chain db oracles instead of cut db
* [x] use validation context (i.e. branch of parent header) for SPV verification in pact service instead of global consensus.
* [x] remove cut db dependency from cut

Most changed lines in this PR are just for deleting the cutdb argument from all pact initialization functions.

This PR supersedes #757.


src/Chainweb/Pact/PactService.hs
```diff
@@ -295,8 +287,23 @@ serviceRequests memPoolAccess reqQ = do
...
+                -- For now grab the last header from the block hash, but that
+                -- can be @Nothing@.
+                -- Or should we use the consensus state?
+                --
+                cp <- view $ psCheckpointEnv . cpeCheckpointer
+                latestHash <- liftIO (_cpGetLatestBlock cp) >>= \case
+                    Nothing -> fmap _blockHash $ genesisBlockHeader
+                        <$> view chainwebVersion
+                        <*> view chainId
+                    Just (_, x) -> return x
```
larskuhtz: @emilypi, @gregorycollins, is this the right thing to do? Please comment.

emilypi: This is costly, but I'll have a fix out immediately after this removing it by tracking the previous block's parent hash in state. For the meantime, this is fine.

src/Chainweb/Pact/SPV.hs
```diff
@@ -114,7 +113,7 @@ verifySPV cid cdbv typ proof = readMVar cdbv >>= go typ proof
             --  3. Extract tx outputs as a pact object and return the
             --  object.
 
-            TransactionOutput p <- verifyTransactionOutputProof cdb u
+            TransactionOutput p <- verifyTransactionOutputProofAt_ bdb u bh
```
emilypi: 👍

PR #762, opened 2019-11-28
Catch decoding errors in /mining/solved

fosskers:
If people send malformed data to `/mining/solved`, this will give shed more light onto why the decoding failed.

### DIFF
--- a/src/Chainweb/Miner/RestAPI/Server.hs
+++ b/src/Chainweb/Miner/RestAPI/Server.hs
@@ -123,7 +123,7 @@ solvedHandler mr (HeaderBytes hbytes) = do
             let err = TL.encodeUtf8 $ TL.fromStrict e
             throwError err400 { errBody = "Decoding error: " <> err }
         Left _ ->
-            throwError err400 { errBody = "General exception occured in /solved handler" }
+            throwError err400 { errBody = "Unexpected encoding exception" }
         Right bh -> liftIO $ do
             publish lf ms (_coordCutDb mr) bh
             let !phash = _blockPayloadHash bh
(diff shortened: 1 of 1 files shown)


src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -115,15 +115,21 @@ workHandler' mr mcid m = do
...
-        !bct = _blockCreationTime bh
-    atomically . modifyTVar' tms . over _Unwrapped $ M.delete (T2 bct phash)
-    pure NoContent
+    ms <- liftIO $ readTVarIO tms
+    liftIO (try $ runGet decodeBlockHeaderWithoutHash hbytes) >>= \case
+        Left (DecodeException e) -> do
+            let err = TL.encodeUtf8 $ TL.fromStrict e
+            throwError err400 { errBody = "Decoding error: " <> err }
+        Left _ ->
+            throwError err400 { errBody = "General exception occured in /solved handler" }
```
larskuhtz: Are you sure that is does what you expect it to do? It will only catch exceptions of type `EncodingExceptions`. It won't catch `SomeException`.

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -190,10 +196,7 @@ miningServer
     .  Logger l
     => MiningCoordination l cas
     -> Server (MiningApi v)
-miningServer mr =
-    workHandler mr
-    :<|> liftIO . solvedHandler mr
-    :<|> updatesHandler mr
+miningServer mr = workHandler mr :<|> solvedHandler mr :<|> updatesHandler mr
```
larskuhtz: 👍

src/Chainweb/Miner/RestAPI/Server.hs
larskuhtz: You may want to use `tryAll` to handle `SomeException` and use `toException` to pattern match on the inner exception types. Or use `catches` with `Handler`.

If you intend to only catch `EncodingException` values here, you may change the "General" error message.

PR #768, opened 2019-11-30
Compile checks

slpopejoy:
- Refactors T2->T3 nonsense to a simple `Either InsertError ChainwebTransaction` vector
- Performs compile checks


slpopejoy:
@emmanueldenloye `RunGas` is an function signature, not a function -- if you look at the code you'll find that it's skipped in validation.

PR #772, opened 2019-12-01
Checkpointer fork testing

marklnichols:


test/Chainweb/Test/Pact/ForkTest.hs
```diff
@@ -0,0 +1,342 @@
...
+        ( [ PactTransaction { _pactCode = T.pack moduleStr , _pactData = d } ] )
+
+txsFromHeight mvar _h = do
+    newCount <- modifyMVar mvar (\n -> return (n + 1, n + 1))
+    V.fromList <$> tailTransactions newCount
+
+toCWTransactions :: P.ChainId -> Vector PactTransaction -> IO (Vector ChainwebTransaction)
+toCWTransactions pactCid txs = do
+    ks <- testKeyPairs sender00KeyPair Nothing
+    mkTestExecTransactions "sender00" pactCid ks "1" 100000 0.01 1000000 0 txs
```
gregorycollins: lower gas price just in case

test/Chainweb/Test/Pact/ForkTest.hs
```diff
@@ -0,0 +1,342 @@
...
+    -> ChainId
+    -> BlockHeader
+    -> Property
+prop_forkValidates pdb bhdb cid genBlock =
+    again $ ioProperty $ do
+        (trunk, left, right) <- generate genForkLengths
+        mVar <- newMVar (0 :: Int)
+        withPactProp testVer Warn pdb bhdb (testMemPoolAccess cid mVar) (return Nothing) $ \reqQ -> do
+            db <- bhdb
+            putStrLn $ "Testing fork lengths:"
```
gregorycollins: change `putStrLn` to `debug` so you can have

```
debug :: String -> IO ()
#define DEBUG_TEST 0
#if DEBUG_TEST
debug = putStrLn
#else
debug = const $ return ()
#endif
```

Everyone grouses about chatty tests but the debug output is useful when you're trying to diagnose an issue

test/Chainweb/Test/Pact/ForkTest.hs
```diff
@@ -0,0 +1,342 @@
...
+
+    stopPact :: (Async a, TBQueue a2) -> IO ()
+    stopPact (a, _) = cancel a
+
+    logger = genericLogger logLevel T.putStrLn
+    cid = someChainId version
+
+genForkLengths :: Gen (Int, Int, Int)
+genForkLengths = do
+    let maxTotalLen = 12
```
gregorycollins: These forks, and the total branch depth, aren't very deep. I don't know whether that matters for us here but e.g. for the new checkpointer, this mightn't even be deep enough to trigger the compaction logic

marklnichols: I had to control the length here to avoid hitting the max rewind of 8 on newBlock -- is that not going to be an issue with the new version?

marklnichols: As per discussion, I'm going to make the depth configurable in another PR

PR #773, opened 2019-12-01
[Bugfix] Reward Miner For Failed Payloads

emilypi:
This addresses a bug in the tx exec logic which failed to reward miners in the case of payloads that failed for whatever reason.


emilypi:
i can't approve my own PR's, but this LGTM

PR #786, opened 2019-12-03
run apt-get update in github-ci builds

larskuhtz:


.github/workflows/applications.yml
```diff
@@ -59,7 +59,9 @@ jobs:
          cabal-version: ${{ matrix.cabal }}
     - name: Install non-Haskell dependencies (ubuntu)
       if: contains(matrix.os, 'ubuntu')
-      run: sudo apt-get install -y git librocksdb-dev zlib1g-dev libtinfo-dev libsqlite3-dev libz3-dev z3
+      run: |
+          sudo apt-get update
```
gregorycollins: Does this also need `-y`?

larskuhtz: I don't think that it has any user interaction. I think, I never put it it there.

PR #788, opened 2019-12-04
prune peer db and update peer selection

larskuhtz:
* [x] prune peer db on each peer db sync (is that too often?)
* [x] limit number of _new_ peers on each sync to at most 32
* [x] process up to 16 new peers concurrently (most of the overhead is in the connection check which is IO bound and thus there is no need to limit it to the number of capabilities)
* [x] on new peers selection only use two priority classes and sort the second class by successive failures (ascending).

### DIFF
--- a/src/P2P/Node.hs
+++ b/src/P2P/Node.hs
@@ -345,15 +345,12 @@ guardPeerDb
 guardPeerDb v nid peerDb pinf = do
     peers <- peerDbSnapshot peerDb
     if
-        | isKnown peers -> return $ Right pinf
+        | isKnown peers pinf -> return $ Right pinf
         | isReserved -> return $ Left $ IsReservedHostAddress pinf
         | otherwise -> canConnect >>= \case
             Left e -> return $ Left $ IsNotReachable pinf (sshow e)
             Right _ -> return $ Right pinf
   where
-    isKnown :: PeerSet -> Bool
-    isKnown peers = not . IXS.null $ IXS.getEQ (_peerAddr pinf) peers
-
     isReserved :: Bool
     isReserved = case v of
         Mainnet01 -> isReservedHostAddress (_peerAddr pinf)
@@ -369,6 +366,9 @@ guardPeerDb v nid peerDb pinf = do
         mgr <- getNewPeerManager
         let env = peerInfoClientEnv mgr pinf
         runClientM (peerGetClient v nid (Just 0) Nothing) env
+
+isKnown :: PeerSet -> PeerInfo -> Bool
+isKnown peers pinf = not . IXS.null $ IXS.getEQ (_peerAddr pinf) peers
 
 guardPeerDbOfNode
     :: P2pNode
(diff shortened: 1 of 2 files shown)


src/P2P/Node.hs
```diff
@@ -345,15 +345,12 @@ guardPeerDb
 guardPeerDb v nid peerDb pinf = do
     peers <- peerDbSnapshot peerDb
     if
-        | isKnown peers -> return $ Right pinf
+        | isKnown peers pinf -> return $ Right pinf
```
larskuhtz: When called via `syncFromPeer` this is redundant, but this function is also called from elsewhere so we leave this check in here.

src/P2P/Node.hs
```diff
@@ -397,30 +397,36 @@ peerClientEnv node = peerInfoClientEnv (_p2pNodeManager node)
...
+                logg node Warn $ "failed to sync peers from " <> showInfo info <> ": " <> sshow e
+                return False
+        Right p -> do
+            peers <- peerDbSnapshot peerDb
+            goods <- fmap catMaybes
+                $ traverseConcurrently (ParN 16) (guardPeerDbOfNode node)
+                $ take 32
+                    -- limit the maximum number of new unknown peers to 32
+                $ filter (\i -> me /= _peerId i)
+                $ filter (not . isKnown peers)
```
larskuhtz: We want to take at most 32 _new_ peers, so we filter for unknown peers.

src/P2P/Node.hs
```diff
@@ -501,22 +513,39 @@ findNextPeer conf node = do
...
+
+    let p0 = toList
+            $ IXS.getGT (ActiveSessionCount 0)
+            $ IXS.getLTE (SuccessiveFailures 1) base
+        p1 = fmap snd
+            $ IXS.groupAscBy @SuccessiveFailures
+            $ IXS.union
+                (IXS.getEQ (ActiveSessionCount 0) base)
+                (IXS.getGT (SuccessiveFailures 1) base)
+    searchSpace <- concat <$> traverse shiftR (p0: p1)
```
larskuhtz:
```suggestion
    searchSpace <- concat <$> traverse shiftR (p0 : p1)
```

src/P2P/Node/PeerDB.hs
```diff
@@ -357,6 +358,25 @@ peerDbDelete (PeerDb _ lock var) i = withMVar lock
...
+        let cutoff = Just $ addUTCTime ((-60) * 60 * 12) now
+        atomically $ modifyTVar' var $ \s ->
+            (getGT (ActiveSessionCount 0) s)
+            |||
+            (getLTE (SuccessiveFailures 5) s)
+            |||
+            (getGT (LastSuccess cutoff) s)
+            |||
+            (fromList $ filter _peerEntrySticky $ toList s)
+
```
larskuhtz: Remove redundant parenthesis and use `|||` as line prefix

src/P2P/Node.hs
```diff
@@ -397,30 +397,36 @@ peerClientEnv node = peerInfoClientEnv (_p2pNodeManager node)
...
+                logg node Warn $ "failed to sync peers from " <> showInfo info <> ": unknown certificate. Deleting peer from peer db"
+                peerDbDelete (_p2pNodePeerDb node) info
+                return False
+            | otherwise -> do
+                logg node Warn $ "failed to sync peers from " <> showInfo info <> ": " <> sshow e
+                return False
+        Right p -> do
+            peers <- peerDbSnapshot peerDb
+            goods <- fmap catMaybes
+                $ traverseConcurrently (ParN 16) (guardPeerDbOfNode node)
```
fosskers: How about a multiple of `caps`?

src/P2P/Node.hs
fosskers: Combine the two `filter` lines?

src/P2P/Node.hs
```diff
@@ -490,6 +496,12 @@ findNextPeer conf node = do
             check (pid /= myPid)
             return n
 
+    -- Classify the peers by priority
+    --
+    let base = IXS.getEQ (_p2pNodeNetworkId node) peers
+
+#if 0
```
fosskers: Still relevant?

src/P2P/Node.hs
larskuhtz: I don't think it that it makes sense to bind this to caps. This is mostly about how many parallel network connections are established.

PR #790, opened 2019-12-04
Debugging with Header dump

emmanueldenloye:


chainweb.cabal
```diff
@@ -702,6 +702,7 @@ executable cwtool
         , QuickCheck >= 2.12.6
         , aeson >= 1.4
         , aeson-pretty >= 0.8
+        , aeson-lens >= 0.5.0.0
```
mightybyte: I'd go with `lens-aeson` instead.

PR #791, opened 2019-12-04
Primed Mining Coordination

fosskers:
This PR allows a Node to set a list of "blessed" Miner identities. Each such miner will have Payloads preformed for it, such that new work requests will return (almost) immediately without incurring the usual costs of going through `_pactNewBlock`.

**Note:** As is, this PR breaks existing configurations! What was previously:
```yaml
      miners:
      - abc123  # just an account name
      - zxy098
```
Must now be:

```yaml
      miners:
      - account: abc123
        predicate: keys-all
        public-keys:
        - 3438e5bcfd086c5eeee1a2f227b7624df889773e00bd623babf5fc72c8f9aa63
      - account: zxy098
        predicate: keys-all
        public-keys:
        - cfd7816f15bd9413e5163308e18bf1b13925f3182aeac9b30ed303e8571ce997
```

It honestly should have been this shape from the beginning.

Closes #598

### DIFF
--- a/src/Chainweb/Chainweb/MinerResources.hs
+++ b/src/Chainweb/Chainweb/MinerResources.hs
@@ -69,6 +69,7 @@ import Chainweb.Version
 import Chainweb.WebPactExecutionService (_webPactExecutionService)
 
 import Data.LogMessage (JsonLog(..), LogFunction)
+import Utils.Logging.Trace (trace)
 
 -- -------------------------------------------------------------------------- --
 -- Miner
--- a/src/Chainweb/Chainweb/MinerResources.hs
+++ b/src/Chainweb/Chainweb/MinerResources.hs
@@ -39,7 +39,7 @@ import Data.Tuple.Strict (T2(..), T3(..))
 import qualified Data.Vector as V
 
 import Control.Concurrent (threadDelay)
-import Control.Concurrent.Async (Concurrently(..), runConcurrently)
+import Control.Concurrent.Async
 import Control.Concurrent.STM (atomically)
 import Control.Concurrent.STM.TVar
 import Control.Lens (at, over, view, (&), (?~), (^?!))
@@ -106,7 +106,7 @@ withMiningCoordination logger conf cdb inner
         l <- newIORef (_coordinationUpdateStreamLimit conf)
         fmap thd . runConcurrently $ (,,)
             <$> Concurrently (prune t c503 c403)
-            <*> Concurrently (primeWork miners m cut $ unsafeChainId 0)  -- TODO per chain
+            <*> Concurrently (mapConcurrently_ (primeWork miners m cut) cids)
             <*> Concurrently (inner . Just $ MiningCoordination
                 { _coordLogger = logger
                 , _coordCutDb = cdb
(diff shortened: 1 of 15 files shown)


fosskers:
Not quite ready, I have one thing to fix and one thing to add.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -85,24 +95,96 @@ withMiningCoordination
...
+    primeWork :: [Miner] -> TVar PrimedWork -> Cut -> ChainId -> IO ()
+    primeWork miners tpw c cid = runForever (logFunction logger) "primeWork" $ go c
+      where
+        go :: Cut -> IO a
+        go cut = do
+            -- Even if the `cut` passed to the function is old, this call here will
+            -- immediately detect the newest `BlockHeader` on the given chain.
+            new <- awaitNewCutByChainId cdb cid cut
+            -- Temporarily block this chain from being considered for queries --
+            atomically $ modifyTVar' tpw (silenceChain cid)
```
fosskers: Temporary block on an updating chain.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -85,24 +95,96 @@ withMiningCoordination
...
+            -- Generate new payloads, one for each Miner we're managing --
+            let !newParent = ParentHeader . fromJuste . HM.lookup cid $ _cutMap new
+            payloads <- traverse (\m -> T2 m <$> getPayload newParent m) miners
+            -- Update the cache in a single step --
+            atomically $ modifyTVar' tpw (\pw -> foldl' (updateCache cid) pw payloads)
```
fosskers: Update the chain and unblock it.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -85,24 +95,96 @@ withMiningCoordination
...
+      :: Miner
+      -> [T2 ChainId ParentHeader]
+      -> IO (HM.HashMap ChainId (Maybe (T2 PayloadWithOutputs BlockCreationTime)))
+    fromCut m cut =
+        HM.fromList <$> traverse (\(T2 cid bh) -> (cid,) . Just <$> getPayload bh m) cut
+
+    getPayload :: ParentHeader -> Miner -> IO (T2 PayloadWithOutputs BlockCreationTime)
+    getPayload (ParentHeader parent) m = do
+        creationTime <- BlockCreationTime <$> getCurrentTimeIntegral
+        payload <- _pactNewBlock pact m parent creationTime
```
fosskers: @larskuhtz This isn't `trace`d like the original call. Should it be?

larskuhtz: Yes, please trace all calls of `_pactNewBlock`. These is important information for measuring overall performance of the system and catching performance regressions.

You may provide the full qualified path of the function. That would allow to distinguish the different code paths in the telemetry data, if needed.

src/Chainweb/Utils.hs
```diff
@@ -1149,6 +1152,14 @@ withLink act = do
...
   return a
 
+
+-- -------------------------------------------------------------------------- --
+-- Tuples
+
+thd :: (a,b,c) -> c
+thd (_,_,c) = c
+{-# INLINE thd #-}
+
```
larskuhtz: `view _3` isn't that much longer. Though I am fine with keeping this.

PR #793, opened 2019-12-05
Enforce lower bound on acceptable node versions

larskuhtz:


src/Chainweb/NodeVersion.hs
```diff
@@ -0,0 +1,137 @@
...
+
+isAcceptedVersion :: NodeVersion -> Bool
+isAcceptedVersion = (<=) minAcceptedVersion
+{-# INLINE isAcceptedVersion #-}
+
+-- -------------------------------------------------------------------------- --
+-- Query Remote Version
+
+requestTimeout :: Int
+requestTimeout = 4 * 1000000
```
fosskers: Units?

src/Chainweb/NodeVersion.hs
```diff
@@ -0,0 +1,137 @@
...
+-- Chainweb Node Version
+
+newtype NodeVersion = NodeVersion { _getNodeVersion :: [Int] }
+    deriving (Show, Eq, Ord, Generic, NFData)
+
+instance HasTextRepresentation NodeVersion where
+    toText (NodeVersion l) = T.intercalate "." $ sshow <$> l
+    fromText = fmap NodeVersion . traverse treadM . T.splitOn "."
+    {-# INLINE toText #-}
+    {-# INLINE fromText #-}
```
fosskers: Consider instead [my versions library](http://hackage.haskell.org/package/versions) for parsing these.

larskuhtz: I think it's not needed in this case. We don't have to represent arbitrary version numbers but only chainweb-node version numbers that will always be a lexicographical list of `Int`. Taking a new dependency and using a more complex type seems overkill for that.

larskuhtz: I think, the versions library is great for representing heterogenous sets of version from different packages. But here we deal with just a single package and we have full control over the format.

PR #806, opened 2019-12-09
Raise minimum accepted chainweb node version to 1.3

larskuhtz:
Version 1.3 will perform a centralized hard fork at 2019-12-10T21:00:00Z. After that date nodes with version <= 1.2 are considered not being part of the network any more.

Nodes <= 1.2 are vulnerable to the coinbase injection attack. New blocks resulting from this attacks will cause validation failures in nodes >= 1.3 after the fork date.

### DIFF
--- a/src/Chainweb/NodeVersion.hs
+++ b/src/Chainweb/NodeVersion.hs
@@ -60,8 +60,11 @@ instance FromJSON NodeVersion where
     parseJSON = parseJsonFromText "NodeVersion"
     {-# INLINE parseJSON #-}
 
+-- | Node verison 1.3 performs a centralized hard fork at 2019-12-10T21:00:00Z.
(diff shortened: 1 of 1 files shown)


larskuhtz:
I am not sure if we should merge this or not. If the remaining nodes 1.2 have only a fraction of the hash power and bad block that they would mine would be rejected quickly and the nodes would still be valid members of the network otherwise.

Only if the bad nodes would have enough hash power to repeatedly inject bad nodes and win forks of non-trivial length, this could cause issues.

The largest risk would be if validation of bad blocks (and the resulting failure conditions) would create so much overhead in the 1.3 nodes such that the old nodes would get an advantage, which causes them to win more blocks, thus creating an avalanche effect.

PR #808, opened 2019-12-09
[Tooling] Add decoding tool to cwtool

emmanueldenloye:


tools/cwtool/CwTool.hs
```diff
@@ -91,6 +92,11 @@ topLevelCommands =
       "header-dump"
       "Dump Block Headers to a JSON array"
       HeaderDump.main
+  , CommandSpec
+      "decode" -- TODO: there is probably a better title to use here
```
mightybyte: Since the variants are nested beneath this top-level command, maybe use `b64`?  And below possibly `Command line utils for Chainweb's base64 encode/decode`?

tools/cwtool/CwTool.hs
```diff
@@ -91,6 +92,11 @@ topLevelCommands =
...
+      "decodeB64UrlNoPaddingText" -- TODO: there is probably a better title to use here
+      EncodeDecodeB64Util.main
+      -- TODO: Might want to put in all variants here?
```
mightybyte: I think it's fine to leave them in the subcommand.

PR #809, opened 2019-12-09
Coin Contract v2

emilypi:


pact/coin-contract/v2/coin.pact
```diff
@@ -48,6 +48,10 @@
     "Magic capability constraining genesis transactions"
     true)
 
+  (defcap SLASH ()
```
slpopejoy: We might as well bring in the new Pact and make this a auto-managed cap with the amount

pact/coin-contract/v2/fungible-v2.pact
```diff
@@ -39,6 +39,21 @@
...
 
+
+  (defun transfer:string
+    ( sender:string
+      receiver:string
+      amount:decimal
+    )
+    @doc " Transfer AMOUNT between accounts SENDER and RECEIVER. \
+         \ Fails if either SENDER or RECEIVER does not exist."
+    @model [ (property conserves-mass)
```
slpopejoy: This doesn't exist ...

pact/coin-contract/v2/coin.pact
```diff
@@ -326,6 +330,20 @@
       (credit account account-guard amount))
     )
 
+  (defun slash:string (account:string amount:decimal)
+    @doc "Internal function for the termination of coins. This function \
+         \cannot be used outside of the coin contract."
```
slpopejoy: Probably better to say something about it being protected by SLASH rather than "internal" and "cannot be used"

pact/coin-contract/v2/coin.repl
```diff
@@ -813,3 +813,22 @@
...
+  "not granted: (coin.SLASH)"
+  (slash "brandon" 1.0))
+
+(test-capability (SLASH))
+
+(env-keys ["brandon"])
+(expect
+  "Slashes succeed in the presence of SLASH"
+  "Write succeeded"
+  (slash "brandon" 1.0))
```
slpopejoy: Should cover negative amount, unit, valid account

pact/coin-contract/v2/coin.pact
```diff
@@ -326,6 +330,20 @@
...
+
+    @model [ (property (valid-account account)) ]
```
slpopejoy: Both this and `coinbase` should have more properties, namely amount > 0, `valid-account`.

Also, `transfer-create` is missing a lot of properties that `transfer` has.

pact/coin-contract/v2/coin.repl
emilypi: Addressed all of these

buckie:
talked to @slpopejoy, can we please add constants that we know are inviolable. e.g. `slash` can only run on or before the fork occurs and coinbase can never create more than 2.3 coins.

pact/coin-contract/v2/coin.pact
```diff
@@ -245,16 +254,16 @@
...
+    (with-capability (ROTATE new-guard)
+      (with-read coin-table account
+        { "guard" := old-guard }
 
-    (with-read coin-table account
-      { "guard" := old-guard }
-
-      (enforce-guard old-guard)
-      (enforce-guard new-guard)
+        (enforce-guard old-guard)
```
emilypi: note: only enforce old guard

pact/coin-contract/v2/coin.pact
```diff
@@ -57,6 +61,11 @@
     "Capability for managing crediting operations"
     (enforce (!= receiver "") "valid receiver"))
 
+  (defcap ROTATE (guard:guard)
```
slpopejoy: I don't think this will work in signatures, change this to just `account:string`

pact/coin-contract/v2/coin.pact
```diff
@@ -326,6 +337,21 @@
       (credit account account-guard amount))
     )
 
+  (defun remediate:string (account:string amount:decimal)
+    @doc "Slash allows the termination of coins. This function \
```
slpopejoy:
```suggestion
    @doc "Remediate an incorrect coinbase or other credit. This function \
```

pact/coin-contract/v2/coin.pact
```diff
@@ -326,6 +337,21 @@
...
+         \is protected by the SLASH capability"
```
slpopejoy:
```suggestion
         \is protected by the magic REMEDIATE capability to prevent calling in regular transactions."
```

pact/coin-contract/v2/coin.pact
```diff
@@ -326,6 +337,21 @@
...
+    @model [ (property (valid-account account))
+             (property (> amount 0.0))
+           ]
+
+    (validate-account account)
+    (enforce-unit amount)
+
+    (require-capability (REMEDIATE account amount))
+    (with-capability (DEBIT account)
```
slpopejoy: This isn't going to work, as we already discussed. Please fix.

pact/coin-contract/v2/coin.repl
```diff
@@ -556,7 +559,20 @@
...
 
 ; account details will now feature rotated guard
-(rotate 'emily (read-keyset 'doug))
+(env-keys ["keys1", "keys2"])
+(expect-failure
+  "guard rotation fails when ROTATE is not scoped"
+  "Managed capability not installed"
+  (rotate 'emily (read-keyset 'doug)))
+
+(env-sigs [{'key: "keys1", 'caps: [(coin.ROTATE (read-keyset 'doug))]}])
```
slpopejoy: So, the issue with sig caps is they can only be primitive pact values. I don't know how we resolve this with having `PactValue` in the `UserCapability` type, but it's something to deal with in Pact. Maybe we need to enforce this in `env-sigs`?

pact/coin-contract/v2/coin.repl
```diff
@@ -813,3 +831,44 @@
...
+(begin-tx)
+
+(use coin)
+
+(expect-failure
+  "Remediations fail without the presence of REMEDIATE"
+  "not granted"
+  (remediate "brandon" 1.0))
+
+(env-keys ["brandon"])
```
slpopejoy: This is precisely what we _cannot_ do in remediations.

pact/coin-contract/v2/coin.pact
emilypi: Odd - I think this was merged incorrectly and overwrote my doc changes

pact/coin-contract/v2/coin.repl
emilypi: ahhhhh `DEBIT` enforces the guard. Right. We'll need to do our own debiting operation

pact/coin-contract/v2/coin.repl
emilypi: We should probably parse and case on the pact value type

test/Chainweb/Test/Pact/TransactionTests.hs
```diff
@@ -0,0 +1,253 @@
...
+
+    doCoinbase pdb mc postForkTime cmd' True $ \pr -> case pr of
+      Left _ -> assertFailure "local call to get-balance failed"
+      Right (PLiteral (LDecimal d))
+        | d == 0.3 -> return ()
+        | otherwise -> assertFailure $ "miner balance is incorrect: " <> show d
+      Right l -> assertFailure $ "wrong return type: " <> show l
+
+  where
+    doCoinbase pdb mc t localCmd precompile testResult = do
```
slpopejoy:
```suggestion
    doCoinbaseExploit pdb mc t localCmd precompile testResult = do
```

test/Chainweb/Test/Pact/TransactionTests.hs
```diff
@@ -0,0 +1,253 @@
...
+  (r, rst) <- execScript' (Script False coinRepl) coinRepl
+  either fail (const $ return ()) r
+  let pdb = PactDbEnv
+            (view (rEnv . eePactDb) rst)
+            (view (rEnv . eePactDbVar) rst)
+      mc = view (rEvalState . evalRefs . rsLoadedModules) rst
+  return (pdb,mc)
+
+
+testCoinbase :: Assertion
```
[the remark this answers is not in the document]
emilypi: Removed

PR #810, opened 2019-12-10
Use block time instead of tx creation time

emilypi:
This was not caught in testing because `DevNet` fork time was set to `epoch`, and `def` for public meta's tx creation time is 0 - also `epoch`. These changes will make use of whatever the block time in the public data is from the header of the previous block of the current one.

### DIFF
--- a/src/Chainweb/Pact/TransactionExec.hs
+++ b/src/Chainweb/Pact/TransactionExec.hs
@@ -76,7 +76,7 @@ import Pact.Gas.Table
 import Pact.Interpreter
 import Pact.Native.Capabilities (evalCap)
 import Pact.Parse (parseExprs)
-import Pact.Parse (ParsedDecimal(..), ParsedInteger(..))
+import Pact.Parse (ParsedDecimal(..))
 import Pact.Runtime.Capabilities (popCapStack)
 import Pact.Types.Capability
 import Pact.Types.Command
@@ -257,10 +257,7 @@ applyCoinbase v logger dbEnv (Miner mid mks) reward@(ParsedDecimal d) pd ph
     forkTime = vuln797FixDate v
     fork1_3InEffect = blockTime >= forkTime
     throwCritical = fork1_3InEffect || enfCBFailure
-    blockTime =
-      let (TxCreationTime (ParsedInteger !bt)) =
-            view (pdPublicMeta . pmCreationTime) pd
-      in Time (TimeSpan (Micros (fromIntegral bt)))
+    blockTime = Time (TimeSpan (Micros $ _pdBlockTime pd))
 
     tenv = TransactionEnv Transactional dbEnv logger pd noSPVSupport
            Nothing 0.0 rk 0 restrictiveExecutionConfig


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -257,10 +257,7 @@ applyCoinbase v logger dbEnv (Miner mid mks) reward@(ParsedDecimal d) pd ph
     forkTime = vuln797FixDate v
     fork1_3InEffect = blockTime >= forkTime
     throwCritical = fork1_3InEffect || enfCBFailure
-    blockTime =
-      let (TxCreationTime (ParsedInteger !bt)) =
-            view (pdPublicMeta . pmCreationTime) pd
-      in Time (TimeSpan (Micros (fromIntegral bt)))
+    blockTime = Time (TimeSpan (Micros $ _pdBlockTime pd))
```
emilypi: This is set in `setBlockData` in `PactService.hs` using the parent header. The public data used for coinbases is a call to `mkPublicData "coinbase" def`, which constructs a public data object from default public meta with no modifications to the block data set in `setBlockData`. So `_pdBlockTime` is exactly the block time from the parent header.

larskuhtz: `mkPublicData "coinbase" def` How many record fields are in `PublicData`? Would it help to remove the `Default` instance and force explicit construction (using record initialization syntax with field names), so that it's obvious in the code what is constructed?

larskuhtz: I wonder if this is an instance of "Code readability should have precedency over writing-convenience and minimizing LOC"?

emilypi: @larskuhtz there are 4 fields in `PublicData`, and yes, we should probably do away with the `Default` instances. But that's not what was at play here. What was at play was that the entire team was stressed, especially me, and I made a stupid error conflating `txCreationTime` with `blockCreationTime`.

slpopejoy: Yeah I don't think this is a `def` issue either. `PublicMeta` has no meaningful values for coinbase, so using a default makes sense, as we don't want `PublicMeta` to be a `Maybe` in `PublicData`

slpopejoy: @emilypi can you document (in this PR, not necc. in code) where `_pdBlockTime` is set so we can confirm it is the right format for direct comparison with `vuln797FixDate`?

emilypi: - It's set here: https://github.com/kadena-io/chainweb-node/blob/e4d797827bc8f5578729ab9c1a0dd7cf0b960373/src/Chainweb/Pact/PactService.hs#L868

- and added to a `PublicData` object here: https://github.com/kadena-io/chainweb-node/blob/e4d797827bc8f5578729ab9c1a0dd7cf0b960373/src/Chainweb/Pact/PactService.hs#L1094


There are only 3 points where block data is ever set: 

- In `execNewBlock`: https://github.com/kadena-io/chainweb-node/blob/e4d797827bc8f5578729ab9c1a0dd7cf0b960373/src/Chainweb/Pact/PactService.hs#L733

- In `execLocal`: https://github.com/kadena-io/chainweb-node/blob/e4d797827bc8f5578729ab9c1a0dd7cf0b960373/src/Chainweb/Pact/PactService.hs#L845

- and in `playOnBlock` (which means it's used for genesis and non genesis blocks): https://github.com/kadena-io/chainweb-node/blob/e4d797827bc8f5578729ab9c1a0dd7cf0b960373/src/Chainweb/Pact/PactService.hs#L929

PR #815, opened 2019-12-11
Log size of primed work cache

fosskers:
Some extra debugging information for the new Primed Coordination.

### DIFF
--- a/src/Chainweb/Chainweb/MinerResources.hs
+++ b/src/Chainweb/Chainweb/MinerResources.hs
@@ -107,7 +107,7 @@ withMiningCoordination logger conf cdb inner
         c403 <- newIORef 0
         l <- newIORef (_coordinationUpdateStreamLimit conf)
         fmap thd . runConcurrently $ (,,)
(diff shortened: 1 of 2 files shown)


src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -199,9 +199,15 @@ withMiningCoordination logger conf cdb inner
...
+        PrimedWork pw <- readTVarIO tpw
         atomicWriteIORef c503 0
         atomicWriteIORef c403 0
-        logFunction logger Info . JsonLog $ MiningStats (M.size ms) count503 count403 (avgTxs m)
+        logFunction logger Info . JsonLog $ MiningStats
+            { _statsCacheSize = M.size ms
+            , _stats503s = count503
+            , _stats403s = count403
+            , _statsAvgTxs = avgTxs m
+            , _statsPrimedSize = HM.foldl' (\acc xs -> acc + HM.size xs) 0 pw }
```
fosskers: `HM.size` is `O(n)` but we expect these caches to be small.

PR #819, opened 2019-12-11
Primed Coordination for all Whitelisted Miners

fosskers:
This PR alters the coordinator to give preferential treatment to any whitelisted miner in the `miners` list, regardless of the public/private mining scheme.

### DIFF
--- a/src/Chainweb/Miner/Coordinator.hs
+++ b/src/Chainweb/Miner/Coordinator.hs
@@ -29,6 +29,7 @@ module Chainweb.Miner.Coordinator
   , ChainChoice(..)
   , PrimedWork(..)
   , CachedPayload
+  , Whitelisted(..), whitelisted
     -- * Functions
   , newWork
   , publish
@@ -121,18 +122,27 @@ newtype PrevTime = PrevTime BlockCreationTime
 
 data ChainChoice = Anything | TriedLast ChainId | Suggestion ChainId
 
+-- | A `Miner` defined in the nodes `miners` list. Should be subject to Primed
+-- Coordination.
+--
+newtype Whitelisted = Whitelisted Miner
+
+whitelisted :: Either Miner Whitelisted -> Miner
+whitelisted (Left m) = m
+whitelisted (Right (Whitelisted m)) = m
+
 -- | Construct a new `BlockHeader` to mine on.
 --
 newWork
     :: LogFunction
     -> CoordinationMode
     -> ChainChoice
-    -> Miner
+    -> Either Miner Whitelisted
     -> PactExecutionService
     -> TVar PrimedWork
     -> Cut
     -> IO (T3 PrevTime BlockHeader PayloadWithOutputs)
-newWork logFun mode choice miner@(Miner mid _) pact tpw c = do
+newWork logFun mode choice eminer pact tpw c = do
     -- Randomly pick a chain to mine on, unless the caller specified a specific
     -- one.
     --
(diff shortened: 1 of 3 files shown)


src/Chainweb/Miner/Coordinator.hs
```diff
@@ -121,18 +122,27 @@ newtype PrevTime = PrevTime BlockCreationTime
 
 data ChainChoice = Anything | TriedLast ChainId | Suggestion ChainId
 
+-- | A `Miner` defined in the nodes `miners` list. Should be subject to Primed
+-- Coordination.
+--
+newtype Whitelisted = Whitelisted Miner
```
larskuhtz: I think the code below in this PR would be more readable if this would be something like:

```Haskell
data MiningClient = Primed Miner | Public Miner

miningClient :: MiningClient -> Miner
miningClient (Primed m) = m
miningClient (Public m) = m
```

instead of passing around `Either Miner Whitelisted` (which is isomorphic to above type but less explicit and more complex).

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -144,10 +154,10 @@ newWork logFun mode choice miner@(Miner mid _) pact tpw c = do
     --
     let !p = ParentHeader (c ^?! ixg cid)
 
-    bool (public p) (private cid p <$> readTVarIO tpw) (mode == Private) >>= \case
+    either (public p) (\m -> primed m cid p <$> readTVarIO tpw) eminer >>= \case
         -- The proposed Chain wasn't mineable, either because the adjacent
         -- parents weren't available, or because the chain is mid-update.
-        Nothing -> newWork logFun mode (TriedLast cid) miner pact tpw c
+        Nothing -> newWork logFun mode (TriedLast cid) eminer pact tpw c
```
larskuhtz: Unrelated to this PR: I wonder what happens if all but one chain is blocked. Would this code would spin while pact computes newBlock. That might cause performance issues. I think that this is rare. But I wonder if we should detect and log that condition.

In the primed case, I wonder if the next chain could be randomly chosen from the available chains in primed work. If no work is available in `PrimedWork` the code could await a change to the PrimedWork variable, thus avoiding to spin.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -121,18 +122,27 @@ newtype PrevTime = PrevTime BlockCreationTime
...
     -> CoordinationMode
     -> ChainChoice
-    -> Miner
+    -> Either Miner Whitelisted
     -> PactExecutionService
     -> TVar PrimedWork
     -> Cut
     -> IO (T3 PrevTime BlockHeader PayloadWithOutputs)
-newWork logFun mode choice miner@(Miner mid _) pact tpw c = do
+newWork logFun mode choice eminer pact tpw c = do
```
larskuhtz: Is `mode` used anywhere in this function (except for the recursive call)?

src/Chainweb/Miner/Coordinator.hs
fosskers: > I wonder what happens if all but one chain is blocked. 

Can you describe when such a thing would happen?

src/Chainweb/Miner/Coordinator.hs
fosskers: No we can take that out now, good catch.

src/Chainweb/Miner/Coordinator.hs
larskuhtz: I don't know how often this happens, but just think of a process that mines on all chains but one. At some point all chains would be blocked waiting for the one chain. I hope that this is very rare with random chain selection, but it has a non-zero probability.

PR #820, opened 2019-12-11
Disable charging gas for module lookups

LindaOrtega:
Modifies the table gas model to charge Gas 0 for module lookups. Previously, module lookups cost Gas 1. Thus, depenalizing the first transaction to load a module into a block's module cache.


src/Chainweb/Pact/PactService.hs
```diff
@@ -1248,3 +1248,20 @@ findLatestValidBlock = getCheckpointer >>= liftIO . _cpGetLatestBlock >>= \case
...
+
+-- | Modified table gas module with free module loads
+--
+freeModuleLoadGasModel :: P.GasModel
+freeModuleLoadGasModel = modifiedGasModel
+  where
+    defGasModel = tableGasModel defaultGasConfig
+    fullRunFunction = P.runGasModel defGasModel
+    modifiedRunFunction name ga = case ga of
+      P.GPostRead (P.ReadModule {}) -> 0
```
slpopejoy: Why do we need this? Why isn't changing the config enough?

LindaOrtega: because making `_gasCostConfig_readColumnCost = 0` would make _all_ `GPostRead` Gas 0.
https://github.com/kadena-io/pact/blob/5c25e8369917eaf07ef66ed3b265f15cb2190378/src/Pact/Gas/Table.hs#L209

slpopejoy: oh, crap

PR #821, opened 2019-12-11
[PactService] Inspecting the module cache on node restart.

emmanueldenloye:
This PR includes a test to make sure that the module cache in PactService is the same when the node is restarted. This restart assumes that the SQLite database from the previous node run has not been touched before the initiation of the restart.

(Currently, the test is failing).
EDIT: Now the test passes.


src/Chainweb/Pact/PactService.hs
```diff
@@ -262,37 +262,54 @@ initPactService' ver cid chainwebLogger bhDb pdb sqlenv reorgLimit act = do
...
+      then do
+          pdb <- asks _psPdb
+          bhdb <- asks _psBlockHeaderDb
+          reloadedCache <- liftIO $
+              withTempSQLiteConnection chainwebPragmas $ \sqlenv ->
+                  initPactService' v cid logger bhdb pdb sqlenv defaultReorgLimit $ do
+                      -- it is reasonable to assume genesis doesn't exist here
+                      validateGenesis
+                      gets _psInitCache
+          psInitCache .= reloadedCache
```
emilypi: these two lines setting the cache should be unnecessary if we're doing a full play through, since the state is updated in `applyPactCmd`. Have you dumped the cache without setting it explicitly like this to confirm?

test/Chainweb/Test/Pact/ModuleCacheOnRestart.hs
```diff
@@ -0,0 +1,159 @@
...
+                  modifyMVar_ mcache (const (pure (_psInitCache pstate)))
+              Check ioa -> do
+                  a <- ioa
+                  assertBool "The module cache on restart is not the same" (a == (_psInitCache pstate))
+
+    stopPact (sqlEnv, _) = stopSqliteDb sqlEnv
+    logger = genericLogger logLevel T.putStrLn
+    cid = someChainId version
+
+-- we want a special version of initPactService'
```
emilypi: Do you mind leaving a more explicit note?

test/Chainweb/Test/Pact/ModuleCacheOnRestart.hs
```diff
@@ -0,0 +1,159 @@
...
+      , after AllSucceed "initial-start" $
+        withPact' testVer Quiet pdb bhdb dir defaultReorgLimit (Check (iom >>= readMVar)) action (testCase "restart")
+
+      ]
+  where
+    label = "Chainweb.Test.Pact.ModuleCacheOnRestart"
+    genblock = genesisBlockHeader testVer testChainId
+    action = initialPayloadState dummyLogger testVer testChainId
+
+data R a = Supply !(IO (MVar a)) | Check (IO a)
```
emilypi: needs either docs or a more illuminating name

test/Chainweb/Test/Pact/ModuleCacheOnRestart.hs
emmanueldenloye: done!

test/Chainweb/Test/Pact/ModuleCacheOnRestart.hs
emmanueldenloye: done

src/Chainweb/Pact/PactService.hs
emilypi: Confirmed that it does not because `initPactService'` here is executing with separate state and returning `()`

PR #825, opened 2019-12-12
Increased strictness for pact service.

gregorycollins:


fosskers:
We've concluded that these changes don't have a performance impact.

PR #830, opened 2019-12-13
Upgrades and Remediations [branch off Master]

emilypi:
Continuation of #827


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -330,6 +339,58 @@ applyLocal logger dbEnv gasModel pd spv cmdIn mc =
...
+    testBlock1 | parentBlockHeight == 0 = go
+               | otherwise = noop
+    parentBlockHeight = _blockHeight parentHeader
+
+    testUpgradeTime t | isUpgradeBlockTime t = go
+                      | otherwise = noop
+    isUpgradeBlockTime upgradeTime =
+      parentTime < upgradeTime
+      &&
+      upgradeTime <= currCreationTime
```
emilypi: Aren't we guaranteed that `parentTime < currCreationTime`? If so, then shouldn't we just test that `upgradeTime <= currCreationTime`?  Also, I don't believe `runCoinbase` is called for genesis transactions, so we won't need to check against whether a particular parent header was at genesis as it's implied.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -272,16 +277,21 @@ applyCoinbase v logger dbEnv (Miner mid mks) reward@(ParsedDecimal d) pd ph
...
 
-          return $! CommandResult rk (_erTxId er) (PactResult (Right (last $ _erOutput er)))
-           (_erGas er) (Just $ _erLogs er) (_erExec er) Nothing
+          upgradedModuleCache <-
+            applyUpgrades v parentHeader currCreationTime
+          logs <- use txLogs
+
+          return $! (CommandResult rk (_erTxId er) (PactResult (Right (last $ _erOutput er)))
+                     (_erGas er) (Just $ logs) (_erExec er) Nothing
+                    ,upgradedModuleCache)
```
emilypi: Let's use `T2` here to be safe

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: @larskuhtz can you weigh in here? This is your proposed logic (as implemented by me)

slpopejoy: > If so, then shouldn't we just test that upgradeTime <= currCreationTime? 
We need just one block, not every block ever before currCreationTime

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: done

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: @emilypi I don't understand the second comment, which seems to refer to `PactService`

emilypi: @slpopejoy It is referring to `PactService`. `runCoinbase` in `PactService` will not run coinbases for genesis blocks, so every block that hits `applyCoinbase` will by definition be a non-genesis block. So rather than `parentBlockHeight == 0`, really what we're looking for is `currBlockHeight == 1`. 

I'm trying to make the argument that passing in a parent header here is not really going to contribute anything.

src/Chainweb/Version.hs
```diff
@@ -578,21 +580,22 @@ txEnabledDate TimedCPM{} = Nothing
 txEnabledDate FastTimedCPM{} = Nothing
 txEnabledDate Development = Just [timeMicrosQQ| 2019-11-30T05:00:00.0 |]
 txEnabledDate Testnet04 = Nothing
-txEnabledDate Mainnet01 = Just [timeMicrosQQ| 2019-12-17T01:00:00.0 |]
+txEnabledDate Mainnet01 = Just [timeMicrosQQ| 2019-12-17T15:30:00.0 |]
```
mightybyte: This looks good.  I just wanted to double check that we want this to be *before* the transferActivationDate.  @fosskers ?

src/Chainweb/Version.hs
```diff
@@ -578,21 +580,22 @@ txEnabledDate TimedCPM{} = Nothing
...
 transferActivationDate Test{} = Nothing
 transferActivationDate TimedConsensus{} = Nothing
 transferActivationDate PowConsensus{} = Nothing
 transferActivationDate TimedCPM{} = Nothing
 transferActivationDate FastTimedCPM{} = Nothing
-transferActivationDate Development = Just [timeMicrosQQ| 2019-11-30T07:00:00.0 |]
+transferActivationDate Development = Just [timeMicrosQQ| 2019-12-13T23:25:00.0 |]
 transferActivationDate Testnet04 = Nothing
-transferActivationDate Mainnet01 = Just [timeMicrosQQ| 2019-12-17T01:00:00.0 |]
+transferActivationDate Mainnet01 = Just [timeMicrosQQ| 2019-12-17T16:00:00.0 |]
```
mightybyte: I agree with this date.  We want this to be late morning NYC time so it will be a convenient time for us to be on top of any potential issues.

src/Chainweb/Version.hs
```diff
@@ -611,3 +614,15 @@ vuln797FixDate Mainnet01 = [timeMicrosQQ| 2019-12-10T21:00:00.0 |]
...
+-- | Must be BEFORE 'txEnabledDate' in mainnet.
+upgradeCoinV2Date :: ChainwebVersion -> Maybe (Time Micros)
+upgradeCoinV2Date Test{} = Nothing
+upgradeCoinV2Date TimedConsensus{} = Nothing
+upgradeCoinV2Date PowConsensus{} = Nothing
+upgradeCoinV2Date TimedCPM{} = Nothing
+upgradeCoinV2Date FastTimedCPM{} = Nothing
+upgradeCoinV2Date Development = Just [timeMicrosQQ| 2019-12-13T23:15:00.0 |]
+upgradeCoinV2Date Testnet04 = Nothing
+upgradeCoinV2Date Mainnet01 = Just [timeMicrosQQ| 2019-12-17T15:00:00.0 |]
```
mightybyte: This looks good too.

slpopejoy:
Leaving branch open for now

PR #831, opened 2019-12-13
Pact service: recover from fatal tx error in newBlock.

### LINKED ISSUE #829: Tx fails in newBlock BuyGas but isn't removed from mempool
It seems that certain tx can cause `newBlock` to fail but are not removed from the mempool. Instead it seems that they get gossiped around and included in each new block, thus block creation of any new blocks on all nodes.

* [ ] try to reproduce the issue and double check that the described symptoms are correct.
* [ ] Every tx that triggers an `PactInternalException` by calling `fatal` is banned from the mempool.

Some background:

> Lars 2:21 PM
> I see errors like this one:
> `Received exception running pact service (execNewBlock): {"tag":"PactInternalError","contents":"tx failure for requestKey when buying gas: (enforce-guard (at 'guard (rea...: Failure: Tx Failed: Keyset failure (keys-all): [caec74f8...]"}`
> Expected?
> `Chainweb.Miner.Miners.localPOW failed: {"tag":"PactInternalError","contents":"{\"tag\":\"PactInternalError\",\"contents\":\"tx failure for requestKey when buying gas: (enforce-guard (at 'guard (rea...: Failure: Tx Failed: Keyset failure (keys-all): [caec74f8...]\"}"}. Restarting ...`
> `critical transaction failure: "1La7kZdo-U0e2XYFnthIG62msapZ7YlKpV6aA_y73HY": tx failure for requestKey when buying gas: (enforce-guard (at 'guard (rea...: Failure: Tx Failed: Keyset failure (keys-all): [caec74f8...]`

> Lars 2:25 PM
> I see them on all nodes on an ongoing basis.

> Hee Kyun 2:37 PM
> nvm my earlier comment. I rotated the keyset to another key and back, which I confirmed was caec74f8... (edited) 
> The tx should have been rejected right away as a validation failure since test-sender was not > signed properly, but it wasn’t rejected right away and gave me back a requestKey

> Lars 2:43 PM
> So, that’s a bug? (edited) 

> Hee Kyun 2:45 PM
> Yeah, but other txs were rejected when sender wasn’t valid (edited) 

> Lars 2:33 PM
> I get 121,098 hits for the following log message in the last 10min
> `critical transaction failure: "1La7kZdo-U0e2XYFnthIG62msapZ7YlKpV6aA_y73HY": tx failure > for requestKey when buying gas: (enforce-guard (at 'guard (rea...: Failure: Tx Failed: Keyset failure (keys-all): [caec74f8...]`

> Linda 2:49 PM
> This might be a regression to back when buyGas was failing and causing the nodes to spin. >buyGas’s validation should have caught it

gregorycollins:
Sometimes we run the pre-block check on a set of transactions and it reports that they all have enough gas to be playable. However, sometimes a transaction in the middle of the block can cause a subsequent transaction to fail the buy-gas check. In this case, we were getting caught in a fatal loop, since the offending txs keep getting re-selected instead of being bad-listed as intended.

Fixes #829.

Changes:

  1. add a mempoolAddToBadList function to Mempool

  2. propagate it to PactService's MemPoolAccess

  3. add a new exception constructor to PactException that contains the failing
     tx hash, and one for coinbase exception

  4. modify TransactionExec to throw the new exception constructor on fatal
     error instead of "Internal error", and fixup the other caller to "fatal"
     to throw the coinbase exception instead

  5. catch the new constructor in `execNewBlock` and use it to bad-list the bad
     transaction so it isn't re-selected.

  6. test for new mempoolAddToBadList

  7. test for badlisting behavior inside PactService.


test/Chainweb/Test/Pact/ChainData.hs
```diff
@@ -228,12 +228,10 @@ data TestException = TestException
...
 
 testMemPoolAccess :: T.Text -> IO (Time Integer) -> MemPoolAccess
-testMemPoolAccess t iotime = MemPoolAccess
+testMemPoolAccess t iotime = mempty
     { mpaGetBlock = \validate bh hash _parentHeader -> do
         time <- f bh <$> iotime
         getTestBlock t time validate bh hash
-    , mpaSetLastHeader = \_ -> return ()
-    , mpaProcessFork = \_ -> return ()
     }
```
larskuhtz: I am not a fan of this kind of changes, since it makes reading the code more difficult, since there isn't is an obvious (at least to me) canonical Monoid instance for MempoolAccess.

test/Chainweb/Test/Mempool.hs
```diff
@@ -219,6 +220,35 @@ propBadlist (txs, badTxs) _ mempool = runExceptT $ do
...
+    block <- getBlock
+    when (block /= [tx]) $ fail "expected to get our tx back"
+
+    liftIO $ mempoolAddToBadList mempool $ hash tx
+    block' <- getBlock
+    when (block' /= []) $ fail "expected to get an empty block"
+    liftIO (lookup [tx]) >>= V.mapM_ lookupIsMissing
+    liftIO (insert [tx])
+    liftIO (lookup [tx]) >>= V.mapM_ lookupIsMissing
+
```
larskuhtz: Would it be possible to extend the test scenario to also check that the bad-list gets cleaned up and isn't growing forward?

src/Chainweb/Pact/PactService.hs
```diff
@@ -753,16 +753,22 @@ execNewBlock
...
+          setBlockData parentHeader
+          rewindTo newblockRewindLimit target
+          newTrans <- withCheckpointer target "preBlock" doPreBlock
+          withCheckpointer target "execNewBlock" (doNewBlock newTrans)
+
+    onTxFailure e@(PactTransactionExecError rk _) = do
+        -- add the failing transaction to the mempool bad list, so it is not
+        -- re-selected for mining.
+        liftIO $ mpaBadlistTx mpAccess rk
+        throwM e
```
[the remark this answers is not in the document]
gregorycollins: Yes, it's fine, I checked

test/Chainweb/Test/Pact/ChainData.hs
gregorycollins: This is test code -- it's mainly to avoid future updates to these regions if we add another field to `MemPoolAccess`, I've already had to go through them all several times and I'm tired of it. I can change this to `emptyMempoolAccess` with `emptyMempoolAccess = mempty` if you'd prefer, and I actually in principle agree with you. I'd make the same argument about `Default`

test/Chainweb/Test/Mempool.hs
gregorycollins: Maybe later, you're right that we aren't testing badmap pruning currently. Testing it will probably require deeper changes to some of the code (in particular to pass the clock around as an IO action for testing) so I'd like to do it later.

To answer your other question in case you're worried about this though, we do have telemetry on the size of the badmap so we'll be able to see if it is growing indefinitely.

test/Chainweb/Test/Pact/ChainData.hs
larskuhtz: I agree, for test code it doesn't matter that much and coding convenience is of higher value than in production code. Let's leaf it as is.

PR #834, opened 2019-12-14
Mining Work Stream

fosskers:


fosskers:
Buggy still.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -90,7 +91,7 @@ import Utils.Logging.Trace (trace)
 -- made as often as desired, without clogging the Pact queue.
 --
 newtype PrimedWork =
-    PrimedWork (HM.HashMap MinerId (HM.HashMap ChainId (Maybe CachedPayload)))
+    PrimedWork (HM.HashMap MinerId (HM.HashMap ChainId (TVar (Maybe CachedPayload))))
```
fosskers: Each leaf is now a `TVar`, so that the producer and consumer threads can each have fine-grained access to these chain-specific updates.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -99,7 +100,7 @@ type CachedPayload = T2 PayloadWithOutputs BlockCreationTime
 -- `publish`.
 --
 newtype MiningState = MiningState
-    (M.Map (T2 BlockCreationTime BlockPayloadHash) (T3 Miner PrevTime PayloadWithOutputs))
+    (M.Map (T2 BlockCreationTime BlockPayloadHash) (T3 MinerId PrevTime PayloadWithOutputs))
```
fosskers: The entire `Miner` structure was not needed, only the account name.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -109,8 +110,7 @@ data MiningStats = MiningStats
     { _statsCacheSize :: !Int
     , _stats503s :: !Int
     , _stats403s :: !Int
-    , _statsAvgTxs :: !Int
-    , _statsPrimedSize :: !Int }
```
fosskers: These stats aren't useful anymore.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -274,8 +275,8 @@ estimatedHashes (PrevTime p) b = floor $ (d % t) * 1000000
     d = case targetToDifficulty $ _blockTarget b of
         HashDifficulty (PowHashNat w) -> int w
 
-getAdjacentParents :: Cut -> BlockHeader -> Maybe BlockHashRecord
-getAdjacentParents c p = BlockHashRecord <$> newAdjHashes
+getAdjacentParents :: Cut -> ParentHeader -> Maybe BlockHashRecord
+getAdjacentParents c (ParentHeader p) = BlockHashRecord <$> newAdjHashes
```
fosskers: If we increase the usage of this pattern, we would likely reduce confusion in deep code locations where both the parent and child headers are passed around beside one another.

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -197,12 +200,60 @@ updatesHandler mr (ChainBytes cbytes) = Tagged $ \req respond -> withLimit respo
...
+        let prevTime = PrevTime . _blockCreationTime $ coerce p
+            cacheHdr = over _Unwrapped . M.insert (T2 bct phash) $ T3 mid prevTime pl
+        atomically $ modifyTVar' (_coordState mr) cacheHdr
+        atomically . writeTVar prv . Just $ poph pl
+        -- Encode and send the Header --
+        pure $ event header
+
+    event :: BlockHeader -> ServerEvent
+    event bh =
+        ServerEvent (Just $ fromByteString "New Work") Nothing [fromByteString $ BX.encode wb]
```
fosskers: Hex encoded, as the Event Stream must be ASCII.

src/Chainweb/Miner/RestAPI.hs
```diff
@@ -55,6 +55,10 @@ type MiningApi_ =
     :<|> "mining" :> "updates"
                   :> ReqBody '[OctetStream] ChainBytes
                   :> Raw
+    :<|> "mining" :> "stream"
+                  :> Capture "chain" ChainId
+                  :> Capture "miner" MinerId
+                  :> Raw
```
fosskers: Implicit `servant` behaviour: When no `ReqBody` is required here, `Raw` defaults to `GET`. If there is instead a `ReqBody`, it becomes `POST`.

fosskers:
I'll reopen if this becomes needed.

PR #838, opened 2019-12-14
Load default modules from pact db

slpopejoy:
Caveats:
- only loads `ns` and `coin`, whereas the previous solution also loaded `gas-payer-v1` and `fungible-v1`. Since this changes at upgrade, thought to go conservative for now. We can consider an `Interpreter`-based approach that pokes around more deeply, loading any module in the default namespace.
- Simple equality didn't work for testing the modules loaded from the PactDb vs loaded from genesis. However the module hashes match, so the test checks that.

### DIFF
--- a/src/Chainweb/Pact/TransactionExec.hs
+++ b/src/Chainweb/Pact/TransactionExec.hs
@@ -28,6 +28,7 @@ module Chainweb.Pact.TransactionExec
 , applyContinuation
 , applyContinuation'
 , runPayload
+, readInitModules
 
   -- * Gas Execution
 , buyGas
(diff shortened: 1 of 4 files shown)


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -340,6 +341,38 @@ applyLocal logger dbEnv gasModel pd spv cmdIn mc =
...
+    tenv = TransactionEnv Local dbEnv logger pd noSPVSupport nid 0.0
+           rk 0 permissiveExecutionConfig
+    txst = TransactionState mempty mempty 0 Nothing (_geGasModel freeGasEnv)
+    interp = defaultInterpreter
+
+    go :: TransactionM p ModuleCache
+    go = do
+      readExec <-
+        liftIO $ buildExecParsedCode Nothing
+        "coin.MINIMUM_PRECISION ns.GUARD_SUCCESS"
```
emilypi: ha. Clever. This should only summon `coin` and `ns` yeah?

slpopejoy: I don't know if it summons `fungible-v2` as Emmanuel's test didn't cover that

slpopejoy: But yes, just `coin` and `ns`.

PR #841, opened 2019-12-16
Supplemented Stale Solution Statistics

fosskers:
This PR adds a code to the `OrphanedBlock` log message, so that it's easier to discern programmatically what the cause of staling was.


fosskers:
No longer needed. The same information can be divined from the log messages.

PR #842, opened 2019-12-17
pact: poll: report when a transaction is on the bad list.

gregorycollins:


fosskers:
Soft approval - I'm going to test this on a node.

PR #845, opened 2019-12-18
More Stale Solution Data

fosskers:


src/Chainweb/Logging/Miner.hs
```diff
@@ -46,7 +46,10 @@ data NewMinedBlock = NewMinedBlock
 
 data OrphanedBlock = OrphanedBlock
     { _orphanedHeader :: !(ObjectEncoded BlockHeader)
+    , _orphanedBestOnCut :: !(ObjectEncoded BlockHeader)
```
fosskers: The block that this stale solution lost to.

PR #850, opened 2019-12-27
/local Endpoint Validations

emilypi:
Per today's discussion, this PR cases on whether `NetworkId` is defined, and if not, runs the transactions naively without reporting gas, and if it is, check gas. 

Also, it looked like some of our nondet tests needed longer polling timeouts, presumably because tx polling  (lookups in general) have gotten slower.


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -339,7 +343,69 @@ applyLocal logger dbEnv gasModel pd spv cmdIn mc =
...
+    -> NetworkId
+    -> TransactionM p (CommandResult [TxLog Value])
+validateLocalCmd v pscid now gas0 t cmd (NetworkId nid)
+    | chainIdToText pscid /= cid =
+      evalErr $ "invalid chain id: " <> cid
+    | not (timingsCheck bct cmd) =
+      evalErr "invalid ttl or creation time"
+    | (decimalPlaces gp) > 12 =
+      gasErr $ "gas price should be rounded to at most 12 places: " <> sshow gp
+    | fromText @CW.ChainwebVersion nid /= Just v =
```
emilypi: ChainId + Network Id vs. Chainweb Version + Sigcount not > 100  is checked in https://github.com/kadena-io/chainweb-node/blob/ba81dbd677ee9c40b8a5e7bda5d383534889fe70/src/Chainweb/Chainweb.hs#L535

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -339,7 +343,69 @@ applyLocal logger dbEnv gasModel pd spv cmdIn mc =
...
+    -> TransactionEnv p
+      -- ^ the tx environment
+    -> Command (Payload PublicMeta ParsedCode)
+      -- ^ user command
+    -> NetworkId
+    -> TransactionM p (CommandResult [TxLog Value])
+validateLocalCmd v pscid now gas0 t cmd (NetworkId nid)
+    | chainIdToText pscid /= cid =
+      evalErr $ "invalid chain id: " <> cid
+    | not (timingsCheck bct cmd) =
```
emilypi: This is the same time check as in https://github.com/kadena-io/chainweb-node/blob/4b6a2d330aec8c74567cc24721bf8ea085753e20/src/Chainweb/Pact/PactService.hs#L709

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -339,7 +343,69 @@ applyLocal logger dbEnv gasModel pd spv cmdIn mc =
...
+      evalErr "invalid ttl or creation time"
+    | (decimalPlaces gp) > 12 =
```
emilypi: This check is usually performed in mempool here: https://github.com/kadena-io/chainweb-node/blob/8bc47a59f9a3e7fa33e6c3269058d769a50a206e/src/Chainweb/Mempool/InMem.hs#L318

src/Chainweb/Pact/TransactionExec.hs
gregorycollins: We should not duplicate check code in this way -- let's unify these

emilypi:
@gregorycollins @slpopejoy i've consolidated our validations checks in a single module.

PR #852, opened 2019-12-28
Enrich CommandResult metadata 

slpopejoy:
Forking change to add block height, block time, and prev block hash.


mightybyte:
Can this be closed now?

PR #857, opened 2020-01-02
 Swagger Json does not include /cut endpoint #800 

### LINKED ISSUE #800: Swagger Json does not include /cut endpoint
For some reason `/swagger.json` does not include `/chainweb/0.0/mainnet01/cut/` endpoint.
Node version: 1.2

emmanueldenloye:
This addresses the concerns listed in issue #800 .

- [x] add the missing /cut endpoint

- [x] clean up redundant endpoints for every chain

closes #800


mightybyte:
I think this should be good to merge.  However, Chainweaver currently uses `/swagger.json` to get info about a node, so I think we should hold off merging this until Chainweaver switches to using `/info`.  Once that happens, we should do one more round of testing and then merge.

fosskers:
Reiterating that we need to test this against the newest version of Chainweaver.

PR #860, opened 2020-01-03
create minimal spv proofs by default

### LINKED ISSUE #854: Fix SPV proof size explosion
Coins transferred from ea564a0a040f74650f03f140c34980388eb183612bf3f09fd2f0b84accabc9fe on chain 3 to 3f00302f6a44308a94268a9c4ac5e89e08a01570d2a8c510371c9cbd9f10a868 on chain 8.

larskuhtz:
closes #854

### DIFF
--- a/test/Chainweb/Test/SPV.hs
+++ b/test/Chainweb/Test/SPV.hs
@@ -23,15 +23,27 @@ module Chainweb.Test.SPV
 , apiTests
 ) where
 
-import Control.Lens ((^?!))
+import Control.Lens ((^?!), view)
+import Control.Monad.Catch
+import Control.Monad.IO.Class
 
 import Data.Aeson
+import qualified Data.ByteString as B
+import qualified Data.ByteString.Lazy as BL
+import Data.CAS
 import Data.Foldable
+import Data.Functor.Of
+import qualified Data.List as L
 import Data.Reflection hiding (int)
+import qualified Data.Vector.Unboxed as V
 
 import Numeric.Natural
 
 import Servant.Client
+
+import Statistics.Regression
+
+import qualified Streaming.Prelude as S
 
 import Test.QuickCheck
 import Test.Tasty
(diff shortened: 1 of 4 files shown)


test/Chainweb/Test/SPV.hs
```diff
@@ -101,6 +118,140 @@ targetChain c srcBlock = do
...
+                    , int $ B.length (_transactionBytes tx)
+                    ]
+
+        isReachable <- reachable cutDb h trgChain
+        try inner >>= \case
+            Right x -> do
+                assertBool "SPV proof creation succeeded although target chain is not reachable" isReachable
+                return (Just x)
+            Left SpvExceptionTargetNotReachable{} -> do
+                assertBool "SPV proof creation failed although target chain is reachable" (not isReachable)
```
slpopejoy: Is this a success case? It's not clear why this would fail the test

test/Chainweb/Test/SPV.hs
```diff
@@ -101,6 +118,140 @@ targetChain c srcBlock = do
...
+    reachable cutDb h trgChain = do
+        m <- maxRank $ cutDb ^?! cutDbBlockHeaderDb trgChain
+        return $ (int m - int (_blockHeight h)) >= (distance cutDb h trgChain)
+
+    -- regression model with @createTransactionProof@. Proof size doesn't depend on target height.
+    --
+    regress r
+        | [proofSize, blockSize, _heightDiff, chainDist, txSize] <- V.fromList <$> L.transpose r
+            = olsRegress [V.map (logBase 2) blockSize, chainDist, txSize] proofSize
+        | otherwise = error "Chainweb.Test.SPV.spvTest.regress: fail to match regressor list. This is a bug in the test code."
```
slpopejoy: if there is any chance a maintainer can screw this up, it should be an `assertionFailed` (I can't remember the exact name) with output that gives a clue what went wrong, same for `regressWithHeightDiff`

test/Chainweb/Test/SPV.hs
slpopejoy: also, again, if there's a chance a maintainer can screw this up, please have the `assertXXX` message output relevant data

test/Chainweb/Test/SPV.hs
larskuhtz: If this fails, it's about the test itself and not about the tested code. 

It's really just about preventing an incomplete pattern match warning in the code for computing the statistics. The regressor value rows are collected in a list (instead of a tuple) because that's what `L.transpose` expects.

test/Chainweb/Test/SPV.hs
larskuhtz: The exception may be thrown in the tested code, so it's part of the tested `createTransactionProof` API. The assertion checks that the exception is only thrown when it is expected.

I can add the height difference and chain distance data to the failure message, which determines if a proof exists or not.

larskuhtz:
There are already existing tests for both transaction and output proofs. The implementation for both is very similar, so I think, it is enough to do these additional tests just for one case. I can switch the code to use outputs instead of transactions.

PR #863, opened 2020-01-06
Deletes blockheight log in header-dump tool

LindaOrtega:
We're currently logging the block height periodically in the header-dump tool. This makes it difficult to transform these header outputs into useful JSON values using command line tools like `grep` and `jq`.


larskuhtz:
Just use `--log-level=quiet`

Or, alternatively, you can use `--logger-backend-handle=stderr`, in which case the logs would go to stderr while the regular output goes to stdin (and to the pipe to `jq`).

LindaOrtega:
Closing because running the following works: 
```
cwtool header-dump --chain-id 0 -d ./rocksDb/ -o "transaction" --log-level=quiet | jq
```

PR #865, opened 2020-01-07
Additional Repl tool commands for new module testing, etc.

marklnichols:


chainweb.cabal
```diff
@@ -743,6 +743,7 @@ executable cwtool
         , filepath
         , errors >= 2.3
         , exceptions >= 0.8
+        , extra >= 1.6
```
mightybyte: Can we get rid of this dependency?  I'd rather not add someone's personal grab-bag of random stuff into our dependency tree.  I'm fine with inlining whatever function we're using.

marklnichols: That one is Neil Mitchell's extra library -- it's very useful and not quite so random.  But I can if you want...

mightybyte: Now that we've launched I think we should avoid new deps without a really compelling reason and this one doesn't quite seem to make the bar in my mind.

marklnichols: I was just going to take it out, but I just noticed its already a dependency in quite a few modules.  My addition was just for the cwtool executable.

PR #868, opened 2020-01-08
Enable User Contracts + Bump Killswitch

### LINKED ISSUE #856: Enable modules, update kill switch dates, etc
Leaning towards keeping a kill switch in place.  How far we push it out is still TBD.

Proposed date: March 15

emilypi:
This PR bumps the killswitch on nodes to `2020-02-15-00T00:00:00.0`, and establishes turning on user contracts for Mainnet and Devnet at `2020-01-15-00T16:00:00.0`. 

closes #856


node/ChainwebNode.hs
```diff
@@ -496,7 +496,7 @@ pkgInfoScopes =
 -- KILLSWITCH for version 1.4
 --
 killSwitchDate :: Maybe String
-killSwitchDate = Just "2020-01-15T00:00:00Z"
+killSwitchDate = Just "2020-02-15T00:00:00Z"
```
mightybyte: February 15 is a Saturday.  Should we move this to a Wednesday like we've been doing?  @slpopejoy

mightybyte: How about Feb 13 or Feb 20?

PR #866, opened 2020-01-08
Enrich CommandResult in poll/listen

slpopejoy:
closes #805

PR #883, opened 2020-01-14
Fix SPV Test Nondeterminism

emilypi:
The mining calculations were wrong - this makes sure that enough blocks are mined between phases of the SPV txs so that consensus is achieved. `RemotePactTest` bumped its timeout backoff rate in #866, which should fix the polling timeouts that were contributing to the tests failing. 

Unfortunately, I have no clue how to fix the `PactReplay` tests. It looks like a wait time problem.

### DIFF
--- a/.github/workflows/applications.yml
+++ b/.github/workflows/applications.yml
@@ -135,7 +135,7 @@ jobs:
     - name: Build
       run: cabal v2-build chainweb-node
     - name: Run Tests
-      run: cabal v2-run chainweb-tests -- --hide-successes --pattern='!/remote pact tests/ && !/Chainweb.Test.Pact.SPV/'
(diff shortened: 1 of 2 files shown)


.github/workflows/applications.yml
```diff
@@ -135,7 +135,7 @@ jobs:
     - name: Build
       run: cabal v2-build chainweb-node
     - name: Run Tests
-      run: cabal v2-run chainweb-tests -- --hide-successes --pattern='!/remote pact tests/ && !/Chainweb.Test.Pact.SPV/'
```
emilypi: @larskuhtz the remote pact tests were fixed with the latest PR bumping the retry rate to accommodate for the increase in polling time.

PR #886, opened 2020-01-16
improve header-dump tool

larskuhtz:
* [x] when no chain is provided it will dump data from all chains (instead of picking a random one)
* [x] include chain-id with miner data
* [x] use the lens pattern in stream processor for extracting miner data


mightybyte:
Looks good at a quick glance.

Issue #890, opened 2020-01-22
[Validation Error] Report why tx failed the buy gas attempt


LindaOrtega:
Potentially related to #753 

Users often get 
```
"Validation failed for hash "6tpv9gbfFuEK57XDzB07OZuUGF1ngY3helQYnDpbpy0": Sender account has insufficient gas." 
```
when they submit a transaction that fails a dry-run attempt to buy gas.

We should improve this error message to explain _why_ the transaction failed the buy gas check.

Relevant code sections:
Error message specified.
https://github.com/kadena-io/chainweb-node/blob/master/src/Chainweb/Pact/PactService.hs#L675

Error type thrown in `attemptBuyGas` function.
https://github.com/kadena-io/chainweb-node/blob/master/src/Chainweb/Mempool/Mempool.hs#L216

PR #893, opened 2020-01-23
More detailed errors when attempt buy gas fails

LindaOrtega:
Closes https://github.com/kadena-io/chainweb-node/issues/890


src/Chainweb/Pact/PactService.hs
```diff
@@ -88,6 +88,7 @@ import Pact.Compile (compileExps)
 import qualified Pact.Gas as P
 import Pact.Gas.Table
 import qualified Pact.Interpreter as P
+import Pact.Types.PactError()
```
slpopejoy: necessary?

LindaOrtega: deleted it

PR #894, opened 2020-01-28
Infrastructure for chainging the hash function

larskuhtz:


src/Chainweb/BlockHeader.hs
```diff
@@ -104,6 +100,10 @@ module Chainweb.BlockHeader
 , blockFlags
 , _blockPow
 , blockPow
+, _blockPowHashAlg
+, blockPowHashAlg
+, _blockPowMultiplyer
```
larskuhtz:
```suggestion
, _blockPowMultiplier
```

PR #616, opened 2019-10-30
Try to fix mempool flaky test

gregorycollins:
- generate fake transactions with appropriate metadata
  - don't check inserts on highwater test

PR #895, opened 2020-01-30
High Water Test

fosskers:
Closes #616


fosskers:
I ran this through CI several times, and it never failed. Perhaps the issue has solved itself.

larskuhtz:
Should we also try to run through github-ci? Sometimes timing is different for the ubuntu builds.

fosskers:
I ran the tests locally with `stack` many times and had no problems.

PR #901, opened 2020-02-05
PowHeader Endpoint

fosskers:
This PR adds a new endpoint which yields JSON-encoded `BlockHeader`s along with their POW hash, for usage with external tools.

### DIFF
--- a/src/Chainweb/BlockHeaderDB/RestAPI.hs
+++ b/src/Chainweb/BlockHeaderDB/RestAPI.hs
@@ -72,6 +72,11 @@ module Chainweb.BlockHeaderDB.RestAPI
 , headerStreamApi
 , someHeaderStreamApi
 
+-- * PowHeader API
+, PowHeader(..)
+, PowHeaderApi
+, powHeaderApi
+
 -- * Sub APIs
 , BranchHashesApi
 , branchHashesApi
@@ -87,7 +92,7 @@ module Chainweb.BlockHeaderDB.RestAPI
 , hashesApi
 ) where
 
-import Control.Monad.Identity
+import Control.Lens hiding ((.=))
 
 import Data.Aeson
 import Data.Bifunctor
@@ -95,6 +100,7 @@ import Data.Bytes.Get
 import Data.Bytes.Put
 import qualified Data.ByteString.Lazy as BL
 import Data.Proxy
+import Data.Swagger
 import Data.Text (Text)
 
 import Network.HTTP.Media ((//), (/:))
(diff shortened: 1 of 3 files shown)


fosskers:
Works as intended and seems to have next to no performance overhead. Tested by this PR: https://github.com/kadena-io/chainweb-data/pull/6

larskuhtz:
There is already an endpoint for JSON encoded block header. I don't think we should overload the consensus node API with endpoints that provide special encodings for particular tools.

I think, providing those endpoints is the job of the chainweb-data API.

larskuhtz:
I think, before we add more APIs we should take some time and do a proper API design. Otherwise we risk ending up with a kitchen sink API with lots of overlapping functionality. That would be both confusing for users and hard to maintain for devs.

fosskers:
This endpoint adds the POW hash, which is missing from other endpoints. This isn't an arbitrary encoding.

src/Chainweb/BlockHeaderDB/RestAPI.hs
```diff
@@ -360,6 +366,7 @@ type BlockHeaderDbApi v c
     :<|> HeaderPutApi v c
     :<|> BranchHashesApi v c
     :<|> BranchHeadersApi v c
+    :<|> PowHeaderApi v c
```
larskuhtz: If we want to return the pow hash (which I see no value doing), I think, it would make more sense to always include it in the object encoded header. Adding a query parameter for it would be an alternative, but it has the disadvantage that it would apply only to the object encoding.

src/Chainweb/BlockHeaderDB/RestAPI.hs
```diff
@@ -408,3 +415,47 @@ headerStreamApi = Proxy
...
+    declareNamedSchema _ = return $ NamedSchema (Just "PowHeader") $ mempty
+        & title ?~ "POW Header"
+        & description ?~ "BlockHeader with its POW Hash"
+        & type_ .~ Just SwaggerObject
+        & required .~ [ "header", "powHash" ]
+
+type PowHeaderApi_
+    = "header"
+    :> Capture "BlockHash" BlockHash
+    :> "pow"
```
larskuhtz: I think, by making `pow` a resource on its own, one would expect to get only the pow hash without the header here. 

Generally, I don't think that the pow hash deserves to be a resource on its own in the consensus API. This hash is really is just a sequence of random bytes that has no purpose out side pow target validation.

fosskers:
No longer needed.

PR #902, opened 2020-02-07
[WIP] Consistently use blockCreationTime of parent as reference for tx processing

larskuhtz:
The current status is that the existing mainnet history validates.

This is work in progress.

* [x] Remove `creationTime` parameter from `newBlock` calls
* [x] Consistently use `blockCreationTime` of `parentHeader` as reference time for all tx processing
    * [x] for tx times validation
    * [x] for fork times
* [x] Simplify some definitions of past chainweb forks by using the actual block heights
* [x] make sure existing history validates
    * [x] add grace period of 5min to validation of tx creation time. Ensures that
        - existing tx are enabled with respect to the blockCreationTime of parent header
        - new tx can be included in a block without back-dating even if created after the parent
           header was already mined.

Test:
* [x] Improve some test messages
* [x] Add `InternalPactServiceException` for adding some additional debugging information to
        internal pact service failures (where "internal" means, that its a bug in the code).
* [x] All CI test suites pass
* [x] Validate existing `mainnet01` history


src/Chainweb/Version.hs
```diff
@@ -615,14 +615,31 @@ vuln797FixDate Testnet04 = epoch
...
+--
+-- This function provides the block heights when coin v2 became effective on the
+-- respective chains.
+--
+coinV2Upgrade
+    :: (Ord a, Num a)
+    => ChainwebVersion
+    -> ChainId
+    -> a
+        -- ^ FIXME this should be BlockHeight
```
larskuhtz: I have another draft PR that moves `BlockHeight` into a module on its own, which is a prerequisite for this.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -261,13 +258,26 @@ applyCoinbase v logger dbEnv (Miner mid mks) reward@(ParsedDecimal d) pd parentH
...
+    bh = fromIntegral $ _pdBlockHeight pd
+
+    cid = V._chainId parentHeader
+        -- FIXME it should hold that
+        -- @cid == unsafeFromText $ P._chainId $ _pmChainId $ _pdPublicMeta pd@
+        -- but in some unit test runs that is not the case. The chain id in public data is empty.
+
+    chash = Pact.Hash $ encodeToByteString $ _blockHash parentHeader
+        -- FIXME ceck that @ _pdPrevBlockHash pd == encode _blockHash@
+        -- FIXME chash includes the quoted text of the parent header. Is that correct?
```
larskuhtz: @emilypi do you have thoughts on these FIXME's? Can we do some cleanup here or add some clarifying comments, or just leave it as is?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -261,13 +258,26 @@ applyCoinbase v logger dbEnv (Miner mid mks) reward@(ParsedDecimal d) pd parentH
...
 
     tenv = TransactionEnv Transactional dbEnv logger pd noSPVSupport
            Nothing 0.0 rk 0 restrictiveExecutionConfig
     txst = TransactionState mc mempty 0 Nothing (_geGasModel freeGasEnv)
     initState = setModuleCache mc $ initCapabilities [magic_COINBASE]
-    chash = Pact.Hash (sshow $ _blockHash parentHeader)
```
larskuhtz: `show` on `BlockHash` returns the encoding as JSON string, which includes `"` characters. Is that indented? I guess, even if it unintentional, it's probably too late (and not worth the hassle) to change at this point?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -261,13 +258,26 @@ applyCoinbase v logger dbEnv (Miner mid mks) reward@(ParsedDecimal d) pd parentH
...
     blockTime = blockTimeOf pd
+        -- FIXME it should hold that @blockTime == _blockCreationTime parentHeader@, but
+        -- in some unit test runs that is not the case
+
+    bh = fromIntegral $ _pdBlockHeight pd
+
+    cid = V._chainId parentHeader
+        -- FIXME it should hold that
+        -- @cid == unsafeFromText $ P._chainId $ _pmChainId $ _pdPublicMeta pd@
+        -- but in some unit test runs that is not the case. The chain id in public data is empty.
```
emilypi: Up until now, we have being using a default value for public metadata for coinbase transactions. We can fix this in the upcoming hardfork, but not much is gained, since it is a miner-only transaction, and will never make use of any metadata values.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -261,13 +258,26 @@ applyCoinbase v logger dbEnv (Miner mid mks) reward@(ParsedDecimal d) pd parentH
     forkTime = vuln797FixDate v
     fork1_3InEffect = blockTime >= forkTime
     throwCritical = fork1_3InEffect || enfCBFailure
+
     blockTime = blockTimeOf pd
+        -- FIXME it should hold that @blockTime == _blockCreationTime parentHeader@, but
+        -- in some unit test runs that is not the case
```
emilypi: The Fixme here is correct. Let's check the unit tests

src/Chainweb/Pact/TransactionExec.hs
emilypi: This fix me is corret

src/Chainweb/Pact/Utils.hs
```diff
@@ -70,15 +72,22 @@ maxTTL = ParsedInteger $ 2 * 24 * 60 * 60
...
+    && toMicrosFromSeconds txOriginationTime < (blockOriginationTime .+^ graceTime)
     && toMicrosFromSeconds (txOriginationTime + ttl) >= blockOriginationTime
     && ttl <= maxTTL
   where
     (TTLSeconds ttl) = timeToLiveOf tx
     toMicrosFromSeconds = Time . TimeSpan . Micros . fromIntegral . (1000000 *)
     (TxCreationTime txOriginationTime) = creationTimeOf tx
+
+    -- FIXME apply this only to old blocks?
+    graceTime = TimeSpan (Micros 300000000)
```
larskuhtz: Should we remove this going forward?

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: No, this is a bug we should fix. Can you file an issue? We need to fork it unfortunately.

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: Agree, let's standardize on parentBlock and stop using `PublicData` altogether, and fix the tests

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: If we stop using PublicData this problem goes away correct?

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: I believe the change here addresses both FIXMEs doesn't it?

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: As @emilypi notes, this is coinbase so this is unused. We can fix

src/Chainweb/Pact/Utils.hs
```diff
@@ -70,15 +73,22 @@ maxTTL = ParsedInteger $ 2 * 24 * 60 * 60
...
-    && toMicrosFromSeconds txOriginationTime < blockOriginationTime
+    && toMicrosFromSeconds txOriginationTime < (blockOriginationTime .+^ graceTime)
     && toMicrosFromSeconds (txOriginationTime + ttl) >= blockOriginationTime
     && ttl <= maxTTL
   where
     (TTLSeconds ttl) = timeToLiveOf tx
     toMicrosFromSeconds = Time . TimeSpan . Micros . fromIntegral . (1000000 *)
     (TxCreationTime txOriginationTime) = creationTimeOf tx
+
+    -- TODO apply this only to old blocks?
```
slpopejoy: Yes please

src/Chainweb/Pact/PactService.hs
```diff
@@ -329,7 +349,10 @@ initializeCoinContract _logger v cid pwo = do
         (Just !p) -> return p
       let target = Just (succ bhe, bhash)
       bhdb <- asks _psBlockHeaderDb
-      parentHeader <- liftIO $! lookupM bhdb bhash
+      parentHeader <- liftIO $! lookupM bhdb bhash `catch` \e -> throwM $ InternalPactServiceException
```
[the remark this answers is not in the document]
larskuhtz: That change is part of a follow up PR.

src/Chainweb/Pact/PactService.hs
```diff
@@ -1000,6 +1023,10 @@ execLocal cmd = withDiscardedBatch $ do
     bhDb <- asks _psBlockHeaderDb
 
     parentHeader <- liftIO $! lookupM bhDb bhash
+      `catch` \e -> throwM $ InternalPactServiceException
```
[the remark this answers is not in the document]
larskuhtz: done in an already existing follow up PR

src/Chainweb/Pact/PactService.hs
```diff
@@ -1044,19 +1071,35 @@ setBlockData bh = do
...
     cp <- getCheckpointer
-    let creationTime = _blockCreationTime currHeader
+
+    -- FIXME don't do this twice
+    bhDb <- asks _psBlockHeaderDb
+    parentCreationTime <- if isGenesisBlockHeader currHeader
+      then
+        return $! _blockCreationTime currHeader
+      else do
+        ph <- liftIO $! lookupM bhDb (_blockParent currHeader)
```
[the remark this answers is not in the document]
larskuhtz: see comment above

src/Chainweb/Pact/PactService.hs
```diff
@@ -1044,19 +1071,35 @@ setBlockData bh = do
...
+          `catch` \e -> throwM $ InternalPactServiceException
+            { _pactServiceExceptionMsg = "failed lookup of parent header in playOneBlock"
+            , _pactServiceExceptionInner = Just e
+            }
+        return $! _blockCreationTime ph
+
     -- prop_tx_ttl_validate
     oks <- withEnableUserContracts currHeader $ do
```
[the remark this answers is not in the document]
larskuhtz: Also addressed in the next PR

src/Chainweb/Pact/TransactionExec.hs
emilypi: Yes

slpopejoy:
Closing but not deleting branch

PR #903, opened 2020-02-08
Add configuration options for checking payload hashes during pact replay

larskuhtz:
Normally, payload hashes are validated only for blocks for which validation is requested, but not for possible checkpointer replays. Normally that is fine because normally we trust the existing databases. But sometimes we want to be extra pedantic.

Moreover, when rebuilding a pact database from an existing chain database (or each time we start with a pact db that is "behind" the latest cut of the chaindb), the pact does an initial database sync during which the payload hashes must be revalidated. 

Ideally, we would always enable pedantic hash validation during the initial sync and possibly disable it afterwards. But for that we would have to extend the pact service API to support enabling and disabling pedantic checks. Which which would be a large change that would involve some API design choices. It's something we may consider for a follow up PR. 

### Follow-Up PRs (for a post February 20 release)

1. Collect all pact-service/checkpointer related configuration options in a subsection in the configuration file

2. Consider allowing to enable and disable the `_psValidateHashesOnReplay` in pact service and enable it during initial sync and disable it after that.

### DIFF
--- a/test/Chainweb/Test/TreeDB.hs
+++ b/test/Chainweb/Test/TreeDB.hs
@@ -44,9 +44,9 @@ import Chainweb.Utils (len)
 
 treeDbInvariants
     :: (TreeDb db, IsBlockHeader (DbEntry db), Ord (DbEntry db), Ord (DbKey db))
-    -- | Given a generic entry, should yield a database for testing, and then
-    -- safely close it after use.
     => (DbEntry db -> (db -> IO Bool) -> IO Bool)
+        -- ^ Given a generic entry, should yield a database for testing, and then
+        -- safely close it after use.
     -> RunStyle
     -> TestTree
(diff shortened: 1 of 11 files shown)


test/Chainweb/Test/TreeDB.hs
```diff
@@ -44,9 +44,9 @@ import Chainweb.Utils (len)
 
 treeDbInvariants
     :: (TreeDb db, IsBlockHeader (DbEntry db), Ord (DbEntry db), Ord (DbKey db))
-    -- | Given a generic entry, should yield a database for testing, and then
-    -- safely close it after use.
     => (DbEntry db -> (db -> IO Bool) -> IO Bool)
+        -- ^ Given a generic entry, should yield a database for testing, and then
+        -- safely close it after use.
```
larskuhtz: This change is unrelated, but without it `ghci` fails to parse this module on my laptop ...

PR #907, opened 2020-02-09
Cleanup module dependencies

larskuhtz:
This cleans up the module dependency graph by moving some types into modules on their own.

This is preparatory work for some upcoming feature PRs.

### DIFF
--- a/src/Chainweb/BlockHeight.hs
+++ b/src/Chainweb/BlockHeight.hs
@@ -1,0 +1,79 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- |
+-- Module: Chainweb.BlockHeight
+-- Copyright: Copyright © 2020 Kadena LLC.
+-- License: MIT
+-- Maintainer: Lars Kuhtz <lars@kadena.io>
+-- Stability: experimental
+--
+-- TODO
+--
+module Chainweb.BlockHeight
+(
+-- * Block Height
+  BlockHeight(..)
+, encodeBlockHeight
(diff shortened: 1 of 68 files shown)


src/Chainweb/Payload.hs
```diff
@@ -488,12 +483,22 @@ coinbaseOutputFromText t = either (throwM . TextFormatException . sshow) return
...
+        ]
+    , "reqKey" .= ("DldRwCblQ7Loqy6wYJnaodHl30d3j3eH-qtFzfEv46g" :: String)
+        -- this is the unique hash value define in @Pact.Types.Hash.initialHash@
+    , "logs" .= Null
+    , "metaData" .= Null
+    , "continuation" .= Null
+    , "txId" .= Null
+    ]
+{-# NOINLINE noCoinbaseOutput #-}
+
```
fosskers: Why did this change?

larskuhtz: `Chainweb.Payload` was supposed to not depend on pact internals. `noCoinBaseOutput` is a constant that will never change. So it's fine to directly provide the content and avoiding  using constructors from pact types. The PR adds a unit test to ensure that both are in sync.

This simplifies the module graph a lot.

larskuhtz: Oh, and this value is only used in test-version of chainweb. It's not used in production.

src/Chainweb/BlockHeight.hs
```diff
@@ -0,0 +1,79 @@
...
+encodeBlockHeight (BlockHeight h) = putWord64le h
+
+decodeBlockHeight :: MonadGet m => m BlockHeight
+decodeBlockHeight = BlockHeight <$> getWord64le
+
+encodeBlockHeightBe :: MonadPut m => BlockHeight -> m ()
+encodeBlockHeightBe (BlockHeight r) = putWord64be r
+
+decodeBlockHeightBe :: MonadGet m => m BlockHeight
+decodeBlockHeightBe = BlockHeight <$> getWord64be
```
emilypi: can we a get a note describing under what circumstances we need big and little endian? Also, while we're shuffling things around, the convention would be to disambiguate with `encodeBlockHeight` with `encodeBlockHeightLe` to contrast the other encoding (ditto for decoding)

src/Chainweb/Payload.hs
emilypi: Yes, thank you. The pact dependency here should definitely go away.

src/Chainweb/BlockHeight.hs
larskuhtz: Good point. I think, we don't use bigendian anywhere any more. I think we can just remove this function.

Issue #899, opened 2020-02-04
Updates and fixes to GAS_PAYER


ggobugi27:
1. BUG: `(read-msg)` currently contains`miner-keyset` and `total` field. This should be removed.

2.  Render useful info about the transaction as data that can be read using`(read-msg)` from `GAS_PAYER`
     - For regular pact commands: `pact code` 
     - For cont commands: `is-cont` (or a bool that specifies if the tx is a cont),  `step`, `pact-id`, `pact code` that will actually run

3. Charge gas for computation inside `GAS_PAYER`. 

4. Implement a `gas-payer-v2` or `gas-payer-redeem` interface to allow gas redemption once GAS_PAYER gets charged for computation

ggobugi27:
Resolved with 1.6(https://github.com/kadena-io/chainweb-node/pull/925) release

PR #912, opened 2020-02-12
Adds enriched message body to gas payer cap

LindaOrtega:
Addresses #1 and #2 in https://github.com/kadena-io/chainweb-node/issues/899


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -627,14 +632,48 @@ findPayer = runMaybeT $ do
...
+        evalCap i CapCallStack False $ mkApp i capRef as
+
       case ar of
         NewlyAcquired -> do
           r <- input
           popCapStack (const (return ()))
           return r
         _ -> evalError' i "Internal error, GAS_PAYER already acquired"
 
+enrichedMsgBody :: Command (Payload PublicMeta ParsedCode) -> Eval e Value
```
slpopejoy: This looks like it can be pure no?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -627,14 +632,48 @@ findPayer = runMaybeT $ do
...
+enrichedMsgBody cmd = case (_pPayload $ _cmdPayload cmd) of
+  Exec (ExecMsg (ParsedCode _ exps) userData) -> return $
+    object [ "tx-type" A..= ( "exec" :: Text)
+           , "exec-code" A..= map renderCompactText exps --TODO should we do both?
```
slpopejoy: I'm going to say just this and not `exec-exprs`.

`exec-exprs` is very well done so I regret ditching it, the justification is however that it's actually too low-level, because `Exp` is so low-level ~, so doing something as simple as matching `(coin.transfer "foo" "bar" 10.0)` will mean checking~

~1) `"list"` type 
2) first element is `(` separator 
3) 2nd is `"coin"` literal
4) then `.` separator
5) then `"transfer"` literal~ EDIT I since saw how it's done in the tests, it's not this bad

Note that this is brittle, in that the expressions `(use coin) (transfer "foo" "bar" 10.0)` is the same but will fail the above. 

Going with just code means the gas-payer will have to provide explicit instructions about code requirements, ie "you must directly reference calls to `coin.transfer` to qualify for gas payment". So let's start with just that.

EDIT it's not as bad as above, but forcing gas-payer to just compare strings is cheaper and less powerful which I think is actually good. If somebody submits a module to a gas payer the Exp JSON will be huge.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -627,14 +632,48 @@ findPayer = runMaybeT $ do
...
+        object [ "type" A..= ("literal" :: Text)
+               , "expr" A..= toJSON l ]
+      EAtom ae ->
+        object [ "type" A..= ("atom" :: Text)
+               , "expr" A..= (toJSON $ renderCompactText ae) ]
+      EList (ListExp li _ _) ->
+        object [ "type" A..= ("list" :: Text)
+               , "expr" A..= (toJSON (map renderExp li)) ]
+      ESeparator se ->
+        object [ "type" A..= ("seperator" :: Text)
```
slpopejoy:
```suggestion
        object [ "type" A..= ("separator" :: Text)
```

src/Chainweb/Pact/TransactionExec.hs
LindaOrtega: Yep, the `Eval` was a left over from when I was attempting to read from the state. Fixed in the upcoming commits.

src/Chainweb/Pact/TransactionExec.hs
LindaOrtega: yeah just having the code simplifies things. I fix it in the latest commit

PR #913, opened 2020-02-13
Simplify Forking Logic

emilypi:
This PR was requested to make forking logic explicit and to simplify our blocktime comparisons

N.B. There are no changes made to the vuln797 code in `TransactionExec.hs` because the tests break due to now depending on the blocktime of the parentheader instead of the public data.

I've left the code in so that we can incorporate it into future refactors and write a new test as a separate PR in the future

### DIFF
--- a/src/Chainweb/Pact/Backend/Types.hs
+++ b/src/Chainweb/Pact/Backend/Types.hs
@@ -75,6 +75,7 @@ module Chainweb.Pact.Backend.Types
     , MemPoolAccess(..)
 
     , PactServiceException(..)
+    , WithCheckpointerResult(..)
     ) where
 
 import Control.Exception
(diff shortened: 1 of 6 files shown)


src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -327,3 +328,7 @@ instance Show PactServiceException where
              ]
 
 instance Exception PactServiceException
+
+data WithCheckpointerResult a
+    = Discard !a
+    | Save BlockHeader !a
```
emilypi: Moved from `PactService.hs` - we had a backend datatype declared in the service code.

emilypi:
Closed to branch of master

PR #915, opened 2020-02-14
Explicit Forking (Redux)

emilypi:


slpopejoy:
My hopes that this stuff was going to get simpler have vanished, plus a lot of the code in question is changed now. Closing

PR #919, opened 2020-02-15
grandfather non-zero feature flags for block heights < 340000

larskuhtz:
Currently master is broken for mainet01. We must merge either this PR or #920 to unbreak it.

This is a technical non-user facing fork.  It affects only miners which would have to migrate "buggy" miners before that block height.

We therefor avoid the extra complexity involved with basing it on a date and use a block height instead -- which is also easier to handle for miners.

The block height is chosen to give miners about 5 days after the release of 1.6 to adapt their mining clients as needed. It is also expected to occur on a Monday during US office hours.

### DIFF
--- a/src/Chainweb/BlockHeader.hs
+++ b/src/Chainweb/BlockHeader.hs
@@ -37,12 +37,6 @@ module Chainweb.BlockHeader
   ParentHeader(..)
 , ParentCreationTime(..)
 
--- * Validation Guards
---
--- $guards
-, slowEpochGuard
-, skipFeatureFlagValidationGuard
-
 -- * Block Payload Hash
 , BlockPayloadHash(..)
 , encodeBlockPayloadHash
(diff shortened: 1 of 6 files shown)


src/Chainweb/BlockHeader.hs
```diff
@@ -216,6 +217,23 @@ slowEpochGuard (ParentHeader p)
...
+-- in chainweb-node versions <= 1.5. There is a large number of blocks in the
+-- history of mainnet before 2020-02-20, that have non-zero feature flags. In
+-- order to prepare future use of fleature flag feature flag validation will
+-- start at block height 340000.
+--
+-- This guard grandfathers the this behavior up to block height 340000.
+--
+-- Blockheight 340000 is expected to occur on mainnet01 on 2019-02-24.
+--
+skipFeatureFlagValidationGuard :: BlockHeader -> Bool
```
slpopejoy: This should be in `ChainwebVersion` with the goal of keeping our time-based switches in one place.

src/Chainweb/BlockHeader/Validation.hs
```diff
@@ -403,7 +403,9 @@ prop_block_current :: Time Micros -> BlockHeader -> Bool
 prop_block_current t b = BlockCreationTime t >= _blockCreationTime b
 
 prop_block_featureFlags :: BlockHeader -> Bool
-prop_block_featureFlags b = _blockFlags b == mkFeatureFlags
+prop_block_featureFlags b
+    | skipFeatureFlagValidationGuard b = _blockFlags b == mkFeatureFlags
```
slpopejoy: This function (and perhaps this whole `prop_XXX`) is extremely confusing -- looking at the old code, one would think that `prop_block_featureFlags` was doing validation, after all it is testing that flags == 0. But instead, it's a `not` in `validateInstrinsic`, which is presumably the bug and why we weren't validating. In general, it's confusing to phrase rules in the negative, which is also why the predicate `skipFeature...` is also undesirable: we now have an even more confusing double-negative. Suggest instead:
- this property should be positive: it returns `True` if the flags are set _properly_, such that an error is `False`, updating `validateIntrinsic` accordingly.
- `skipFeatureFlagValidationGuard` should be positive: `enableFeatureFlagValidationGuard`.

Going forward please endeavor to follow this as it is widely agreed that reasoning in the negative is confusing. At a minimum, if something is in the negative the name MUST indicate it, ie `prop_invalid_feature_flags`. But I vastly prefer positive reasoning.

src/Chainweb/BlockHeader.hs
larskuhtz: Agreed, this guard section along with the documentation, originally pre-dates the dates in `Chainweb.Version`, we just forgot to move it when we started to put guards in `Chainweb.Version`.

I'll move it.

PR #926, opened 2020-02-18
Put pact v1.6 changes behind guard

slpopejoy:


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -148,15 +150,17 @@ applyCmd
...
     second _txCache <$!>
       runTransactionM cenv txst applyBuyGas
   where
     txst = TransactionState mcache0 mempty 0 Nothing (_geGasModel freeGasEnv)
 
     executionConfigNoHistory = mkExecutionConfig $
       [ FlagDisableHistoryInTransactionalMode ] ++
-      if ecMod then [] else [ FlagDisableModuleInstall ]
+      [ FlagDisableModuleInstall | not ecMod ] ++
+      [ FlagOldReadOnlyBehavior | isPactBackCompatV16 ]
```
LindaOrtega: What does `FlagOldReadOnlyBehavior` do?

slpopejoy: It's what fixes `enforce-one`, which attempts "full read only" within `enforce-guard`'s  "system read only". Before this resulted in an error, now `runReadOnly` runs with reduced privileges and does not fail (unless of course what's being run hits a user table)

slpopejoy: See https://github.com/kadena-io/pact/blob/master/src/Pact/Types/Purity.hs#L135

LindaOrtega: Should we add a line in https://pact-language.readthedocs.io/en/latest/pact-reference.html#pure-execution about the different types of purity present in pact (i.e. "full read only" vs. "system read only")?

PR #931, opened 2020-02-19
Replace date base coinbaseV2 guard by block height based guard

larskuhtz:
Second PR in a series with the goal to reduce #902 to a minimal forking subset.

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -881,6 +881,7 @@ checkEnableUserContracts
      -- ^ is block genesis
     -> BlockCreationTime
        -- ^ block creation time, in new block we can't get this from header
+       -- TODO: replace guard by block height
     -> ChainwebVersion
     -> Bool
 checkEnableUserContracts isGenesis (BlockCreationTime blockTime) v =
@@ -951,7 +952,7 @@ execNewBlock mpAccess parentHeader miner creationTime = go
 
         -- NEW BLOCK COINBASE: Reject bad coinbase, always use precompilation
         results <- withEnableUserContracts' False creationTime parentHeader $
-          execTransactions (Just (parentHeader,creationTime)) miner newTrans
+          execTransactions (Just parentHeader) miner newTrans
             (EnforceCoinbaseFailure True)
             (CoinbaseUsePrecompiled True)
             pdbenv
@@ -1071,6 +1072,10 @@ setBlockData bh = do
 playOneBlock
     :: (PayloadCas cas)
     => BlockHeader
+        -- ^ this is the current header. We may consider changing this to the parent
+        -- header to avoid confusion with new block and prevent using data from this
+        -- header when we should use the respective values from the parent header
+        -- instead.
     -> PayloadData
     -> PactDbEnv'
     -> PactServiceM cas (T2 Miner Transactions)
@@ -1117,8 +1122,6 @@ playOneBlock currHeader plData pdbenv = do
 
     isGenesisBlock = isGenesisBlockHeader currHeader
 
-    currCreationTime = _blockCreationTime currHeader
-
     go m txs = if isGenesisBlock
       then do
         setBlockData currHeader
@@ -1135,7 +1138,7 @@ playOneBlock currHeader plData pdbenv = do
         setBlockData ph
         -- VALIDATE COINBASE: back-compat allow failures, use date rule for precompilation
         withEnableUserContracts currHeader $
-          execTransactions (Just (ph,currCreationTime)) m txs
+          execTransactions (Just ph) m txs
             (EnforceCoinbaseFailure False) (CoinbaseUsePrecompiled False) pdbenv
 
 -- | Rewinds the pact state to @mb@.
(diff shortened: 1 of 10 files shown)


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -269,17 +265,29 @@ applyCoinbase v logger dbEnv (Miner mid mks) reward@(ParsedDecimal d) pd parentH
...
            Nothing 0.0 rk 0 ec
     txst = TransactionState mc mempty 0 Nothing (_geGasModel freeGasEnv)
     initState = setModuleCache mc $ initCapabilities [magic_COINBASE]
-    chash = Pact.Hash (sshow $ _blockHash parentHeader)
     rk = RequestKey chash
+    bh = fromIntegral $ _pdBlockHeight pd
+    blockTime = blockTimeOf pd
+        -- NOTE it should hold that @blockTime == _blockCreationTime parentHeader@, but
+        -- in some unit test runs that is not the case. That is fine because coinbase
+        -- is a special case
```
emilypi: I figured this out btw. Stuart's test for the forking change validations makes use of blocktime in the publidata _only_, without using a parent header to generate the data as I do in `PactService.runCoinbase`. This leads to the error message when we remove this code and only rely on the parent header for time data. Technically, this should be fine, since its generated via `mkPublicData`, which makes use of the parent header in service state. I'm fine with not poking it if we don't have to, removing it when we move to blockheights

test/Chainweb/Test/Pact/TransactionTests.hs
```diff
@@ -250,52 +245,6 @@ testCoinbaseEnforceFailure = do
...
-    v = Development
-    miner = Miner (MinerId "abcd") (MinerKeys $ mkKeySet [] "<")
-    pubData = PublicData def blockHeight' (toInt64 blockTime) ""
-    upgradeTime = fromJuste $ upgradeCoinV2Date v
-    blockTime = add (TimeSpan (Micros (- 1000))) upgradeTime
-    creationTime = BlockCreationTime $ add (TimeSpan (Micros 1000)) upgradeTime
-    toInt64 (Time (TimeSpan (Micros m))) = m
-    blockHeight' = 123
-    logger = newLogger neverLog "" -- set to alwaysLog to debug
-    devnetHeader = setBlockTime blockTime $ someBlockHeader v (BlockHeight blockHeight')
```
emilypi: We're removing this because now the `PublicData` won't determine the upgrades, yeah? I'm okay with this as long as we get Stuart to also sign off on this and we all 👍

larskuhtz: yes, we could write a test for the new code, which would be different and much simpler. However, we test this each time we do a mainnet catchup, this gets covered. Also we now have CI tests for validating against the actual mainnet history. So, I don't think we need this test any more.

src/Chainweb/Pact/TransactionExec.hs
larskuhtz: Agree, that we can leave it like this for now. I just put the note in there to not get confused again in the future.

test/golden/create-accounts-expected.txt
```diff
@@ -1,4 +1,4 @@
-- output: HLIEBEUgf32fElyvoriTOISAKAdnV0EGOp6xnyHRb38
+- output: npos8cElfalyz9qQL-nVwTaWKXUYlSk_QX7X4CujGwE
```
larskuhtz: This is expected. Before the coin base upgrade check for test versions was based on `_blockHeight parentHeader == 0`. Since the test mines a block at height 11, the upgrade was not applied. 

Now it depends on the the block height of the current header as stored in `PublicData`, namely `_pdBlockHeight`, to be 1. The value of `_pdBlockHeight` is incremented by pact service on execution of each block. For tests that skip ahead in the history, e.g. by using `someTestVersionHeader`, this value doesn't match the actual height of the header. For this test the value is `1` and the upgrade is applied during the test.

The synchronization of public data and the actual data for the current chain context is an issue that should be resolved in a separate PR.

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: see my review https://github.com/kadena-io/chainweb-node/pull/902#pullrequestreview-362793443 , we should fix this now/standardize on basing all logic on injected `parentHeader` and remove comments.

test/Chainweb/Test/Pact/TransactionTests.hs
slpopejoy: Let's keep it please for future reference when we need to test remediations.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -429,19 +423,14 @@ applyUpgrades v parentHeader (BlockCreationTime currCreationTime) =
       initCapabilities [mkMagicCapSlot "REMEDIATE"]
 
     applyTx tx = do
-
       infoLog $ "Running upgrade tx " <> sshow (_cmdHash tx)
```
slpopejoy: what's up with the whitespace deletions here, please undo

src/Chainweb/Pact/PactService.hs
```diff
@@ -1071,6 +1072,10 @@ setBlockData bh = do
 playOneBlock
     :: (PayloadCas cas)
     => BlockHeader
+        -- ^ this is the current header. We may consider changing this to the parent
```
slpopejoy: See review https://github.com/kadena-io/chainweb-node/pull/902#pullrequestreview-362793443, we should change this now (review usages of `currHeader` below to switch to using parent header.

src/Chainweb/Version.hs
```diff
@@ -624,17 +624,42 @@ vuln797FixDate Testnet04 = epoch
...
-upgradeCoinV2Date Development = Just [timeMicrosQQ| 2019-12-14T18:50:00.0 |]
-upgradeCoinV2Date Testnet04 = Nothing
-upgradeCoinV2Date Mainnet01 = Just [timeMicrosQQ| 2019-12-17T15:00:00.0 |]
+-- | Mainnet upgrade to coin v2 at time at @[timeMicrosQQ| 2019-12-17T15:00:00.0 |]@,
+-- which was BEFORE 'txEnableDate'.
+--
+-- This function provides the block heights when coin v2 became effective on the
+-- respective chains.
+--
+coinV2Upgrade
```
slpopejoy: Is this really better? We're replacing a single date with a bunch of constants.

larskuhtz: I think, the logic at the use site becomes simpler.

src/Chainweb/Pact/TransactionExec.hs
larskuhtz: It helped me reading the code. Distinguishing which lines belong to the same function.

larskuhtz: I can put back in but the newlines here seem excessive

src/Chainweb/Pact/PactService.hs
larskuhtz: There is an upcoming PR that addresses the usage of `currHeader` vs. `parentHeader` more in depth. Let's revisit this in the review of that PR.

src/Chainweb/Pact/TransactionExec.hs
larskuhtz: Some of these code lines are going to disappear in one of the follow up PRs. Let's post pone the discussion about current vs. parent header to that PR.

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: You can leave them in but going forward avoid needless/opinionated changes please

src/Chainweb/Version.hs
slpopejoy: How did you find the numbers? This is not reviewable code

slpopejoy: We can keep this in the interests of expediency, but I don't see much value in this change so I question why you bothered to do the necessary work. I'm also concerned that there is no way to validate these heights except "run it in mainnet".

test/golden/create-accounts-expected.txt
slpopejoy: We should probably move these to more than just a hash anyway as they are of questionable value as is.

PR #932, opened 2020-02-21
Remove trivial guards

larskuhtz:
This PR removes outdated kill switch logic and outdated guards that will always evaluate to the same value.

Removing this redundant logic makes room for new guards related to the upcoming forking changes (creation time fix, hash function changes, 20 chains).

Some of the removed code was already marked for removal.


src/Chainweb/Mempool/InMem.hs
```diff
@@ -162,10 +161,10 @@ withInMemoryMempool_ :: Logger logger
                      -> ChainwebVersion
                      -> (MempoolBackend t -> IO a)
                      -> IO a
-withInMemoryMempool_ l cfg v f = do
+withInMemoryMempool_ l cfg _v f = do
```
emilypi: why not just remove the version from the signature? Same for below

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -145,19 +142,16 @@ applyCmd
...
   where
     txst = TransactionState mcache0 mempty 0 Nothing (_geGasModel freeGasEnv)
 
-    executionConfigNoHistory = mkExecutionConfig $
-      [ FlagDisableHistoryInTransactionalMode ] ++
-      [ FlagDisableModuleInstall | not ecMod ] ++
-      [ FlagOldReadOnlyBehavior | isPactBackCompatV16 ]
+    executionConfigNoHistory = mkExecutionConfig
+      $ FlagDisableHistoryInTransactionalMode
+      : [ FlagOldReadOnlyBehavior | isPactBackCompatV16 ]
```
emilypi: is this... a list comprehension? I've never seen that style before!

test/Chainweb/Test/Pact/TransactionTests.hs
```diff
@@ -226,17 +226,23 @@ testCoinbase797DateFix = testCaseSteps "testCoinbase791Fix" $ \step -> do
...
+    postForkHeight = 121452
+
     logger = newLogger neverLog ""
 
+    -- | someBlockHeader is a bit slow for the vuln797Fix to trigger. So, instead
+    -- of mining a full chain we fake the height.
+    --
+    mkTestHeader :: BlockHeight -> BlockHeader
+    mkTestHeader h = (someBlockHeader (FastTimedCPM singleton) 10)
+        { _blockHeight = h }
```
emilypi: 👍

src/Chainweb/Mempool/InMem.hs
larskuhtz: This is part of the main mempool API, which I did't want to change. It also seems reasonable that the mempool is aware of the chainweb version, even if the current implementation doesn't use it.

src/Chainweb/Pact/TransactionExec.hs
larskuhtz: yes, list comprehension are desugared to `MonadPlus` computations. A boolean expression `exp` becomes `guard exp`. It's quite useful to filter the results of the comprehension.

slpopejoy: I switched to comprehensions in a recent PR, but the credit goes entirely to Dr Kuhtz (used in block header validation)

PR #935, opened 2020-02-24
Creation time fix [Phase 1]

larskuhtz:
In the current code base the block creation time of the *current* header is used as reference for tx timing validation. The current time during pact validation is the block creation time of the *parent* header is used.

This PR provides the first step of resolving this inconsistency. The PR

* [x] makes it explicit throughout the code which header value is used as reference by labeling block headers with the `ParentHeader` newtype when the header is used as parent header in the respective context.

* [x] adds a validation guard for switching between using the *current* and the *parent* header as reference time for tx timing validation.

* [x] adjusts unit tests to be compatible with both behaviors (using the new behavior as default).

The line of changes from this PR is continued in #942

### DIFF
--- a/src/Chainweb/Pact/Backend/ForkingBench.hs
+++ b/src/Chainweb/Pact/Backend/ForkingBench.hs
@@ -401,7 +401,7 @@ testRocksDb l = rocksDbNamespace (const prefix)
   where
     prefix = (<>) l . sshow <$> (randomIO @Word64)
 
-toTxCreationTime :: Time Integer -> TxCreationTime
+toTxCreationTime :: Integral a => Time a -> TxCreationTime
 toTxCreationTime (Time timespan) = case timeSpanToSeconds timespan of
           Seconds s -> TxCreationTime $ ParsedInteger s
 
@@ -607,7 +607,7 @@ makeMetaWithSender sender c =
 -- hardcoded sender (sender00)
 makeMeta :: ChainId -> IO PublicMeta
 makeMeta c = do
-    t <- toTxCreationTime <$> getCurrentTimeIntegral
+    t <- toTxCreationTime @Int <$> getCurrentTimeIntegral
     return $ PublicMeta
         {
           _pmChainId = Pact.ChainId $ chainIdToText c
--- a/src/Chainweb/Pact/Backend/ForkingBench.hs
+++ b/src/Chainweb/Pact/Backend/ForkingBench.hs
@@ -145,12 +145,12 @@ bench = C.bgroup "PactService" $
         name = "block-new" ++ (if validate then "-valid" else "") ++
                "[" ++ show txCount ++ "]"
 
-testMemPoolAccess :: IORef Int -> MVar (Map Account (NonEmpty SomeKeyPairCaps)) -> Time Integer -> MemPoolAccess
+testMemPoolAccess :: IORef Int -> MVar (Map Account (NonEmpty SomeKeyPairCaps)) -> Time Int -> MemPoolAccess
 testMemPoolAccess txsPerBlock accounts t = mempty
     { mpaGetBlock = \validate bh hash _header -> getTestBlock accounts t validate bh hash }
   where
 
-    setTime time = \pb -> pb { _pmCreationTime = toTxCreationTime time }
+    setTime time pb = pb { _pmCreationTime = toTxCreationTime time }
 
     getTestBlock mVarAccounts txOrigTime validate bHeight@(BlockHeight bh) hash
         | bh == 1 = do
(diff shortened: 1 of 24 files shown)


src/Chainweb/Pact/Utils.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE NumericUnderscores #-}
```
slpopejoy: this looks unused

src/Chainweb/Pact/Backend/ForkingBench.hs
```diff
@@ -401,7 +401,7 @@ testRocksDb l = rocksDbNamespace (const prefix)
   where
     prefix = (<>) l . sshow <$> (randomIO @Word64)
 
-toTxCreationTime :: Time Integer -> TxCreationTime
+toTxCreationTime :: Integral a => Time a -> TxCreationTime
```
slpopejoy: Since this is only used here would prefer specialization to `Int` and avoid type annotation below which is noisy IMO

src/Chainweb/Pact/PactService.hs
```diff
@@ -1,5 +1,5 @@
 {-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
```
slpopejoy: I just checked and `DeriveGeneric` wasn't needed so I assume you can remove it instead of changing it.

src/Chainweb/Pact/PactService.hs
```diff
@@ -864,17 +881,21 @@ execNewBlock mpAccess parentHeader miner creationTime = go
...
-          validate bhi _bha txs =
+          validate bhi _bha txs = do
             -- note that here we previously were doing a validation
             -- that target == cpGetLatestBlock
             -- which we determined was unnecessary and was a db hit
             --
-            -- TODO: propagate the underlying error type?
-            V.map (either (const False) (const True)) <$>
-              validateChainwebTxs cp creationTime bhi txs runDebitGas
+            -- TODO: propagate the underlying error type? Yes, please!
```
slpopejoy: :) this is done in #940

src/Chainweb/Pact/PactService.hs
```diff
@@ -1293,14 +1326,21 @@ execPreInsertCheckReq txs = do
     b <- liftIO $ _cpGetLatestBlock cp
     case b of
         Nothing -> return $! V.map Right txs
-        Just (h, ha) ->
-            withCheckpointer (Just (h+1, ha)) "execPreInsertCheckReq" $ \pdb -> do
-                now <- liftIO getCurrentTimeIntegral
+        Just (parentHeight, parentHash) -> do
```
slpopejoy: I don't understand why the pre-insert check needs to change? Current time would seem to make sense here.

src/Chainweb/Version.hs
```diff
@@ -637,6 +637,29 @@ pactBackCompat_v16 Mainnet01 h = h < 328000
 pactBackCompat_v16 Development h = h < 120
 pactBackCompat_v16 _ _ = False
 
+-- | If this is true the creation time of the current header is used for pact tx
+-- creation time and ttl validation.
+--
+-- Once the block height for triggering this on mainnet is in the past, the
```
slpopejoy: ... but below, `Mainnet01` is always `True` so this para seems wrong

src/Chainweb/Version.hs
```diff
@@ -637,6 +637,29 @@ pactBackCompat_v16 Mainnet01 h = h < 328000
 pactBackCompat_v16 Development h = h < 120
 pactBackCompat_v16 _ _ = False
 
+-- | If this is true the creation time of the current header is used for pact tx
```
slpopejoy: I'm concerned that it seems we're introducing this complexity for `Development` only. Mainnet is always `True`, and other versions are basically duplicative of switches on `isGenesis`. Wouldn't it make more sense to have a special case for "first non-genesis block" to use the current header time?

src/Chainweb/Version.hs
larskuhtz: The current value for `Mainnet01` is just a placeholder, since it's not yet decided when to do the decision.

src/Chainweb/Version.hs
larskuhtz: We are going to need the switch for Mainnet. We'll probably use the value for `Development` for testing. All other test version it does't really matter, as long as tests pass.

As we did with other guards, we might be able to remove it after the transition is completed and replace it by a hard-coded timing validation grace period, which would allow to also remove related code logic.

src/Chainweb/Pact/PactService.hs
larskuhtz: The old check was just an approximation, because the true reference time for tx timing validation wasn't yet known. With the new behavior we can make the check precise.

Once the transition is complete, the code paths for the old check can be removed.

src/Chainweb/Version.hs
larskuhtz: I have an upcoming PR, that makes the timing behavior in several tests deterministic (avoiding use if getCurrentTime). With that change (possibly extending it to more test) it may be possible to remove the special case for test versions.

PR #938, opened 2020-02-28
Allow reads in local via config

slpopejoy:
Allows configuration to make direct table reads in local

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -31,6 +31,7 @@ module Chainweb.Pact.PactService
     , execNewBlock
     , execValidateBlock
     , execTransactions
+    , execLocal
     , initPactService
     , readCoinAccount
     , readAccountBalance
@@ -73,9 +74,6 @@ import qualified Data.Text.Encoding as T
 import Data.Tuple.Strict (T2(..))
 import Data.Vector (Vector)
 import qualified Data.Vector as V
-import Data.Word
-
-import GHC.Generics
 
 import System.Directory
 import System.IO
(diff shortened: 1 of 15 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -137,23 +135,6 @@ import Data.CAS (casLookupM)
 import Data.LogMessage
 import Utils.Logging.Trace
 
--- -------------------------------------------------------------------------- --
```
slpopejoy: This was per my review https://github.com/kadena-io/chainweb-node/pull/902#discussion_r382709468

Basically
- let's try not to add new types to PactService but in Types files
- let's try to reuse existing Exception types instead of bespoke ones

src/Chainweb/Pact/PactService.hs
```diff
@@ -345,12 +323,7 @@ initializeCoinContract _logger v cid pwo = do
         Nothing -> throwM NoBlockValidatedYet
         (Just !p) -> return p
       let target = Just (succ bhe, bhash)
-      bhdb <- asks _psBlockHeaderDb
-      parentHeader <- liftIO $! lookupM bhdb bhash
-        `catch` \e -> throwM $ InternalPactServiceException
-          { _pactServiceExceptionMsg = "failed lookup of parent header in initializeCoinContract"
-          , _pactServiceExceptionInner = Just e
-          }
+      parentHeader <- lookupBlockHeader bhash "initializeCoinContract"
```
slpopejoy: Also per https://github.com/kadena-io/chainweb-node/pull/902#discussion_r382709468 

- if the same code block shows up 4 times, please consider making a single function for it

larskuhtz: That one is addressed here: https://github.com/kadena-io/chainweb-node/pull/935/commits/889fe0e754f5dc21279ff8e9c40b65f57def2013

PR #940, opened 2020-03-03
[bug] chain data on rewind issue

slpopejoy:


src/Chainweb/Pact/PactService.hs
```diff
@@ -1012,17 +1014,13 @@ playOneBlock currHeader plData pdbenv = do
...
-                        msg = [ (hash, validationErr hash) ]
-                    in throwM $ TransactionValidationException msg
-    -- transactions are now successfully validated.
+    valids <- V.zip trans <$> liftIO
+      (validateChainwebTxs cp creationTime (_blockHeight currHeader) trans skipDebitGas)
+
+    case foldr handleValids [] valids of
+      [] -> return ()
+      errs -> throwM $ TransactionValidationException $ errs
+
```
slpopejoy: This shouldn't break anything, wanted to see the actual errors when validation fails. The errors blow up and don't end up in payload outputs.

src/Chainweb/Pact/PactService.hs
```diff
@@ -1284,6 +1278,14 @@ transactionsFromPayload plData = do
     toCWTransaction bs = evaluate (force (codecDecode chainwebPayloadCodec $
                                           _transactionBytes bs))
 
+debugResult :: A.ToJSON a => Text -> a -> PactServiceM cas ()
```
slpopejoy: I'm assuming debugging coinbase and tx outputs won't be a perf problem if `Debug` is not enabled, assuming the logger is lazy?

test/Chainweb/Test/Pact/PactInProcApi.hs
```diff
@@ -41,57 +41,120 @@ import Test.Tasty.HUnit
...
 tests :: ScheduledTest
 tests = ScheduledTest label $
-    withRocksResource $ \rocksIO ->
-    withTemporaryDir $ \dir ->
-    withBlockHeaderDb rocksIO genblock $ \bhdb ->
-    withPayloadDb $ \pdb ->
+    withResource (newMVar "1") (const $ return()) $ \noncer ->
     testGroup label
-    [ withPact testVersion Warn pdb bhdb testMemPoolAccess dir 100000
+    [ withPactTestBlockDb testVersion cid Warn testMemPoolAccess conf
```
slpopejoy: All pact tests are run `Sequential` so the `after AllSucceed` stuff can go away

PR #944, opened 2020-03-05
Timeless Payload Reassociation

fosskers:
Following from @larskuhtz 's recent work involving block creation times, this PR updates the Mining Coordination logic not to make any assumptions about `BlockCreationTime` w.r.t. how blocks and cached payloads are associated.


src/Chainweb/Logging/Miner.hs
```diff
@@ -37,7 +32,6 @@ data NewMinedBlock = NewMinedBlock
     { _minedBlockHeader :: !(ObjectEncoded BlockHeader)
     , _minedBlockTrans :: {-# UNPACK #-} !Word
     , _minedBlockSize :: {-# UNPACK #-} !Word   -- ^ Bytes
-    , _minedHashAttempts :: !Natural
```
fosskers: No longer useful, and supporting it complicated some types.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -92,16 +86,17 @@ import Utils.Logging.Trace (trace)
...
-
 -- | Data shared between the mining threads represented by `newWork` and
 -- `publish`.
 --
+-- The key is a unique pair of `BlockHash` of the parent block with the hash of
+-- the current block's payload.
+--
 newtype MiningState = MiningState
-    (M.Map (T2 BlockCreationTime BlockPayloadHash) (T3 Miner PrevTime PayloadWithOutputs))
+    (M.Map (T2 BlockHash BlockPayloadHash) (T3 Miner PayloadWithOutputs (Time Micros)))
```
fosskers: @emilypi and I discussed this, and we believe that this pair is a sufficiently unique key for cached payloads.

fosskers:
Still needs to be tested.

fosskers:
Appears to be working. I'm mining to it on devnet and successfully seeing rewards.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -263,19 +256,6 @@ publish lf (MiningState ms) cdb bh = do
...
-estimatedHashes :: PrevTime -> BlockHeader -> Natural
-estimatedHashes (PrevTime p) b = floor $ (d % t) * 1000000
-  where
-    t :: Integer
-    t = case timeBetween (_blockCreationTime b) p of Micros t' -> int t'
-
-    d :: Integer
-    d = case targetToDifficulty $ _blockTarget b of
-        HashDifficulty (PowHashNat w) -> int w
-
```
emilypi: Can you say why this went away?

src/Chainweb/Logging/Miner.hs
```diff
@@ -15,9 +12,9 @@
 -- TODO
 --
 module Chainweb.Logging.Miner
-( NewMinedBlock(..)
-, OrphanedBlock(..)
-) where
+  ( NewMinedBlock(..)
+  , OrphanedBlock(..)
+  ) where
```
emilypi: Unnecessary whitespace changes.

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -87,8 +86,7 @@ data MiningCoordination logger cas = MiningCoordination
     , _coord403s :: !(IORef Int)
     , _coordConf :: !CoordinationConfig
     , _coordUpdateStreamCount :: !(IORef Int)
-    , _coordPrimedWork :: !(TVar PrimedWork)
-    }
+    , _coordPrimedWork :: !(TVar PrimedWork) }
```
emilypi: unnecessary

src/Chainweb/Miner/Coordinator.hs
fosskers: It was used to estimate hash power when we were in the early days of CPU mining, before we had better tools like the Block Explorer. This is only consumed on Kibana, and even then only on one widget that isn't useful.

emilypi: Cool, thanks!

PR #949, opened 2020-03-11
Stack CI

fosskers:
It compiles without optimizations and doesn't run the test suites.
Not intended for producing production artifacts.


.github/workflows/stack-ci.yaml
```diff
@@ -0,0 +1,32 @@
+name: Build
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
larskuhtz:
```suggestion
      - master
      - ci/*
```

Issue #954, opened 2020-03-12
Feature flags


larskuhtz:
The goal of this feature is to require that the first 8 bytes of the block header are set to zero for future use as feature flags.

* [x] Change CPU miners to use last 8 bytes as nonce and set first 8 bytes to 0
* [x] Switch nonce and feature flags in the binary serialization of the code
* [x] configure existing feature flag block header validation to enforce flag being 0
       (guarded by block height, I think, the guard may already exist)
* [x] work with pools and external miners to be aware of the change and adjust their software.

## Optional (possibly second PR):

* [x] improve performance of CPU miner by only iterating the final blake2s chunk (compress step) when searching for the nonce.

mightybyte:
It might be a good idea to hold off merging these larger changes to master so we can keep master clear for intermediate releases if necessary.

@larskuhtz 
@slpopejoy

larskuhtz:
Item (1.) and (2.) above are safe to merge. Item (3.) depends on (4.).

This can be released as soon as the big pools and miners set the first 8 bytes to 0. If we are certain that this will be the case when we release 1.7 we should merge to master now, in order to get test coverage. Otherwise we should not merge now.

I propose we do (1.) and (2.) and (4.). I think, (3.) would be a one line change that we can do whenever (4.) is complete.

fosskers:
(1) is already in `master`. (2) is queued in https://github.com/kadena-io/chainweb-node/pull/960

larskuhtz:
Just to confirm: #960 doesn't yet enforce the first 8 bytes to be zero? Is that right?

fosskers:
Correct.

PR #957, opened 2020-03-12
CPU Mining: Inject nonce bytes at the end

fosskers:
In the place of the (soon to be) former feature flags.

See #954


fosskers:
`chainweb-miner` also relies on this code, so I'm going to test that now.

fosskers:
This appears to work. I'm doing CPU mining to devnet, and I see my balance increasing. So, my byte offset value must be correct.

src/Chainweb/Miner/Core.hs
```diff
@@ -169,8 +169,10 @@ mine _ orig@(Nonce o) (TargetBytes tbytes) (HeaderBytes hbytes) = do
 -- hashed `BlockHeader`. If that layout changes, this functions need to be
 -- updated. The assumption allows us to iterate on new nonces quickly.
 --
+-- Recall: `Nonce` contains a `Word64`, and is thus 8 bytes long.
+--
 injectNonce :: Nonce -> Ptr Word8 -> IO ()
-injectNonce (Nonce n) buf = poke (castPtr buf) n
+injectNonce (Nonce n) buf = pokeByteOff (castPtr buf) 278 n
```
fosskers: Seems to be the correct magic number.

larskuhtz: Maybe adding a comment with a reference to https://github.com/kadena-io/chainweb-node/wiki/Block-Header-Binary-Encoding ?

PR #959, opened 2020-03-13
Fmelp/allocation tests

slpopejoy:


mightybyte:
I updated this today with new shared Chainweaver keys.  Should be good to merge once it passes CI.

mightybyte:
Added a comment mentioning the chainweaver recovery phrase

slpopejoy:
I'm the creator of the PR so can't "approve" but it's not my work -- so recording approval here and noting that CI passed. Forcing merge.

PR #960, opened 2020-03-13
Updated binary encoding for BlockHeader

fosskers:
See https://github.com/kadena-io/chainweb-node/issues/954

### DIFF
--- a/src/Chainweb/BlockHeader.hs
+++ b/src/Chainweb/BlockHeader.hs
@@ -752,7 +752,7 @@ instance ToJSON (ObjectEncoded BlockHeader) where
 
 parseBlockHeaderObject :: Object -> Parser BlockHeader
 parseBlockHeaderObject o = BlockHeader
-    <$> o .: "nonce"
+    <$> o .: "featureFlags"
     <*> o .: "creationTime"
     <*> o .: "parent"
     <*> o .: "adjacents"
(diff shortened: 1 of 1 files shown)


fosskers:
This needs to be ran on devnet, and possibly also mainnet, to prove that assumptions about hashing aren't being broken.

src/Chainweb/BlockHeader.hs
```diff
@@ -648,28 +648,29 @@ decodeBlockHeaderWithoutHash = do
...
+    ht <- decodeHashTarget
+    pa <- decodeBlockPayloadHash
+    ci <- decodeChainId
+    wt <- decodeBlockWeight
+    hg <- decodeBlockHeight
+    cv <- decodeChainwebVersion
+    es <- decodeEpochStartTime
+    no <- decodeNonce
+    bh <- decodeBlockHash
+    pure $ BlockHeader no ct ph hr ht pa ci wt hg cv es ff bh
```
larskuhtz: I think changing the order of the fields in `BlockHeader` wouldn't break anything and would avoid these changes.

fosskers: Ah, yes, perhaps that's better.

fosskers:
I'm currently doing a full mainnet catch up on my node with this.

PR #963, opened 2020-03-17
Better test utilities for building Pact commands

slpopejoy:
with `PactExec` moved over to them (even passing goldens!)

### DIFF
--- a/test/Chainweb/Test/Pact/Utils.hs
+++ b/test/Chainweb/Test/Pact/Utils.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE NumericUnderscores #-}
@@ -19,10 +21,13 @@
 module Chainweb.Test.Pact.Utils
 ( -- * Exceptions
   PactTestFailure(..)
-, ChainwebKeyPair
   -- * test data
+, ChainwebKeyPair
+, SimpleKeyPair
 , sender00KeyPair
+, sender00
 , sender01KeyPair
+, sender01
 , allocation00KeyPair
 , allocation01KeyPair
 , allocation02KeyPair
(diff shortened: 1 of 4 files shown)


test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -47,6 +52,40 @@ module Chainweb.Test.Pact.Utils
 , withPayloadDb
 , withBlockHeaderDb
 , withTemporaryDir
+-- * 'PactValue' helpers
```
mightybyte: It's not worth holding up this PR for, but this stuff looks fairly similar to some of the transaction building stuff in the tx generator.  I guess we recently pulled that code out into a separate repo and maybe these are chainweb-specific enough that it doesn't make sense, but it might be worth keeping in mind for consolidation at some point down the road if possible.

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -201,6 +245,10 @@ sender00KeyPair =
     , ED25519
     )
 
+sender00 :: SimpleKeyPair
```
mightybyte: For instance, the txg had some infrastructure for reading these keys directly from the JSON files.

https://github.com/kadena-io/txg/blob/38c1112ff74c07d41e5924fb252b8eda28eda2b0/lib/TXG/Simulate/Contracts/Common.hs#L120

slpopejoy: That I've deliberately avoided as it means keyset manipulation is in `IO` ...

test/Chainweb/Test/Pact/Utils.hs
slpopejoy: Unless I'm missing something, TXG uses Pact's `ApiReq` which while it is a builder, it's optimized for YAMLs, so there's a million `Maybe`s. The approach here is to provide a principled un-signed builder.

What are you referring to in TXG? https://github.com/kadena-io/txg

mightybyte: Nothing specific.  It was just a notion of vaguely similar things floating around, and I wanted to mention it if for nothing more than to increase awareness...especially since it's in a different repo now.

PR #965, opened 2020-03-18
Get rid of MonadFail from IO

larskuhtz:
Fix incomplete pattern matches in `IO` and get the code base ready for GHC-8.8.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -27,3 +27,22 @@ constraints:
       base-unicode-symbols < 0.2.4,
       megaparsec < 8,
       neat-interpolation < 0.4
+
+-- GHC-8.8
+
+package pact
+    ghc-options: -Wwarn
+    haddock-options: --optghc=-Wwarn
+
+allow-newer:
+    *:sbv
+
+package vault
+    documentation: false
+
+-- Currently PR status
+source-repository-package
+    type: git
+    location: https://github.com/stepcut/ixset-typed
+    tag: 745634a16d34ff8cb3abf7d5b40fffca95ac36fc
+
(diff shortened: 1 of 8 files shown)


cabal.project
```diff
@@ -27,3 +27,22 @@ constraints:
...
       neat-interpolation < 0.4
+
+-- GHC-8.8
+
+package pact
+    ghc-options: -Wwarn
+    haddock-options: --optghc=-Wwarn
+
+allow-newer:
+    *:sbv
```
emilypi: this clause may be unnecessary once https://github.com/kadena-io/pact/pull/783 goes in.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -694,11 +698,14 @@ handlePossibleRewind bRestore hsh = do
...
         assign bsBlockHeight bRestore
-        SInt txid <- callDb "getting txid" $ \db ->
+        r <- callDb "getting txid" $ \db ->
           expectSingleRowCol msg =<< qry db
               "SELECT endingtxid FROM BlockHistory WHERE blockheight = ?;"
               [SInt (fromIntegral bCurrent)]
               [RInt]
+        txid <- case r of
+            SInt x -> return x
+            _ -> error "Chainweb.Pact.ChainwebPactDb.handlePossibleRewind.newChildBlock: failed to match SInt"
```
emilypi: These look like undetected partial patmatches? Yikes 😬

cabal.project
larskuhtz: 👍 nice

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
larskuhtz: I assume that the result being an `SInt` is a code invariant? This code comes from the db, so there is a serialization step and we'll have to reintroduce the invariant as a type at runtime. So the code is unsafe and we have to do something if our invariant doesn't hold, which would be a bug. I think using `error` is fine in that case.

PR #968, opened 2020-03-18
remove reflection from WebBlockHeaderDb

larskuhtz:


src/Chainweb/SPV/CreateProof.hs
```diff
@@ -435,26 +434,25 @@ crumbsOnChain trgHeader srcHeight
...
-    path = shortestPath (_chainId trgHeader) srcCid graph
+    path = shortestPath (_chainId trgHeader) srcCid (_chainGraph db)
 
     go
        :: BlockHeader
        -> [ChainId]
        -> [(Int, BlockHeader)]
        -> IO (BlockHeader, [(Int, BlockHeader)])
-    go !cur [] !acc = return $! (cur, acc)
+    go !cur [] !acc = return (cur, acc)
```
emilypi: we sure we don't want to force this?

larskuhtz: `(,)` is a constructor. There's nothing to force.

PR #969, opened 2020-03-19
Better test mempool fixture; overhaul PactInProcAPI

slpopejoy:
Gets rid of `adminData` and `goldenTestTransactions` in Pact test utils too

### DIFF
--- a/test/Chainweb/Test/Pact/Utils.hs
+++ b/test/Chainweb/Test/Pact/Utils.hs
@@ -34,12 +34,10 @@ module Chainweb.Test.Pact.Utils
 , allocation02KeyPair'
 , testPactFilesDir
 , testKeyPairs
-, adminData
 , mkKeySetData
   -- * helper functions
 , mergeObjects
 , formatB16PubKey
-, goldenTestTransactions
 , mkTestExecTransactions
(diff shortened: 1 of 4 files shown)


test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -846,6 +837,31 @@ toTxCreationTime (Time timespan) = case timeSpanToSeconds timespan of
...
+  }
+  where
+    call :: (MemPoolAccess -> f) -> (f -> IO a) -> IO a
+    call f g = readIORef r >>= g . f
+
+-- | use a "delegate" which you can dynamically reset/modify
+withDelegateMempool
+  :: (IO (IORef MemPoolAccess, MemPoolAccess) -> TestTree)
+  -> TestTree
+withDelegateMempool = withResource start (const mempty)
```
larskuhtz: Isn't `(const mempty :: a -> IO ()) == (mempty :: a -> IO())`?
 ```suggestion
withDelegateMempool = withResource start mempty
```

slpopejoy: Went around and fixed this elsewhere in the module too :)

PR #974, opened 2020-03-20
Feature Flag Validation

fosskers:
The Nonce and Feature Flag bytes have already switched places on `master`. This PR enforces those Feature Flag bytes to be `0` for certain `ChainwebVersion` and `BlockHeight` pairs.

When the value for `Mainnet01` is set, we need to ensure that miners are aware that they need to set the "old" nonce bytes (new FF bytes) to 0.


src/Chainweb/Version.hs
```diff
@@ -710,21 +710,17 @@ slowEpochGuard _ _ = False
...
+-- Feature Flag bytes and Nonce bytes have switched places in `BlockHeader`. For
+-- live chains, enforcing the following condition must be ignored for the
+-- historical blocks for which both the Nonce and Flags could be anything.
 --
 skipFeatureFlagValidationGuard
     :: ChainwebVersion
     -> BlockHeight
         -- ^ height of header
     -> Bool
 skipFeatureFlagValidationGuard Mainnet01 _ = True
```
fosskers: We should change the value for `Mainnet01` only after the release.

larskuhtz, line 723: This should be set to a value shortly after the next killswitch triggers.

test/Chainweb/Test/BlockHeader/Validation.hs
```diff
@@ -157,58 +157,58 @@ prop_fail_validate = testCase "validate invalid BlockHeaders" $ do
...
       )
     , ( hdr & testHeaderHdr . blockFlags .~ fromJuste (runGet decodeFeatureFlags badFlags)
-      , [IncorrectHash, IncorrectPow]
+      , [IncorrectHash, IncorrectPow, InvalidFeatureFlags]
       )
     , ( hdr & testHeaderHdr . blockAdjacentHashes .~ BlockHashRecord mempty
-      , [IncorrectHash, IncorrectPow]
+      , [IncorrectHash, IncorrectPow, InvalidFeatureFlags]
       )
     ]
```
larskuhtz, line 213: Instead of failing with `InvalidFeatureFlags` (almost) everywhere, wouldn't it be better to fix the test header or replace it with one that has correct feature flags?

There are utils in `Test.Chainweb.Utils.TestHeader` for picking a TestHeader from the history of an existing Chainweb instance.

test/Chainweb/Test/BlockHeader/Validation.hs
```diff
@@ -157,58 +157,58 @@ prop_fail_validate = testCase "validate invalid BlockHeaders" $ do
 validationFailures :: [(TestHeader, [ValidationFailureType])]
 validationFailures =
```
larskuhtz, line 158: Please add an entry that specifically fails for feature flag validation by setting it to an invalid value.

src/Chainweb/Version.hs
```diff
@@ -750,20 +750,17 @@ oldTargetGuard _ _ = False
...
+-- live chains, enforcing the following condition must be ignored for the
+-- historical blocks for which both the Nonce and Flags could be anything.
 --
 skipFeatureFlagValidationGuard
     :: ChainwebVersion
     -> BlockHeight
         -- ^ height of header
     -> Bool
-skipFeatureFlagValidationGuard Mainnet01 _ = True
+skipFeatureFlagValidationGuard Mainnet01 _ = False
```
larskuhtz, line 763: Wouldn't this reject existing Mainnet history? This should toggle to `False` only from a certain block height onward after the next kill switch.

src/Chainweb/Version.hs
```diff
@@ -750,20 +750,17 @@ oldTargetGuard _ _ = False
...
 skipFeatureFlagValidationGuard Development _ = True
```
larskuhtz, line 764: For testing purposes, will probably want to set this to `True` only up to a certain block height, say, 2000.

test/Chainweb/Test/BlockHeader/Validation.hs
```diff
@@ -242,15 +242,6 @@ validationFailures =
...
-            , t "AAARACjupkPfZFz8v54FAIxbIHQdq4grSLNI-NL6Lxhn3BSGZij58RFxHpHFdvHlAwABAAAAfdHDK_Q8xoD-W0nBPPBPMOgs1VukuCImYwCNnaBUwOMFAAAA4eefM0SUltzJ0Qszo3N0R9B4w_ap2_M2e6nlKEqJmkoHAAAAHLF8WtH6kE_yIwsM0o7RJK36e1MSHUhVPYTiZI-SrTbiMNKcS7VzGITdCwrGSYWrFNQvGP7KAzjbLQIAAAAAABlK0LefdM1J4t_Qeg6xAVNNDKEOhiEmNKe6SK9N6TAZBgAAALFBPU7YDgRoAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAOdsEAAAAAAAFAAAA2eJ0y7-eBQAAAORH5OUAAGvj8zv81JpNcizsrdpA94iOnHoSqkPynSvA-qmfl-l-"
-            , t "AABPAIw5kEeHUtD7v54FAH2hiBHu2vsBk_DnRF8_Ub5cyVZCYJhD5s5ONm840DotAwAAAAAAPYZZ2yg5iXsMOyKqKKUhrGaboexUhUVK8e-fhn3FzNkEAAAAu_A9WCeRoLM17g_jc0A2UnhvCQFe5LCtTnaze9LqajQHAAAAHLF8WtH6kE_yIwsM0o7RJK36e1MSHUhVPYTiZI-SrTbP1aVtUvTRaiRyg9hCVSPXuIpf3IjuwHaBKgIAAAAAABQWMBli4UbscIslyPPH2ItcNaY2_Fm7yFucQM86oqojAgAAAFy5ttLGN19tAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAOdsEAAAAAAAFAAAAddFMzL-eBQAAALuSPmYAAMu0EgB6CHhY9CV4nXv3KPjmhM82Fm3mYQH6YVyPPiEd"
-            ]
-        ]
-    , testHeader
```
larskuhtz, line 253: Why is this test case removed? The header is an actually header from Mainnet and should succeed validation.

fosskers: It failed, at least with the `skipFeatureFlagValidationGuard` as it is (without a height).

src/Chainweb/Version.hs
fosskers: Yes you're right.

src/Chainweb/Version.hs
fosskers: For which `ChainwebVersion`?

test/Chainweb/Test/BlockHeader/Validation.hs
```diff
@@ -205,7 +205,8 @@ validationFailures =
       , [IncorrectHash, IncorrectPow]
       )
     , ( hdr & testHeaderHdr . blockFlags .~ fromJuste (runGet decodeFeatureFlags badFlags)
-      , [IncorrectHash, IncorrectPow]
+            & testHeaderHdr . blockHeight .~ 550000
```
larskuhtz, line 208: Let's have a test case for both cases, with and without this line. Maybe one case for `testHeaderHdr . blockHeight .~ 530500` and one for `testHeaderHdr . blockHeight .~ 530499`?

fosskers: Good idea.

src/Chainweb/Version.hs
larskuhtz: Devnet, but we can still do it later when we reset devnet the next time.

PR #977, opened 2020-03-24
compute epoch time based on parent header

larskuhtz:
This shifts the computation of the epoch time back one block, so that the computation of the target doesn’t depend on the creation time of the current block.

Legacy behavior is guarded by `oldTargetGuard` and switched on for `Mainnet01` and `Testnet04`. It's on for `Development` for block `< 360`.

When the guard switches of there is one epoch that has only 119 blocks. There is no need to adjust target computation, because change in the overall epoch time are minimal.

### DIFF
--- a/src/Chainweb/BlockHeader.hs
+++ b/src/Chainweb/BlockHeader.hs
@@ -272,19 +272,24 @@ powTarget
         -- ^ parent header
     -> BlockCreationTime
         -- ^ block creation time of new block
+        --
+        -- This parameter is used only when @oldTargetGuard@ is @True@.
+        --
     -> HashTarget
         -- ^ POW target of new block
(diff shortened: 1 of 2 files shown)


larskuhtz:
I tested the PR on devnet:

1. deploy changes to existing testnet (without reset) with the trigger set to a future block height
2. verify that the PR version is compatible with the old version
3. verify that the devnet advances past the trigger
4. wait for first DA to happen and verify that the devnet is healthy and has correct block rate
5. verify that old code fails to validate blocks after the trigger with

    ```
    The given target for the block is incorrect for its history
    The epoch start time of the block is incorrect
    ```

6. verify that the failing node recovers after switching to new version.
7. verify that new version can do a complete catchup on the devenet. Which double check backward compatibility with blocks mined by old versions.

PR #982, opened 2020-03-27
Omit joining initial cut from CutDB with genesis cut

larskuhtz:
We generally trust the chain db. Let's also trust the cut that is persisted in the DB.

### DIFF
--- a/src/Chainweb/Cut.hs
+++ b/src/Chainweb/Cut.hs
@@ -43,6 +43,7 @@ module Chainweb.Cut
 , forkDepth
 , limitCut
 , limitCutHeaders
+, unsafeMkCut
 
 -- * Exceptions
 , CutException(..)
@@ -210,6 +211,12 @@ cutAdjs
 cutAdjs c cid = HM.intersection
     (_cutHeaders c)
     (HS.toMap (adjacentChainIds (_chainGraph c) cid))
+
+unsafeMkCut :: ChainwebVersion -> HM.HashMap ChainId BlockHeader -> Cut
+unsafeMkCut v hdrs = Cut
+    { _cutHeaders = hdrs
+    , _cutChainwebVersion = v
+    }
 
 -- -------------------------------------------------------------------------- --
 -- Limit Cut Hashes By Height
@@ -434,7 +441,7 @@ join_ wdb prioFun a b = do
         db <- getWebBlockHeaderDb wdb cid
         (q' :> h) <- S.fold g q id $ branchDiff_ db x y
         let !h' = q' `seq` h `seq` HM.insert cid h m
-        return $! (h', q')
+        return (h', q')
 
     g :: JoinQueue a -> DiffItem BlockHeader -> JoinQueue a
     g q x = foldl' maybeInsert q $ zip (biList x) (biList (prioFun x))
--- a/src/Chainweb/CutDB.hs
+++ b/src/Chainweb/CutDB.hs
@@ -373,6 +373,7 @@ startCutDb config logfun headerStore payloadStore cutHashesStore = mask_ $ do
   where
     logg = logfun @T.Text
     wbhdb = _webBlockHeaderStoreCas headerStore
+    v = _chainwebVersion headerStore
 
     processor :: PQueue (Down CutHashes) -> TVar Cut -> IO ()
     processor queue cutVar = runForever logfun "CutDB" $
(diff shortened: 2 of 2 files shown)


larskuhtz:
Tested and confirmed that the fix works.

src/Chainweb/CutDB.hs
```diff
@@ -407,15 +408,13 @@ startCutDb config logfun headerStore payloadStore cutHashesStore = mask_ $ do
...
-                    hm' <- case _cutDbParamsInitialHeightLimit config of
+                Right hm -> unsafeMkCut v
+                    <$> case _cutDbParamsInitialHeightLimit config of
                         Nothing -> return hm
                         Just h -> limitCutHeaders wbhdb h hm
-                    logg Debug $ "joining intial cut from cut db configuration with cut that was loaded from the database " <> sshow hm'
-                    joinIntoHeavier_
-                        (_webBlockHeaderStoreCas headerStore)
-                        hm'
-                        (_cutMap $ _cutDbParamsInitialCut config)
```
larskuhtz: This was joining the persisted cut with the genesis cut. By joining the cuts, it was validated that the stored cut was actually a cut (correctly braided) and all headers in the history of the cut exist in the chain db.

The algorithm for joining cuts creates a heap the includes all headers in the symmetric difference of the involved cuts. In the case of the join with the genesis cut, this is all headers in the history of the chain, which at the current block height in mainnet requires > 8GB. Nodes with <=8GB would run out of memory and crash.

By removing this step we just trust the stored cut in the DB. Which should be fine, since we trust the db in any case.

There are more efficient (and more complete) ways to validate the integrity of the DB, which can be performed offline. It would also be possible to add support to do this in the background in a chainweb node. But that's beyond the scope of the PR.

src/Chainweb/CutDB.hs
```diff
@@ -407,15 +408,13 @@ startCutDb config logfun headerStore payloadStore cutHashesStore = mask_ $ do
...
+                            -- FIXME: this requires that the pact db is deleted because it
+                            -- interfers with rewind limit. As a fix temporarily disable
+                            -- rewind limit during initialization.
```
larskuhtz: This FIXME is unrelated to the change in this PR. But we discovered the issue while working on this PR. It's not urgent, since the initial cut limit is rarely used.

src/Chainweb/CutDB.hs
slpopejoy: The old check would seem not just to be RAM intensive but probably take a while, no?

Also is it truly redundant? Are there any modes where the check would alert of a corrupted node?

src/Chainweb/Cut.hs
```diff
@@ -434,7 +441,7 @@ join_ wdb prioFun a b = do
         db <- getWebBlockHeaderDb wdb cid
         (q' :> h) <- S.fold g q id $ branchDiff_ db x y
         let !h' = q' `seq` h `seq` HM.insert cid h m
-        return $! (h', q')
+        return (h', q')
```
slpopejoy: why remove the strictness?

src/Chainweb/CutDB.hs
larskuhtz: The check would capture some database corruptions but not all. I see little value in it. In particular, it the failure is just an invalid braiding, it would be corrected as soon as the next remote cut is received. Only, if all nodes would have invalidly braided cuts it would be an issue.

If a block header is missing, the node would either work normally, if the header isn't used or it would fail if it tries to access the header. Depending on the situation the node may or may not recover from it by querying the missing header from a remote node.

larskuhtz: On nodes with sufficient RAM, the test is surprisingly fast -- still it takes some time. The runtime seems to scales super-linear under RAM pressure, probably due to caching issues.

However, DB pruning seems to be the main runtime bottleneck during startup. In particular on nodes with low number of cores (since it's concurrent for chains).

src/Chainweb/Cut.hs
larskuhtz: It's redundant. `(,)` is a constructor and constructor application is already in weak head normal form, so it's always strict.

slpopejoy: but ... strictness theater! 😄

src/Chainweb/CutDB.hs
slpopejoy: I asked about time, b/c if nodes were taking this time, we could probably design a memory-efficient iterative version of this check, but it sounds like a different/more exhaustive check (and probably offline) is better.

fosskers:
I'm going to test this on my personal node this morning.

PR #985, opened 2020-03-30
Switch flags and nonce in Merke tree

larskuhtz:
This fixes #960. The former PR switched the names for feature flags and nonce. This was done in a way that the binary encoding and the Merkel root for old headers was unchanged.

However, the types of the Merkel tree entries guid the selection of leaf position of proof subject during proof creation. Thus proof creation would look for the subject in the wrong place in the tree.

This PR also switches the types of flags and nonce (and the tags in the MerkleUniverse) in the Merkel tree.

### DIFF
--- a/test/Chainweb/Test/Orphans/Internal.hs
+++ b/test/Chainweb/Test/Orphans/Internal.hs
@@ -123,7 +123,7 @@ instance Arbitrary BlockHeader where
     arbitrary = fromLog . newMerkleLog <$> entries
       where
         entries
-            = liftA2 (:+:) (Nonce <$> chooseAny)
+            = liftA2 (:+:) arbitrary
             $ liftA2 (:+:) arbitrary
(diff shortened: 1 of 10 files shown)


test/Chainweb/Test/Orphans/Internal.hs
```diff
@@ -133,7 +133,7 @@ instance Arbitrary BlockHeader where
             $ liftA2 (:+:) (BlockHeight . int @Int . getPositive <$> arbitrary)
             $ liftA2 (:+:) arbitrary
             $ liftA2 (:+:) arbitrary
-            $ liftA2 (:+:) arbitrary
+            $ liftA2 (:+:) (Nonce <$> chooseAny)
```
slpopejoy: Maybe an `Arbitrary` instance for `Nonce`?

PR #994, opened 2020-04-01
Remove redundant $!

fosskers:
`$!` has no (practical) affect when applied to a constructor.


fosskers:
The stack cache is being frequently evicted, I'm going to look into it next.

PR #995, opened 2020-04-01
Hlint suggestions to ignore

fosskers:
This reduces the output of `hlint` on our code base from 586 warnings to 421 (~30%).

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,10 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+- ignore: {name: "Redundant multi-way if"}
+- ignore: {name: "Redundant do"}
+- ignore: {name: "Reduce duplication"}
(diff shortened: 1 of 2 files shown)


larskuhtz:
I am not a big fan of adding .hlint to the repository. Isn't this something that everybody should configure themself, like editor integrations, emacs keybindings, etc?

.hlint.yaml
```diff
@@ -0,0 +1,10 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+- ignore: {name: "Redundant multi-way if"}
+- ignore: {name: "Redundant do"}
+- ignore: {name: "Reduce duplication"}
+- ignore: {name: "Eta reduce"}
```
larskuhtz, line 8: I like to be reminded of these -- even though I often ignore those. Really, I think, hlint is a tool for personal use. It isn't a good replacement for a developers of projected converging towards a common coding style.

PR #1000, opened 2020-04-08
LTS 15

fosskers:
This PR bumps our Stack config to use `lts-15`, and also validates https://github.com/kadena-io/pact/pull/769.


.github/workflows/stack-ci.yaml
```diff
@@ -31,9 +31,9 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ~/.stack
-          key: ${{ runner.os }}-stack-${{ hashFiles('stack.yaml') }}
+          key: ${{ runner.os }}-stack883-${{ hashFiles('stack.yaml') }}
```
larskuhtz, line 34: maybe use the LTS version as key?

chainweb.cabal
```diff
@@ -351,7 +351,7 @@ library
         , monad-control >= 1.0
         , mtl >= 2.2
         , mwc-random >= 0.13
-        , mwc-probability >= 2.0 && <2.2
+        , mwc-probability >= 2.0 && <2.3
```
larskuhtz, line 354: Is this upper bound needed at all? I am not aware of any issue with that package.

stack.yaml
```diff
@@ -1,47 +1,44 @@
...
-  - lens-aeson-1.1
-  - tls-1.5.4
-  - tls-session-manager-0.0.4
-  - warp-3.3.9
-  - warp-tls-3.2.11
-
   # --- Custom Pins --- #
   - github: kadena-io/pact
-    commit: f628486a99cdd22e635b98a2d8c631d303c3d4da
+    commit: 8ae08fda34b2e2f11c49f108659ade68451beeb6
```
larskuhtz, line 42: Could you also update the pin in `cabal.project` and for nix, if required?

chainweb.cabal
fosskers: It was just me lifting the restriction, I'm not sure that there was any API-related issue.

PR #1002, opened 2020-04-09
remove insert from from BlockHeaderDb

larskuhtz:
This PR deletes a lot of Code. It restricts insertion into the BlockHeaderDb, by 

* deleting the unused Rest API endpoint,
* removing the insert method from `TreeDb`, and
* moving inserting into `BlockHeaderDb` into `BlockHeaderDB.Internal`.

With these changes only "authorized" code is able to insert block headers into the `BlockHeaderDb` after fully validating the inserted header. Test code still can gain access by importing `BlockHeaderDB.Internal`.


src/Chainweb/BlockHeaderDB.hs
```diff
@@ -21,7 +16,10 @@ module Chainweb.BlockHeaderDB
 , closeBlockHeaderDb
 , withBlockHeaderDb
 
--- internal
+-- * insertion
+-- , insertBlockHeaderDb
+
+-- * internal
```
fosskers, line 22: Can these comments be removed?

src/Chainweb/BlockHeaderDB/Internal.hs
```diff
@@ -143,8 +145,6 @@ newtype BlockRank = BlockRank { _getBlockRank :: BlockHeight }
 -- -------------------------------------------------------------------------- --
 -- Internal
 
-type E = BlockHeader
-
```
fosskers, line 147: An ancient artifact from Chainweb Prehistory :)

PR #1003, opened 2020-04-10
Update chainweb-storage dependency and use HasCasLookup

larskuhtz:
This PR restricts the payload db to be readonly under most circumstances. It also fixes a bug in the `fakePact` test mock a few minor issues along the way.

It depends on https://github.com/kadena-io/chainweb-storage/pull/4. The nix build fails currently, because the chainweb-storage pin hasn't yet been updated.

PR #1004, opened 2020-04-12
Make BlockHeader validation aware of adjacent parents

larskuhtz:
This PR seems bigger than it actually is, since it includes #1002 and #1003. Once those PRs are merged into master, this PR will become smaller and will exit Draft status.

* [x] make block validation aware of adjacent headers. The validation rules themself are unchanged.
* [x] Make block validation safer and more maintainable.
    * [x] Move all block header validation logic into `BlockHeader.Validation`.
    * [x] delete module `TreeDB.Validation`.
    * [x] Enforce chain db invariants on type level (as long as no "unsafe" functions are used, which exist for tests)
* [x] Move `ChainValue` type into a module on its own
* [x] Improve WebBlockHeaderDB API
    * [x] add `HasCasLookUp` instance for WebBlockHeaderDB, that uses `ChainValue BlockHeader` as value type.
* [x] Improve test coverage for BlockHeader validation
* [x] Provide better tools for adding validation test cases for special block headers, like, for instance fork points.


src/Chainweb/BlockHeader/Validation.hs
```diff
@@ -273,12 +575,9 @@ validateInductiveM p e = unless (null failures)
 isValidBlockHeader
     :: Time Micros
         -- ^ The current clock time
-    -> ParentHeader
-        -- ^ parent block header. The genesis header is considered its own parent.
-    -> BlockHeader
-        -- ^ The block header to be checked
+    -> WebStep
```
fosskers, line 578: Nice.

src/Chainweb/BlockHeader/Validation.hs
```diff
@@ -414,67 +690,61 @@ prop_block_featureFlags b
...
+-- in the cut validation is stronger than the braiding property that we could
+-- check here (which is the property that is described in the chainweb paper).
+-- So, I check here would add some redundancy, but it's not clear of how much
+-- actual value that would be.
+--
+-- If we want to enforce it here, too, we would have to look back 2 steps in
+-- history.
+--
+prop_block_braiding :: WebStep -> Bool
+prop_block_braiding _ = True
```
fosskers, line 749: Remove if not checked?

larskuhtz: I kept it because of the comment above. It would be easy to include all grandparents into a WebStep.

Performance with it shouldn't make a big difference here in validation. The size of a WebStep is `O(diameter graph) == O(log(number_of_chains))` in any case. But it might affect performance of database lookups (because caches would grow by `O(order graph) == O(number_of_chains)` with respect to the depth in history. Although I am not sure how much it matters. It would also add overhead in creating test cases. 

Given that it's an redundant check, I am not sure if the added robustness pays off.

PR #1011, opened 2020-04-20
Move cut tests to Chainweb-tests and remove reflection from it

larskuhtz:


chainweb.cabal
```diff
@@ -491,6 +491,7 @@ test-suite chainweb-tests
         , data-ordlist >= 0.4.7
         , deepseq >= 1.4
         , directory >= 1.2
+        , errors >= 2.3
```
fosskers, line 494: I don't see this imported anywhere...

larskuhtz: It's used in `Chainweb.Test.Cut` (which was `Chainweb.Cut.Test` in the chainweb library)

PR #1017, opened 2020-04-21
get rid of reflection

larskuhtz:
The only remaining use of reflection was actually redundant. 

Most changes in this PR are just whitespace changes do to reindentation of a function body `test/Chainweb/Test/SPV.hs`.

### DIFF
--- a/test/Chainweb/Test/SPV.hs
+++ b/test/Chainweb/Test/SPV.hs
@@ -34,7 +34,6 @@ import Data.Foldable
 import Data.Functor.Of
 import qualified Data.List as L
 import Data.LogMessage
-import Data.Reflection hiding (int)
 import qualified Data.Vector.Unboxed as V
 
 import Numeric.Natural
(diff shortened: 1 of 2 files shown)


test/Chainweb/Test/SPV.hs
```diff
@@ -343,12 +342,12 @@ spvTransactionOutputRoundtripTest rdb v step = do
 
 type TestClientEnv_ cas = TestClientEnv MockTx cas
 
-apiTests :: RocksDb -> Bool -> ChainwebVersion -> TestTree
-apiTests rdb tls v = withTestPayloadResource rdb v 100 (\_ _ -> return ()) $ \dbIO ->
```
larskuhtz, line 347: This change is unrelated. The `tls` parameter was always `True` and thus both test case groups were actually doing the same.

test/Chainweb/Test/SPV.hs
```diff
@@ -359,49 +358,48 @@ apiTests rdb tls v = withTestPayloadResource rdb v 100 (\_ _ -> return ()) $ \db
 
 txApiTests :: PayloadCasLookup cas => IO (TestClientEnv_ cas) -> Step -> IO ()
 txApiTests envIO step = do
-    PayloadTestClientEnv env cutDb payloadDbs v <- envIO
-    give (snd . head $ payloadDbs) $ do
```
larskuhtz, line 363: Removing this line is the only real change here. The rest is just reindenting the function body.

PR #1024, opened 2020-04-24
slow down startup of devnet

larskuhtz:
slow down start of devnet. This is currently based 1.7 for testing with devnet bootstrapped at 1.7. After testing it can be merged into master.

### DIFF
--- a/src/Chainweb/Version.hs
+++ b/src/Chainweb/Version.hs
@@ -623,21 +623,18 @@ coinV2Upgrade Mainnet01 cid h
     | cid == unsafeChainId 8 = h == 140808
     | cid == unsafeChainId 9 = h == 140808
     | otherwise = error $ "invalid chain id " <> sshow cid
-coinV2Upgrade Development cid h
-    | cid == unsafeChainId 0 = h == 3
-    | otherwise = h == 4
 coinV2Upgrade _ _ 1 = True
 coinV2Upgrade _ _ _ = False
 
 -- | Preserve Pact bugs pre 1.6 chainweb version
 -- Mainnet 328000 ~ UTC Feb 20 15:36, EST Feb 20 10:56
 -- Devnet 1 hour of blocks
+--
 pactBackCompat_v16
   :: ChainwebVersion
   -> BlockHeight
   -> Bool
 pactBackCompat_v16 Mainnet01 h = h < 328000
-pactBackCompat_v16 Development h = h < 120
 pactBackCompat_v16 _ _ = False
 
 -- | If this is true the creation time of the current header is used for pact tx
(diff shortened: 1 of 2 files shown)


src/Chainweb/Version.hs
```diff
@@ -623,21 +623,18 @@ coinV2Upgrade Mainnet01 cid h
     | cid == unsafeChainId 8 = h == 140808
     | cid == unsafeChainId 9 = h == 140808
     | otherwise = error $ "invalid chain id " <> sshow cid
-coinV2Upgrade Development cid h
-    | cid == unsafeChainId 0 = h == 3
-    | otherwise = h == 4
```
larskuhtz, line 628: This breaks two test cases in `test/Chainweb/Test/Pact/TransactionTests.hs`. Those test should probably be ported to use chainweb version `Mainnet01`.

src/Chainweb/Version.hs
```diff
@@ -653,8 +650,6 @@ useLegacyCreationTimeForTxValidation
     -> BlockHeight
     -> Bool
 useLegacyCreationTimeForTxValidation Mainnet01 h = h < 449940 -- ~ 2020-04-03T00:00:00Z
-useLegacyCreationTimeForTxValidation Testnet04 h = h < 286231 -- 2020-04-02T02:00:00Z
-useLegacyCreationTimeForTxValidation Development h = h < 150
```
larskuhtz, line 657: Having this in devnet was annoying and slowing down testing. I don't thing we need to cover this in devnet any more.

src/Chainweb/Version.hs
```diff
@@ -736,8 +726,6 @@ slowEpochGuard _ _ = False
 --
 oldTargetGuard :: ChainwebVersion -> BlockHeight -> Bool
 oldTargetGuard Mainnet01 h = h < 452820 -- ~ 2020-04-04T00:00:00Z
-oldTargetGuard Testnet04 h = h < 286352 -- ~ 2020-04-02T03:00:00Z
-oldTargetGuard Development h = h < 360
```
larskuhtz, line 740: This isn't needed any more. We should simplify the logic behind this guard further in a follow up PR, by just skipping target validation for mainnet blocks below above height. Those blocks are grandfathered into the history of mainnet as witnessed by the top-level Merkle hash.

src/Chainweb/Version.hs
```diff
@@ -756,7 +744,6 @@ skipFeatureFlagValidationGuard
     -> BlockHeight
         -- ^ height of header
     -> Bool
-skipFeatureFlagValidationGuard Mainnet01 _ = True
-skipFeatureFlagValidationGuard Development _ = True
-skipFeatureFlagValidationGuard Testnet04 _ = True
+skipFeatureFlagValidationGuard Mainnet01 h = h < 530500  -- ~ 2020-05-01T00:00:xxZ
+skipFeatureFlagValidationGuard Development h = h < 140
```
larskuhtz, line 748: This causes validation failures. It's here only for testing the respective mainnet behavior. Those tests have been completed by now.

```suggestion
```

PR #1025, opened 2020-04-24
Devnet updates

larskuhtz:
Remove legacy logic from devnet. The changes only affect devnet.

### DIFF
--- a/src/Chainweb/Version.hs
+++ b/src/Chainweb/Version.hs
@@ -633,12 +633,14 @@ coinV2Upgrade Mainnet01 cid h
     | cid == unsafeChainId 8 = h == 140808
     | cid == unsafeChainId 9 = h == 140808
     | otherwise = error $ "invalid chain id " <> sshow cid
+coinV2Upgrade Development cid h
+    | cid == unsafeChainId 0 = h == 3
+    | otherwise = h == 4
 coinV2Upgrade _ _ 1 = True
 coinV2Upgrade _ _ _ = False
 
 -- | Preserve Pact bugs pre 1.6 chainweb version
 -- Mainnet 328000 ~ UTC Feb 20 15:36, EST Feb 20 10:56
--- Devnet 1 hour of blocks
 --
 pactBackCompat_v16 :: ChainwebVersion -> BlockHeight -> Bool
 pactBackCompat_v16 Mainnet01 h = h < 328000
--- a/src/Chainweb/BlockHeader.hs
+++ b/src/Chainweb/BlockHeader.hs
@@ -34,6 +34,7 @@ module Chainweb.BlockHeader
 (
 -- * Newtype wrappers for function parameters
   ParentHeader(..)
+, parentHeader
 , ParentCreationTime(..)
 
 -- * Block Payload Hash
@@ -363,7 +364,6 @@ newtype ParentHeader = ParentHeader
     deriving (Show, Eq, Ord, Generic)
     deriving anyclass (NFData)
 
-
 instance HasChainId ParentHeader where
     _chainId = _chainId . _parentHeader
     {-# INLINE _chainId #-}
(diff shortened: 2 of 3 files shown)


src/Chainweb/Version.hs
```diff
@@ -633,21 +633,15 @@ coinV2Upgrade Mainnet01 cid h
     | cid == unsafeChainId 8 = h == 140808
     | cid == unsafeChainId 9 = h == 140808
     | otherwise = error $ "invalid chain id " <> sshow cid
-coinV2Upgrade Development cid h
-    | cid == unsafeChainId 0 = h == 3
-    | otherwise = h == 4
```
larskuhtz, line 638: This breaks two test cases in test/Chainweb/Test/Pact/TransactionTests.hs. Those test should probably be ported to use chainweb version Mainnet01.

src/Chainweb/Version.hs
```diff
@@ -658,12 +652,8 @@ pactBackCompat_v16 _ _ = False
 -- dependency of `newBlock` on the creation time of the current header can be
 -- removed.
 --
-useLegacyCreationTimeForTxValidation
-    :: ChainwebVersion
-    -> BlockHeight
-    -> Bool
+useLegacyCreationTimeForTxValidation :: ChainwebVersion -> BlockHeight -> Bool
 useLegacyCreationTimeForTxValidation Mainnet01 h = h < 449940 -- ~ 2020-04-03T00:00:00Z
-useLegacyCreationTimeForTxValidation Development h = h < 150
```
larskuhtz, line 666: Having this in devnet was annoying and slowing down testing. I don't thing we need to cover this in devnet any more.

src/Chainweb/Version.hs
```diff
@@ -744,7 +726,6 @@ slowEpochGuard _ _ = False
 --
 oldTargetGuard :: ChainwebVersion -> BlockHeight -> Bool
 oldTargetGuard Mainnet01 h = h < 452820 -- ~ 2020-04-04T00:00:00Z
-oldTargetGuard Development h = h < 360
```
larskuhtz, line 747: This isn't needed any more. We should simplify the logic behind this guard further in a follow up PR, by just skipping target validation for mainnet blocks below above height. Those blocks are grandfathered into the history of mainnet as witnessed by the top-level Merkle hash.

src/Chainweb/Version.hs
```diff
@@ -755,12 +736,7 @@ oldTargetGuard _ _ = False
...
-skipFeatureFlagValidationGuard
-    :: ChainwebVersion
-    -> BlockHeight
-        -- ^ height of header
-    -> Bool
+skipFeatureFlagValidationGuard :: ChainwebVersion -> BlockHeight -> Bool
 skipFeatureFlagValidationGuard Mainnet01 h = h < 530500  -- ~ 2020-05-01T00:00:xxZ
-skipFeatureFlagValidationGuard Development _ = True
-skipFeatureFlagValidationGuard Testnet04 _ = False
+skipFeatureFlagValidationGuard Development h = h < 140
```
larskuhtz, line 741: This causes validation failures. It's here only for testing the respective mainnet behavior. Those tests have been completed by now.

```suggestion
```

mightybyte:
Since we've already tested the 1.8 release, let's save this one for 1.9.

larskuhtz:
This contains some of the changes that we used during testing on devnet

fosskers:
Approved, this doesn't affect mainnet codepaths.

src/Chainweb/Version.hs
```diff
@@ -755,12 +738,6 @@ oldTargetGuard _ _ = False
...
 -- historical blocks for which both the Nonce and Flags could be anything.
 --
-skipFeatureFlagValidationGuard
-    :: ChainwebVersion
-    -> BlockHeight
-        -- ^ height of header
-    -> Bool
+skipFeatureFlagValidationGuard :: ChainwebVersion -> BlockHeight -> Bool
 skipFeatureFlagValidationGuard Mainnet01 h = h < 530500  -- ~ 2020-05-01T00:00:xxZ
-skipFeatureFlagValidationGuard Development _ = True
```
larskuhtz, line 764: This change brings devnet in line with mainnet behavior > 530500

larskuhtz: The current devnet uses `h < 140` for testing the fork behavior. Those tests are done, so we can drop this.

PR #1028, opened 2020-04-25
Adds logging of transaction's gas logs

LindaOrtega:


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -726,6 +728,7 @@ gasInterpreter :: Gas -> TransactionM db (Interpreter p)
 gasInterpreter g = do
     mc <- use txCache
     return $ initStateInterpreter
+        $ set evalLogGas (Just [("GTxSize",g)]) -- enables gas logging
```
larskuhtz, line 731: Is this fine to merge or does it create a lot of perf overhead?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -493,6 +493,7 @@ applyExec
     -> TransactionM p (CommandResult [TxLog Value])
 applyExec interp em senderSigs hsh nsp = do
     EvalResult{..} <- applyExec' interp em senderSigs hsh nsp
+    debug $ "gas logs: " <> sshow _erLogGas
```
larskuhtz, line 496: How large can these get? Using JSON format instead of Haskell `Show` would make it easier to debug large logs, because there many tools for processing and pretty printing json, but (almost) none for Haskell `Show` outputs.

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: This doesn't. Although @LindaOrtega why is it `set` and not `over` with `(++)` or something? Are you sure `evalGasLog` is empty here?

src/Chainweb/Pact/TransactionExec.hs
slpopejoy: Not that big, example: 
```
gas logs: Just [("module:GWrite:WriteModule",46),("some-two-step-pact:GModuleMember",1),("G:GModuleMember",1),("module:GModuleDecl",1),("module:GWrite:WriteModule",240),("gas-payer-guard:GModuleMember",1),("create-gas-payer-guard:GModuleMember",1),("ALLOW_GAS:GModuleMember",1),("GAS_PAYER:GModuleMember",1),("ledger:GModuleMember",1),("gas:GModuleMember",1),("G:GModuleMember",1),("module:GModuleDecl",1),("success:GUserApp",1),("namespace:GPostRead:ReadNamespace",1),("namespace:GUnreduced",12),("GTxSize",18)]: "Cj7gyAFf-26TYpeng0u_NI0bA3NLcvgLenJkalfKak8"
```

PR #1029, opened 2020-04-25
Fix init module cache for interfaces

slpopejoy:


test/Chainweb/Test/Pact/PactReplay.hs
```diff
@@ -12,40 +12,32 @@ import Control.Concurrent.MVar
 import Control.Monad.Catch
```
slpopejoy, line 12: Modernized this test for no particular reason. Hoping to get rid of `withPact` and `mkTestExecTransactions` at some point.

test/Chainweb/Test/Pact/SPV.hs
```diff
@@ -32,6 +32,7 @@ import Control.Arrow ((***))
 import Control.Concurrent.MVar
```
slpopejoy, line 32: More test modernization, here to get rid of `mkTestExecXXX`

test/Chainweb/Test/Pact/ModuleCacheOnRestart.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE RecordWildCards #-}
```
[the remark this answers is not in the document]
slpopejoy: apparently not

PR #1035, opened 2020-04-28
Move benchmarks into bench component

larskuhtz:
This moves the benchmark modules into the bench component. It removes the associated compile time overhead and also removes some non-production dependencies from the the library.

The PR also updates the benchmark CI job to upload the benchmark results to S3 to make them available for analysis.

Some dead code is removed along the way and the dependency on fake is dropped.


larskuhtz:
The benchmarks are the same. Just the location of the modules changed. This remove some compile time overhead form the library and also removes some non-production dependencies from the library.

PR #1037, opened 2020-04-29
Adds Rosetta's network endpoints

LindaOrtega:
- [x] full `/network/list` endpoint
- [x] scaffolding for `network/options` endpoint
- [x] full `/network/status` endpoint

Follow-up work:
- [x] Make sure to validate the NetworkId in the relevant endpoint handlers.
- [ ] Make the version of the Rosetta API a variable in the Rosetta repo: https://github.com/kadena-io/rosetta/issues/7
- [x] ~~Potentially better structuring of the `RosettaFailure` type: Currently made a manual Enum/Bounded instance to iterate through it since `/network/options` requires a list of all possible errors the api will return.~~ EDIT: Colin made `RosettaFailure` all single constructor making the Enum and Bounded class derivable.
- [ ] Add in all `operation_statuses` the node allows for the `network/options` endpoint. Currently just an empty list.
- [ ] Add in all `operation_types` the node allows for the `network/options` endpoint. Currently just an empty list.

### DIFF
--- a/src/Chainweb/Rosetta/RestAPI.hs
+++ b/src/Chainweb/Rosetta/RestAPI.hs
@@ -22,7 +22,7 @@ module Chainweb.Rosetta.RestAPI
   , validateNetwork
   ) where
 
-import Control.Monad (unless)
+import Control.Monad (when)
 import Control.Monad.Except (throwError)
 import Control.Monad.Trans.Except (ExceptT)
 
@@ -87,8 +87,9 @@ data RosettaFailure
     | RosettaMempoolBadTx
     | RosettaUnparsableTx
     | RosettaInvalidTx
-    | RosettaBadNetwork
-    deriving (Enum, Bounded)
+    | RosettaInvalidBlockchainName
+    | RosettaMismatchNetworkName
+    deriving (Show, Enum, Bounded)
 
 -- TODO: Better grouping of rosetta error index
 rosettaError :: RosettaFailure -> RosettaError
@@ -97,11 +98,13 @@ rosettaError RosettaInvalidChain = RosettaError 1 "Invalid chain value" False
 rosettaError RosettaMempoolBadTx = RosettaError 2 "Transaction not present in mempool" False
 rosettaError RosettaUnparsableTx = RosettaError 3 "Transaction not parsable" False
 rosettaError RosettaInvalidTx = RosettaError 4 "Invalid transaction" False
-rosettaError RosettaBadNetwork = RosettaError 5 "Invalid network identifier" False
+rosettaError RosettaInvalidBlockchainName = RosettaError 5 "Invalid blockchain name" False
+rosettaError RosettaMismatchNetworkName = RosettaError 6 "Invalid Chainweb network name" False
 
 throwRosetta :: RosettaFailure -> Handler a
 throwRosetta e = throwError err500 { errBody = encode $ rosettaError e }
 
 validateNetwork :: Monad m => ChainwebVersion -> NetworkId -> ExceptT RosettaFailure m ()
-validateNetwork v (NetworkId bc n _) = unless (bc == "kadena" && Just v == fromText n)
-    $ throwError RosettaBadNetwork
+validateNetwork v (NetworkId bc n _) = do
+    when (bc /= "kadena") $ throwError RosettaInvalidBlockchainName
+    when (Just v /= fromText n) $ throwError RosettaMismatchNetworkName
--- a/src/Chainweb/Rosetta/RestAPI.hs
+++ b/src/Chainweb/Rosetta/RestAPI.hs
@@ -90,52 +90,6 @@ data RosettaFailure
     | RosettaBadNetwork
     deriving (Enum, Bounded)
 
--- instance Enum RosettaFailure where
---   succ RosettaChainUnspecified = RosettaInvalidChain defChainIdErrMsg
---   succ (RosettaInvalidChain _) = RosettaMempoolBadTx
---   succ RosettaMempoolBadTx = RosettaInvalidBlockchainName defBlockchainNameErrMsg
---   succ RosettaUnparsableTx = RosettaInvalidTx
---   succ RosettaInvalidTx = errorWithoutStackTrace "Prelude.Enum.Bool.succ: bad argument"
-
---   pred RosettaInvalidTx = RosettaUnparsableTx
---   pred RosettaUnparsableTx = RosettaMismatchNetworkName defChainwebVerErrMsg defNetworkNameErrMsg
---   pred RosettaMempoolBadTx = RosettaInvalidChain defChainIdErrMsg
---   pred (RosettaInvalidChain _) = RosettaChainUnspecified
---   pred RosettaChainUnspecified = errorWithoutStackTrace "Prelude.Enum.Bool.pred: bad argument"
-
---   toEnum x
---     | x == 0 = RosettaChainUnspecified
---     | x == 1 = RosettaInvalidChain defChainIdErrMsg
---     | x == 2 = RosettaMempoolBadTx
---     | x == 3 = RosettaInvalidBlockchainName defBlockchainNameErrMsg
---     | x == 4 = RosettaMismatchNetworkName defChainwebVerErrMsg defNetworkNameErrMsg
---     | x == 5 = RosettaUnparsableTx
---     | x == 6 = RosettaInvalidTx
---     | otherwise = errorWithoutStackTrace "Prelude.Enum.().toEnum: bad argument"
-
---   fromEnum RosettaChainUnspecified = 0
---   fromEnum (RosettaInvalidChain _) = 1
---   fromEnum RosettaMempoolBadTx = 2
---   fromEnum RosettaUnparsableTx = 5
---   fromEnum RosettaInvalidTx = 6
-
--- -- NOTE: Must update when new rosetta errors are added
--- instance Bounded RosettaFailure where
---   minBound = RosettaChainUnspecified
---   maxBound = RosettaInvalidTx
-
--- defChainIdErrMsg :: Text
--- defChainIdErrMsg = "someInvalidChainId"
-
--- defBlockchainNameErrMsg :: Text
--- defBlockchainNameErrMsg = "someInvalidBlockchainName"
-
--- defChainwebVerErrMsg :: ChainwebVersion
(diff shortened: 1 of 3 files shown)


src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -56,16 +60,16 @@ rosettaServer ms = (const $ error "not yet implemented")
...
     :<|> (const $ error "not yet implemented")
 
 someRosettaServer
     :: ChainwebVersion
     -> [(ChainId, MempoolBackend a)]
     -> SomeServer
-someRosettaServer (FromSingChainwebVersion (SChainwebVersion :: Sing vT)) ms =
-    SomeServer (Proxy @(RosettaApi vT)) $ rosettaServer ms
+someRosettaServer v@(FromSingChainwebVersion (SChainwebVersion :: Sing vT)) ms =
+    SomeServer (Proxy @(RosettaApi vT)) $ rosettaServer ms v
```
LindaOrtega, line 72: Is there a way to get the Chainweb Version from inside the `rosettaServer` function?

src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -119,3 +123,37 @@ mempoolTransactionH mtr ms = runExceptT work >>= either throwRosetta pure
 
 --------------------------------------------------------------------------------
 -- Network Handlers
+
+networkListH :: MetadataReq -> ChainwebVersion -> Handler NetworkListResp
+networkListH _ v
+  | (isRosettaEnabled v) = return $ NetworkListResp (networkIds v)
+  | otherwise = throwRosetta $ RosettaNotSupported v
```
LindaOrtega, line 130: Decided to throw an error here, but wanted to get opinion on returning an empty list instead?

src/Chainweb/Rosetta/RestAPI/Server.hs
fosskers: You seem to have accomplished that? Was there more you were looking for?

src/Chainweb/Rosetta/RestAPI/Server.hs
fosskers: `isRosettaEnabled` is probably unnecessary in general, since the node can decide if they turn these endpoints on or off. They're off by default.

src/Chainweb/Rosetta/RestAPI/Server.hs
LindaOrtega: It was purely an optional thing to not to pass in yet another parameter. Was just wondering if there was a way to access the chainweb node version from the state.

src/Chainweb/Rosetta/RestAPI/Server.hs
LindaOrtega: uhm I see what you mean. If Rosetta is disabled in the config, will trying to access `/rosetta/<endpoint>` return a 404 error?

src/Chainweb/Rosetta/RestAPI/Server.hs
fosskers: We pass is it in like this in the other servers, so no harm done.

src/Chainweb/Rosetta/RestAPI/Server.hs
fosskers: Yes.

src/Chainweb/Rosetta/RestAPI/Server.hs
LindaOrtega: that works then. Resolving comment

src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -151,9 +181,142 @@ mempoolTransactionH ms mtr = runExceptT work >>= either throwRosetta pure
...
+      }
+
+networkOptionsH :: ChainwebVersion -> NetworkReq -> Handler NetworkOptionsResp
+networkOptionsH v (NetworkReq nid _) = do
+  _ <- handle (enforceValidNetworkId v nid)
+  pure $ NetworkOptionsResp version allow
+  where
+    version = RosettaNodeVersion
+      { _version_rosettaVersion = "1.3.1"   -- TODO: Make this a variable in /rosetta repo
+      , _version_nodeVersion = chainwebNodeVersionHeaderValue :: T.Text
```
LindaOrtega, line 210: @larskuhtz do you suggest using the chainweb node version (i.e. 1.8) here? I've included the api version and the chainweb version in the meta data section as well. 
This field is supposed to be:
>The node_version is the canonical version of the node runtime. This can help clients manage deployments.

src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -139,7 +139,7 @@ constructionSubmitH v ms (ConstructionSubmitReq net@(NetworkId _ _ msni) tx) =
         SubNetworkId n _ <- msni ?? RosettaChainUnspecified
         cmd <- command tx ?? RosettaUnparsableTx
         validated <- hoistEither . first (const RosettaInvalidTx) $ validateCommand cmd
-        mp <- (readMaybe (T.unpack n) >>= flip lookup ms) ?? RosettaInvalidChain
```
LindaOrtega, line 117: (related to above comment) lines 139 and lines 142 perform the same checks as the validate network function.

src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -163,7 +163,7 @@ mempoolH v ms (MempoolReq net@(NetworkId _ _ msni)) =
     work = do
         validateNetwork v net
         SubNetworkId n _ <- msni ?? RosettaChainUnspecified
-        _ <- (readMaybe @ChainId (T.unpack n) >>= flip lookup ms) ?? RosettaInvalidChain
```
LindaOrtega, line 140: same as above. duplicate

src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -190,7 +190,7 @@ mempoolTransactionH v ms mtr = runExceptT work >>= either throwRosetta pure
     work = do
         validateNetwork v net
         SubNetworkId n _ <- msni ?? RosettaChainUnspecified
-        mp <- (readMaybe (T.unpack n) >>= flip lookup ms) ?? RosettaInvalidChain
```
LindaOrtega, line 167: duplicate

src/Chainweb/Rosetta/RestAPI.hs
```diff
@@ -85,74 +85,65 @@ data RosettaFailure
     = RosettaChainUnspecified
     | RosettaInvalidChain
     | RosettaMempoolBadTx
-    | RosettaInvalidBlockchainName Text
-    | RosettaMismatchNetworkName ChainwebVersion Text
```
LindaOrtega: If we're not providing dynamic error messages anymore, we should at least keep these two types of bad network errors to improve user debugging.

src/Chainweb/Rosetta/RestAPI.hs
```diff
@@ -80,14 +80,17 @@ type RosettaApi_ =
...
 
+-- TODO: Investigate if Rosetta Erros can be dynamic
 data RosettaFailure
     = RosettaChainUnspecified
     | RosettaInvalidChain
     | RosettaMempoolBadTx
     | RosettaUnparsableTx
     | RosettaInvalidTx
     | RosettaBadNetwork
+    deriving (Enum, Bounded)
```
LindaOrtega, line 91: was the motivation for making `RosettaFailure` single constructors to make enum/bounded derivations simpler?

LindaOrtega, line 91:
```suggestion
    deriving (Enum, Bounded,Show)
```
optional, but it makes debugging easier

src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -273,57 +286,39 @@ networkStatusH v cutDb peerDb (NetworkReq nid _) = do
...
-
-enforceValidNetworkId
-  :: ChainwebVersion
-  -> NetworkId
-  -> ExceptT RosettaFailure Handler ChainId
-enforceValidNetworkId v (NetworkId bn rv sid) = do
-  enforce isValidBlockchainName (RosettaInvalidBlockchainName bn)
-  enforce isValidNetworkVersion (RosettaMismatchNetworkName v rv)
-  SubNetworkId cidt _ <- sid ?? RosettaChainUnspecified
-  (readChainIdText v cidt) ?? (RosettaInvalidChain cidt)
```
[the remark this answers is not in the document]
fosskers: `NetworkId` won't always contain the chain id, and there are some requests that don't require it.

src/Chainweb/Rosetta/RestAPI.hs
fosskers: Sure, let's expand `validateNetwork`.

src/Chainweb/Rosetta/RestAPI.hs
fosskers: Yes :)

I think we should make the messages "total" in this way. We also don't really need to tell them the exact string values that caused the errors. They should know that already, since they were the ones that made the request.

LindaOrtega: gotcha! I was thinking this as well as I was writing the enum. As long as the error message is specific enough, they should be able to debug it.

src/Chainweb/Rosetta/RestAPI.hs
```diff
@@ -104,7 +109,18 @@ rosettaError RosettaMismatchNetworkName = RosettaError 6 "Invalid Chainweb netwo
 throwRosetta :: RosettaFailure -> Handler a
 throwRosetta e = throwError err500 { errBody = encode $ rosettaError e }
 
-validateNetwork :: Monad m => ChainwebVersion -> NetworkId -> ExceptT RosettaFailure m ()
-validateNetwork v (NetworkId bc n _) = do
+-- | Every Rosetta reqest that requires a `NetworkId` also requires a
```
LindaOrtega, line 112:
```suggestion
-- | Every Rosetta request that requires a `NetworkId` also requires a
```

PR #1040, opened 2020-05-01
Support changing the chain graph

larskuhtz:
The following is a list of files with more substantial changes. The changes to the remaining files are mostly straight forward consequences of those changes listed below.

**`Chainweb.Graph`**:

*   Removed "unsafe" function `toChainGraph` from export list.
*   `knownGraph` now returns `DiGraph Int` for a `KnownGraph`.
*   new function `knownChainGraph` returns the (memoized) `ChainGraph` for a `KnownGraph`

**`Chainweb.Version`**:

*   In order to test chain graph changes in `MultiNode` tests the chainweb version`TimedConsensus` now has two graph parameters with a transition occurring at block height 8.

*   Cleaned up and improved maintainability of the code related to the serialization of chainweb versions.

*   new functions:
    * `chainwebGraphs`
    * `genesisGraph`
    * `genesisHeight`
    * `chainGraphAt`
    * `chainGraphAt_`
    * `chainwebGraphsAt`

*   functions moved to `Chainweb.Version.Utils`
    *   `someChainId`
    *   `randomChainId`

**`Chainweb.Version.Utils`**:

*   new module with utilities related to the definitions in `Chainweb.Version` (to keep
     `Chainweb.Version` with its important definitions from growing to big).

**`Chainweb.Miner.Coordinator`**

*   new logic for `getAdjacentParents` that takes into account chain graph changes.
        
    *This is one of the major changes in code logic in of this PR*

**`Chainweb.Cut`**:
   
*   numerous changes to braiding validation (*WORK IN PROGRESS*)
*   new functions:
     *   `chainHeights`
     *   `meanChainHeight`
     *   `maxChainHeight`
     *   `minChainHeight`
     *   `isTransitionCut`

**`Chainweb.SPV.CreateProof`**:

*   A change that makes proof creation succeed even if there are graph changes in the history.

**`Chainweb.BlockHeader.Genesis`**:

*   new function: `genesisBlockHeadersAtHeight`

**`Chainweb.BlockHeight`**:

*   new data type: `CutHeight`

**`Chainweb.TreeDB`**:

*   new function: `minRank`

**`Chainweb.Test.Multinode`**:
  
*   Adjust test to perform graph change at block height 8.


src/Chainweb/SPV/CreateProof.hs
```diff
@@ -484,7 +488,13 @@ minimumTrgHeader headerDb tcid scid bh = do
             }
   where
     trgChain = headerDb ^?! ixg tcid
-    trgHeight = int bh + int (length path)
+    trgHeight
+        | srcGraph == trgGraph = int bh + int distance
+        | otherwise = int bh + 2 * (int distance)
+            -- This assumes that graph changes are at least graph-diameter
+            -- blocks appart.
+
```
larskuhtz, line 496: This changes guarantees that an SPV proofs can be constructed on histories with graph changes.

src/Chainweb/SPV/CreateProof.hs
```diff
@@ -484,7 +488,13 @@ minimumTrgHeader headerDb tcid scid bh = do
...
+    distance = length $ shortestPath tcid scid srcGraph
```
slpopejoy, line 497: This can be less than diameter

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -131,68 +151,69 @@ newWork
     :: LogFunction
     -> ChainChoice
     -> MinerStatus
+    -> WebBlockHeaderDb
+        -- ^ this is used to lookup parent headers that are not in the cut
+        -- itself.
```
larskuhtz, line 156: Before it was possible to mine a new block with just the current cut and a pact service (and assuming that the pact state was in sync with the cut, no db access was needed at all in theory). 

Now, creating a new block requires access to the parent headers of the current cut. That's a major change in the chainweb algorithm. I am not completely happy with it, but I don't see a way how this could be avoided.

The reason for this is that the target computation needs the creation times of the adjacent parent headers. Before chains where autonomous and no concrete information would flow from one chain to another. The only information that was transferred was the Merkel hash. Now the creation times from neighboring chains are used.

src/Chainweb/BlockHeader/Validation.hs
```diff
@@ -723,13 +723,20 @@ prop_block_chainId (ChainStep (ParentHeader p) b)
...
 
 prop_block_creationTime :: WebStep -> Bool
-prop_block_creationTime (WebStep _as (ChainStep (ParentHeader p) b))
-    | isGenesisBlockHeader b = _blockCreationTime b == _blockCreationTime p
-    | otherwise = _blockCreationTime b > _blockCreationTime p
+prop_block_creationTime (WebStep as (ChainStep (ParentHeader p) b))
+    | isGenesisBlockHeader b
+        = _blockCreationTime b == _blockCreationTime p
+    | fixedEpochStartGuard (_chainwebVersion b) (_blockHeight b)
+        = _blockCreationTime b > _blockCreationTime p
```
larskuhtz, line 736: This is the current legacy behavior.

src/Chainweb/BlockHeader/Validation.hs
```diff
@@ -723,13 +723,20 @@ prop_block_chainId (ChainStep (ParentHeader p) b)
...
+    | otherwise
+        = _blockCreationTime b > _blockCreationTime p
+        && all (\x -> _blockCreationTime b > _blockCreationTime (_parentHeader x)) as
```
larskuhtz, line 739: This new behavior is to require that the block creation time is strictly larger than the block creation time of the parent header *and* the block creation time of all adjacent parent headers

src/Chainweb/BlockHeader/Validation.hs
```diff
@@ -723,13 +723,20 @@ prop_block_chainId (ChainStep (ParentHeader p) b)
 -- Mutli chain inductive properties
 
 prop_block_epoch :: WebStep -> Bool
-prop_block_epoch (WebStep _as (ChainStep p b))
-    = _blockEpochStart b == epochStart p (_blockCreationTime b)
+prop_block_epoch (WebStep as (ChainStep p b))
+    = _blockEpochStart b <= EpochStartTime (_bct $ _blockCreationTime b)
+    && _blockEpochStart (_parentHeader p) <= _blockEpochStart b
+    && _blockEpochStart b == epochStart p as (_blockCreationTime b)
```
larskuhtz, line 729: This adds some sanity invariants for the epoch start value. A "sane" DA algorithm should guarantee these. Our current DA certainly assumes these properties.

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -105,15 +105,15 @@ emptyPayload = PayloadWithOutputs mempty miner coinbase h i o
...
-genesisTime Mainnet01 = BlockCreationTime [timeMicrosQQ| 2019-10-30T00:01:00.0 |]
+genesisTime :: ChainwebVersion -> ChainId -> BlockCreationTime
+genesisTime Test{} _ = BlockCreationTime epoch
+genesisTime TimedConsensus{} _ = BlockCreationTime epoch
+genesisTime PowConsensus{} _ = BlockCreationTime epoch
+genesisTime TimedCPM{} _ = BlockCreationTime epoch
+genesisTime FastTimedCPM{} _ = BlockCreationTime epoch
+genesisTime Testnet04 _ = BlockCreationTime [timeMicrosQQ| 2019-07-17T18:28:37.613832 |]
+genesisTime Mainnet01 _ = BlockCreationTime [timeMicrosQQ| 2019-10-30T00:01:00.0 |]
+genesisTime Development _ = BlockCreationTime [timeMicrosQQ| 2019-07-17T18:28:37.613832 |]
```
larskuhtz, line 116: This adds a chainId parameter. Note that the actual value didn't change.

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -201,8 +200,20 @@ genesisBlockHeader' v p ct@(BlockCreationTime t) n = fromLog mlog
...
 -- | This is an expensive call, try not to repeat it.
+--
 genesisBlockHeaders :: ChainwebVersion -> HM.HashMap ChainId BlockHeader
 genesisBlockHeaders v = HM.fromList
     . fmap (id &&& genesisBlockHeader v)
     . toList
     $ chainIds v
+
+    -- TODO define chainIdsAtHeight
+    -- (and use max graph for chainIds)
```
larskuhtz, line 211:
```suggestion
```

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,404 @@
...
+--    other chains.
+--
+-- 3. This is a "normal" transition.
+--
+-- TODO: it is important that the semantics of this function corresponds to the
+-- respective validation in the module "Chainweb.Cut", in particular
+-- 'isMonotonicCutExtension'. It must not happen, that a cut passes validation
+-- that can't be further extended.
+--
+getCutExtension
```
larskuhtz, line 159: The logic of this function should be merged with the respective code used by `Chainweb.Cut.monotonicCutExtension` and similar functions.

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,404 @@
...
+-- 3. Call pact service to create new payload for the new block
+-- 4. Assemble preliminary work header, with preliminary time and nonce value
+--    (time should be current time, nonce should be 0) and send work header
+--    to the miner.
+-- 5. Receive resolved header (with final nonce and creation time) from miner
+-- 6. Create new header by computing the merkle hash for the resolved header
+-- 7. Extend cut with new header by calling 'tryMonotonicCutExtension'
+-- 8. Create new cut hashes with the new header and payload attached.
+-- 9. Submit cut hashes (along with attached new header and payload) to cut
+--    validation pipeline.
```
larskuhtz, line 32: Update this comment. Also, create a woke-age for this.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -94,7 +114,7 @@ newtype PrimedWork =
 -- the current block's payload.
 --
 newtype MiningState = MiningState
-    (M.Map (T2 BlockHash BlockPayloadHash) (T3 Miner PayloadWithOutputs (Time Micros)))
+    (M.Map BlockPayloadHash (T3 Miner PayloadData (Time Micros)))
```
larskuhtz, line 117: It is not clear to me why the parent block hash is used in the key. I am not sure if the parent header (or some other unique value like the block height or creation time) actually gets included in the payload hash (probably not). But even if the payload isn't unique for a block (which may, unfortunately, be the case), we wouldn't care here.

(Also, using it as first component slows down the comparison, because many different payloads share the same parent hash, e.g. each miner has one payload hash for each parent. If we are concerned about memory cache performance in case of very large maps, we should use the block height instead, but I don't think that's an issue at this point.)

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -20,71 +22,89 @@
...
+
 -- -------------------------------------------------------------------------- --
 -- Miner
 
 -- | Precached payloads for Private Miners. This allows new work requests to be
 -- made as often as desired, without clogging the Pact queue.
 --
 newtype PrimedWork =
-    PrimedWork (HM.HashMap MinerId (HM.HashMap ChainId (Maybe PayloadWithOutputs)))
+    PrimedWork (HM.HashMap MinerId (HM.HashMap ChainId (Maybe PayloadData)))
```
larskuhtz, line 107: The outputs of a new block are not needed. All that is used is the payload hash and the `PayloadData`. The latter is attached to the new cut hashes before it is submitted to the cut pipeline.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -20,71 +22,89 @@
...
+  MiningState(..)
+, MiningStats(..)
+, PrevTime(..)
+, ChainChoice(..)
+, PrimedWork(..)
+, MinerStatus(..), minerStatus
+  -- * Functions
+, newWork
+, publish
+) where
```
larskuhtz, line 35: A lot of the logic for creating a new block headers and extending a cut got moved from this module to `Chainweb.Cut.Create`, where all functionality for extending a cut is centralized, now.

src/Chainweb/Sync/WebBlockHeaderStore.hs
```diff
@@ -313,9 +317,11 @@ getBlockHeaderInternal headerStore payloadStore candidateHeaderCas candidatePayl
...
         -- created.
         --
-        let queryAdjacentParent p = Concurrently $ void $ do
-                logg Debug $ taskMsg k $ "getBlockHeaderInternal.getPrerequisteHeader (adjacent) for " <> sshow h <> ": " <> sshow p
-                getBlockHeaderInternal headerStore payloadStore candidateHeaderCas candidatePayloadCas priority maybeOrigin' p
+        let isGenesisParentHash p = _chainValueValue p == genesisParentBlockHash v p
+            queryAdjacentParent p = Concurrently $ unless (isGenesisParentHash p) $ void $ do
+
+                    logg Debug $ taskMsg k $ "getBlockHeaderInternal.getPrerequisteHeader (adjacent) for " <> sshow h <> ": " <> sshow p
+                    getBlockHeaderInternal headerStore payloadStore candidateHeaderCas candidatePayloadCas priority maybeOrigin' p
```
larskuhtz, line 324: Fix indentation?

src/Chainweb/Version.hs
```diff
@@ -483,61 +459,138 @@ instance HasChainwebVersion ChainwebVersion where
...
+-- Invariants:
+--
+-- * Entries are sorted by 'BlockHeight' in decreasing order.
+-- * The last entry is for 'BlockHeight' 0.
+-- * The graphs decrease in order.
+--
+-- The functions provided in 'Chainweb.Version.Utils' are generally more
+-- convenient to use than this function.
+--
+chainwebGraphs :: ChainwebVersion -> NE.NonEmpty (BlockHeight, ChainGraph)
```
slpopejoy, line 500: I'm not understanding why we have `NonEmpty (BlockHeight, ChainGraph)` here and `Map BlockHeight ChainGraph` in Utils. Both approaches lean on invariants to validate unsafe operations: here the usages of `NE.dropWhile`, in Utils `fromJuste . lookup`. 

Wouldn't it be better to have one so that the codebases of this and Utils would overlap more? `Map` seems like the better fit IMO

- the small size (and unsafety acceptance) justifies using `lookupMin`, vs `NE.head` here. 
- `fromJuste . lookupLE` seems far better than `NE.dropWhile ((> h) . fst)`

src/Chainweb/Version/Utils.hs
```diff
@@ -0,0 +1,434 @@
...
+
+-- -------------------------------------------------------------------------- --
+--  Utils
+
+limitHeight :: BlockHeight -> M.Map BlockHeight a -> M.Map BlockHeight a
+limitHeight h = M.takeWhileAntitone (h >=)
+{-# INLINE limitHeight #-}
+
+atHeight :: HasCallStack => BlockHeight -> M.Map BlockHeight a -> a
+atHeight h = snd . fromJuste . M.lookupLE h
```
slpopejoy, line 84: Probably worth documenting unsafety here and `atCutHeight` since the unsafety applies to all the code calling these

src/Chainweb/Version/Utils.hs
```diff
@@ -0,0 +1,434 @@
...
+nextGraphChange v h = fst . fromJuste . M.lookupGT h $ chainGraphs v
+{-# INLINE nextGraphChange #-}
+
+chainCountAt
+    :: HasCallStack
+    => HasChainwebVersion v
+    => v
+    -> BlockHeight
+    -> Natural
+chainCountAt v h = order $ atHeight h $ chainGraphs v
```
slpopejoy, line 150: Suggest refactor `atHeight h $ chainGraphs v` to shared function and use in the 4 places here

src/Chainweb/Version/Utils.hs
```diff
@@ -0,0 +1,434 @@
...
+
+-- | BlockHeight intervals for the chain graphs of a chainweb version
+--
+-- Post-condition:
+--
+-- @
+-- 0 == minimum $ M.keys $ chainGraphs v
+-- @
+--
+chainGraphs
```
slpopejoy, line 102: See note in `Version`, seems like we should standardize on `Map` as otherwise we're creating a `NonEmpty`, destructing it and creating a `Map` for every operation in here.

src/Chainweb/Version/Utils.hs
```diff
@@ -0,0 +1,434 @@
...
+blockCountAt
+    :: HasCallStack
+    => HasChainwebVersion v
+    => HasChainId cid
+    => v
+    -> cid
+    -> BlockHeight
+    -> Natural
+blockCountAt v cid h
+    | h < gh = error $ "Chainweb.Version.Utils.blockCount: requested block height " <> sshow h
```
slpopejoy, line 232: Suggest one of two things:
- return 0 for < genesis heights, keeps function pure 
- make function safe (return `Maybe Natural`)
Depending on usage site requirements and frequency of use. ie if this is just for difficulty and we expect this to never happen, a `0` is acceptable and cleaner.

src/Chainweb/Version/Utils.hs
```diff
@@ -0,0 +1,434 @@
...
+    . chainGraphs
+{-# INLINE chainGraphsByCutHeight #-}
+
+-- chainGraphAtCutHeight = atCutHeight h
+
+-- | The chain graph at the given cut height
+--
+-- Note, that the result isn't accurate during a chain graph change
+--
+chainGraphAtCutHeight :: HasChainwebVersion v => v -> CutHeight -> ChainGraph
```
slpopejoy, line 298: Do we need all the versions of everything for both CutHeight and BlockHeight? Would suggest eliding unused stuff

src/Chainweb/Version.hs
```diff
@@ -483,61 +459,138 @@ instance HasChainwebVersion ChainwebVersion where
...
+-- | Return the Graph History at a given block height in descending order.
+--
+-- The functions provided in 'Chainweb.Version.Utils' are generally more
+-- convenient to use than this function.
+--
+chainwebGraphsAt
+    :: HasCallStack
+    => ChainwebVersion
+    -> BlockHeight
+    -> NE.NonEmpty (BlockHeight, ChainGraph)
```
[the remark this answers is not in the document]
larskuhtz: `chainwebGraphs v` is guaranteed to have an entry for which `((> h) . fst)` is `False`. So, this function is safe.

The invariant is documented above for `chainwebGraphs v`: 

> * The last entry is for 'BlockHeight' 0.

There's also a unit test that checks this invariant for all chainweb versions. And in this case the test provides exhaustive coverage. If it passes it is guaranteed that the invariant holds.

I'll add this to the documentation.

src/Chainweb/Version.hs
```diff
@@ -483,61 +459,138 @@ instance HasChainwebVersion ChainwebVersion where
...
+chainwebGraphs (TimedCPM g) = pure (0, g)
+chainwebGraphs (FastTimedCPM g) = pure (0, g)
+chainwebGraphs Testnet04 = pure (0, petersonChainGraph)
+chainwebGraphs Mainnet01 = pure (0, petersonChainGraph)
+chainwebGraphs Development = pure (0, petersonChainGraph)
+{-# INLINE chainwebGraphs #-}
+
+-- chainwebGlobalBlockRates :: ChainwebVersion -> NE.NonEmpty (BlokHeight, Natural)
+-- chainwebGlobalBlockRates =
+
```
larskuhtz, line 513:
```suggestion
```

src/Chainweb/Version.hs
larskuhtz: I am not so concerned about the invariants here. Those invariants can be tested very well, and there are tests for each of them. Also, I'd rather like the this code to crash hard with an error (most likely during testing or immediately when a node is started) than somehow continuing to run if something goes wrong with these definitions. 

I think, it's an example where partial functions (with `error`) are actually making the code safer. It's the closest we get to enforcing the invariants statically, without using dependent types.

larskuhtz: I am using `List.NonEmpty` here because, I expect that it will give very efficient access to the head of the list, which will probably be 99.99% of all use cases. It also supports a concise human readable encoding in the code without using tricks like Template Haskell. I think, both is important for this module, which I consider the heart of chainweb.

I consider the functions in `Chainweb.Version.Utils` less mission critical. The use of `Map` was based on convenience and concise code, and the shortest path to a working implementation. Otherwise, it's total overkill and probably inefficient. There is a chance that we have replace the use of `Map` if we find that it is too inefficient, which would be fine.

src/Chainweb/Version/Utils.hs
larskuhtz: This is just an internal helper, which is supposed to be called only when it's safe. It's not exported. I hope, the unsafely doesn't leak to any exported function.

I'll add a comment that clarifies it.

src/Chainweb/Version/Utils.hs
larskuhtz: Yeah, it's possibly inefficient. However, I think the implementation is optimized and fuses well. This module uses `Map` not as a data structure but more as an efficient search algorithm.

After writing a lot of specialized code for dealing with intervals I found that using map was just the fastest and safest way to get to a working implementation quickly.

Map handles a lot of nasty corner cases of the required traversals. The explicit implementations in my first approach sometimes got pretty complex due to those corner cases and there is some chance that Map isn't that bad in comparison, even performance wise.

But there is some chance that after profiling we'll have to revisit this module and replace the use of Map by more efficient implementations.

src/Chainweb/Version/Utils.hs
larskuhtz: Yes, 0 seems like the correct result.

On the other hand, I sometimes prefer a clean crash with an informative error message over computations continuing under possibly unexpected assumptions. At least during development and debugging, it was very helpful having those errors, that would trigger when some invariant was violated.

src/Chainweb/Version/Utils.hs
larskuhtz: Probably not all of it. However, I wrote this module after going through several cycles of adding, deleting, and rewriting utils while working downstream in the code base. So, I made this more exhaustive to note have to go back and forth.

I propose that we keep it until the graph-transition project finished and come back and clean it up when we are done with it. I think, that may also be a good time to revisit the decision of using `Map` (and `List.NonEmpty` in `Chainweb.Version`).

src/Chainweb/Version/Utils.hs
larskuhtz: After thinking a bit more about it, I think, a good solution could be, to keep using `Data.List.NonEmpty` in `Chainweb.Version` for specifying the graphs and to memoize the result of this function for the three production chainweb versions `mainet01`, `development`, and `testnet04` here.

Something like:

```Haskell
chainGraphs :: HasChainwebVersion v => v -> M.Map BlockHeight ChainGraph
chainGraphs v = case _chainwebVersion v of
    Mainnet01 -> mainnet01GraphMap
    Testnet04 -> testnet04GraphMap
    Development -> developmentGraphMap
    x -> M.fromList . toList $ chainwebGraphs x

mainnet01GraphMap :: M.Map BlockHeight ChainGraph
mainnet01GraphMap = M.fromList . toList $ chainwebGraphs Mainnet01
{-# NOINLINE mainnet01GraphMap #-}

testnet04GraphMap :: M.Map BlockHeight ChainGraph
testnet04GraphMap = M.fromList . toList $ chainwebGraphs Testnet04
{-# NOINLINE testnet04GraphMap #-}

developmentGraphMap :: M.Map BlockHeight ChainGraph
developmentGraphMap = M.fromList . toList $ chainwebGraphs Development
{-# NOINLINE developmentGraphMap #-}
```

That has the benefit of keeping the specification in `Chainweb.Version` simple and maintainable and providing an efficient Map in this module. It would also allow us to do further optimization (e.g. using arrays/vector instead of Map) without "disturbing" `Chainweb.Version`.

src/Chainweb/Miner/Coordinator.hs
slpopejoy: :+1:

src/Chainweb/Miner/Coordinator.hs
slpopejoy: We should think about whether "nodes supporting multiple miners" is a feature we want to continue to support. In this case I support moving to whatever the lookup needs, which this looks good for.

slpopejoy: As a general note, I don't prefer to use `T3` in a core datastructure. It's one thing to use tuples in processing code but in this case a proper type with a constructor and fields would be self-documenting.

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,442 @@
...
+--
+getCutExtension
+    :: HasCallStack
+    => HasChainId cid
+    => Cut
+        -- ^ the cut which is to be extended
+    -> cid
+        -- ^ the header onto which the new block is created. It is expected
+        -- that this header is contained in the cut.
+    -> Maybe CutExtension
```
slpopejoy, line 171: Consider making this an `Either XXX CutExtension` so we can see why it failed, given there are multiple failure modes.

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,442 @@
...
+    --
+    newAdjHashes :: ChainGraph -> Maybe (HM.HashMap ChainId BlockHash)
+    newAdjHashes g =
+        imapM (\xcid _ -> hashForChain xcid)
+        $ HS.toMap
+        $ adjacentChainIds g p
+
+    hashForChain acid
+        -- existing chain
+        | Just b <- c ^? ixg acid = tryAdj b
```
slpopejoy, line 211:
```suggestion
        | [something legible] = tryAdj b
```
1. I had to look up `<-`. Not a fan, although that's not what's motivating my complaint.
2. Can we get a utility function for the ubiquitous `c ^? ixg foo`? It makes reading this codebase painful.

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,442 @@
...
+    isGraphTransitionPre = isGraphChange c targetHeight
+    isGraphTransitionPost = isGraphChange c parentHeight
+    isGraphTransitionCut = isTransitionCut c
+
+    -- | Try to get all adjacent hashes dependencies for the given graph.
+    --
+    newAdjHashes :: ChainGraph -> Maybe (HM.HashMap ChainId BlockHash)
+    newAdjHashes g =
+        imapM (\xcid _ -> hashForChain xcid)
+        $ HS.toMap
```
slpopejoy, line 206:
```suggestion
        $ HS.toMap -- converts to Map Foo ()
```
Consider, as this is an unfamiliar/weird library function

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -20,71 +22,89 @@
...
+-- error if the lookup fails.
+--
+-- Generally, failing lookup in a cut is a code invariant violation. In almost
+-- all circumstances there should be a invariant in scope that guarantees that
+-- the lookup succeeds. This function is useful when debugging corner cases of
+-- new code logic, like graph changes.
+--
+lookupInCut :: HasCallStack => HasChainId cid => Cut -> cid -> BlockHeader
+lookupInCut c cid
+    | Just x <- c ^? ixg (_chainId cid) = x
```
slpopejoy, line 94: See comment elsewhere about a utility function for this, as well as it's `^?!` variant

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,442 @@
...
+            <> ". Parent: " <> encodeToText (ObjectEncoded p)
+
+    tryAdj :: BlockHeader -> Maybe BlockHash
+    tryAdj b
+
+        -- When the block is behind, we can move ahead
+        | _blockHeight b == targetHeight = Just $! _blockParent b
+
+        -- if the block is ahead it's blocked
+        | _blockHeight b + 1 == parentHeight = Nothing -- chain is blocked
```
slpopejoy, line 226: The comment above about returning an `Either XXX` would mean a lot of these comments could instead be code ie `Left "chain is blocked"`

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,442 @@
...
+
+        -- If this is not a graph transition cut we can move ahead
+        | _blockHeight b == parentHeight = Just $! _blockHash b
+
+        -- The cut is invalid
+        | _blockHeight b > targetHeight = error $ T.unpack
```
slpopejoy, line 232: ... and these could be `Left` instead of `error`

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,442 @@
...
+-- the cut and matches the given payload data. It also checks a few other
+-- invariants. It does't do a full validation.
+--
+-- The resulting 'CutHashes' value contains the new header and payload as
+-- attachement and can be submitted into a cut pipeline.
+--
+-- The result is 'Nothing' if the given cut can't be extended with the solved
+-- work.
+--
+extend
```
slpopejoy, line 399:
```suggestion
extendCut
```

slpopejoy: (and obviously rename internal function to `extendCutInternal` or whatever)

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,442 @@
...
+--    to the miner.
+-- 5. Receive resolved header (with final nonce and creation time) from miner
+-- 6. Create new header by computing the merkle hash for the resolved header
+-- 7. Extend cut with new header by calling 'tryMonotonicCutExtension'
+-- 8. Create new cut hashes with the new header and payload attached.
+-- 9. Submit cut hashes (along with attached new header and payload) to cut
+--    validation pipeline.
+--
+--
+module Chainweb.Cut.Create
```
slpopejoy, line 37: Perhaps `Mine` instead of `Create`? It would seem cut creation occurs (in production) strictly in the context of mining, and it would motivate including mining-specific types like `WorkHeader` and `SolvedWork` in here.

src/Chainweb/Cut.hs
```diff
@@ -185,42 +201,115 @@ cutWeight :: Getter Cut BlockWeight
...
+cutAdjChainIds :: HasChainId cid => Cut -> cid -> HS.HashSet ChainId
+cutAdjChainIds c cid = c ^?! ixg (_chainId cid) . blockAdjacentChainIds
+{-# INLINE cutAdjChainIds #-}
+
+-- | /Directed/ adjacent pairs of a cut. Note that block headers in a cut can be
+-- of different block height and can thus use different chain graphs.
+--
+_cutAdjPairs :: Cut -> [(BlockHeader, BlockHeader)]
+_cutAdjPairs c =
+    [ (h, x)
```
slpopejoy, line 250: The multiline list comprehension is too weird here. Maybe use `do`?

src/Chainweb/Cut.hs
```diff
@@ -185,42 +201,115 @@ cutWeight :: Getter Cut BlockWeight
...
+
+-- | Returns whether a chain graph transition occurs within the cut.
+--
+isTransitionCut :: Cut -> Bool
+isTransitionCut c = minChainHeight c < lastGraphChange c (maxChainHeight c)
 
 -- -------------------------------------------------------------------------- --
 -- Limit Cut Hashes By Height
 
+-- | This an internal function. The result is meaningful only if the input
```
slpopejoy, line 289: Should it be `projectChainsUnsafe`?

src/Chainweb/Cut.hs
```diff
@@ -305,48 +401,86 @@ instance Exception CutException
...
+-- @
+--
+-- The actual implementation is a it more complex because headers of different
+-- height in a cut may use a diferent chain graph during the transition from one
+-- graph to another graph: instead of undirected adjacent pairs the condition
+-- consideres directed adjacent pairs.
+--
+-- A corresponding function is implemente in 'Chainweb.Miner.Coordinator'
+-- TODO: provide test that the implementation in the miner matches with this
+-- function.
```
slpopejoy, line 468: I couldn't find this in `Coordinator`, is this comment still true?

src/Chainweb/Cut.hs
```diff
@@ -408,19 +555,32 @@ data Join a = Join
     , _joinQueue :: !(JoinQueue a)
     }
 
+-- | This computes the join for cuts accross all chains.
```
slpopejoy, line 558: Can we get some docs on the `Join` datatype?

src/Chainweb/Cut.hs
```diff
@@ -497,21 +660,38 @@ applyJoin m = foldM
 -- the intersection. 'tryMonotonicCutExtension' is guaranteed to maintain that
 -- property.
 --
+-- Chains that aren't yet initialized are included in the join and later
+-- stripped from the result.
+--
+-- If you want to compute a join for cuts that include only a subset of all
```
slpopejoy, line 666: Suggest a `see comment on 'join' regarding subsets` here and below to save on comment maintenance

src/Chainweb/BlockHeader/Genesis.hs
slpopejoy, line 116: Can you move this back to L114 ie above `Testnet04` as before? 
Confirmed times did not change

src/Chainweb/BlockHeader/Validation.hs
```diff
@@ -723,13 +723,20 @@ prop_block_chainId (ChainStep (ParentHeader p) b)
 -- Mutli chain inductive properties
```
slpopejoy, line 723:
```suggestion
-- Multi chain inductive properties
```

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -6,7 +6,9 @@
 {-# LANGUAGE ScopedTypeVariables #-}
```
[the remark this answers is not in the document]
larskuhtz: That code was unmaintainable in it's previous state. I refactored it in the process of trying to understand the code. Indentation was completely inconsistent and there was almost no obvious structure in the organization of the code. I first tried to edit without refactoring, but I wasn't able to comfortably make changes to it -- which I had to do. 

I spent a lot of time understanding and carefully refactoring that file and I am confident that the changes are correct.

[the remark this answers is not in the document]
larskuhtz: I guess, that just too many people have made different changes and additions to this module over time and it had grown to a state where it was simply not possible any more to understand what was going on.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -205,7 +205,7 @@ localTest iot nio = do
     mv <- newMVar 0
     SubmitBatch batch <- testBatch iot mv gp
     let cmd = head $ toList batch
-    sid <- mkChainId v (0 :: Int)
+    sid <- mkChainId v maxBound (0 :: Int)
```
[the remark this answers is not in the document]
larskuhtz: I made this change just mechanically. Yeah, I think, `unsafeChainId 0` would make sense.

test/golden/development-block-hashes-expected.txt
```diff
@@ -1 +1 @@
-28TrKiSqox6bt0dId6XIp_YZYL44m3DfvP_wNrXMMe0=mbZ9UUdGdpR3kvc5h1ZaXEwSEUvrHe2uj0KoTB7HAZQ=-DCYOM_qGqbIiAS_Uju61RAJX4JqmlWyeXVZZIsv7bI=YDcGkFNEauWqmYSaBQ0t6exbUVfoOi09ggal6RBr1U0=-eMD-a3DlmMBjVmLqqIhN3U7dxU1SQX_aWh2XfqFBAI=Ew3Oh-wbrZTlyOpTLneizNIK_n0DHTCmUIBvlsXavpc=yg25Ke7Ztuu08twyc6-XnAGx3NJVIUFyW_imtfTMq2Y=oFWsZDdyzAtxWG_gX-JT2aJoI905LTK85LOdvEN8qCc=p32ZnyiLcvIcUpJ4V4EOEPKcErNI30AxA9kfG3f8v7E=PPpKjQQgOEp0QEr-mqLta7kaezdrllM63TLm3KDttEU=
```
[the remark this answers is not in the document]
larskuhtz: The PR adds a graph change to devnet, which adds additional genesis block hashes.

test/Chainweb/Test/Mempool/Consensus.hs
```diff
@@ -342,7 +344,7 @@ header' h = do
             :+: BlockWeight (targetToDifficulty target) + _blockWeight h
             :+: succ (_blockHeight h)
             :+: v
-            :+: epochStart (ParentHeader h) t'
+            :+: epochStart (ParentHeader h) mempty t'
```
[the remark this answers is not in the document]
larskuhtz: No, it's not clear to me. It seems that this module is creating its own chain with a fork. There are test utils for this. Not sure if there is a reason for not using those other than that they are a bit difficult to discover and use.

test/Chainweb/Test/Pact/Checkpointer.hs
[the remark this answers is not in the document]
slpopejoy: You don't seem to get my point. You chose to reformat _a massive regression test for a major component_ from top to bottom, rendering it unreviewable. Any arguments about how much you disliked it are not relevant. The code is not "unmaintainable", a fact proven by numerous devs who have maintained it up until now.

> I spent a lot of time understanding and carefully refactoring that file and I am confident that the changes are correct.

That's nice, but that doesn't help review, and review is not optional. I have no idea what changed, or even what _needed to change_, given that this PR shouldn't have to disturb Checkpointer too much. How do we resolve this? 

I'm considering reverting the changes myself to see what broke, but you must agree that is a slow and painful review process.

test/golden/development-block-hashes-expected.txt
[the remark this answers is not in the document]
slpopejoy: Is it additional hashes? Or is it adjacency changes? I would think additional alone wouldn't affect a single-chain test.

test/Chainweb/Test/Mempool/Consensus.hs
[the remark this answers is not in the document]
slpopejoy: OK well maybe let's flag this with a TODO to replace with more maintainable mechanisms.

src/Chainweb/BlockHeader.hs
```diff
@@ -276,53 +279,106 @@ powTarget
...
-            | otherwise -> _blockTarget ph
+powTarget p@(ParentHeader ph) bct = case effectiveWindow ph of
+    Nothing -> maxTarget
+    Just w
+        -- A special case for starting a new devnet. Using maxtarget results in
+        -- a too high block production and consecutively orphans and network
+        -- congestion. The consequence are osciallations to take serval hundred
+        -- blocks before the system stabilizes. This code cools down initial
+        -- block production.
+        | ver == Development && _blockHeight ph == (genesisHeight ver cid + 1) -> HashTarget (maxBound `div` 100000)
```
slpopejoy, line 290:
```suggestion
        | ver == Development && _blockHeight ph == ((genesisHeight ver cid) + 1) -> HashTarget (maxBound `div` 100000)
```

slpopejoy: Also this seems like the only change yes? The rest is reformatting/reordering of cases?

src/Chainweb/BlockHeader.hs
```diff
@@ -276,53 +279,106 @@ powTarget
...
+    --
+    --    How can we adjust for that?
+    --
+    --    a) using a constant factor that is based on properties of the
+    --       geometric distribution.
+    --    b) adjusting timeBlocked in both directions, possibly based off the
+    --       avg?
+    --
+    timeBlocked = maximum adjCreationTimes .-. _blockCreationTime p
+        -- the maximum is at least @_blockCreationTime p@ and thus the result is
```
slpopejoy, line 375: Shouldn't this comment go above `timeBlocked` with the previous comment?

src/Chainweb/BlockHeader.hs
```diff
@@ -276,53 +279,106 @@ powTarget
...
     | isLastInEpoch p = EpochStartTime (_bct $ _blockCreationTime p)
-    | ver == Development && _blockHeight p == 1 = EpochStartTime (_bct $ _blockCreationTime p)
-        -- this is a special case for starting a new devnet. The creation time
-        -- of the development genesis blocks way in the past which would cause
-        -- the first /and/ second epochs to be at max target. By using the first
-        -- block instead of the genesis block for computing the first epoch time
-        -- we slow down mining eariler.
-    | otherwise = _blockEpochStart p
+
+    -- Within epoch
```
slpopejoy, line 346:
```suggestion
    -- Within epoch but before fixed epoch start guard
```

src/Chainweb/BlockHeader.hs
```diff
@@ -276,53 +279,106 @@ powTarget
...
+    | fixedEpochStartGuard ver (_blockHeight p + 1) = _blockEpochStart p
+    | otherwise = _blockEpochStart p .+^ timeBlocked
```
slpopejoy, line 348: Add comment e.g. `-- After fixed epoch start: adjust for blocked time`

src/Chainweb/BlockHeader.hs
```diff
@@ -710,6 +766,10 @@ _blockAdjacentChainIds =
 blockAdjacentChainIds :: Getter BlockHeader (HS.HashSet ChainId)
 blockAdjacentChainIds = to _blockAdjacentChainIds
 
+-- | @getAdjacentHash cid h@ returns the adjacent hash of h for chain cid. It
+-- throws a @ChainNotAdjacentException@ if @cid@ is not adajcent with @_chainId
+-- h@ in the chain graph of @h@.
```
slpopejoy, line 771: Haddocks quesstion: is `@h@` the same as `'h'`? If so I find the latter much more readable in code ...

src/Chainweb/BlockHeader.hs
```diff
@@ -722,7 +782,8 @@ computeBlockHash h = BlockHash $ MerkleLogHash $ computeMerkleLogRoot h
 {-# INLINE computeBlockHash #-}
 
 isGenesisBlockHeader :: BlockHeader -> Bool
-isGenesisBlockHeader b = _blockHeight b == BlockHeight 0
+isGenesisBlockHeader b
```
slpopejoy, line 785: please move `=` to this line

src/Chainweb/BlockHeader.hs
```diff
@@ -745,6 +806,13 @@ timeBetween after before = f after - f before
     f :: BlockCreationTime -> Micros
     f (BlockCreationTime (Time (TimeSpan ts))) = ts
 
+-- | Absolute BlockHeight Difference
+--
+absBlockHeightDiff :: BlockHeader -> BlockHeader -> BlockHeight
+absBlockHeightDiff a b
```
slpopejoy, line 812:
```suggestion
absBlockHeightDiff a b = abs $ _blockHeight a - _blockHeight b
```
unless I'm missing something?

test/Chainweb/Test/Cut.hs
```diff
@@ -142,92 +205,64 @@ testMine'
...
     -> Cut
-    -> Either MineFailure (T2 BlockHeader Cut)
-createNewCut n t pay i c = do
-    h <- note BadAdjacents $ newHeader . BlockHashRecord <$> newAdjHashes
-    unless (checkTarget (_blockTarget h) $ _blockPow h) $ Left BadNonce
-    c' <- first (\e -> error $ "Chainweb.Cut.createNewCut: " <> sshow e)
-        $ monotonicCutExtension c h
+    -> m (T2 BlockHeader Cut)
+createNewCut hdb n t pay i c = do
+    extension <- fromMaybeM BadAdjacents $ getCutExtension c i
```
slpopejoy, line 243: I must say `fromMaybeM` is a terrible name. `fromMaybe` doesn't cause/throw an error. Please change this to a more descriptive name. (It's sibling `???` is also pretty bad, and used like twice, so I would prefer we just ditch it).

test/Chainweb/Test/Cut.hs
```diff
@@ -142,92 +205,64 @@ testMine'
...
-    tryAdj :: BlockHeight -> BlockHeader -> Maybe BlockHash
-    tryAdj h b
-        | _blockHeight b == h = Just $! _blockHash b
-        | _blockHeight b == h + 1 = Just $! _blockParent b
-        | otherwise = Nothing
 
 -- | Create a new cut where the new block has a creation time of one second
 -- after its parent.
 --
-createNewCutWithoutTime
```
slpopejoy, line 218: This is only used once, here. Maybe we should ditch this alternate and inline the call to `createNewCut`.

test/Chainweb/Test/Cut.hs
larskuhtz: The name is consistent with the convention throughout the chainweb code base that functions that end with `M` handle failures within `MonadThrow` (I think the `M` was originally inspired by the `M` in `throwM`).

This function has been in the code base since the very beginning. I think changing it is beyond the scope of this PR, but go into a PR on its own.

test/Chainweb/Test/Cut.hs
```diff
@@ -88,31 +91,93 @@ import qualified Test.QuickCheck.Monadic as T
...
+            --
+            -- A "real" miner would inject the nonce and time without first
+            -- decoding the header and would hand over the header in serialized
+            -- form.
+            --
+            & encodeBlockHeaderWithoutHash
+            & runPut
+            & runGet decodeBlockHeaderWithoutHash
+            & fromJuste
+            & SolvedWork
```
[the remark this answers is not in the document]
larskuhtz: `&` is defined in `base`. 

I find this sequence easier to read in top down direction. Using `$` would result in the exact same code lines just in reversed order (which makes inlining comments less readable because the reader has to change direction and it's not that obvious if the comment refers to the line above or below.

src/Chainweb/Miner/Core.hs
```diff
@@ -177,6 +183,9 @@ injectNonce :: Nonce -> Ptr Word8 -> IO ()
 injectNonce (Nonce n) buf = pokeByteOff buf 278 n
 {-# INLINE injectNonce #-}
 
+injectTime :: Time Micros -> Ptr Word8 -> IO ()
```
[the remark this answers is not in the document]
larskuhtz: https://github.com/kadena-io/chainweb-node/wiki/Block-Header-Binary-Encoding#work-header-binary-format

src/Chainweb/BlockHeader.hs
larskuhtz: Why should the parenthesis be added? There is no other operator that could cause a precedence issue. Also, I think hlint would probably complain (although, I don't care that much about hlint...).

> Also this seems like the only change yes? The rest is reformatting/reordering of cases?
Yes

src/Chainweb/BlockHeader.hs
larskuhtz, line 348:
```suggestion
    -- Within an epoch when legacy 'fixedEpochStartGuard' does not apply
    | otherwise = _blockEpochStart p .+^ timeBlocked
```

src/Chainweb/BlockHeader.hs
larskuhtz: In the code base we use the convention that comments that are indented refer to the code above. I put this comment separate, because the comment above is about the overall algorithm, while this comment is an implementation detail.

In any case, this code, including the comment will change in the next iteration of the 20-chain work.

src/Chainweb/BlockHeader.hs
larskuhtz: `@...@` and `'...'` have different meaning in Haddock. `'...'` is for references to functions defined in the code , for which a hyper link is created. `@...@` is for arbitrary text that is formatted verbatim. 

I think, recent versions of Haddock all support markdown style back-tick quotation for verbatim text, but I haven't tried that yet and have been sticking to the backward compatible `@...@` notation. Maybe it's time to try  out back-tick quotes and switch to that syntax if it works for our builds?

src/Chainweb/BlockHeader.hs
larskuhtz: `BlockHeight` is a newtype wrapper around an unsigned type. Haskell uses "unsafe" C semantics for primitive types:

```
λ> abs $ BlockHeight 1 - BlockHeight 2
18446744073709551615
λ> abs $ BlockHeight 2 - BlockHeight 1
1
```

src/Chainweb/BlockHeader.hs
```diff
@@ -722,7 +782,8 @@ computeBlockHash h = BlockHash $ MerkleLogHash $ computeMerkleLogRoot h
 {-# INLINE computeBlockHash #-}
 
 isGenesisBlockHeader :: BlockHeader -> Bool
-isGenesisBlockHeader b = _blockHeight b == BlockHeight 0
+isGenesisBlockHeader b
+    = _blockHeight b == genesisHeight (_blockChainwebVersion b) (_blockChainId b)
```
larskuhtz, line 786:
```suggestion
isGenesisBlockHeader b =
    _blockHeight b == genesisHeight (_blockChainwebVersion b) (_blockChainId b)
```

src/Chainweb/BlockHeader/Genesis.hs
larskuhtz, line 116:
```suggestion
genesisTime Development _ = BlockCreationTime [timeMicrosQQ| 2019-07-17T18:28:37.613832 |]
genesisTime Testnet04 _ = BlockCreationTime [timeMicrosQQ| 2019-07-17T18:28:37.613832 |]
genesisTime Mainnet01 _ = BlockCreationTime [timeMicrosQQ| 2019-10-30T00:01:00.0 |]
```

bench/Chainweb/Pact/Backend/ForkingBench.hs
```diff
@@ -237,16 +241,20 @@ mineBlock parentHeader nonce pdb bhdb r = do
...
+            { _workHeaderBytes = BS.toShort $ runPut $ encodeBlockHeaderWithoutHash bh
+            , _workHeaderChainId = _chainId bh
+            , _workHeaderTarget = _blockTarget bh
+            }
 
-     T2 (HeaderBytes new) _ <- usePowHash testVer (\p -> mine p (_blockNonce bh) tbytes) hbytes
-     newHeader <- runGet decodeBlockHeaderWithoutHash new
+     T2 (SolvedWork newHeader) _ <- usePowHash testVer $ \(_ :: Proxy a) -> mine @a (_blockNonce bh) work
 
      mv' <- validateBlock newHeader (payloadWithOutputsToPayloadData payload) r
```
slpopejoy, line 259:
```suggestion
     mv' <- validateBlock (newHeader { _blockCreationTime = creationTime}) (payloadWithOutputsToPayloadData payload) r
```

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -192,7 +191,7 @@ genesisBlockHeader' v p ct@(BlockCreationTime t) n = fromLog mlog
         :+: genesisBlockPayloadHash v cid
         :+: cid
         :+: BlockWeight 0
-        :+: BlockHeight 0
+        :+: genesisHeight v cid
```
larskuhtz, line 194:
```suggestion
        :+: genesisHeight v cid -- because of chain graph changes (new chains) not all chains start at 0
```

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -205,7 +205,7 @@ localTest iot nio = do
...
     res <- local sid cenv cmd
```
larskuhtz, line 209:
```suggestion
    res <- local (unsafeChainId 0) cenv cmd
```

test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -56,7 +56,8 @@ import Chainweb.Payload.PayloadStore.InMemory (newPayloadDb)
 import Chainweb.Test.Pact.Utils
 import Chainweb.Test.Utils
 import Chainweb.Transaction
-import Chainweb.Version (ChainwebVersion(..), someChainId)
+import Chainweb.Version (ChainwebVersion(..))
+import Chainweb.Version.Utils (someChainId)
```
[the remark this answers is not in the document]
larskuhtz: 👍 let's keep reorganizing the content of `Chainweb.Version` and `Chainweb.Version.Utils` in a follow up PR.

test/Chainweb/Test/MultiNode.hs
```diff
@@ -34,7 +35,7 @@
 module Chainweb.Test.MultiNode ( test ) where
 
 #ifndef DEBUG_MULTINODE_TEST
-#define DEBUG_MULTINODE_TEST 0
+#define DEBUG_MULTINODE_TEST 1
```
larskuhtz, line 38:
```suggestion
#define DEBUG_MULTINODE_TEST 0
```

src/Chainweb/Version/Utils.hs
larskuhtz: hmm, I tried it, but decided that it wasn't worth it: `chainGraphsAtHeight v h` is not that much different than `atHeight h $ chainGraphs v` but adds an level of indirection when understanding the code.

src/Chainweb/Version/Utils.hs
```diff
@@ -0,0 +1,460 @@
...
+blockCountAt
+    :: HasCallStack
+    => HasChainwebVersion v
+    => HasChainId cid
+    => v
+    -> cid
+    -> BlockHeight
+    -> Natural
+blockCountAt v cid h
+    | h < gh = error $ "Chainweb.Version.Utils.blockCount: requested block height " <> sshow h
```
larskuhtz, line 258:
```suggestion
    | h < gh = 0
```

src/Chainweb/Version.hs
larskuhtz: As discussed offline, this will most likely be replaced by plain `[]` in a future PR that is going to address the relation of `Chainweb.Version` and `Chainweb.Version.Utils` in more detail.

src/Chainweb/Sync/WebBlockHeaderStore.hs
```diff
@@ -141,8 +145,8 @@ instance HasChainwebVersion WebBlockHeaderStore where
     _chainwebVersion = _chainwebVersion . _webBlockHeaderStoreCas
     {-# INLINE _chainwebVersion #-}
 
-instance HasChainGraph WebBlockHeaderStore where
-    _chainGraph = _chainGraph . _webBlockHeaderStoreCas
+instance HasChainGraph (WebBlockHeaderStore, BlockHeight) where
```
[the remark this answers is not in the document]
larskuhtz: I just deleted it.

src/Chainweb/Sync/WebBlockHeaderStore.hs
```diff
@@ -313,9 +317,11 @@ getBlockHeaderInternal headerStore payloadStore candidateHeaderCas candidatePayl
         -- prerequesite in the memo-table it is awaited, otherwise a new job is
         -- created.
         --
-        let queryAdjacentParent p = Concurrently $ void $ do
-                logg Debug $ taskMsg k $ "getBlockHeaderInternal.getPrerequisteHeader (adjacent) for " <> sshow h <> ": " <> sshow p
-                getBlockHeaderInternal headerStore payloadStore candidateHeaderCas candidatePayloadCas priority maybeOrigin' p
+        let isGenesisParentHash p = _chainValueValue p == genesisParentBlockHash v p
+            queryAdjacentParent p = Concurrently $ unless (isGenesisParentHash p) $ void $ do
```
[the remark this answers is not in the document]
larskuhtz: Yes. It would require careful refactoring of the code around line 345. 

The overhead of creating a thread in the Haskell runtime is really minimal and there is some chance that the guard can be floated out during optimization, so, I think, it's not worth refactoring the code at this point.

src/Chainweb/Pact/Backend/Utils.hs
```diff
@@ -55,7 +55,7 @@ import Chainweb.Pact.Service.Types
 
 runBlockEnv :: MVar (BlockEnv SQLiteEnv) -> BlockHandler SQLiteEnv a -> IO a
 runBlockEnv e m = modifyMVar e $
-  \(BlockEnv  dbenv bs)  -> do
+  \(BlockEnv dbenv bs)  -> do
```
larskuhtz, line 58:
```suggestion
  \(BlockEnv dbenv bs) -> do
```

src/Chainweb/Miner/Miners.hs
```diff
@@ -84,28 +84,33 @@ localTest lf v tpw m cdb gen miners = runForever lf "Chainweb.Miner.Miners.local
     loop :: IO a
     loop = do
         c <- _cut cdb
-        T2 bh pl <- newWork lf Anything (Plebian m) pact tpw c
```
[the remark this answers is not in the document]
larskuhtz: It's only used for testing. I just simulates mining:

```
work height w = do
         MWC.geometric1 t gen >>= threadDelay
         runGet decodeSolvedWork $ BS.fromShort $ _workHeaderBytes w
```

src/Chainweb/Miner/Core.hs
```diff
@@ -118,38 +109,53 @@ usePowHash Mainnet01 f = f $ Proxy @Blake2s_256
...
+                        -- Compute POW hash for the nonce
+                        injectNonce n buf
+                        hash ctx buf pow
+
+                        -- check whether the nonce meets the target
+                        fastCheckTarget trgPtr (castPtr pow) >>= \case
+                            True -> Nothing <$ writeIORef nonces (nv - o)
+                            False -> go1 (i - 1) (Nonce $ nv + 1)
+
+                -- outer loop
```
larskuhtz, line 138:
```suggestion
                -- outer loop
                -- Estimates how many iterations of the inner loop run in one second. It runs the inner loop
                -- that many times and injects an updated creation time in each cycle.
```

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -201,70 +222,52 @@ chainChoice c choice = case choice of
...
 
-        -- Fail Early: If the `BlockHeader` is already stale and can't be
-        -- appended to the best `Cut` we know about, reject it.
-        --
-        c' <- tryMonotonicCutExtension c bh
-            !? T2 miner "Mined block for outdated Cut"
+        -- Publish CutHashes to CutDb and log success
+        Right (bh, Just ch) -> do
+            addCutHashes cdb ch
+            let bytes = foldl' (\acc (Transaction bs) -> acc + BS.length bs) 0 $
```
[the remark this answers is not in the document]
larskuhtz: Yeah, strictness doesn't seem required here. The result is forced in line 246.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -233,14 +233,15 @@ doLookupBlock dbenv (bheight, bhash) = runBlockEnv dbenv $ do
...
-doGetBlockParent dbenv (bh, hash) =
-    if bh == 0 then return Nothing else
-      doLookupBlock dbenv (bh, hash) >>= \blockFound ->
-        if not blockFound then return Nothing else runBlockEnv dbenv $ do
-          r <- callDb "getBlockParent" $ \db -> qry db qtext [SInt (fromIntegral (pred bh))] [RBlob]
-          case r of
-            [[SBlob blob]] -> either (internalError . T.pack) (return . return) $! Data.Serialize.decode blob
-            _ -> internalError "doGetBlockParent: output mismatch"
+doGetBlockParent dbenv (bh, hash) = do
+    blockFound <- doLookupBlock dbenv (bh, hash)
```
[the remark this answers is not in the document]
larskuhtz: Actually, not sure. I think, I hoped that for genesis blocks `doLookupBlock` would return `not blockFound`.

Should, we call `error` instead? Well, it would probably result in `internalError` in any case. So, far this hasn't triggered.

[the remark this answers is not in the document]
larskuhtz: The problem is that within the checkpointer we don't have the chainweb version and/or the chain id available, so we don't know what the genesis height is.

Since the checkpointer is just and function dictionary, with no explicit state, there's no place where we could store this kind of information.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -801,8 +801,9 @@ initSchema = do
 
 getEndingTxId :: BlockHeight -> BlockHandler SQLiteEnv TxId
 getEndingTxId bh = callDb "getEndingTxId" $ \db -> do
-    if bh == 0
-      then return 0
+    if bh == 0 -- FIXME: what if genesisHeight > 0 for the chain?
+      then error "Chainweb.Pact.Backend.ChainwebPactDb.getEndingTxId: called with block height 0"
```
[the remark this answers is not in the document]
larskuhtz: I turned this into an error, to figure out if the code relies on it. If it does, it now would be a failure, because genesis height isn't necessarily 0.

Actually, we always had explicit `isGenesisBlock` and `isGenesisHash` predicates, because we were anticipating new chains. But, we didn't enforce their use throughout the code base.

We now have the problem that within pact service backend we don't have access to the chainweb version and the chain id, so there's no way to figure out the genesis height.

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -123,23 +130,38 @@ workHandler' mr mcid m = do
...
+            -- Fail Early: If a `BlockHeader` comes in that isn't associated with any
+            -- Payload we know about, reject it.
+            --
+            let key = _blockPayloadHash hdr
+            MiningState ms <- liftIO $ readTVarIO tms
+            T3 m pd _ <- case M.lookup key ms of
+                Nothing -> throwError err404 { errBody = "No associated Payload" }
+                Just x -> return x
+
+            liftIO $ NoContent <$ do
```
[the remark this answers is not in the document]
larskuhtz: `NoContent` is the servant constructor for empty HTTP bodies. The solve API and point is a POST call that returns an empty response.

src/Chainweb/CutDB.hs
```diff
@@ -495,10 +494,10 @@ processCuts conf logFun headerStore payloadStore cutHashesStore queue cutVar = q
...
     hdrStore = _webBlockHeaderStoreCas headerStore
 
-    graph = _chainGraph headerStore
-
-    threshold :: Int
-    threshold = int $ 2 * diameter graph * order graph
+    threshold :: Cut -> Int
+    threshold c = int $ 2 * diameter graph * order graph
+      where
+        graph = chainGraphAt_ headerStore (meanChainHeight c)
```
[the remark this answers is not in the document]
larskuhtz: It doesn't really matter here, since it's a heuristics and it only makes a difference during a transition. But given how the new transition algorithm works, minimum height would actually make more sense I think. I'll change it.

src/Chainweb/Cut/Create.hs
```diff
@@ -0,0 +1,442 @@
...
+    { _cutExtensionCut' :: !Cut
+        -- This is overly restrictive, since the same cut extension can be
+        -- valid for more than one cut. It's fine for now.
+    , _cutExtensionParent' :: !ParentHeader
+    , _cutExtensionAdjacentHashes' :: !BlockHashRecord
+    }
+
+makeLenses ''CutExtension
+
+_cutExtensionCut :: CutExtension -> Cut
```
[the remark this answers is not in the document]
larskuhtz: There is a validation related invariant associated with CutExtension, which, I think, justifies to hide the constructors and force the use of smart constructors. I usually don't mind boilerplate that much. It's a small one-time overhead when writing the code.

I wish the lens package would provide TH splices for deriving just getters. Not sure why it's not included with lens.

src/Chainweb/Cut/Create.hs
larskuhtz: I think, internally, this function is actually just a coercion. I think, `HashSet a` is implemented as `HashMap a ()`.

src/Chainweb/Cut/Create.hs
larskuhtz: This code is meant to be the first draft of what should become the consensus side of API for mining code.

The mining code should not need to now about the internal of block headers. In particular it shouldn't have to deal with the construction of Merkel trees and Cut braidings. The knowledge of the Merkel trees and braidings should be limited to a small core of the consensus code.

The mining API only receives an `Work` item and all it knows is 

1. where to inject the nonce,
2. where to inject the creation time,
3. how to compute the POW hash,
4. compare the POW hash with the target. 

Otherwise the work item should be opaque to the miner. Ideally, I `Work` item would be just 

1. Initial Nonce,
2. Initial time,
3. some random bytes (a cryptographic hash of the remaining data in the header), and
4. possibly some feature flags.

At this point the (3.) part is spread out in between the other parts -- something we may want to fix at some point.

When a miner has found a valid nonce, it returns the result as `SolvedWork` item.

The code in the module will change in upcoming PRs.

src/Chainweb/Cut/Create.hs
larskuhtz: Changing the result to `Either` makes sense, although mostly for debugging and logging purposes.

This function will be refactored in an upcoming PR. Most likely it will be merged with the respective function in the Cut module. For now it returns `Maybe` because that's what the existing Coordinator was expecting.

src/Chainweb/Cut/Create.hs
larskuhtz: At this point these cases are intentionally `error` / assertions. If these trigger, it would be a bug in the code. At runtime, if a caller would trigger any of these cases, all they could do would be to throw an error themself. The detailed error message were only for me when I was debugging the chain graph transition algorithm during development.

This may change, when this code get integrated with the the braiding validation logic in the Cut module. There it may be possible that an attacker injects bad cuts and triggers these cases, in which case we want to raise more meaningful braiding validation exceptions.

So, this will be addressed in upcoming PRs.

src/Chainweb/Cut/Create.hs
larskuhtz: I like to defer this to the above mentioned follow up PR.

src/Chainweb/Cut.hs
larskuhtz: hmm, I find `do` notation for lists much harder to read -- and I imaging that most non-Haskeller programmers would find the same. But in the end it's pretty similar syntax that desugars to the same monadic code. I'll change it to `do` notation.

src/Chainweb/Cut.hs
larskuhtz: I think that might be overkill. This module is full of invariants about the internal structure of cuts. Most likely, we would have to prefix all non-exported functions with "unsafe" :-)

src/Chainweb/Cut.hs
larskuhtz, line 468:
```suggestion
-- A corresponding function is currently implemente in 'Chainweb.Cut.Create'
-- TODO: unify the implementation of both. In particular make sure that the function here that is used
-- during validation is not more lenient than the function used for extending cuts for mining.
```

larskuhtz: It's not true any more. I updated the comment.

src/Chainweb/Cut.hs
larskuhtz: Let's postpone this. It's a rabbit whole. But feel free to remind me to not forget about it. Better documenting the theory behind cuts and the implementation of it has been on my TODO for a while now.

Actually, `Chainweb.Test.Cut` has quite a bit of documentation around the properties of cuts and in particular the lattice structure. But it's very technical and hard to understand without more background at this point.

larskuhtz: I added brief comments to the data type and its constructor fields. But I am not sure how helpful that actually is.

src/Chainweb/Cut.hs
larskuhtz: I tent duplicate Haddock documentation, because many tools, like `ghci`s `:doc` command, or also pop-up windows in editors only show the documentation for the function itself and don't offer ways to easily follow cross-refs.

But I am not sure how important that is for these details...

src/Chainweb/BlockHeader.hs
slpopejoy: It's just confusing to read, but I can live with it. In general I don't like arithmetic code that relies heavily on precedence for comprehension. In this case it wasn't clear if it was `(cid + 1)` as opposed to `(genesisHeight... + 1)`. Resolving

src/Chainweb/TreeDB.hs
```diff
@@ -389,6 +390,10 @@ root :: TreeDb db => db -> IO (DbEntry db)
 root db = fromJuste <$> entries db Nothing (Just 1) Nothing Nothing S.head_
 {-# INLINE root #-}
 
+minRank :: TreeDb db => db -> IO Natural
```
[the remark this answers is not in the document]
slpopejoy: Please don't resolve a review question without an answer. `minRank` is unused outside of this module, so you need to justify the export, as if it were to not be used later in this module, it would be dead code.

chainweb.cabal
```diff
@@ -467,6 +468,8 @@ test-suite chainweb-tests
         Chainweb.Test.TreeDB
         Chainweb.Test.TreeDB.RemoteDB
         Chainweb.Test.Utils
+        Chainweb.Test.Version
+        Chainweb.Test.Utils.BlockHeader
```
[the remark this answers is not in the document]
slpopejoy: Was this resolved?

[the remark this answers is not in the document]
larskuhtz: yes, let me double check ...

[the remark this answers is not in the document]
larskuhtz: yes, it's resolved

src/Chainweb/Miner/Miners.hs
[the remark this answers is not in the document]
larskuhtz: There is already a Haddock comment, which indicates that it just simulates mining:

```
-- | Artificially delay the mining process to simulate Proof-of-Work.
--
```

It's also not related to this PR. Anyways, I'll add a comment that makes it more clear.

src/Chainweb/TreeDB.hs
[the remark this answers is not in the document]
larskuhtz: I think, the review asked for a Haddock comment, that I added before marking it resolved.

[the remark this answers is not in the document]
larskuhtz: It's exported because `TreeDB` exports and API, and `minRank` fits conceptually in that API. There is `maxEntry, `maxRank`, and `root` (`minEntry`). It makes sense to also have `minRank`.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -801,8 +801,9 @@ initSchema = do
 
 getEndingTxId :: BlockHeight -> BlockHandler SQLiteEnv TxId
 getEndingTxId bh = callDb "getEndingTxId" $ \db -> do
-    if bh == 0
-      then return 0
+    if bh == 0 -- FIXME: what if genesisHeight > 0 for the chain?
```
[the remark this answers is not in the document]
larskuhtz: The GitHub CI tests run `ea` and check that the result matches what it currently checked in. I'll double check that those tests succeed.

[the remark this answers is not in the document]
larskuhtz: We run this as part of every GitHub ci build:

```shell
cabal v2-run cwtool -- ea
if ! git diff --exit-code; then
    echo "Inconsistent genesis headers detected. Did you forget to run ea?" 1>&2
    exit 1
fi
```

PR #1042, opened 2020-05-01
Robust Network Validation

fosskers:
This also make the `RosettaFailure` type a pure Sum type, so that it's associated Int/Message pairs can be easily extracted.

### DIFF
--- a/src/Chainweb/Rosetta/RestAPI/Server.hs
+++ b/src/Chainweb/Rosetta/RestAPI/Server.hs
@@ -52,18 +52,19 @@ import Chainweb.Version
 
 rosettaServer
     :: forall (v :: ChainwebVersionT)
-    . [(ChainId, MempoolBackend ChainwebTransaction)]
+    . ChainwebVersion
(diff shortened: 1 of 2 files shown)


src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -73,8 +74,8 @@ someRosettaServer
     :: ChainwebVersion
     -> [(ChainId, MempoolBackend ChainwebTransaction)]
     -> SomeServer
-someRosettaServer (FromSingChainwebVersion (SChainwebVersion :: Sing vT)) ms =
-    SomeServer (Proxy @(RosettaApi vT)) $ rosettaServer ms
+someRosettaServer v@(FromSingChainwebVersion (SChainwebVersion :: Sing vT)) ms =
+    SomeServer (Proxy @(RosettaApi vT)) $ rosettaServer v ms
```
larskuhtz, line 78: I think, I would pass `v` as only type parameter to `rosettaServer`.

PR #1043, opened 2020-05-01
Unify PactService on parent header

slpopejoy:
Cleanup PactService state to simply track parent header, use new `TxContext` to pair this with `PublicMeta` from tx, `PublicData` only assembled at actual tx execution


src/Chainweb/Pact/PactService.hs
```diff
@@ -1298,17 +1285,17 @@ runCoinbase
     -> ModuleCache
     -> PactServiceM cas (P.CommandResult [P.TxLog A.Value])
 runCoinbase Nothing _ _ _ _ _ = return noCoinbase
-runCoinbase (Just parentHeader) dbEnv miner enfCBFail usePrecomp mc = do
+runCoinbase (Just _) dbEnv miner enfCBFail usePrecomp mc = do
```
emilypi, line 1288: Do we even need this parameter now? It could just be `Bool` if we're not going to use the header case.

PR #1046, opened 2020-05-12
Pact block history query

slpopejoy:
Provide full txlog history for a block via `WebPactExecutionService`

### DIFF
--- a/src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
+++ b/src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
@@ -67,6 +67,7 @@ initInMemoryCheckpointEnv loggers logger = do
                     , _cpGetBlockParent = doGetBlockParent inmem
                     , _cpRegisterProcessedTx = doRegisterSuccessful inmem
                     , _cpLookupProcessedTx = doLookupSuccessful inmem
+                    , _cpGetBlockHistory = undefined -- TODO
                 }
             , _cpeLogger = logger
             })
--- a/src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
+++ b/src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
@@ -67,7 +67,7 @@ initInMemoryCheckpointEnv loggers logger = do
                     , _cpGetBlockParent = doGetBlockParent inmem
                     , _cpRegisterProcessedTx = doRegisterSuccessful inmem
                     , _cpLookupProcessedTx = doLookupSuccessful inmem
-                    , _cpGetBlockHistory = undefined -- TODO
+                    , _cpGetBlockHistory = \_ _ -> error "unimplemented"
                 }
             , _cpeLogger = logger
             })
--- a/src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
+++ b/src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
@@ -27,6 +27,7 @@ import Data.ByteString (ByteString)
 import Data.Aeson hiding (encode,(.=))
 import qualified Data.DList as DL
 import Data.Foldable (toList,foldl')
+import Data.Int
 import qualified Data.Map.Strict as M
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.List as List
@@ -36,7 +37,7 @@ import Data.Tuple.Strict
 import qualified Data.Vector as V
 import qualified Data.Vector.Algorithms.Tim as TimSort
 
-import Database.SQLite3.Direct (Utf8(..))
+import Database.SQLite3.Direct
 
 import Prelude hiding (log)
 
@@ -290,11 +291,12 @@ doGetBlockHistory dbenv blockHeader d = runBlockEnv dbenv $ do
               sshow (bhi,bha)
         _ -> internalError $ "doGetBlockHistory: expected single-row int result, got " <> sshow r
 
+    queryHistory :: Database -> Utf8 -> Int64 -> Int64 -> IO [(TxId,TxLog Value)]
     queryHistory db tableName s e = do
       let sql = "SELECT txid, rowkey, rowdata FROM [" <> tableName <>
                 "] WHERE txid > ? AND txid <= ?"
       r <- qry db sql
-           [SInt $ fromIntegral s,SInt $ fromIntegral e]
+           [SInt s,SInt e]
            [RInt,RText,RBlob]
       forM r $ \case
         [SInt txid, SText key, SBlob value] -> (fromIntegral txid,) <$> toTxLog d key value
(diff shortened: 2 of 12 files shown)


src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -256,3 +264,45 @@ doLookupSuccessful dbenv (TypedHash hash) = runBlockEnv dbenv $ do
...
         return $! T2 (fromIntegral h) hsh
     go _ = fail "impossible"
+
+doGetBlockHistory :: FromJSON v => Db -> BlockHeader -> Domain k v -> IO BlockTxHistory
+doGetBlockHistory dbenv blockHeader d = runBlockEnv dbenv $ do
+  callDb "doGetBlockHistory" $ \db -> do
+    endTxId <- getEndTxId db bHeight (_blockHash blockHeader)
+    startTxId <- getEndTxId db (pred bHeight) (_blockParent blockHeader)
+    history <- queryHistory db (domainTableName d) startTxId endTxId
+    return $! BlockTxHistory $ V.fromList $ M.toList $ foldl groupByTxid mempty history
```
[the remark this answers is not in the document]
slpopejoy: > why call insertWith (++) when we can :?
We're using a map to group by so not clear how `:` is relevant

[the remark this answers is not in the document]
slpopejoy: Used `foldl'`

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -256,3 +264,45 @@ doLookupSuccessful dbenv (TypedHash hash) = runBlockEnv dbenv $ do
...
+                "] WHERE txid > ? AND txid <= ?"
+      r <- qry db sql
+           [SInt $ fromIntegral s,SInt $ fromIntegral e]
+           [RInt,RText,RBlob]
+      readHistoryResult' d r
+
+
+    readHistoryResult' :: (FromJSON v,FromJSON l) =>
+                         Domain k v -> [[SType]] -> IO [(TxId,TxLog l)]
+    readHistoryResult' _d rows = forM rows $ \case
```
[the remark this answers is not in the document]
slpopejoy: I inlined the function as the original idea to reuse code from ChainwebPactDb wasn't working

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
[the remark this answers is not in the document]
emilypi: If we don't care about ordering, then instead of 

```haskell
M.insertWith (++) t [l] r
```

You could 

```haskell
M.insertWith (flip (:)) t l r
```

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
```diff
@@ -67,6 +67,7 @@ initInMemoryCheckpointEnv loggers logger = do
                     , _cpGetBlockParent = doGetBlockParent inmem
                     , _cpRegisterProcessedTx = doRegisterSuccessful inmem
                     , _cpLookupProcessedTx = doLookupSuccessful inmem
+                    , _cpGetBlockHistory = undefined -- TODO
```
LindaOrtega, line 70: Is this for future work?

src/Chainweb/WebPactExecutionService.hs
```diff
@@ -41,27 +41,54 @@ import Pact.Types.Hash
...
-        :: ChainId
+        )
+      -- ^ Lookup pact hashes as of a block header to detect duplicates
+    , _pactPreInsertCheck :: !(
+        ChainId
         -> Vector ChainwebTransaction
-        -> IO (Either PactException (Vector (Either InsertError ())))
+        -> IO (Either PactException (Vector (Either InsertError ()))))
+      -- ^ Run speculative checks to find bad transactions (ie gas buy failures, etc)
+    , _pactBlockTxHistory :: !(
```
LindaOrtega, line 78: A quick comment on the purpose of `_pactBlockTxHistory` would be helpful, like the comment in the other functions of `PactExecutionService`

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -256,3 +264,40 @@ doLookupSuccessful dbenv (TypedHash hash) = runBlockEnv dbenv $ do
...
+        "SELECT endingtxid FROM BlockHistory WHERE blockheight = ? and hash = ?;"
+        [SInt $ fromIntegral $ bhi, SBlob $ encode $ bha]
+        [RInt]
+      case r of
+        [[SInt tid]] -> return tid
+        [] -> throwM $ BlockHeaderLookupFailure $ "doGetBlockHistory: not in db: " <>
+              sshow (bhi,bha)
+        _ -> internalError $ "doGetBlockHistory: expected single-row int result, got " <> sshow r
+
+    queryHistory db tableName s e = do
```
LindaOrtega, line 293: can you add type information for `queryHistory`? and a comment on whether the start and end index are inclusive?

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -256,3 +264,40 @@ doLookupSuccessful dbenv (TypedHash hash) = runBlockEnv dbenv $ do
...
+    getEndTxId db bhi bha = do
+      r <- qry db
+        "SELECT endingtxid FROM BlockHistory WHERE blockheight = ? and hash = ?;"
+        [SInt $ fromIntegral $ bhi, SBlob $ encode $ bha]
+        [RInt]
+      case r of
+        [[SInt tid]] -> return tid
+        [] -> throwM $ BlockHeaderLookupFailure $ "doGetBlockHistory: not in db: " <>
+              sshow (bhi,bha)
+        _ -> internalError $ "doGetBlockHistory: expected single-row int result, got " <> sshow r
```
LindaOrtega, line 291: when might this error occur?

slpopejoy: pretty much what the comment says (and it's an internal error that shouldn't happen) -- we were expecting 1 row only with only 1 `SInt` field in it

src/Chainweb/Pact/Backend/InMemoryCheckpointer.hs
slpopejoy: changed to "unimplemented", this checkpointer is not used in production

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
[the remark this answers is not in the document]
slpopejoy: Order isn't necessarily unimportant, and per-txid lists are small ... I'm good to leave it

PR #1047, opened 2020-05-12
Adds /account/balance endpoint

LindaOrtega:
- [x] blocker: have /local return the block hash and block height of the block where the local command was run in. Currently, it returns the parents of the block and (+1 of the block height).
- [ ] ~~Research possible bug (checkpointer error keeps being thrown when I hit this endpoint in non-genesis case)~~ Can't replicate and it seems to work fine with a local mining node now.

### DIFF
--- a/src/Chainweb/Rosetta/RestAPI.hs
+++ b/src/Chainweb/Rosetta/RestAPI.hs
@@ -149,10 +149,9 @@ readChainIdText v c = do
 kdaToRosettaAmount :: Decimal -> Amount
 kdaToRosettaAmount k = Amount (sshow amount) currency Nothing
   where
-    -- Value of the transaction in atomic units represented as an
-    -- arbitrary-sized signed integer.
+    -- Value in atomic units represented as an arbitrary-sized signed integer.
     amount :: Integer
-    amount = floor $ k * (realToFrac $ (10 :: Integer) ^ numDecimals)
+    amount = floor $ k * (realToFrac ((10 :: Integer) ^ numDecimals))
 
     -- How to convert from atomic units to standard units
     numDecimals = 12 :: Word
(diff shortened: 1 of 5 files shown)


src/Chainweb/Rosetta/RestAPI.hs
```diff
@@ -118,9 +137,24 @@ validateNetwork v (NetworkId bc n msni) = do
...
   mkChainId v cid
+
+
+kdaToRosettaAmount :: Decimal -> Amount
+kdaToRosettaAmount k = Amount (sshow amount) currency Nothing
+  where
+    -- Value of the transaction in atomic units represented as an
+    -- arbitrary-sized signed integer.
+    amount :: Integer
+    amount = floor $ k * (realToFrac $ (10 :: Integer) ^ numDecimals)
```
LindaOrtega, line 155: Is this the correct approach?

src/Chainweb/Rosetta/RestAPI.hs
```diff
@@ -118,9 +137,23 @@ validateNetwork v (NetworkId bc n msni) = do
...
+
 -- | Guarantees that the `ChainId` given actually belongs to this
 -- `ChainwebVersion`.
 readChainIdText :: ChainwebVersion -> T.Text -> Maybe ChainId
 readChainIdText v c = do
   cid <- readMaybe @Word (T.unpack c)
   mkChainId v cid
+
+
+kdaToRosettaAmount :: Decimal -> Amount
```
slpopejoy, line 149: Suggest a quick unit test for just this function

LindaOrtega: I merged this PR into the #1050 and added a unit test for `kdaToRosettaAmount` there

LindaOrtega:
Closing this as it was merged into #1050

PR #1050, opened 2020-05-14
Finish Rosetta Endpoints

LindaOrtega:
- [x] Add /account/balance endpoint (merge https://github.com/kadena-io/chainweb-node/pull/1047 into this one and close)
- [x] Add /block endpoint
- [x] Add /block/transaction endpoint
- [x] Finish up /network/options (add enumeration of operation statuses and types allowed)
- [x] Refactor helper functions to Rosetta Utils module
- [x] Investigate how continuations and rollbacks affect coin accounts. **Findings**: We already see this in `fundTx`.
- [x] Address review comments
- [x] Add unit test to check logic of matching logs to all block transactions
- [x] Add unit test to check logic for matching logs to single transaction
- [x] Add unit test to check conversion from KDA to Rosetta Amount
- [x] Add unit test to check validating networkId correctly
- [x] Fix failing txLog history test
- [x] Merge in /mempool endpoint once merged into master
- [x] Verify that Rosetta Transaction Id being parsed and created in standard way across endpoints
- [x] Successfully run rosetta validator on local development node db
- [x] Adds historical balance lookup internal API using similar approach to https://github.com/kadena-io/chainweb-node/pull/1046
- [ ] Fix account address sanitization

Future Work:
- [ ] Add gas price and gas limit in metadata for /mempool/transaction and transaction in general.
- [ ] When creating Operations figure out how to efficiently add "related operations" (an optional Rosetta field in the Operation type). Already including TxId and operation id is generated based on the order the logs appeared in.
- [ ] Investigate how Rosetta validator tool handles continuations and cross-chain transactions. Especially if cross-chain wrapped inside of another function.
- [ ] https://github.com/kadena-io/chainweb-node/issues/1073
- [x] https://github.com/kadena-io/chainweb-node/issues/1072
- [x] Add Show instances to Rosetta types for better debugging: https://github.com/kadena-io/rosetta/issues/9


src/Chainweb/Pact/Templates.hs
```diff
@@ -18,6 +18,10 @@ module Chainweb.Pact.Templates
 ( mkBuyGasTerm
 , mkCoinbaseTerm
 , mkCoinbaseCmd
+, app
```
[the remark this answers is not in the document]
LindaOrtega: Made them longer in the latest commit.

src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -93,20 +123,379 @@ rosettaServer v ms peerDb cutDb = (const $ error "not yet implemented")
...
+  where
+    readBal :: PactValue -> Maybe Decimal
+    readBal (PLiteral (LDecimal d)) = Just d
+    readBal _ = Nothing
+
+    readBlock :: Maybe Value -> Maybe (Word64, T.Text)
+    readBlock (Just (Object meta)) = do
+      hi <- (HM.lookup "blockHeight" meta) >>= (hushResult . fromJSON)
+      hsh <- (HM.lookup "prevBlockHash" meta) >>= (hushResult . fromJSON)
+      pure $ (pred hi, hsh)
```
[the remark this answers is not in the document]
LindaOrtega: Added

src/Chainweb/Rosetta/RestAPI.hs
```diff
@@ -94,21 +96,57 @@ data RosettaFailure
...
 rosettaError RosettaMempoolBadTx = RosettaError 2 "Transaction not present in mempool" False
 rosettaError RosettaUnparsableTx = RosettaError 3 "Transaction not parsable" False
 rosettaError RosettaInvalidTx = RosettaError 4 "Invalid transaction" False
 rosettaError RosettaInvalidBlockchainName = RosettaError 5 "Invalid blockchain name" False
 rosettaError RosettaMismatchNetworkName = RosettaError 6 "Invalid Chainweb network name" False
+rosettaError RosettaHistBalCheckUnsupported =
+  RosettaError 7 "Historical account balance lookup is not supported." False
+rosettaError RosettaPactExceptionThrown =
+  RosettaError 7 "A pact exception was thrown" False
+rosettaError RosettaPactErrorThrown = RosettaError 8 "Transaction failed with a pact error" False
```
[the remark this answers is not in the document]
LindaOrtega: This is only thrown when looking up account balance fails with a pact error, which (assuming I constructed the local command correctly) should only happen if the account doesn't exist. I changed it to something more user friendly: `RosettaCouldNotRetrieveBalance`

src/Chainweb/Rosetta/RestAPI.hs
```diff
@@ -94,21 +96,57 @@ data RosettaFailure
...
+    | RosettaUnparsableTxOut
     deriving (Show, Enum, Bounded)
 
--- TODO: Better grouping of rosetta error index
+
+-- TODO: Better grouping of rosetta error index?
 rosettaError :: RosettaFailure -> RosettaError
 rosettaError RosettaChainUnspecified = RosettaError 0 "No SubNetwork (chain) specified" False
-rosettaError RosettaInvalidChain = RosettaError 1 "Invalid chain value" False
+rosettaError RosettaInvalidChain = RosettaError 1 "Invalid sub-network chain id value" False
```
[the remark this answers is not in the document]
LindaOrtega: Changed

src/Chainweb/Rosetta/Util.hs
```diff
@@ -0,0 +1,686 @@
...
+    d = (Domain' (UserTables "coin_coin-table"))
+
+
+getHistoricalLookupBalance
+    :: PactExecutionService
+    -> BlockHeader
+    -> T.Text
+    -> ExceptT RosettaFailure Handler Decimal
+getHistoricalLookupBalance cr bh k = do
+  someHist <- liftIO $ (_pactHistoricalLookup cr) bh d key
```
LindaOrtega, line 301: Could this be subject to sql injection? Is it okay to use the bare text submitted by the user when its going to be used to form sql statements?

src/Chainweb/Rosetta/Util.hs
```diff
@@ -0,0 +1,686 @@
...
+                   map (operation Successful CoinbaseReward tid) l
+         in pure $ (f cr ops)
+    else Left "First log's txId does not match coinbase tx's TxId"
+
+
+------------------------
+-- NonGenesis Helpers --
+------------------------
+
+-- | Algorithm:
```
slpopejoy, line 414: Start this comment with a haddocks e.g. "accumulates a triple of transactions representing a tx and it's bracketing gas logs" or whatever.

src/Chainweb/Rosetta/Util.hs
```diff
@@ -0,0 +1,686 @@
...
+--   it peeks at the next logs in the list (i.e. idx + 1).
+--   If the log's txId matches the tx's txId, then these
+--   logs are associated with the tx and the logs that
+--   paid for gas are retrieved from following index (i.e. idx + 1 + 1).
+--   If the txIds don't match OR if the tx failed, the gas logs
+--   are retrieved from idx + 1 instead.
+-- |
+gasTransactionAcc
+    :: V.Vector (TxId, [AccountLog])
+    -> (a -> Maybe TxId)
```
slpopejoy, line 427: Again wondering about indirection here

slpopejoy: Also it's bewildering trying to understand what roles `a`, `b` and `c` have. Please consider specializing to concrete types for legibility, and if not, use type variables that give some idea what is going on.

src/Chainweb/Rosetta/Util.hs
```diff
@@ -0,0 +1,686 @@
...
+    -> RequestKey
+    -> NonGenesisMatchFunction Transaction
+nonGenesisTransaction logs initial rest target =
+  NonGenesisMatchFunction $
+    (noteOptional RosettaTxIdNotFound .
+     overwriteError RosettaMismatchTxLogs) $
+       nonGenesisTransaction'
+       logs _crReqKey _crTxId rosettaTransaction initial rest target
+
+nonGenesisTransaction'
```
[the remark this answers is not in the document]
slpopejoy: Also same comment as above regarding polymorphism and roles of `a` and `b`, it makes it hard to understand what is being accomplished here.

src/Chainweb/Rosetta/Util.hs
```diff
@@ -0,0 +1,686 @@
...
+
+-- |
+-- Module: Chainweb.Rosetta.Util
+-- Copyright: Copyright © 2018 - 2020 Kadena LLC.
+-- License: MIT
+-- Maintainer: Linda Ortega <linda@kadena.io>
+-- Stability: experimental
+--
+--
+module Chainweb.Rosetta.Util where
```
slpopejoy, line 14: This isn't really `Util` as this is where "the sausage is made" as opposed to some grab-bag of utilities. `Internal` is probably more illustrative, although it could also be `Impl` or just `Rosetta`

src/Chainweb/Rosetta/Util.hs
slpopejoy: Also, in the future you should consider possibly making `_pactHistoricalLookup` return the _last two_ transactions (since your SQL is already a `LIMIT 1` affair). Why? Because we can do deltas with that! 😎

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -308,3 +302,42 @@ doGetBlockHistory dbenv blockHeader d = runBlockEnv dbenv $ do
...
+doGetHistoricalLookup
+    :: FromJSON v
+    => Db
+    -> BlockHeader
+    -> Domain k v
+    -> RowKey
+    -> IO (Maybe (TxLog Value))
+doGetHistoricalLookup dbenv blockHeader d k = runBlockEnv dbenv $ do
+  callDb "doGetHistoricalLookup" $ \db -> do
+    endTxId <- getEndTxId db bHeight (_blockHash blockHeader)
```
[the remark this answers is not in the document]
LindaOrtega: We needed to guard `doGetBlockHistory` because it searches within a range of txIds and `startTxId` tried to get the ending txId of block height -1 (which caused `getEndTxId` to blow up). 
For `doGetHistoricalLookup` we don't have to worry about that because we're searching for balances below an upper bound (i.e. the ending txId of the block we received).

src/Chainweb/Rosetta/Util.hs
LindaOrtega: @slpopejoy regarding returning the last two transactions, I've made an issue for it here: https://github.com/kadena-io/chainweb-node/issues/1078

@slpopejoy and @emilypi regarding the potential sql injection via account address: I've added the following code to check that the account address is a valid Pact `Name` (since I believe row keys need to be valid `Name` correct?)
https://github.com/kadena-io/chainweb-node/pull/1050/files#diff-04e165a290c33e67270413e0ae8c3d85R515-R519

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -275,31 +276,25 @@ doGetBlockHistory :: FromJSON v => Db -> BlockHeader -> Domain k v -> IO BlockTx
 doGetBlockHistory dbenv blockHeader d = runBlockEnv dbenv $ do
   callDb "doGetBlockHistory" $ \db -> do
     endTxId <- getEndTxId db bHeight (_blockHash blockHeader)
-    startTxId <- getEndTxId db (pred bHeight) (_blockParent blockHeader)
+    startTxId <- if (bHeight == genesisHeight v cid)
```
LindaOrtega, line 279: @slpopejoy using `genesisHeight` here

src/Chainweb/Rosetta/Internal.hs
```diff
@@ -0,0 +1,519 @@
...
+-- | Retrieve the coin contract logs for transaction(s) in a block.
+matchLogs
+    :: LogType tx
+    -> BlockHeader
+    -> M.Map TxId [AccountLog]
+    -> CoinbaseTx (CommandResult Hash)
+    -> V.Vector (CommandResult Hash)
+    -> Either RosettaFailure tx
+matchLogs typ bh logs coinbase txs
+  | bheight == genesisHeight v cid = matchGenesis
```
LindaOrtega, line 124: @slpopejoy changed to `genesisHeight v cid` here

src/Chainweb/Rosetta/Util.hs
```diff
@@ -0,0 +1,224 @@
...
+
+--------------------------------------------------------------------------------
+-- Functions to create Rosetta types --
+--------------------------------------------------------------------------------
+
+-- | If its the genesis block, Rosetta wants the parent block to be itself.
+--   Otherwise, fetch the parent header from the block.
+parentBlockId :: BlockHeader -> BlockId
+parentBlockId bh
+  | (bHeight == genesisHeight v cid) = blockId bh  -- genesis
```
LindaOrtega, line 75: @slpopejoy changed to using `genesisHeight v cid` here as well

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -275,7 +276,9 @@ doGetBlockHistory :: FromJSON v => Db -> BlockHeader -> Domain k v -> IO BlockTx
 doGetBlockHistory dbenv blockHeader d = runBlockEnv dbenv $ do
   callDb "doGetBlockHistory" $ \db -> do
     endTxId <- getEndTxId db bHeight (_blockHash blockHeader)
-    startTxId <- getEndTxId db (pred bHeight) (_blockParent blockHeader)
+    startTxId <- if (_blockHeight blockHeader == 0)
```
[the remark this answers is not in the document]
LindaOrtega: I'm now using `genesisHeight v cid` instead of checking for height `0`. See the following comments for the exact places where these changes occurred.

src/Chainweb/Rosetta/Util.hs
```diff
@@ -0,0 +1,686 @@
...
+
+newtype GenesisMatchFunction a = GenesisMatchFunction
+  { _genesisMatchFunction :: Either RosettaFailure a }
+newtype RemediationMatchFunction a = RemediationMatchFunction
+  { _remediationMatchFunction :: Either RosettaFailure a }
+newtype NonGenesisMatchFunction a = NonGenesisMatchFunction
+  { _nonGenesisMatchFunction :: Either RosettaFailure a }
+
+-- | Retrieve the coin contract logs for transaction(s) in a block.
+matchLogs
```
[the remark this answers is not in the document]
LindaOrtega: The new `matchLogs` functions get rid of this indirection: https://github.com/kadena-io/chainweb-node/pull/1050/files#diff-04e165a290c33e67270413e0ae8c3d85R123

src/Chainweb/Rosetta/Util.hs
LindaOrtega: Added

src/Chainweb/Rosetta/Util.hs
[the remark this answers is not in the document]
LindaOrtega: Got rid of the `getXXX` indirection. And yes part of the indirection was there for the tests because debugging is easier with `Either String rosettaTx`. But I added the conversion of the `Either String rosettaTx` to the `matchLogs` function.

@slpopejoy @emilypi I tried to simplify the matching logic and cut down on the polymorphism. See this commit for exactly what changed: https://github.com/kadena-io/chainweb-node/pull/1050/commits/ee355f55305ed9a6637e9603572189dab48b1172

src/Chainweb/Rosetta/Util.hs
LindaOrtega: See comment: https://github.com/kadena-io/chainweb-node/pull/1050#discussion_r440313717

src/Chainweb/Rosetta/Util.hs
LindaOrtega: Moved block functions and tx matching functions to `Chainweb.Rosetta.Internal`.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -275,7 +276,9 @@ doGetBlockHistory :: FromJSON v => Db -> BlockHeader -> Domain k v -> IO BlockTx
...
+      then pure 0  -- genesis block
```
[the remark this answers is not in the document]
LindaOrtega: @slpopejoy we discussed this briefly on Slack, but do I need to change this assumption? i.e. is it okay to assume that genesis' transaction's TxIds start at zero even if the genesis block height is not necessarily zero any more?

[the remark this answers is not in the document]
slpopejoy: I think so yes

src/Chainweb/Rosetta/Util.hs
LindaOrtega: Actually, I need to reconsider this approach. It throws an errors with `"address": "POTENT AF MINER"` on mainnet. This seems to be to strict

PR #1051, opened 2020-05-16
Create "chainweb-portable" package, which can run on other linuxes

ryantrinkle:
This has not been extensively tested, but it did work on the one Ubuntu machine I tried.

To build, do this: nix-build ./project.nix -A proj.ghc.chainweb-portable

To install, take all the files in the output and copy them to /opt/chainweb on the target machine.  This prefix directory can be changed by supplying a different argument in `project.nix`, but it cannot be made fully relocatable easily.  Perhaps something like [rtldi](http://bitwagon.com/rtldi/rtldi.html) could help, but I haven't evaluated it.  It would also be possible to supply a script based on `patchelf` to change the prefix without rebuilding.

### DIFF
--- a/make-portable.nix
+++ b/make-portable.nix
@@ -1,0 +1,57 @@
+{ runCommand, patchelf, coreutils, binutils-unwrapped, find-libs }:
+
+targetPrefix: inputPackage:
+
+runCommand (inputPackage.pname + "-portable") {
+  inherit inputPackage targetPrefix;
+} ''
+  set -euo pipefail
+
+  cp -a "$inputPackage" "$out"
+  chmod -R u+w "$out"
+  mkdir -p "$out/lib"
+
+  processLib() {
+    lib="$1"
+    target="$2"
+
+    cp "$lib" "$target"
+    chmod u+w "$target"
+    if [ ! -z "$("${patchelf}/bin/patchelf" --print-rpath "$target")" ] ; then
+      "${patchelf}/bin/patchelf" --set-rpath "$targetPrefix/lib" "$target"
+    fi
+  }
+
+  addLib() {
+    lib="$1"
+
+    libname="$(${coreutils}/bin/basename "$lib")"
+    target="$out/lib/$libname"
+    if [ -e "$target" ] ; then
+      tmpfile="$("${coreutils}/bin/mktemp" "make-portable.XXXXXXXX")"
+      processLib "$lib" "$tmpfile"
+      diff "$target" "$tmpfile" >/dev/null
+      libsAreIdentical=$?
+      rm "$tmpfile"
+
+      if [ "$libsAreIdentical" -ne 0 ] ; then
+        2>&1 echo "error: multiple libraries named $libname are required, and they are not identical files"
+        exit 1
+      fi
+      echo "Library $lib already present - skipping"
+    else
+      echo "Adding library $lib"
+      processLib "$lib" "$target"
+    fi
+  }
+
+  for exe in $out/bin/* ; do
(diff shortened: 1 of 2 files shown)


ryantrinkle:
To clarify: I have tested that chainweb-node successfully starts, and provides output that looks like it's running, on Ubuntu 17.10 (Linux kernel 4.13.0).  I haven't tested that the program works beyond that.

ryantrinkle:
Converted to draft because it still depends on a fork of nixpkgs.  PR to nixpkgs is up here: https://github.com/NixOS/nixpkgs/pull/88070

project.nix
```diff
@@ -14,11 +14,19 @@ proj = kpkgs.rp.project ({ pkgs, hackGet, ... }: with pkgs.haskell.lib;
...
+    repo = "nixpkgs";
+    rev = "a9a141a10f32691228ae7668e571359f2aaf82e4";
+    sha256 = "0ncqzsq3rwkh4g68bxsxmznq4z4pcqy6wclslnrsnqbq1b2lk2gl";
+  }) {};
   in {
     name = "chainweb";
     overrides = self: super: {
       chainweb = enableCabalFlag (
         justStaticExecutables (enableDWARFDebugging (convertCabalTestsAndBenchmarksToExecutables super.chainweb))) "use_systemd";
+      chainweb-portable = nixpkgsWithFindLibs.callPackage ./make-portable.nix {} "/opt/chainweb" self.chainweb;
```
matthewbauer, line 29: You can use `$ORIGIN` instead of `/opt/chainweb` here so that ld.so will look for libraries in the same directory as the executable. That way, the libraries don't have to be in `/opt/chainweb`.

make-portable.nix
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  for exe in $out/bin/* ; do
+    echo "Processing executable $exe"
+    for lib in $("${find-libs}/bin/find-libs" "$exe") ; do
+      addLib "$lib"
+    done
+    interpreter="$("${patchelf}/bin/patchelf" --print-interpreter "$exe")"
+    addLib "$interpreter"
+    "${patchelf}/bin/patchelf" --set-interpreter "$targetPrefix/lib/$("${coreutils}/bin/basename" "$interpreter")" --set-rpath "$targetPrefix/lib" "$exe"
```
matthewbauer, line 55: Unfortunately the interpreter can't use $ORIGIN, mentioned below, so that keeps you stuck with the /opt/chainweb absolute path. `/lib/ld-linux-${arch}.so.${ver}` is a safe bet for most Linux distros. Unfortunately that doesn't work on NixOS, though. See https://github.com/NixOS/nixpkgs/pull/69057/files#diff-aca5d4e8a2b0e3c27da452f0505a5a9dR224

ryantrinkle: I tried using `/lib/ld-linux...` but it didn't seem to work for me - I got this issue: https://stackoverflow.com/questions/49245547/version-glibc-private-not-defined-in-file-ld-linux-x86-64-so-2

project.nix
ryantrinkle: Yep, I originally used "$ORIGIN/.." as the prefix, but that didn't work due to the ld-linux issues; since I needed a prefix anyway, I just kept it consistent.

mightybyte: Oh, using `$ORIGIN` would be much nicer IMO.

make-portable.nix
matthewbauer: Oh right - we would have to link against an older glibc to support older versions, which isn't very easy.

PR #1053, opened 2020-05-18
Make miner reward calculation graph order aware

emilypi:
Instead of just pre-calculating miner rewards, expecting the order of the chain graph to remain fixed, this PR moves the graph-order calculation out to the `minerReward` function, so that we can calculate the division on the fly.

### DIFF
--- a/src/Chainweb/Miner/Pact.hs
+++ b/src/Chainweb/Miner/Pact.hs
@@ -29,7 +29,6 @@ module Chainweb.Miner.Pact
   -- * Optics
 , minerId
 , minerKeys
-, minerRewardHeights
 , minerRewards
   -- * Defaults
 , noMiner
@@ -49,9 +48,8 @@ import qualified Data.Csv as CSV
 import Data.Decimal (Decimal)
 import Data.FileEmbed (embedFile)
 import Data.Hashable
-import Data.HashMap.Strict (HashMap)
-import qualified Data.HashMap.Strict as HM
-import qualified Data.List as L (sort)
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as M
 import Data.String (IsString(..))
 import Data.Text (Text)
 import qualified Data.Vector as V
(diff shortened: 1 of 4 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -902,14 +902,17 @@ readAccountGuard pdb account
...
 --
 minerReward
-    :: MinerRewards
+    :: ChainwebVersion
+    -> MinerRewards
     -> BlockHeight
     -> IO P.ParsedDecimal
-minerReward (MinerRewards rs q) bh =
-    case V.find (bh <=) q of
+minerReward v (MinerRewards rs q) bh = case V.find (bh <=) q of
```
slpopejoy, line 909: I suggest replacing this whole Vector+HashMap with just a `Map` --
- `V.find` is O(n) vs. `Map` `split` which is O(log n). 
- Less error prone (here we have to do two error checks)
- already sorted on key

slpopejoy: http://hackage.haskell.org/package/containers-0.6.0.1/docs/Data-Map-Strict.html#v:split

emilypi: As discussed in Slack, we made the `HashMap` a `Map.Strict`, and we're just making use of `lookupGE` now, which is O(log n)

src/Chainweb/Miner/Pact.hs
```diff
@@ -147,42 +143,29 @@ fromMinerData :: MonadThrow m => MinerData -> m Miner
...
+    { _minerRewards :: Map BlockHeight Decimal
       -- ^ The map of blockheight thresholds to miner rewards
-    , _minerRewardHeights :: !(V.Vector BlockHeight)
-      -- ^ A (sorted) vector of blockheights (head is most recent)
     } deriving (Eq, Ord, Show, Generic)
 
 -- | A getter into the map of heights to rewards
 --
-minerRewards :: Getter MinerRewards (HashMap BlockHeight ParsedDecimal)
+minerRewards :: Getter MinerRewards (Map BlockHeight Decimal)
```
slpopejoy, line 153: This is unused, let's ditch it

emilypi: 👍

PR #1057, opened 2020-05-20
Add Rosetta /mempool endpoint 

emilypi:
This implements the following endpoint: 

```
/mempool
Get All Mempool Transactions

Get all Transaction Identifiers in the mempool
```

### DIFF
--- a/src/Chainweb/Rosetta/RestAPI/Server.hs
+++ b/src/Chainweb/Rosetta/RestAPI/Server.hs
@@ -166,9 +166,9 @@ mempoolH v ms (MempoolReq net) = runExceptT work >>= either throwRosetta pure
         r <- liftIO $ newIORef mempty
         -- TODO: This will need to be revisited once we can add
         -- pagination + streaming the mempool
-        _ <- liftIO $ mempoolGetPendingTransactions mp Nothing $ writeIORef r
-        txs <- liftIO $ fmap f <$> readIORef r
-        return $ MempoolResp $ V.toList txs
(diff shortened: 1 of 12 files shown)


src/Chainweb/Rosetta/RestAPI/Server.hs
```diff
@@ -156,10 +157,18 @@ mempoolH
...
+
     work = do
         cid <- validateNetwork v net
-        _ <- lookup cid ms ?? RosettaInvalidChain
-        error "not yet implemented"  -- TODO!
+        mp <- lookup cid ms ?? RosettaInvalidChain
+        r <- liftIO $ newIORef mempty
+        -- TODO: This will need to be revisited once we can add
+        -- pagination + streaming the mempool
+        _ <- liftIO $ mempoolGetPendingTransactions mp Nothing $ writeIORef r
```
LindaOrtega, line 169: why are we ignoring these results?

emilypi: there's no useful result to ignore. `mempoolGetPendinTransactions` is

```haskell
Maybe HighwaterMark -> (Vector TransactionHash -> IO ()) -> IO HighwaterMark
```

We do not need the high water mark returned, just the vector of hashes in the pending pool. The only way to extract these hashes is via stateful callback (so we use an `IORef` to extract them).

emilypi:
@LindaOrtega Yeah, i'm working on the best way to do that now. I may just go with the "dumb" way through the JSON instances for now just to get the thing finished.

PR #1060, opened 2020-05-27
Add retry to lookupBlockHeader and improve exception handling

larskuhtz:


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -98,9 +100,19 @@ import Chainweb.Pact.Templates
...
+-- -------------------------------------------------------------------------- --
+-- NFData orphans for Pact types
+--
+-- TODO: these should be moved to Pact!
+
+deriving instance NFData PactResult
+deriving instance NFData RequestKey
+deriving instance NFData a => NFData (CommandResult a)
+
+-- -------------------------------------------------------------------------- --
```
larskuhtz, line 115: @emilypi could we include these in pact?

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -429,12 +441,11 @@ applyUpgrades v cid height
...
 
-      r <- try $ runGenesis tx permissiveNamespacePolicy interp
-
-      case r of
+      tryAllSynchronous (runGenesis tx permissiveNamespacePolicy interp) >>= \case
         Right _ -> return ()
-        Left (e :: SomeException) -> do
+        Left e -> do
           logError $ "Upgrade transaction failed! " <> sshow e
+            -- TODO: is it really OK, to just swallow the exception?
```
larskuhtz, line 448: @emilypi What is your though on this comment?

src/Chainweb/Pact/PactService.hs
```diff
@@ -332,12 +334,51 @@ initializeCoinContract _logger v cid pwo = do
...
-                      "failed lookup of parent header in " <> ctx <> ": " <> sshow (e :: SomeException)
+    bhdb <- asks _psBlockHeaderDb
+    liftIO $! (retryOnNothing (lookup bhdb bhash)) >>= \case
+        Nothing -> throwM $ BlockHeaderLookupFailure $
+            "failed lookup of parent header in " <> ctx <> ". Parent header hash: " <> encodeToText bhash
+        Just x -> return x
+  where
+    -- Lookup failures are presumably expensive (involve disk IO).
+    -- We start with 500us, in total 3.5ms
+    policy = exponentialBackoff 500 <> limitRetries 3
```
emilypi, line 380: is this bound reasoned, or just a rough estimate?

src/Chainweb/Pact/TransactionExec.hs
emilypi: Yeah, i'll open a PR with it today.

src/Chainweb/Pact/TransactionExec.hs
emilypi: It would be pretty catastrophic if the upgrades didn't work, but other parts of the service would immediately begin to fail block validation as a result, so we end up with an error being thrown either way. We may as well just throw an internal error here.

mightybyte:
I've been running this branch and it does seem to almost completely eliminate the errors.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -429,12 +434,11 @@ applyUpgrades v cid height
 
       infoLog $ "Running upgrade tx " <> sshow (_cmdHash tx)
 
-      r <- try $ runGenesis tx permissiveNamespacePolicy interp
-
-      case r of
+      tryAllSynchronous (runGenesis tx permissiveNamespacePolicy interp) >>= \case
```
slpopejoy, line 437: Why does this PR need this fix?

src/Chainweb/Pact/PactService.hs
```diff
@@ -332,12 +334,51 @@ initializeCoinContract _logger v cid pwo = do
         psInitCache .= mc
         return $! Discard ()
 
+-- | Lookup a header in the BlockHeader db.
```
slpopejoy, line 337: I'm worried we're going the wrong direction here. The question is, when are these failings occurring and why?

- With local and pre-checks, we seem to be trying to use the last-validated block. Why do we even need a lookup at this point? Shouldn't we instead be caching the values (parent header only) instead of doing a lookup, eliminating failures by construction?

- I don't know if we're experiencing failures in other cases. If not though retries are adding a lot of operational complexity here.

larskuhtz: > Why do we even need a lookup at this point? 

I think, the pact db doesn’t have the block header data that is needed to evaluate pact code in the context of chainweb. But I am not 100%. If we would store the bits of the parent header that are relevant for the public metadata in the BlockHistory, maybe we wouldn’t need the lookup for those calls.

> I don't know if we're experiencing failures in other cases. If not though retries are adding a lot of operational complexity here.

In almost all cases, the retry doesn’t happen. In particular, in latency critical operations, i.e. mining and consensus calls to `newBlock` and `execValidateBlock`, the caller provides the context explicitly via a parent header. 

The failure can occur in two cases:
1. actual validation failure or data corruption due to hard crashes (where finalizers aren’t exectuted).
2. a race, where a local/pre-check call (calls that don’t provide a parent header context) comes in before the block is written to the db.

In the first case the retry won’t help and we raise an error. The error is now raised with a few ms delay, but that’s fine, I think, since the service is in a bad state in any case. If it's a validation failure it will probably resolve when the next block is validated (or mined) which will take at least several ms.

The second case is rare (we know that, from the number of exceptions that we observed in the past), and it is very likely that the first retry resolves it. I choose a relatively high initial retry value of 500ms and a low number of retries, to minimize overhead.

> Shouldn't we instead be caching the values (parent header only) instead of doing a lookup, eliminating failures by construction?

Yes, if the lookup is really only needed to populate the public meta (or otherwise initialize the execution environment from the parent header) than that is the proper solution, because it would eliminate the race in first place.

But, it would be probably more work. Most likely (but I am not 100% sure) we would have to change the schema and store additional data in BlockHistory or elsewhere, which seems non-trivial too me.

I think at this point BlockHistory only has enough information to do the lookup in the chain db. But maybe the relevant data is stored elsewhere in the sqlite db?

src/Chainweb/Pact/TransactionExec.hs
larskuhtz: I'd say, if it is catastrophic, let's throw here so here and don't wait for other parts falling apart. That should help diagnosing the issue, if it ever happens.

slpopejoy:
Closing but leaving branch

PR #1062, opened 2020-05-30
Remove header lookup from local, pre-insert check

slpopejoy:
Resolves `TreeDbKeyNotFound` issues in `local` and mempool pre-insert check, by using `PactService` parent header in cache. 

Adds an extra update to the cache after block validation to set the "parent" to the just-validated current block.

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -629,9 +629,9 @@ withCurrentCheckpointer
     -> (PactDbEnv' -> PactServiceM cas (WithCheckpointerResult a))
     -> PactServiceM cas a
 withCurrentCheckpointer caller act = do
-  ph <- _parentHeader <$> use psParentHeader
-  let target = Just (succ $ _blockHeight ph, _blockHash ph)
-  withCheckpointer target caller act
+    ph <- _parentHeader <$> use psParentHeader
+    let target = Just (succ $ _blockHeight ph, _blockHash ph)
+    withCheckpointer target caller act
 
 -- | Same as 'withCheckpointer' but rewinds the checkpointer state to the
 -- provided target.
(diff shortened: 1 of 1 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -622,6 +622,17 @@ withCheckpointer target caller act = mask $ \restore -> do
...
         psStateValidated .= Just header
 
+-- | 'withCheckpointer' but using the cached parent header for target.
+withCurrentCheckpointer
+    :: PayloadCasLookup cas
+    => String
+    -> (PactDbEnv' -> PactServiceM cas (WithCheckpointerResult a))
+    -> PactServiceM cas a
+withCurrentCheckpointer caller act = do
+    ph <- _parentHeader <$> use psParentHeader
```
emilypi, line 632: The big question I'm wondering when staring at this: when the cached header will ever be different from the one we find with the checkpointer?

slpopejoy: Yes, it's a valid concern. There is a semantic mismatch between the two, in that `setParentHeader` sets information pertaining to the _parent block_, while checkpointer head is the last-saved _current block_. 

Nonetheless, these are in sync when playing blocks, as checkpointer "restores" to head which is considered the parent to the current block playing. This matches the usage of `setParentHeader`. 

If the checkpointer is discarded then these stay in sync. Only when the checkpointer is saved does its head advance, at which point `setParentHeader` would be one behind checkpointer head. 

The save at the end of `execValidateBlock` seeks to resolve this, and explicitly for the "dirty read" case (`local` et al), so that `execLocal` and friends will find a parent header that matches checkpointer. 

There is one remaining case, which is checkpointer saves during rewind -- these are not conformed with `setParentHeader`. But this is OK, because rewind always precedes some other operation (new block or validate block), so it can't affect `local` and friends.

Here's all of the places these interactions happen:

### Checkpointer `Save` 
  - in `rewindTo`/`fastForward`, saving on each replayed block: https://github.com/kadena-io/chainweb-node/blob/0151945ccebca95849218306ed6c86e759772080/src/Chainweb/Pact/PactService.hs#L1216
 - in `execValidateBlock`, saving the validated block: https://github.com/kadena-io/chainweb-node/blob/0151945ccebca95849218306ed6c86e759772080/src/Chainweb/Pact/PactService.hs#L1235

### `setParentHeader` 
- in initialization, setting it to the genesis header: https://github.com/kadena-io/chainweb-node/blob/0151945ccebca95849218306ed6c86e759772080/src/Chainweb/Pact/PactService.hs#L252
- in `initializeCoinContract`->`readContracts`, ie when not starting from genesis; this sets checkpointer head as parent: https://github.com/kadena-io/chainweb-node/blob/0151945ccebca95849218306ed6c86e759772080/src/Chainweb/Pact/PactService.hs#L326-L327
- in `playOneBlock`, handling both genesis and normal. This is what is called by `rewindTo` and `execValidateBlock` both: https://github.com/kadena-io/chainweb-node/blob/0151945ccebca95849218306ed6c86e759772080/src/Chainweb/Pact/PactService.hs#L1151-L1159
- in `execNewBlock`, both before rewind (for buyGasCheck maybe?) and immediately before playing the block. https://github.com/kadena-io/chainweb-node/blob/0151945ccebca95849218306ed6c86e759772080/src/Chainweb/Pact/PactService.hs#L941 https://github.com/kadena-io/chainweb-node/blob/0151945ccebca95849218306ed6c86e759772080/src/Chainweb/Pact/PactService.hs#L991
- in `execValidateBlock`->`setCurrAsParent`, the new operation mentioned above: https://github.com/kadena-io/chainweb-node/blob/0151945ccebca95849218306ed6c86e759772080/src/Chainweb/Pact/PactService.hs#L1245-L1246

larskuhtz: I am also a bit concerned about the maintainability of these invariants. This isn't an issue of this PR, and the PR is an improvement.

I wonder if it would be possible to make the use of block headers during pact validation more consistent. I find the checkpoint target parameter (`let target = Just (succ $ _blockHeight ph, _blockHash ph)`) particularly confusing. 

I think pact validation only depend on the parent header, and the current payload. If these two are available from the environment, there should be no need to query the block header db. It seems that only during rewind these must be queried.

src/Chainweb/Pact/PactService.hs
```diff
@@ -622,6 +622,17 @@ withCheckpointer target caller act = mask $ \restore -> do
...
+    let target = Just (succ $ _blockHeight ph, _blockHash ph)
```
larskuhtz, line 633: It's probably beyond the scope of this PR, but I find this parameter highly confusing, because the types don't indicate that the block height is't the height of the hash.

PR #1063, opened 2020-05-30
Publish service APIs on separate HTTP port

larskuhtz:
With the PR APIs that are not required to be public and are not required for consensus and don't use P2P are published on a separate HTTP port (no HTTPS):

* Mining API
* Pact API
* Block Header stream
* Rosetta

By default these are published on port 1848 (which can be change in the configuration file). Since no P2P and HTTPs is used users can use standard techniques, like load balancers or HTTP proxies to integrate the APIs into their infrastructure.

This is a breaking change for users of those local APIs. It should thus be released only when users have sufficient time to upgrade their infrastructure.

## Open Questions

* [ ] make CORS configurable (at least on local APIs and possibly disable it on peer APIs)
* [ ] Add middleware to peer api that redirects service API routes to new chainweb api domain.
    * [ ] disable by default and enable on all bootstrap nodes.


src/Chainweb/Chainweb/PeerResources.hs
```diff
@@ -130,19 +130,13 @@ peerServerSettings peer
...
-    return (conf', sock)
-
-deallocateSocket :: (P2pConfiguration, Socket) -> IO ()
-deallocateSocket (_, sock) = close sock
-
-withSocket :: P2pConfiguration -> ((P2pConfiguration, Socket) -> IO a) -> IO a
-withSocket conf = bracket (allocateSocket conf) deallocateSocket
+withPeerSocket :: P2pConfiguration -> ((P2pConfiguration, Socket) -> IO a) -> IO a
+withPeerSocket conf act
+    = bracket (allocateSocket port interface) deallocateSocket $ \(p, s) ->
```
slpopejoy, line 135: use `withSocket`

slpopejoy:
So as discussed, let's govern which server ("peer" or local) serves Pact API with a config flag. I'm leaning to have the flag default to putting it in "peer" as that is most backward-compatible but not dead-set on it.

sirlensalot (approved):

PR #1064, opened 2020-05-30
Use base* libs + clarify encoding primitives

emilypi:
@larskuhtz


src/Chainweb/Utils.hs
```diff
@@ -581,19 +581,16 @@ encodeB64UrlText = T.decodeUtf8 . B64U.encode
 --
 decodeB64UrlNoPaddingText :: MonadThrow m => T.Text -> m B.ByteString
 decodeB64UrlNoPaddingText = fromEitherM
-    . first (Base64DecodeException . T.pack)
-    . B64U.decode
+    . first Base64DecodeException
+    . B64U.decodeBase64Unpadded
```
emilypi, line 585: This used to support arbitrary padding. Now it's strict in the padding type

PR #1065, opened 2020-06-01
fix some exception handling

larskuhtz:


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -1,11 +1,14 @@
...
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
larskuhtz, line 11:
```suggestion
```

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -1,11 +1,14 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveAnyClass #-}
```
larskuhtz, line 3:
```suggestion
```

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -429,12 +434,11 @@ applyUpgrades v cid height
...
 
-      r <- try $ runGenesis tx permissiveNamespacePolicy interp
-
-      case r of
+      tryAllSynchronous (runGenesis tx permissiveNamespacePolicy interp) >>= \case
         Right _ -> return ()
-        Left (e :: SomeException) -> do
+        Left e -> do
           logError $ "Upgrade transaction failed! " <> sshow e
+            -- TODO: is it really OK, to just swallow the exception?
```
larskuhtz, line 441:
```suggestion
          throwM e
```

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -429,12 +431,11 @@ applyUpgrades v cid height
...
 
-      r <- try $ runGenesis tx permissiveNamespacePolicy interp
-
-      case r of
+      tryAllSynchronous (runGenesis tx permissiveNamespacePolicy interp) >>= \case
         Right _ -> return ()
-        Left (e :: SomeException) -> do
+        Left e -> do
           logError $ "Upgrade transaction failed! " <> sshow e
+          throwM e
```
larskuhtz, line 438:
```suggestion
         void $  throwM e
```

emilypi: Do you want to to force this? `($!)`?

larskuhtz:
There are two different test failures:

1. `(interface fungible-v2    " St...: Failure: Existing interface found (interfaces cannot be upgraded)`
2. `user error (<interactive>:1:0:Unable to locate executable for Z3 Executable specified: "z3"`

The first is caused by rethrowing in the case that the update fails. The second is related to the upgrade of pact. It seems that the pact code repl tests now depend on `z3` being available in the test environment.

mightybyte:
I just pushed a commit that should fix the Z3 failure.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -429,13 +431,11 @@ applyUpgrades v cid height
...
-      r <- try $ runGenesis tx permissiveNamespacePolicy interp
-
-      case r of
+      tryAllSynchronous (runGenesis tx permissiveNamespacePolicy interp) >>= \case
         Right _ -> return ()
-        Left (e :: SomeException) -> do
+        Left e -> do
           logError $ "Upgrade transaction failed! " <> sshow e
-          return ()
+          void $ throwM e
```
larskuhtz, line 438: I consider this line the only actual change to production of this PR (beside the update of the pact pin).

test/Chainweb/Test/Pact/PactExec.hs
```diff
@@ -70,23 +70,32 @@ testVersion = FastTimedCPM petersonChainGraph
...
+    , withPactCtxSQLite testVersion (bhdbIO rocksIO) pdb defaultPactServiceConfig $
+        \ctx -> _schTest $ execTxsTest ctx "testTfrGas" testTfrGas
+    , withPactCtxSQLite testVersion (bhdbIO rocksIO) pdb defaultPactServiceConfig $
+        \ctx -> _schTest $ execTxsTest ctx "testGasPayer" testGasPayer
+    , withPactCtxSQLite testVersion (bhdbIO rocksIO) pdb defaultPactServiceConfig $
+        \ctx -> _schTest $ execTxsTest ctx "testContinuationGasPayer" testContinuationGasPayer
+    , withPactCtxSQLite testVersion (bhdbIO rocksIO) pdb defaultPactServiceConfig $
+        \ctx -> _schTest $ execTxsTest ctx "testExecGasPayer" testExecGasPayer
+    , withPactCtxSQLite testVersion (bhdbIO rocksIO) pdb defaultPactServiceConfig $
+      \ctx -> _schTest $ execTest ctx testReq6
```
emilypi, line 98: Now, originally, these tests were meant to share a ctx so that we can limit their resource consumption. This seems like a costly change. Can we add a TODO to sort these out so that we can make use of a single ctx as needed like we do in `Tests/Pact/SPV.hs`?

larskuhtz:
I'll move ahead and merge. It will be subject to intense testing for the upcoming release.

PR #1071, opened 2020-06-09
Lars/devnet 20 transition

larskuhtz:


larskuhtz:
this PR is subsumed by #1075

PR #1084, opened 2020-06-19
Remove `these` dependency

slpopejoy:
Aka "`these`? Pleeze"

Motivated by dislike of `These a b` as a primarily a stylistic maneuver over `(Maybe a,Maybe b)`, as well as ambient nervousness/distrust of the maintenance of the `these` package. 

You can argue for a lawful union of `NonEmpty (Either a b)` and `These`, in that `These` elides the `(Nothing,Nothing)` case when iterating, which does add some annoying (but harmless) cases to this PR. 

However, this can be countered by aesthetic dislike of both: `NonEmpty a` for its rampant API leakage into unsafe `[a]` via `init`, `drop` etc, and generally horrible UX; `These` because "really, I need to depend on a library for a hipster handling of `(Maybe a, Maybe b)`"? I personally dislike `This` and `That` imbued with a polarity.

The change of `alignWith f rs txs` to `f <$> padZip rs txs` is an example of "not having to learn a whole new DSL for a common need", with `padZip` in particular being one of those "where have you been all my life?" utilities. However, `semialign` makes the mistake that `these` does of seeing the 2-tuple as so primary: I would _love_ to have `padZip3`, `padZip4` but nooooooo.

In any case this is what happens when an opinionated programmer can't sleep. These? Deez.

NB: I think `NonEmpty` is primarily useful in APIs. In implementation code it's a drag.

NB: I found `padZip` with yet another awesome hoogle type search https://hoogle.haskell.org/?hoogle=%5Ba%5D%20-%3E%20%5Bb%5D%20-%3E%20%5B(Maybe%20a%2CMaybe%20b)%5D

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -391,7 +391,6 @@ library
         , temporary >= 1.3
         , text >= 1.2
         , trifecta >= 2.1
-        , these >= 1 && < 1.2
         , time >= 1.8
         , tls >=1.5.2
         , token-bucket >= 0.1
--- a/src/Chainweb/Chainweb.hs
+++ b/src/Chainweb/Chainweb.hs
@@ -128,7 +128,7 @@ import Control.Lens hiding ((.=), (<.>))
 import Control.Monad
 import Control.Monad.Catch (throwM)
 
-import Data.Align (alignWith)
+import Data.Align (padZip)
 import Data.Bifunctor (second)
 import Data.CAS (casLookupM)
 import Data.Foldable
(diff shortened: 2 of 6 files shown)


emilypi:
So, just some comments for Dr. Kuhtz when he reads this PR: there are a few alternatives to a bare tuple of `Maybe` values that we can consider: 

1. [these-skinny](https://hackage.haskell.org/package/these-skinny) - this is a drop-in replacement for `these` in most respects. It exports the main datatype and a bunch of auxiliary functions with _no unnecessary deps_!!!

2. [smash](https://hackage.haskell.org/package/smash) - this is my mix-tape and it is 🔥. This package similarly has no unnecessary deps, but has a few more bells and whistles in the sense that there are bifunctor instances and related combintors - all from deps bunded with GHC. The `Can` datatype is isomorphic to `(Maybe a, Maybe b)`, with a nicer UX. We'd also control maintenance for this package.

slpopejoy:
Closing in favor of #1085

PR #1085, opened 2020-06-21
Migrate to `these-skinny` and eliminate `semialign`

slpopejoy:


chainweb.cabal
```diff
@@ -391,7 +390,7 @@ library
         , temporary >= 1.3
         , text >= 1.2
         , trifecta >= 2.1
-        , these >= 1 && < 1.2
+        , these-skinny >= 0.7.4
```
larskuhtz, line 393: 👍

chainweb.cabal
```diff
@@ -527,7 +526,6 @@ test-suite chainweb-tests
         , retry >= 0.7
         , rosetta >= 1.0
         , scheduler >= 1.4
-        , semialign >= 1
```
larskuhtz, line 530: 👍

src/Chainweb/Utils.hs
```diff
@@ -391,12 +389,13 @@ roundBy :: Integral a => a -> a -> a
...
-partitionEithersNEL :: NonEmpty (Either a b) -> These (NonEmpty a) (NonEmpty b)
-partitionEithersNEL (h :| es) = case bimap NEL.nonEmpty NEL.nonEmpty $ partitionEithers es of
-    (Nothing, Nothing) -> either (This . pure) (That . pure) h
-    (Just as, Nothing) -> This as
-    (Nothing, Just bs) -> That bs
-    (Just as, Just bs) -> These as bs
+-- | Elide 'semialign' import with this simple Vector-specialized version.
+-- O(n)-ish -- O(min (m,n) + 2*max(m-n,n-m))
+alignWithV :: (These a b -> c) -> V.Vector a -> V.Vector b -> V.Vector c
+alignWithV f a b = V.zipWith (\a' -> f . These a') a b <> case (V.length a,V.length b) of
```
emilypi, line 395: It would make sense to use `splitAt (max (V.length a) (V.length b))` for both vectors here because it's constant time in all operations. I believe that should shave off the `2*` multiplier and just leave you at `O(max(n,m))`.

larskuhtz: `O(max(n,m)) == O(2*max(n,m))` :-)

Aren't `V.length` and `V.drop` `O(1)`?

emilypi: `length` is, but i didn't know `drop` is too! So we should be fine then. Regardless, `splitAt` does the dropping for you :)

slpopejoy: I know O(n) is true for all linear but sometimes I write the pedantic version to document the linearity. In this case, the `min` is the zip and the `2*max diff` is `mappend` and `mapM`.

PR #1086, opened 2020-06-22
Add devnet payloads for chains 10-19

emilypi:


src/Chainweb/Pact/Transactions/MainnetKADTransactions.hs
```diff
@@ -0,0 +1,18 @@
...
+import Chainweb.Transaction
+import Chainweb.Utils
+
+transactions :: IO [ChainwebTransaction]
+transactions =
+  let decodeTx t =
+        fromEitherM . (first (userError . show)) . codecDecode chainwebPayloadCodec =<< decodeB64UrlNoPaddingText t
+  in mapM decodeTx [
+    "eyJoYXNoIjoiUEZfMUtLMmJPZnlqSDU5ZUo0aGUzcmpMWlBTbzFzeGFWRHUzaGxRVDZmdyIsInNpZ3MiOltdLCJjbWQiOiJ7XCJuZXR3b3JrSWRcIjpudWxsLFwicGF5bG9hZFwiOntcImV4ZWNcIjp7XCJkYXRhXCI6bnVsbCxcImNvZGVcIjpcIihjb2luLnJlbWVkaWF0ZSBcXFwiZTdmNzYzNGU5MjU1NDFmMzY4YjgyN2FkNWM3MjQyMTkwNTEwMGY2MjA1Mjg1YTc4YzE5ZDdiNGEzODcxMTgwNVxcXCIgMTAwLjApXCJ9fSxcInNpZ25lcnNcIjpbXSxcIm1ldGFcIjp7XCJjcmVhdGlvblRpbWVcIjowLFwidHRsXCI6MTcyODAwLFwiZ2FzTGltaXRcIjowLFwiY2hhaW5JZFwiOlwiXCIsXCJnYXNQcmljZVwiOjAsXCJzZW5kZXJcIjpcIlwifSxcIm5vbmNlXCI6XCJtYWlubmV0LXJlbWVkaWF0aW9ucy1rYWQtb3BzXCJ9In0"
+    ]
```
emilypi, line 18: This is the single remediation on Chain 0

src/Chainweb/Pact/Transactions/UpgradeTransactions.hs
```diff
@@ -38,3 +42,16 @@ upgradeTransactions Mainnet01 cid = case cidInt of
         cidInt = chainIdInt cid
 upgradeTransactions Development _ = Devnet.transactions
 upgradeTransactions _ _ = Other.transactions
+
+twentyChainUpgradeTransactions :: ChainwebVersion -> ChainId -> IO [ChainwebTransaction]
+twentyChainUpgradeTransactions Mainnet01 cid = case chainIdInt @Int cid of
+  0 -> MNKAD.transactions
+  c | c >= 1, c <= 9 -> return [] -- no upgrades or remeds
+  c | c >= 10, c <= 19 -> MN20.transactions
+  c -> internalError $ "Invalid mainnet chain id: " <> sshow c
```
emilypi, line 51: Single remediation tx on chain 0. 


Currently, this setup will immediately upgrade chains 10-19 with upgrades, but this is possibly not the right route to go. We may want to have 10-19 use v2 from genesis (will that even work?). 

If we want v2 to be immediately available on 10-19, it would take 5min to write an alternative workflow generating the proper genesis txs.

tools/ea/Ea/Genesis.hs
```diff
@@ -290,6 +308,15 @@ mainnet9 = mainnet0
     & txChainId .~ Nine
     & allocations .~ (Just mainnetAllocations9)
 
+mainnetKAD :: Genesis
+mainnetKAD = mainnet0
```
slpopejoy, line 312: Let's not base these off of 0 and probably just not use a default. In this case, it includes `mainnetKeysets` which we don't need, but it generally is hard for maintainability when it's so different from `mainnet0`.

tools/ea/Ea/Genesis.hs
```diff
@@ -147,6 +159,12 @@ developmentN = development0
     & txChainId .~ N
     & coinbase .~ (Just devNGrants)
 
+developmentKAD :: Genesis
+developmentKAD = development0
```
slpopejoy, line 163: more forgivable here to use a default but maybe just build it from scratch too for less confusion than using `development0`

pact/coin-contract/v2/coin-temp.pact
```diff
@@ -0,0 +1,586 @@
...
+            , "balance" : 0.0
+            })
+
+          "Allocation successfully released to main ledger")
+    )))
+
+)
+
+(create-table coin-table)
+(create-table allocation-table)
```
slpopejoy, line 586:
```suggestion
(create-table allocation-table)
(enforce 
  (= "ut_J_ZNkoyaPUEJhiwVeWnkSQn9JT9sQCWKdjjVVrWo" 
  (at ('hash (describe-module 'coin)))
  "hash mismatch")
```

emilypi: 👍  inline your change

tools/ea/Ea/Genesis.hs
emilypi: Addressed

tools/ea/Ea/Genesis.hs
emilypi: Addressed

PR #1088, opened 2020-06-22
Add deltas to BlockTxHistory on branch

slpopejoy:


src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -279,31 +282,52 @@ doGetBlockHistory dbenv blockHeader d = runBlockEnv dbenv $ do
     startTxId <- if (bHeight == genesisHeight v cid)
       then pure 0  -- genesis block
       else getEndTxId db (pred bHeight) (_blockParent blockHeader)
-    history <- queryHistory db (domainTableName d) startTxId endTxId
-    return $! BlockTxHistory $ foldl' groupByTxid mempty history
+    let tname = domainTableName d
+    history <- queryHistory db tname startTxId endTxId
+    let (!hkeys,tmap) = foldl' procTxHist (S.empty,mempty) history
+    !prev <- M.fromList . catMaybes <$> mapM (queryPrev db tname startTxId) (S.toList hkeys)
```
LindaOrtega, line 288:
```suggestion
    !prev <- M.fromList . catMaybes <$> mapM (queryPrev db tname endTxId) (S.toList hkeys)
```
I think its supposed to be `endTxId` here since the SQL code is:
```
WHERE rowkey = ? AND txid < ?
```

slpopejoy: "Prev" is everything BEFORE the block, thus before startTxId ...

PR #1089, opened 2020-06-23
Expose graph history so explorer works with 20 chains

mightybyte:
This PR allows the existing block explorer to work with 20 chain graphs.


src/Chainweb/Graph.hs
```diff
@@ -41,6 +41,7 @@ module Chainweb.Graph
 -- * Chain Graph
 
 , ChainGraph
+, _chainGraphGraph
```
larskuhtz, line 44: The constructors of the `ChainGraph` must not be exported. 

There is a lensy `Getter` `chainGraphGraph` that is exported. For a non-lensy version the constructor fields would have to be renamed and `_chainGraphGraph` would have to be exported as separately defined read-only function.

src/Chainweb/RestAPI/NodeInfo.hs
```diff
@@ -38,7 +44,12 @@ data NodeInfo = NodeInfo
     nodeVersion :: ChainwebVersion
   , nodeApiVersion :: Text
   , nodeChains :: [Text]
+  -- ^ Current list of chains
   , nodeNumberOfChains :: !Int
+  -- ^ Current number of chains
+  , nodeGraphHistory :: [(BlockHeight, [(Int, [Int])])]
```
larskuhtz, line 50: This should only include the *known name* of the graph, not the graph structure. 

The info endpoint is hit potentially very often and it should be very efficient, both in terms of server side compute and in terms of payload size. We expect chainweb graphs to grow very large, so we shouldn't include their (well known static) structure into the response from the info endpoint.

slpopejoy:
> The info endpoint is hit potentially very often and it should be very efficient ... We expect chainweb graphs to grow very large

I don't see the performance issue becoming anything worth discussing for a long time. Maybe not ever -- a 1000-chain graph would be 24k uncompressed. 

More importantly, the design of the `info` endpoint is appealing intuitively so that the node can act as a single source of truth about it's configuration, it is easy to maintain, and requires no infrastructure coordination with other projects.

src/Chainweb/Graph.hs
mightybyte: Oh, I didn't notice that.  Fixed.

src/Chainweb/RestAPI/NodeInfo.hs
slpopejoy: As discussed, after this release we will be moving the `info` endpoint out of the `peer` network, to enable putting these kinds of services behind a load balancer etc, and eventually not have to serve this from a node at all potentially.

PR #1093, opened 2020-06-23
[Rosetta] Return Balance Increase/Decrease in Tx Operations

LindaOrtega:


src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -279,31 +282,52 @@ doGetBlockHistory dbenv blockHeader d = runBlockEnv dbenv $ do
...
+                "ORDER BY txid DESC LIMIT 1"
+      r <- qry db sql
+           [SText k,SInt s]
+           [RBlob]
+      case r of
+        [] -> return Nothing
+        [[SBlob value]] -> Just . (RowKey $ toS sk,) <$> toTxLog d k value
+        _ -> internalError $ "queryPrev: expected 0 or 1 rows, got: " <> T.pack (show r)
+
+
```
emilypi, line 330: instead of querying the database twice, would it be possible to just condense this query into a single one, collecting just the first two rows generated by a rowkey for some table ordered by txid?

src/Chainweb/Rosetta/Internal.hs
```diff
@@ -242,37 +244,52 @@ gasTransactionAcc accTyp txa@(TxAccumulator logs' _) ctx = combine logs'
...
 
     listErr expectedMsg logs = Left $
       expectedMsg ++ ": Received logs list " ++ show logs
 
+    -- NOTE: no way to check fund's txId and prev txId seen are sequential
+    -- Assumption: Operations related to a transaction have sequential txIds.
+    assertSequentialTxIds tid1 tid2
+      | tid2 == succ tid1 = pure ()
+      | otherwise = Left $ "Expected sequential txId, found txId1="
+        ++ show tid1 ++ ", txId2=" ++ show tid2
```
emilypi, line 292: i don't think this is going to be true in general for someone's transactions. We enforce this by construction as well. It's probably unnecessary to test

src/Chainweb/Pact/Service/Types.hs
```diff
@@ -91,12 +91,17 @@ instance FromJSON PactException
...
--- for public API use; ToJSONs are for logging output.
-newtype BlockTxHistory = BlockTxHistory { _blockTxHistory :: Map TxId [TxLog Value] }
-  deriving (Eq,Generic, NFData)
+-- | Gather tx logs for a block, along with last tx for each
+-- key in history, if any
+-- Not intended for public API use; ToJSONs are for logging output.
+data BlockTxHistory = BlockTxHistory
+  { _blockTxHistory :: Map TxId [TxLog Value]
+  , _blockPrevHistory :: Map RowKey (TxLog Value)
+  }
```
emilypi, line 100: Let's make sure these are strict so they don't leak

src/Chainweb/Rosetta/Util.hs
```diff
@@ -50,7 +47,17 @@ import Chainweb.Version
...
-type AccountLog = (T.Text, Decimal, Value)
+newtype BalanceDelta = BalanceDelta { _balanceDelta :: Decimal }
+  deriving (Show, Eq)
+data AccountLog = AccountLog
+  { _accountLogKey :: T.Text
+  , _accountLogBalanceTotal :: Decimal
+  , _accountLogBalanceDelta :: BalanceDelta
+  , _accountLogCurrGuard :: Value
+  , _accountLogPrevGuard :: Value
+  }
```
emilypi, line 58: !'s on all of these

src/Chainweb/Rosetta/Internal.hs
LindaOrtega: took it out in latest commit

src/Chainweb/Pact/Service/Types.hs
LindaOrtega: done

src/Chainweb/Rosetta/Util.hs
LindaOrtega: done

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
LindaOrtega: @slpopejoy thoughts?

slpopejoy: I banged my head against that very question and came up empty, I think at a minimum you need a temp table. Note that strategically this stuff will all be in chainweb-data at some point so less-optimum is ok for now.

PR #1094, opened 2020-06-24
Remove QuickCheck from production code and add missing module descriptions

larskuhtz:


src/Chainweb/HostAddress.hs
```diff
@@ -232,7 +219,7 @@ ipV6Parser = p0
 
 portParser :: Parser Port
 portParser = Port
-    <$> (decimal >>= \(d :: Integer) -> int d <$ guard (d < 2^(16 :: Int)))
+    <$> (decimal >>= \(d :: Integer) -> int d <$ guard (d <= 2^(16 :: Int)-1))
```
emilypi, line 222: isn't this an equivalent statement?

src/Chainweb/Pact/NoCoinbase.hs
```diff
@@ -8,7 +8,7 @@
 -- Maintainer: Lars Kuhtz <lars@kadena.io>
 -- Stability: experimental
 --
--- TODO
+-- A fake coin base for testing purposes.
```
emilypi, line 11: It's not fake - it's necessary for genesis transactions

src/Chainweb/HostAddress.hs
larskuhtz: it is for `d :: Integer`. I got bitten by copy and pasting it to a context with `d :: Word16` in which case it isn't equivalent. So, I thought it was better to use the more robust code.

emilypi: ahh interesting alright!

src/Chainweb/Pact/NoCoinbase.hs
larskuhtz, line 11:
```suggestion
-- A noop coin base for genesis transactions and testing purposes.
```

larskuhtz: Oh, I didn't know that. I fixed the comment.

PR #1095, opened 2020-06-25
Fix Arbitrary BlockHeader instance for genesis blocks

larskuhtz:


src/Chainweb/BlockHash.hs
```diff
@@ -209,8 +209,8 @@ decodeBlockHashWithChainId = (,) <$!> decodeChainId <*> decodeBlockHash
 decodeBlockHashRecord :: MonadGet m => m BlockHashRecord
 decodeBlockHashRecord = do
     l <- getWord16le
-    hashes <- mapM (const decodeBlockHashWithChainId) [1 .. l]
-    return $! BlockHashRecord $! HM.fromList hashes
+    hashes <- replicateM (int l) decodeBlockHashWithChainId
+    return $ BlockHashRecord $! HM.fromList hashes
```
larskuhtz, line 213: This is an unrelated change that improves readability and removes a redundant `$!`.

test/Chainweb/Test/Orphans/Internal.hs
```diff
@@ -194,7 +194,9 @@ arbitraryBlockHashRecordVersionHeightChain v h cid
         <$> infiniteListOf arbitrary
     | otherwise = discard
   where
-    graph = chainGraphAt v (max 0 (h - 1))
+    graph
+        | h == genesisHeight v cid = chainGraphAt v h
```
larskuhtz, line 198: This is the actual fix: genesis blocks are treated different than other blocks during graph transitions. The adjacent dependencies of a genesis header are from the graph at the height of the header itself. For other headers the graph of the parent (i.e. the target of the dependency) is used

test/Chainweb/Test/Version.hs
```diff
@@ -142,8 +142,11 @@ prop_headerSizeBytes v = property $ do
...
-    return
-        $ counterexample ("header: " <> sshow h)
-        $ workSizeBytes (_chainwebVersion h) (_blockHeight h) === l
+    if (_blockHeight h == genesisHeight v (_chainId h))
+      then discard
+      else do
+        let l = int $ B.length $ runPut $ encodeBlockHeaderWithoutHash h
+        return
+            $ counterexample ("header: " <> sshow h)
+            $ workSizeBytes (_chainwebVersion h) (_blockHeight h) === l
```
larskuhtz, line 152: genesis headers are never mined, and `workSizeBytes` doesn't return a meaning full result for them.

PR #1098, opened 2020-07-01
don't overlap on 20-chain upgrades

emilypi:


src/Chainweb/Version.hs
```diff
@@ -777,8 +777,8 @@ twentyChainUpgrade
     -> ChainId
     -> BlockHeight
     -> Bool
-twentyChainUpgrade Mainnet01 _ 745000 = True
-twentyChainUpgrade Development _ 150 = True
+twentyChainUpgrade Mainnet01 _ h = h == 745000
+twentyChainUpgrade Development _ h = h == 150
```
larskuhtz, line 781: Could this be derived from `chainGraphs`? If not, does `chainGraphs` put some constraints on this function? If so, could you please add a comment or even a test case?

src/Chainweb/Version.hs
```diff
@@ -777,8 +777,8 @@ twentyChainUpgrade
     -> ChainId
     -> BlockHeight
     -> Bool
-twentyChainUpgrade Mainnet01 _ 745000 = True
-twentyChainUpgrade Development _ 150 = True
+twentyChainUpgrade Mainnet01 _ h = h == 745000
```
larskuhtz, line 780: hmm, is this already decided? Is this safe to put this value here? 

I find it dangerous to put in modification of mainnet01 before we decide on the final fork heights. Otherwise, bad things may happen if the fork gets delayed for whatever reason or we forget about his setting for other reasons.

emilypi: No, this is undecided, and I've raised this. It's just a placeholder

src/Chainweb/Version.hs
emilypi: we could probably re-instrument the way we determine block heights with `chainGraphs`, but that's a separate PR addressing a separate issue.

src/Chainweb/Version.hs
larskuhtz: There's currently no graph transition configured for `mainnet01`. Shouldn't this here just be `False` in order to be consistent?

test/Chainweb/Test/Pact/TransactionTests.hs
```diff
@@ -311,3 +299,55 @@ testCoinbaseUpgradeDevnet cid upgradeHeight = do
...
+      }
+
+    logResults logs =
+      let f l = (_txDomain l, _txKey l, l ^? txValue . _Object . ix "balance")
+      in f <$> logs
+
+    expectedResults =
+      [ ("USER_coin_coin-table","NoMiner",Just (Number 0.1))
+      , ( "USER_coin_coin-table"
+        , "e7f7634e925541f368b827ad5c72421905100f6205285a78c19d7b4a38711805"
```
slpopejoy, line 331: 👍

test/Chainweb/Test/Pact/TransactionTests.hs
```diff
@@ -311,3 +299,55 @@ testCoinbaseUpgradeDevnet cid upgradeHeight = do
...
+        , Just (Number 50.0)
```
slpopejoy, line 332: 👍

PR #1099, opened 2020-07-02
Make info endpoint show currently in use chains

mightybyte:


src/Chainweb/RestAPI/NodeInfo.hs
```diff
@@ -56,15 +60,21 @@ data NodeInfo = NodeInfo
 instance ToJSON NodeInfo
 instance FromJSON NodeInfo
 
-nodeInfoHandler :: ChainwebVersion -> Server NodeInfoApi
-nodeInfoHandler v = return
-  NodeInfo
-  { nodeVersion = v
-  , nodeApiVersion = prettyApiVersion
-  , nodeChains = (chainIdToText <$> (toList $ chainIds v))
```
mightybyte, line 64: @larskuhtz Do we need to add a `BlockHeight` parameter to this `chainIds` function?

mightybyte:
Yeah, agree with the sentiment.  The current behavior is to return the "current" value, so this is mainly a backwards compatibility thing that we will phase out at some point down the road.

PR #1100, opened 2020-07-03
Fix leaking sqlite savepoints

larskuhtz:
This PR currently does the following:

*Pact Service Checkpointer Usage*:

* Remove unsafe (no-bracketed) function `restoreCheckpointer` and inline the code into the save `withCheckpointer` call.
* Do `rewindTo` in `withCheckpointer`
* Add new function `withCheckpointerWithoutRewind` for special internal use-cases; `withCheckpointer` should be preferred where possible.
* Wrap all uses of `withCheckpointer` into `withBatch` or `withDiscardedBatch`.
* Assert that `withCheckpointer` is only used within the scope of a checkpointer batch.
* Add function `syncParentHeader` that synchronizes `_psParentHeader` with the latest header of the checkpointer.


src/Chainweb/Pact/PactService.hs
```diff
@@ -364,41 +364,45 @@ serviceRequests logFn memPoolAccess reqQ = do
...
+                        V.map (() <$) <$> execPreInsertCheckReq txs
                 go
             BlockTxHistoryMsg (BlockTxHistoryReq bh d resultVar) -> do
-              trace logFn "Chainweb.Pact.PactService.execBlockTxHistory" bh 1 $
-                tryOne "execBlockTxHistory" resultVar $
-                execBlockTxHistory bh d
+                trace logFn "Chainweb.Pact.PactService.execBlockTxHistory" bh 1 $
+                    tryOne "execBlockTxHistory" resultVar $
+                        execBlockTxHistory bh d
+                go
```
larskuhtz, line 400: This adds the recursive call to `go`. I guess it was an omission that it wasn't there?

src/Chainweb/Pact/PactService.hs
```diff
@@ -364,41 +364,45 @@ serviceRequests logFn memPoolAccess reqQ = do
...
             HistoricalLookupMsg (HistoricalLookupReq bh d k resultVar) -> do
-              trace logFn "Chainweb.Pact.PactService.execHistoricalLookup" bh 1 $
-                tryOne "execHistoricalLookup" resultVar $
-                execHistoricalLookup bh d k
+                trace logFn "Chainweb.Pact.PactService.execHistoricalLookup" bh 1 $
+                    tryOne "execHistoricalLookup" resultVar $
+                        execHistoricalLookup bh d k
+                go
```
larskuhtz, line 405: Adds recursive call to `go`. See above.

src/Chainweb/Pact/PactService.hs
```diff
@@ -425,6 +429,7 @@ serviceRequests logFn memPoolAccess reqQ = do
                     , "): "
                     , show e
                     ]
+                cleanupDb
```
larskuhtz, line 432: This removes any possibly pending save points from the sqlite context. Generally, this should not be necessary, because all inner logic (in particular `withCheckpointer`) are supposed to do proper cleanup, even in case of exceptions.

However, code can have bugs and we've observed that it can actually happen that there are dangling SQL savepoints which leads to sqlite never committing anything to disk and results in memory leaks.

This change fixes the memory leak by allowing pact service to recover, but doesn't resolve the underlying issue.

src/Chainweb/Pact/PactService.hs
```diff
@@ -320,15 +322,10 @@ initializeCoinContract _logger v cid pwo = do
...
 
-    readContracts cp = do
-      mbLatestBlock <- liftIO $ _cpGetLatestBlock cp
-      (bhe, bhash) <- case mbLatestBlock of
-        Nothing -> throwM NoBlockValidatedYet
-        (Just !p) -> return p
-      let target = Just (succ bhe, bhash)
-      parent <- ParentHeader <$!> lookupBlockHeader bhash "initializeCoinContract"
-      setParentHeader "readContracts" parent
-      withCheckpointer target "readContracts" $ \(PactDbEnv' pdbenv) -> do
```
larskuhtz, line 331: This code got factored out into the new function `syncParentHeader` below.

src/Chainweb/Pact/PactService.hs
```diff
@@ -320,15 +322,10 @@ initializeCoinContract _logger v cid pwo = do
...
+    readContracts = withDiscardedBatch $ do
```
larskuhtz, line 325: wraps the use of `withCheckpointer` info `withDiscardedBatch`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -1031,28 +1108,21 @@ execNewBlock mpAccess parent miner = handle onTxFailure $ do
...
-  where
-    start = do
-        cp <- getCheckpointer
-        liftIO (_cpBeginCheckpointerBatch cp)
-        return cp
-    end = liftIO . _cpDiscardCheckpointerBatch
-
+withDiscardedBatch act = do
+    cp <- getCheckpointer
+    local (set psIsBatch True) $ bracket_
```
emilypi, line 1122:
```suggestion
    locally psIsBatch (const True) $ bracket_
```

src/Chainweb/Pact/PactService.hs
```diff
@@ -573,90 +598,143 @@ validateHashes bHeader pData miner transactions =
...
+-- checkpointer. The checkpointer is rewinded and restored to the state to the
 -- provided target.
 --
-withCheckpointerRewind
+-- The result of the inner action indicates whether the resulting checkpointer
+-- state should be discarded or saved.
+--
+-- If the inner action throws an exception the checkpointer state is discarded.
+--
+withCheckpointer
```
slpopejoy, line 723: Might want to leave this `withCheckpointerRewind` since at this time calling code needs to be aware of semantics (unfortunately).

src/Chainweb/Pact/PactService.hs
```diff
@@ -1449,6 +1542,8 @@ execPreInsertCheckReq txs = do
           cid = _chainId psEnv
       liftIO $ fmap Discard $ V.zipWith (>>)
         <$> validateChainwebTxs v cid cp parentTime lenientCreationTime currHeight txs (runGas pdb psState psEnv)
+            -- FIXME: it seems suspicious that the checkpointer is provided to the inner action.
```
slpopejoy, line 1545: This is because `validateChainwebTxs` runs in `IO` due to the requirements of the mempool API, so `PactServiceM` is not available. 

The requirement is to do a dupe check https://github.com/kadena-io/chainweb-node/blob/3a5a5a73a711e22e61bfb3670ad9d6e3b50f0516/src/Chainweb/Pact/PactService.hs#L846-L851

note that `runGas` is a closure provided for the gas check and has the `pdb`/`psState`/`psEnv` for the current checkpointer. 

I would remove this FIXME (the real problem is the mempool API).

src/Chainweb/Pact/PactService.hs
```diff
@@ -573,90 +598,143 @@ validateHashes bHeader pData miner transactions =
...
+withCheckpointerWithoutRewind target caller act = do
+    checkPointer <- getCheckpointer
+    logInfo $ "restoring (with caller " <> caller <> ") " <> sshow target
+
+    -- check requirement that this must be called within a batch
+    unlessM (asks _psIsBatch) $
+        error $ "Code invariant violation: withCheckpointer called by " <> caller <> " outside of batch. Please report this as a bug."
+    -- we allow exactly one nested call of 'withCheckpointer', which is used
+    -- during fastforward in 'rewindTo'.
+    unlessM ((<= 1) <$> asks _psCheckpointerDepth) $ do
```
slpopejoy, line 681: I'm not satisfied with the implementation of the two new invariants. We can consider merging this PR in interests of expedience as I am satisfied it fixes the issues, but would be happier if we resolve the following issues:

1. `psIsBatch` is only enforced here, which means that we only care about it when we're doing checkpointer restore/finalize. We need to consider whether batching is even meaningful in PactService as a separate operation -- the question being, *do we always pair batch calls with a `withCheckpointerXXX` call*? If so we should just do the batching here and get rid of the flag.

2. If `psIsBatch` is truly necessary as a separate invariant, we should consider enforcing against nested batching in `withBatchXXX` as I'm pretty sure we don't want that. 

2. `psCheckpointerDepth` semantically indicates that we never want to nest `withCheckpointerXXX` calls. Thus it should be a boolean, and cannot be `True` coming into this function.

src/Chainweb/Pact/PactService.hs
larskuhtz: Yes, renaming it to `withCheckpointerRewind` makes sense.

src/Chainweb/Pact/PactService.hs
larskuhtz: oh, yes, that FIXME shouldn't be there. I just forgot to delete it.

src/Chainweb/Pact/PactService.hs
slpopejoy: I know this got merged in the other PR but I'm wondering how we missed this. Wouldn't rosetta tests be failing after the first call?

slpopejoy: Also, just want to note that this is a weakness in this function, the type of the casing should force code to handle the recursion (or not, ie for `CloseMsg`)

src/Chainweb/Pact/PactService.hs
larskuhtz: These invariants can and should both be enforced on the type level. We never expect them to trigger in production. It's just for the case that maintainers of the code forget to nest transactions properly --  which obvious is easy to miss :-)

Implementing it this way, was the least intrusive (smallest) change I could think of. But I fully agree that this should be replaced by an implementation that supports proper scoping more explicitly.

1. Agreed, but I considered it out of scope for this PR to change that. So, given that this invariant exists and that it's unnoticed violation caused bugs, I added this check as a temporary solution to enforce this invariant.

2. Yes, we can add that as a third invariant, too. Also, I don't think anything would break if someone nests it. But not 100%.

3. I can't be Boolean, since it is actually called nested. But only 2 levels deep. This check ensures that we don't accidentally introduce additional nesting. It's a bit like a golden test. If a programer would change the nesting structure, they would be notified here. So it prevents accidentally messing things up.

src/Chainweb/Pact/PactService.hs
```diff
@@ -1074,7 +1144,7 @@ execLocal
     :: PayloadCasLookup cas
     => ChainwebTransaction
     -> PactServiceM cas (P.CommandResult P.Hash)
-execLocal cmd = do
+execLocal cmd = withDiscardedBatch $ do
```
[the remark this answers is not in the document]
larskuhtz: `_cpDiscard` discards only `Block` SAVEPOINTs. It doesn't discard the `PreBlock` SAVEPOINTS.

`_cpRestore` does the following SAVEPOINT operations:

```
1.  open `PreBlock`
2. release/rollback `PreBlock`
3. open `Block`
```

`_cpDiscard` does

```
4. release/rollback `Block`
```

The `Batch` SAVEPOINTs wrap all of the above. Without it aborting within the action given to `withCheckpointer` would only rollback `Block` but not `PreBlock`.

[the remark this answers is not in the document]
larskuhtz: In particular `_cpSave` is a complex operation that even when executed under a mask in a finalizer may fail in different ways. If that happens, it would result in a leaked `SAVEPOINT`. Discarding a batch is much simpler and more robust.

src/Chainweb/Pact/PactService.hs
larskuhtz, line 1545:
```suggestion
```

src/Chainweb/Pact/PactService.hs
```diff
@@ -1449,6 +1542,7 @@ execPreInsertCheckReq txs = do
           cid = _chainId psEnv
       liftIO $ fmap Discard $ V.zipWith (>>)
         <$> validateChainwebTxs v cid cp parentTime lenientCreationTime currHeight txs (runGas pdb psState psEnv)
+            -- withCurrentCheckpointer should abstract over the checkpointer that is used.
```
larskuhtz, line 1545:
```suggestion
 ``

src/Chainweb/Pact/PactService.hs
larskuhtz: Actually, in the first iteration I used `Bool` and only then noticed that there are nested calls :-)

slpopejoy: Oh I see, `0` and `1` are valid states. Yuck ... when is it nested?

src/Chainweb/Pact/PactService.hs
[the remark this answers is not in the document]
slpopejoy: I think the `PreBlock` was originally intended to possibly "keep the rewind but discard the playthrough" on new block? In any case, I think we should ditch the `PreBlock` batch altogether, since it should succeed or fail alongside `Block`. Not for this PR obviously.

src/Chainweb/Pact/PactService.hs
larskuhtz: In `rewindTo` blocks are replayed to fast forward the checkpointer. Each block runs in it's own `BLOCK` SAVEPOINT environment. That the only place where it's nested. I put this check to avoid that we accidentally add another nesting. I would be particularly concerned about recursive nesting.

src/Chainweb/Pact/PactService.hs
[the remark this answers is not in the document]
larskuhtz: Agreed, that this is stuff for another PR. This PR tries to be conservative with respect to lower-level checkpointer and pact execution logic.

PR #1101, opened 2020-07-06
stricter pact service

larskuhtz:
Use strict `modify'` instead of `.=` and `%=` which are lazy.


larskuhtz:
I haven't checked benchmark results. But load tests on devnet did look fine -- though didn't show big improvements neither.

PR #1103, opened 2020-07-06
Later 20 chains fork

mightybyte:


src/Chainweb/Version.hs
```diff
@@ -512,7 +512,7 @@ chainwebGraphs (TimedCPM g) = pure (0, g)
 chainwebGraphs (FastTimedCPM g) = pure (0, g)
 chainwebGraphs Testnet04 = pure (0, petersonChainGraph)
 chainwebGraphs Mainnet01 = pure (0, petersonChainGraph)
-chainwebGraphs Development = (50, twentyChainGraph) NE.:| [ (0, petersonChainGraph) ]
+chainwebGraphs Development = (twentyChainUpgradeBlockHeight Development, twentyChainGraph) NE.:| [ (0, petersonChainGraph) ]
```
larskuhtz, line 515: Isn't this the wrong way around? I think, the upgrades should be specified here and `twentyChainUpgradeBlockHeight` should be derived from it.

src/Chainweb/Version.hs
```diff
@@ -767,6 +767,12 @@ coinV2Upgrade Development cid h
 coinV2Upgrade _ _ 1 = True
 coinV2Upgrade _ _ _ = False
 
+-- | Specify the twenty chain fork block height in one place.
+twentyChainUpgradeBlockHeight :: ChainwebVersion -> BlockHeight
+twentyChainUpgradeBlockHeight Development = 400
+twentyChainUpgradeBlockHeight Mainnet01 = maxBound
+twentyChainUpgradeBlockHeight _ = 2
```
larskuhtz, line 774: Doesn't this test chainweb versions that don't do a transition but use pact, like e.g. `FastTimedCPM`?

larskuhtz: Actually, I think `twentyChainUpgrade` should be derived from `chainwebGraphs` and this function could be dropped. I seems like an unneeded level of indirection. I think `chainGraphs` is the authoritative source for graph transition block heights.

src/Chainweb/Version.hs
```diff
@@ -778,13 +784,9 @@ coinV2Upgrade _ _ _ = False
...
 twentyChainUpgrade
     :: ChainwebVersion
-    -> ChainId
     -> BlockHeight
     -> Bool
-twentyChainUpgrade Mainnet01 _ h = h == maxBound -- TODO: update me with a calculated height
-twentyChainUpgrade Development _ h = h == 150
-twentyChainUpgrade _ _ 2 = True
-twentyChainUpgrade _ _ _ = False
+twentyChainUpgrade v h = h == twentyChainUpgradeBlockHeight v
```
larskuhtz, line 789: This performs the update at the genesis block. Is that correct, which is hard-coded. Before it was done at a non-genesis block, actually 100 blocks after genesis.

larskuhtz: After reading the Haddock comment for this function, it seems that this doesn't depend on the twenty chain transition block height at all. It should trigger early after chainweb-node 1.21 is release on testnet and mainnet and at height 2 for all other versions.

src/Chainweb/Version.hs
larskuhtz: It seems that `twentyChainUpgrade` doesn't depend on the actual height of the graph transition. So, I think, we can just drop this change.

PR #1104, opened 2020-07-06
fix pact-service serivce loop

larskuhtz:
* [x] In `PactService.serviceRequests`, add missing recursive call of the service loop after `BlockTxHistoryMsg` and `HistoricalLookupMsg` and format the function body consistently. Currently pact service exits after each of these calls and is restarted via `initializePactService`, which also resets the checkpointer, but does't reinitialize the sqlite connection. 

* [x] add missing tracing for `execLocal`, to make it consistent with all other pact service requests.

* [x] replace on usage of `catch` for `SomeException` by `catchAllSynchronous`.

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -338,9 +338,9 @@ initializeCoinContract _logger v cid pwo = do
 lookupBlockHeader :: BlockHash -> Text -> PactServiceM cas BlockHeader
 lookupBlockHeader bhash ctx = do
   bhdb <- asks _psBlockHeaderDb
-  liftIO $! lookupM bhdb bhash
-        `catch` \e -> throwM $ BlockHeaderLookupFailure $
-                      "failed lookup of parent header in " <> ctx <> ": " <> sshow (e :: SomeException)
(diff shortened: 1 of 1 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -338,9 +338,9 @@ initializeCoinContract _logger v cid pwo = do
 lookupBlockHeader :: BlockHash -> Text -> PactServiceM cas BlockHeader
 lookupBlockHeader bhash ctx = do
   bhdb <- asks _psBlockHeaderDb
-  liftIO $! lookupM bhdb bhash
-        `catch` \e -> throwM $ BlockHeaderLookupFailure $
-                      "failed lookup of parent header in " <> ctx <> ": " <> sshow (e :: SomeException)
+  liftIO $! lookupM bhdb bhash `catchAllSynchronous` \e ->
+        throwM $ BlockHeaderLookupFailure $
+            "failed lookup of parent header in " <> ctx <> ": " <> sshow e
```
larskuhtz, line 343: `catchAllSynchronous` doesn't catch asynchronous exceptions. Catching asynchronous exception interferes with top-level exception handlers. In particular it interferes with thread management from the `async` package.

PR #1109, opened 2020-07-08
fix thresholds for out of scope cut checks

larskuhtz:
This fixes an issue when the cut height jumps during a graph transition and cuts with the new height are rejected in the cut pipeline because they are considered as "to far ahead".

The solution is to use block heights instead of cut height for calculating thresholds for cuts to be ahead or behind.

### DIFF
--- a/src/Chainweb/CutDB.hs
+++ b/src/Chainweb/CutDB.hs
@@ -200,11 +200,10 @@ defaultCutDbParams v ft = CutDbParams
 -- 'processCuts' below in the module and the node would never be able to join
 -- the consensus of the network.
 --
--- NOTE: THIS NUMBER MUST BE STRICTLY LARGER THAN THE RESPECTIVE LIMIT IN
--- 'CutDB.Sync'
+-- NOTE: this number multiplied by the (current) number of chains must always be
+-- STRICTLY LARGER THAN 'catchupStepSize' in "Chainweb.CutDB.Sync".
 --
-farAheadThreshold :: BlockHeight
-farAheadThreshold = 200
+farAheadThreshold :: BlockHeight farAheadThreshold = 200
 
 -- -------------------------------------------------------------------------- --
 -- CutHashes Table
@@ -528,7 +527,7 @@ processCuts conf logFun headerStore payloadStore cutHashesStore queue cutVar = q
         curMin <- minChainHeight <$> readTVarIO cutVar
         let diam = diameter $ chainGraphAt_ headerStore curMin
             newMin = _cutHashesMinHeight x
-        let r = newMin <= curMin - 2 * int diam
+        let r = newMin + 2 * (1 + int diam) <= curMin
         when r $ loggc Debug x "skip very old cut"
             -- log at debug level because this is a common case during catchup
         return r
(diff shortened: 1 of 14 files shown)


src/Chainweb/CutDB.hs
```diff
@@ -195,16 +195,16 @@ defaultCutDbParams v ft = CutDbParams
...
+-- blocks would be immediately rejected by the cut processing pipeline
 -- 'processCuts' below in the module and the node would never be able to join
 -- the consensus of the network.
 --
 -- NOTE: THIS NUMBER MUST BE STRICTLY LARGER THAN THE RESPECTIVE LIMIT IN
 -- 'CutDB.Sync'
 --
-farAheadThreshold :: Int
-farAheadThreshold = 2000
+farAheadThreshold :: BlockHeight
```
larskuhtz, line 206: Make it explicit that this is based on `BockHeight` now.

src/Chainweb/CutDB/Sync.hs
```diff
@@ -103,11 +103,18 @@ syncSession v useOrigin p db logg env pinf = do
...
+        cur <- _cut db
+
         -- Query cut that is at most 1000 blocks ahead
-        h <- _cutHeight <$> _cut db
-        c <- getCut cenv (h + min catchupStepSize (int farAheadThreshold - 1))
-            -- Cf. documentation of 'fastAheadThreshold' for why this bound is
-            -- needed
+        let curHeight = _cutHeight cur
+            curChainCount = length $ view cutMap cur
+            limit = curHeight + min catchupStepSize (int farAheadThreshold * int curChainCount - 1)
```
larskuhtz, line 111: Ideally, we should stop using `CutHeight` wherever possible, because it jumps during graph transitions. However, we can't easily switch to `BlockHeight` here, because the limit is part of the public REST API. So, we would have to extend the API with a query parameter that limits cuts by 'BlockHeight'. We should address this in a separate PR.

src/Chainweb/CutDB/Sync.hs
```diff
@@ -103,11 +103,18 @@ syncSession v useOrigin p db logg env pinf = do
...
     receive = do
+        cur <- _cut db
+
         -- Query cut that is at most 1000 blocks ahead
-        h <- _cutHeight <$> _cut db
-        c <- getCut cenv (h + min catchupStepSize (int farAheadThreshold - 1))
-            -- Cf. documentation of 'fastAheadThreshold' for why this bound is
-            -- needed
+        let curHeight = _cutHeight cur
+            curChainCount = length $ view cutMap cur
```
slpopejoy, line 110: Is this a question we ask a lot? Wondering if this should be a util function ... fine to do in a later PR

src/Chainweb/CutDB.hs
slpopejoy, line 206: The "MUST BE STRICTLY LARGER THAN" comment is now confusing (and was already a little vague by not explicitly mentioning `catchupStepSize`) as 200 < 1000. Of course 200 is now 2000 or 4000 so it's still true but please update the comment.

src/Chainweb/CutDB/Sync.hs
```diff
@@ -103,11 +103,18 @@ syncSession v useOrigin p db logg env pinf = do
         logg @T.Text Debug $ "put cut " <> encodeToText c
 
     receive = do
+        cur <- _cut db
+
         -- Query cut that is at most 1000 blocks ahead
```
slpopejoy, line 108:
```suggestion
        -- Query cut that is at most 'catchupStepSize' blocks ahead
```

src/Chainweb/CutDB/Sync.hs
larskuhtz: hmm, not sure how often it comes up in the code. The use here won't count, since it should be replaced by a block height based limit. C.f. the comment on the next line.

PR #1110, opened 2020-07-08
Set transition dates

slpopejoy:


src/Chainweb/Version.hs
```diff
@@ -782,7 +794,8 @@ twentyChainUpgrade
     -> ChainId
     -> BlockHeight
     -> Bool
-twentyChainUpgrade Mainnet01 _ h = h == maxBound -- TODO: update me with a calculated height
+twentyChainUpgrade Mainnet01 _ h = h == to20ChainsMainnet
+twentyChainUpgrade Testnet04 _ h = h == to20ChainsTestnet
```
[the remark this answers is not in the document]
larskuhtz: I propose that we use a block height some block height on 2020-07-23, which is them day when the 'oldDaGuard' triggers.

src/Chainweb/Version.hs
```diff
@@ -897,7 +910,7 @@ skipFeatureFlagValidationGuard Mainnet01 h = h < 530500  -- ~ 2020-05-01T00:00:x
 skipFeatureFlagValidationGuard _ _ = False
 
 oldDaGuard :: ChainwebVersion -> BlockHeight -> Bool
-oldDaGuard Mainnet01 _ = True
-oldDaGuard Testnet04 _ = True
+oldDaGuard Mainnet01 h = h < 771_414 -- ~ 2020-07-23 16:00:00
+oldDaGuard Testnet04 h = h < 318_204 -- ~ 2020-07-23 16:00:00
```
larskuhtz, line 914: 👍 double checked and got:

```
2020-07-23 16:02:52 UTC
2020-07-23 15:57:53 UTC
```

src/Chainweb/Version.hs
```diff
@@ -511,11 +511,23 @@ chainwebGraphs (TimedConsensus g1 g2) = (8, g2) NE.:| [ (0, g1) ]
...
+chainwebGraphs Mainnet01 =
+    ( to20ChainsMainnet, twentyChainGraph ) NE.:|
+    [ ( 0, petersonChainGraph ) ]
+chainwebGraphs Development =
+    ( 50, twentyChainGraph ) NE.:|
+    [ ( 0, petersonChainGraph ) ]
 {-# INLINE chainwebGraphs #-}
 
+to20ChainsMainnet :: BlockHeight
+to20ChainsMainnet = 852_054 -- 2020-08-20 16:00:00
```
larskuhtz, line 526: 👍 double checked and got `2020-08-20 16:02:36 UTC`

src/Chainweb/Version.hs
```diff
@@ -511,11 +511,23 @@ chainwebGraphs (TimedConsensus g1 g2) = (8, g2) NE.:| [ (0, g1) ]
...
+
+to20ChainsTestnet :: BlockHeight
+to20ChainsTestnet = 332_604 -- 2020-07-28 16:00:00
```
larskuhtz, line 529: 👍 double checked and got `2020-07-28 15:58:06`

src/Chainweb/Version.hs
larskuhtz, line 798:
```suggestion
twentyChainUpgrade Mainnet01 _ h = h == 771_474 -- ~ 2020-07-23 16:30:00
twentyChainUpgrade Testnet04 _ h = h == 318_264 -- ~ 2020-07-23 16:30:00
```

node/ChainwebNode.hs
```diff
@@ -468,10 +468,10 @@ pkgInfoScopes =
 -- -------------------------------------------------------------------------- --
 -- main
 
--- KILLSWITCH for version 1.9
+-- KILLSWITCH for version 2.0
 --
 killSwitchDate :: Maybe String
-killSwitchDate = Just "2020-07-16T00:00:00Z"
+killSwitchDate = Just "2020-08-13T00:00:00Z"
```
larskuhtz, line 474: 👍

dep/kpkgs/github.json
```diff
@@ -3,6 +3,6 @@
   "repo": "kpkgs",
   "branch": "master",
   "private": false,
-  "rev": "96a60f1d74bd6e18b49eabb89afaddbe8905d247",
+  "rev": "8c15564d689258ff87539bc7609e1fcfee51a2bf",
```
mightybyte, line 6: Verified that this is a commit on `kpkgs` master.  👍

node/ChainwebNode.hs
mightybyte: 👍

src/Chainweb/Version.hs
mightybyte: I was originally thinking doing this a week or so earlier.  That would give a little more time between the 20 chain forks.  But I don't have a strong opinion.

src/Chainweb/Version.hs
mightybyte: 👍 on 2020-08-20.

src/Chainweb/Version.hs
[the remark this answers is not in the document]
mightybyte: Also, if we're going to change it, I would suggest changing the name to something more accurate.

src/Chainweb/Version.hs
mightybyte: Ahhh, I guess this is why the testnet fork was a week later.  👍

src/Chainweb/Version.hs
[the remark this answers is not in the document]
emilypi: I agree with that. There is only a single chain 0 remediation that happens as a result of this fork, so it can certainly happen just prior.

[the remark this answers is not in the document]
slpopejoy: We're not doing the remediations until we fork as the point is to keep coins in circulation stable.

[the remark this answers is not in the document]
slpopejoy: Perhaps `to20ChainsRebalances` to match the naming convention here and point more to the economic intent.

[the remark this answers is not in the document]
slpopejoy: Renamed to `to20ChainsRebalance`

slpopejoy: See previous comment about importance of pairing with transition

PR #1111, opened 2020-07-08
Add test for 20-chain remediation pass-through

emilypi:
This test proves that if e7f7 is underfunded, then the service will not fail at remediation time (tested for Devnet)

### DIFF
--- a/test/Chainweb/Test/Pact/TransactionTests.hs
+++ b/test/Chainweb/Test/Pact/TransactionTests.hs
@@ -20,7 +20,6 @@ import Test.Tasty
 import Test.Tasty.HUnit
 
 import Control.Concurrent (readMVar)
-import Control.Exception (SomeException)
 import Control.Lens hiding ((.=))
 import Control.Monad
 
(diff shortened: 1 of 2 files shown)


test/Chainweb/Test/Pact/TransactionTests.hs
```diff
@@ -343,13 +332,25 @@ testUpgradeScript
...
     -> V.ChainId
     -> BlockHeight
-    -> (Either SomeException (T2 (CommandResult [TxLog Value]) (Maybe ModuleCache)) -> IO ())
+    -> (T2 (CommandResult [TxLog Value]) (Maybe ModuleCache) -> IO ())
     -> IO ()
-testUpgradeScript script cid bh test = do
+testUpgradeScript fp cid bh f
+    = testUpgradeScript' fp v cid bh f
+
+testUpgradeScript'
```
slpopejoy, line 340: It looks like this is never directly called, do you want to leave this version in?

emilypi: Yeah, i could probably remove this. I intended it to allow us to call scripts with different versions, but we only ended up testing dev here.

PR #1114, opened 2020-07-10
define targets for new chains

larskuhtz:
* [x] disable POW for genesis blocks
* [x] define initial targets for new chains on mainnet
* [x] define initial target for new chains on testnet
* [x] define initial devnet targets and remove previous hack to inject lower difficulty on startup


src/Chainweb/BlockHeader.hs
```diff
@@ -285,35 +282,22 @@ powTarget
 powTarget p@(ParentHeader ph) as bct = case effectiveWindow ph of
     Nothing -> maxTarget
     Just w
-        -- A special case for starting a new devnet. Using maxtarget results in
-        -- a too high block production and consecutively orphans and network
-        -- congestion. The consequence are osciallations to take serval hundred
-        -- blocks before the system stabilizes. This code cools down initial
-        -- block production.
-        | ver == Development && _blockHeight ph == (genesisHeight ver cid + 1) -> HashTarget (maxBound `div` 100000)
```
larskuhtz, line 293: This hack is not needed any more, since now we define the initial target in the genesis blocks themself.

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -91,8 +91,88 @@ genesisParentBlockHash v p = BlockHash $ MerkleLogHash
...
+-- osciallations to take serval hundred blocks before the system stabilizes.
+-- This setting cools down initial block production.
+--
+genesisBlockTarget :: ChainwebVersion -> ChainId -> HashTarget
+genesisBlockTarget v@Mainnet01 cid
+    | genesisHeight v cid > 731382 = mainnet20InitialHashTarget
+genesisBlockTarget v@Testnet04 cid
+    | genesisHeight v cid > 278626 = testnet20InitialHashTarget
+genesisBlockTarget Development _ = HashTarget (maxBound `div` 100000)
+genesisBlockTarget _ _ = maxTarget
```
larskuhtz, line 109: We may consider moving this and the following definitions to `Chainweb.Version` (or a submodule of Chainweb.Version`).

test/golden/mainnet01-block-hashes-expected.txt
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+SNAMXHBvZtAbeS19x12bZV2p5d1E4BaxrggceNdUxbw=aVFFqF8PbGu-MZkUhb1Ok4XijgxWUYxBxigBQgZmDFI=Isgjp62lHk0uXgQfu_HZoeuP3udIiwkBlq1hHvXqXCM=tMyy7k6lrQOsGg2oXh20kOC3zudjwsaL9F5Ogx0XNBo=yeWKDE27xM45-LgdwpnVMM6E4pAoYryRrJJFgRiaKgQ=GmCgnnxUtn08N8WtL8Gpn3Hg6AKyP4q0W0u3NLZn0Zo=OXehM2W5YuiV0iWlDGIPMdaRv721zMUbj3caJqZJNBU=CAfkqtRJpdf3rNSKx0SRwgaJlSUBLXwF1OMjn2-eBps=09qk9ulwtq854ogJwlDZKN7NC_Q3lqiy1jdD0_YvWMI=5cwptPDy7u6aMS221qsCr93g6ifDr0WUyvoKhi59-PA=E28feCcSOofOZxVZQLZl9DQbvJxFihdBGGmvZb-9rtg=GkJQrN61WF-Mx_m-jR_jy92IVv1JvSTZM4mqFS5LtQU=Vw8OnBUxIdxuMPd_1s7YHkTutGzc8tuktZ2r9rUTf3E=xSYNJasBhc4xekxa1hs8LkufqldXA4rQLIo90kv4rRk=SRi5nq4kSaFkYv28ROIhB5k1IEYnVcde94LyJ8QfLU8=0HBBPxyNbw41Ozo-enbOorx1FSaPd-qWj79gUtB3Sjg=_J6S2skxFbbGsNaPoHPKFQ2a9qOtDVaRkj2FwKESscE=EnTN8zCxWwTjeOOiyQGWFYR42dSZ6ga1lCF7jQgjri8=E8wsOQYcL9SYP6Aku_lRXETSTrc-KCHmAxHq_JiStOY=kl0X0aATbhasN-a0oV9yUN3qZjkbCUMtbqHaXvR6Hqw=
```
larskuhtz, line 1: I double checked that only the hashes for the new chains changed

src/Chainweb/BlockHeader/Genesis.hs
slpopejoy: Add a TODO for this as we really should keep magic block numbers there

slpopejoy: Also, should this just be `cid > 9` instead of testing for a particular block height threshold? It's the same but would make it ELI5 that this doesn't touch existing blocks.

PR #1120, opened 2020-07-11
fix initial targets for new chains

larskuhtz:
* [x] fix computations of the genesis targets for new chains
* [x] add test the verifies the targets based on the binary representation of the headers onto which the targets are based.


src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -139,43 +148,51 @@ genesisBlockTarget _ _ = maxTarget
 --
 -- It holds that:
 --
--- prop> mainnet20InitialHashTarget == HashTarget . (4 *) <$> (runGet decodePowHashNat "DOordl9cgfs4ZTBdFnbjRW5th-hW-pL33DIAAAAAAAA")
+-- prop> Just mainnet20InitialHashTarget == HashTarget . (4 *) <$> (runGet decodePowHashNat =<< decodeB64UrlNoPaddingText "DOordl9cgfs4ZTBdFnbjRW5th-hW-pL33DIAAAAAAAA")
 --
 mainnet20InitialHashTarget :: HashTarget
-mainnet20InitialHashTarget = HashTarget 92812039490652836170182663013446053204096613861175006070293989356886611016976
+mainnet20InitialHashTarget = HashTarget 326935740379188069500505933882552768991030445202673027909658672
```
mightybyte, line 154: I might suggest defining these constants in hex and padded with leading zeroes to 256 bits to make them easier to inspect by hand.

larskuhtz: Makes sense. I may add a test case for the hex representation, just to make sure that I don't mess up the `printf` format string :-). Or is there a way to tell ghci to print numbers in hex by default?

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -139,43 +148,51 @@ genesisBlockTarget _ _ = maxTarget
 --
 -- It holds that:
 --
--- prop> mainnet20InitialHashTarget == HashTarget . (4 *) <$> (runGet decodePowHashNat "DOordl9cgfs4ZTBdFnbjRW5th-hW-pL33DIAAAAAAAA")
+-- prop> Just mainnet20InitialHashTarget == HashTarget . (4 *) <$> (runGet decodePowHashNat =<< decodeB64UrlNoPaddingText "DOordl9cgfs4ZTBdFnbjRW5th-hW-pL33DIAAAAAAAA")
 --
 mainnet20InitialHashTarget :: HashTarget
-mainnet20InitialHashTarget = HashTarget 92812039490652836170182663013446053204096613861175006070293989356886611016976
+mainnet20InitialHashTarget = HashTarget 0x000000000000cb73de4be95ba21db5b9178dd85974c194e3ee05717dd8afa830
```
mightybyte, line 154: I verified that this correctly converts to what we had in the previous commit.

src/Chainweb/BlockHeader/Genesis.hs
```diff
@@ -139,43 +148,51 @@ genesisBlockTarget _ _ = maxTarget
...
 -- It holds that:
 --
--- prop> testnet20InitialHashTarget == HashTarget . (4 *) <$> (runGet decodePowHashNat "UWGTKb1Jt2oMPuayWSTeOXvcpN4bk2AsyNaMDeMEAAA")
+-- prop> Just testnet20InitialHashTarget == HashTarget <$> (runGet decodePowHashNat =<< decodeB64UrlNoPaddingText "NZIklpW6xujSPrX3gyhXInfxxOS6JDjkW_GbGwAAAAA")
+-- prop> _hashTarget testnet20InitialHashTarget `div` _hashTarget mainnet20InitialHashTarget == PowHashNat 8893
+-- prop> _hashTarget (genesisBlockTarget Development (unsafeChainId 10)) `div` _hashTarget testnet20InitialHashTarget == PowHashNat 38
 --
 testnet20InitialHashTarget :: HashTarget
-testnet20InitialHashTarget = HashTarget 92732593277323743564702843783986955023191593410093893802967421484438852033876
+testnet20InitialHashTarget = HashTarget 0x000000001b9bf15be43824bae4c4f17722572883f7b53ed2e8c6ba9596249235
```
mightybyte, line 195: Also verified this one

src/Chainweb/BlockHeader/Genesis.hs
mightybyte: Also verified that mainnet currently has a target with 12 leading zeroes, so this is the right ball park.

PR #1122, opened 2020-07-14
apply module name fix2 for testnet

larskuhtz:


src/Chainweb/Version.hs
```diff
@@ -836,6 +836,7 @@ enableModuleNameFix _ bh = bh >= 2
 --
 enableModuleNameFix2 :: ChainwebVersion -> BlockHeight -> Bool
 enableModuleNameFix2 Mainnet01 bh = bh >= 752214 -- ~ 2020-07-17 0:00:00 UTC
+enableModuleNameFix2 Testnet04 bh = bh >= 289966 -- ~ 2020-07-13
```
larskuhtz, line 839: This is roughly the block height at which chainweb-node-2.0 was deployed to us1.testnet.chainweb.com, which was the only mining node on testnet at that time.

larskuhtz:
I tested by running an offline pact history validation on the testnet database.

src/Chainweb/Version.hs
slpopejoy: Great catch

PR #1123, opened 2020-07-14
rewind pact service initially to lastest available header in history

larskuhtz:
If the latest block that is stored in the checkpointer gets orphaned after the service is stopped, pact service fails on startup. This PR fixes this by rewinding on startup to the latest block in the checkpointer block history that is available header in the block header db.

Some fixes and cleanups around the use of '_psParentHeader' and the use of the checkpointer are included in this PR, which was helpful for debugging and resolving the issue.

These are the changes in detail:

* [x] Before the checkpointer target was specified in pact service as `(parentHeight + 1, parentHash)`, which was more complicated than needed and could cause confusion. This PR consistently uses the 'ParentHeader' everywhere.

* [x] Add a check to `playOnBlock` that guarantees that the function is called only within the scope of a call to 'withCheckpointer*'.

* [x] Fix `rewindTo` to enforce the post condition that the checkpointer got restored to the requested target header, even if no rewind was needed for that.

* [x] Handle updating of `_psParentHeader` transparently along with the checkpointer state.

* [x] Guarantee that the block header for latest block of the checkpointer is always available.

* [x] Add a lots of new documentation that explains pre- and post-conditions for various checkpointer related functions in pact service.

* [x] Add a regression test for pact service initialization after the latest block of the checkpointer got orphaned.

* [x]  A very small unrelated change that provides version of `pruneForks` that doesn't require a logger. It ended up not being used at this point, but maybe become useful again in the future. So, I think, it's worth keeping it.

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -269,23 +269,23 @@ initPactService' ver cid chainwebLogger bhDb pdb sqlenv config act = do
                 }
         !pst = PactServiceState Nothing mempty initialParentHeader P.noSPVSupport
     runPactServiceM pst pse $ do
+
         -- If the latest header that is stored in the checkpointer was on an
         -- orphaned fork, there is no way to recover it in the call of
         -- 'initalPayloadState.readContracts'. We therefore rewind to the latest
         -- avaliable header in the block header database.
         --
-        initializeParentHeader
+        initializeLatestBlock
         act
   where
     initialBlockState = initBlockState $ genesisHeight ver cid
     loggers = pactLoggers chainwebLogger
     logger = P.newLogger loggers $ P.LogName ("PactService" <> show cid)
 
-initializeParentHeader :: PayloadCasLookup cas => PactServiceM cas ()
-initializeParentHeader = do
-    findLatestValidBlock >>= \case
-        Nothing -> return ()
-        Just b -> rewindTo initialRewindLimit (Just $ ParentHeader b)
+initializeLatestBlock :: PayloadCasLookup cas => PactServiceM cas ()
+initializeLatestBlock = findLatestValidBlock >>= \case
+    Nothing -> return ()
+    Just b -> withBatch $ rewindTo initialRewindLimit (Just $ ParentHeader b)
   where
     initialRewindLimit = Just 1000
 
@@ -602,6 +602,10 @@ validateHashes bHeader pData miner transactions =
           prevOutputsHash newOutputsHash
         ]
 
+-- | When using the checkpointer for evaluating pact code the caller must
+-- indicate whether the result of the evaluation should be persisted or
+-- discarded.
+--
 data WithCheckpointerResult a
     = Discard !a
     | Save BlockHeader !a
@@ -654,13 +658,22 @@ syncParentHeader caller = do
 -- in the current history. It doesn't resolve forks or fast forwards the
 -- checkpointer.
 --
+-- /Assumption:/
+--
+-- This function assumes that '_psParentHeader' has been updated to match the
+-- latest block in the checkpointers. This is guaranteed to be the case after
+-- calling any of 'rewindTo', 'syncParentHeader', 'withCheckPointerRewind',
+-- 'withCheckPointerWithoutRewind', or 'withCurrentCheckpointer'.
+--
 -- /NOTE:/
 --
 -- In most use cases one needs to rewind the checkpointer first to the target
 -- and function 'withCheckpointer' should be preferred.
 --
--- Only use this function when 1. you need the extra performance from skipping
--- the call to 'rewindTo' and 2. you know exactly what you do.
+-- Only use this function when
+--
+-- 1. you need the extra performance from skipping the call to 'rewindTo' and
+-- 2. you know exactly what you do.
 --
 withCheckpointerWithoutRewind
     :: PayloadCasLookup cas
@@ -682,15 +695,19 @@ withCheckpointerWithoutRewind target caller act = do
     unlessM ((<= 1) <$> asks _psCheckpointerDepth) $ do
         error $ "Code invariant violation: to many nested calls of withCheckpointerRewind. Please report this as a bug."
 
+    case target of
+        Just h -> setParentHeader (caller <> ".withCheckpointerWithoutRewind") h
+        Nothing -> return ()
+
     local (over psCheckpointerDepth succ) $ mask $ \restore -> do
         cenv <- restore $ do
-            r <- liftIO $ _cpRestore checkPointer checkpointerTarget
+            r <- liftIO $! _cpRestore checkPointer checkpointerTarget
             return r
 
         try (restore (act cenv)) >>= \case
-            Left e -> discardTx >> throwM @_ @SomeException e
-            Right (Discard !result) -> discardTx >> return result
-            Right (Save header !result) -> saveTx header >> return result
+            Left !e -> discardTx checkPointer >> throwM @_ @SomeException e
+            Right (Discard !result) -> discardTx checkPointer >> return result
+            Right (Save header !result) -> saveTx checkPointer header >> return result
   where
     checkpointerTarget = case target of
         Nothing -> Nothing
@@ -698,13 +715,14 @@ withCheckpointerWithoutRewind target caller act = do
             -- the second argument of _cpRestore expects the hash of the parent
             -- and the height of the parent plus one.
 
-    discardTx = finalizeCheckpointer _cpDiscard
-    saveTx !header = do
+    discardTx checkPointer = liftIO $! _cpDiscard checkPointer
+
+    saveTx checkPointer !header = do
         -- TODO: _cpSave is a complex call. If any thing in there throws
         -- an exception it would result in a pending tx.
-        finalizeCheckpointer (flip _cpSave $ _blockHash header)
+        liftIO $! _cpSave checkPointer $ _blockHash header
         modify' $ set psStateValidated (Just header)
-        setParentHeader (caller <> ".withCheckpointerWithoutRewind") (ParentHeader header)
+        setParentHeader (caller <> ".withCheckpointerWithoutRewind.saveTx") (ParentHeader header)
 
 -- | 'withCheckpointer' but using the cached parent header for target.
 --
@@ -715,6 +733,8 @@ withCurrentCheckpointer
     -> PactServiceM cas a
 withCurrentCheckpointer caller act = do
     ph <- syncParentHeader "withCurrentCheckpointer"
+        -- discover the header for the latest block that is stored in the
+        -- checkpointer.
     withCheckpointerRewind (Just 0) (Just ph) caller act
 
 -- | Execute an action in the context of an @Block@ that is provided by the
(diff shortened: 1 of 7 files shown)


larskuhtz:
A relatively reliable way to reproduce the issue and confirm that the fix is effective: Restart nodes while running devnet on 20 chains under high tx load (~100-120 tps).

src/Chainweb/Pact/PactService.hs
```diff
@@ -273,12 +268,27 @@ initPactService' ver cid chainwebLogger bhDb pdb sqlenv config act = do
...
     initialBlockState = initBlockState $ genesisHeight ver cid
     loggers = pactLoggers chainwebLogger
     logger = P.newLogger loggers $ P.LogName ("PactService" <> show cid)
 
+initializeLatestBlock :: PayloadCasLookup cas => PactServiceM cas ()
+initializeLatestBlock = findLatestValidBlock >>= \case
+    Nothing -> return ()
+    Just b -> withBatch $ rewindTo initialRewindLimit (Just $ ParentHeader b)
+  where
+    initialRewindLimit = Just 1000
```
slpopejoy, line 290: What happens when this is exceeded at startup?

src/Chainweb/Pact/PactService.hs
```diff
@@ -649,22 +658,27 @@ syncParentHeader caller = do
...
 --
 -- In most use cases one needs to rewind the checkpointer first to the target
 -- and function 'withCheckpointer' should be preferred.
 --
--- Only use this function when 1. you need the extra performance from skipping
--- the call to 'rewindTo' and 2. you know exactly what you do.
+-- Only use this function when
+--
+-- 1. you need the extra performance from skipping the call to 'rewindTo' and
+-- 2. you know exactly what you do.
```
slpopejoy, line 676:
```suggestion
-- 2. you know exactly what you are doing.
```

test/Chainweb/Test/Pact/PactReplay.hs
```diff
@@ -143,6 +150,54 @@ dupegenMemPoolAccess = mempty
         return outtxs
     }
 
+-- | This is a regression test for correct initialization of the checkpointer
+-- during pact service initialization.
+--
+-- Removing the call to 'initializeLatestBlock' in 'initPactService' causes
+-- this test to fail.
+--
```
slpopejoy, line 158: 👍

src/Chainweb/Pact/PactService.hs
```diff
@@ -1607,3 +1659,57 @@ freeModuleLoadGasModel = modifiedGasModel
 --
 officialGasModel :: P.GasModel
 officialGasModel = freeModuleLoadGasModel
+
+-- -------------------------------------------------------------------------- --
```
slpopejoy, line 1663: I might argue that we make a new module instead of worrying about formatting this over-large module, a Checkpointer sub-module would be great

src/Chainweb/Pact/PactService.hs
```diff
@@ -1564,11 +1606,21 @@ execLookupPactTxs restorePoint txs
...
-        withCheckpointerRewind Nothing t "lookupPactTxs" $ \_ ->
+      DoRewind parent -> withDiscardedBatch $ do
+        withCheckpointerRewind Nothing (Just $ ParentHeader parent) "lookupPactTxs" $ \_ ->
           liftIO $ Discard <$> V.mapM (_cpLookupProcessedTx cp) txs
 
+-- | Find the latest block stored in the checkpointer for which the respective
+-- block header is available in the block header database.
+--
+-- NOTE: the function doesn't take into consideration the block header stored in
+-- '_psParentHeader' in 'PactServiceM'. It is meant to be used during re-orgs
```
slpopejoy, line 1617: > It is meant to be used during re-orgs

Is this true? It is used in initialization but also in `rewindTo`, so that doesn't sound like re-org only.

src/Chainweb/Pact/PactService.hs
```diff
@@ -1344,23 +1385,24 @@ execValidateBlock
     -> PayloadData
     -> PactServiceM cas PayloadWithOutputs
 execValidateBlock currHeader plData = do
+    -- The parent block header must be available in the block header data base
```
slpopejoy, line 1388:
```suggestion
    -- The parent block header must be available in the block header database
```

src/Chainweb/Pact/PactService.hs
```diff
@@ -1344,23 +1385,24 @@ execValidateBlock
...
-    isGenesisBlock = isGenesisBlockHeader currHeader
-
-    -- On final success, use header as the new/next parent header
-    setCurrAsParent pwos =
-      setParentHeader "setCurrAsParent" (ParentHeader currHeader) >> return pwos
+    getTarget
+        | isGenesisBlockHeader currHeader = return Nothing
+        | otherwise = Just . ParentHeader
+            <$> lookupBlockHeader (_blockParent currHeader) "execValidateBlock"
+                -- It is up to the user of pact serivce to guaranteed that this
```
slpopejoy, line 1403:
```suggestion
                -- It is up to the user of pact service to guaranteed that this
```

src/Chainweb/Pact/PactService.hs
```diff
@@ -1269,66 +1301,75 @@ rewindTo
...
+rewindTo _ Nothing = return ()
+rewindTo rewindLimit (Just (ParentHeader parent)) = do
+
+    -- skip if the checkpointer is already at the target.
+    (_, lastHash) <- getCheckpointer >>= liftIO . _cpGetLatestBlock >>= \case
+        Nothing -> throwM NoBlockValidatedYet
+        Just p -> return p
+
+    if lastHash == parentHash
+      then setParentHeader "rewindTo" (ParentHeader parent)
```
slpopejoy, line 1316: Might be good to note where `setParentHeader` gets called in the `else` case (presumably in `playFork -> withCheckpointerWithoutRewind` right?)

src/Chainweb/Pact/PactService.hs
```diff
@@ -1269,66 +1301,75 @@ rewindTo
...
-        V.mapM_ (fastForward payloadDb) newBlocks
+    playFork lastHeader = do
+        bhdb <- asks _psBlockHeaderDb
+        (!commonAncestor, _, newBlocks) <-
+            liftIO $ collectForkBlocks bhdb lastHeader parent
+
+        if V.null newBlocks
+          then
+            -- If newBlocks is empty the checkpointer isn't restored via
+            -- 'fastForward'. So we do an empty 'withCheckPointerWithoutRewind'.
```
slpopejoy, line 1352: Was this needed to fix something?

src/Chainweb/Pact/PactService.hs
```diff
@@ -681,23 +695,34 @@ withCheckpointerWithoutRewind target caller act = do
     unlessM ((<= 1) <$> asks _psCheckpointerDepth) $ do
         error $ "Code invariant violation: to many nested calls of withCheckpointerRewind. Please report this as a bug."
 
+    case target of
+        Just h -> setParentHeader (caller <> ".withCheckpointerWithoutRewind") h
```
slpopejoy, line 699: We seem to have outgrown `psParentHeader` at this point, especially since `withCurrentCheckpointer` hits the db first. I think we're ready to move it into the environment and out of state, as there is never a case where we invoke PactService and expect it to be set properly already. 

I propose it should be a `Maybe ParentHeader` in `PactServiceEnv`, and be `Nothing` outside of any `local` call (ie in the `serviceRequests` loop), so that code that needs to _read_ it (`execTransactions` etc) will simply fail if it's not set. Then we'll finally get out from under the state problem.

What do you think of including this change in this PR?

src/Chainweb/Pact/PactService.hs
larskuhtz: Good point. I added `exitOnRewindLimitExceeded` that makes it terminate the process and exit with a meaningful error message. It's the same what already happens when `execValidateBlock` hits the rewind limit.

src/Chainweb/Pact/PactService.hs
larskuhtz: Yes, I think, that's the right approach. However, currently, there's one situation when `_psParentHeader` is used from the state: when `withCurrentCheckpointer` is used and the the last saved block isn't yet available in the block header db. 

This can happen, for instance, when an `execLocal` request immediately follows an `execValidateBlock` request. In those situations the '_psParentHeader' is used and it would be inefficient to restore the checkpointer to the previous block, because any subsequent new block or validation request is most likely to happen on top of the current header, which would have to be replayed if we would restore the checkpointer.

One can argue, that it would be the job of the checkpointer to remember the last block header internally. I can see good arguments both way. There is also `_psStateValidated :: Maybe BlockHeader`, which seems unused. Maybe this was used originally for this purpose? I was about it remove it, but maybe I should leave it and instead use it to initialize '_psParentHeader' in 'withCurrentCheckpointer'?

src/Chainweb/Pact/PactService.hs
larskuhtz: hmm, maybe it shouldn't be set here. `withCheckpointerWithoutRewind` sets it only if the result is saved. One could argue that here it doesn't matter, because in the case of `lastHash == parentHash` it doesn't make a difference whether the result is saved or discarded. But, under that assumption, setting '_psParentHash' here should not be needed in first place.

larskuhtz: The current approach is a bit paranoid. It makes sure that it is set before validation starts and also after it ends. In theory the latter should be sufficient (assuming proper initialization). The former is added as a second line of defense in case of failures slipping through existing finalizers.

src/Chainweb/Pact/PactService.hs
larskuhtz: Yes, this, along with 'initializeLatestBlock' is actually the main fix in this PR. For instance, 'initializeLatestBlock' alone doesn't fix the issue that restarts with an orphaned latest block fail, because `rewindTo` would be a noop, and the latest block would still point to the orphaned header. `_cpRestore' is only called in `withCheckpointerWithoutRewind`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -1564,11 +1590,21 @@ execLookupPactTxs restorePoint txs
...
 
+-- | Find the latest block stored in the checkpointer for which the respective
+-- block header is available in the block header database.
+--
+-- NOTE: the function doesn't take into consideration the block header stored in
+-- '_psParentHeader' in 'PactServiceM'. It is meant to be used during re-orgs
+-- (recovering from forks).
+--
+-- First the result of '_cpGetLatestBlock' is checked. If the respective block
+-- header isn't available, the fucntion recursively checks the result of
```
larskuhtz, line 1610:
```suggestion
-- header isn't available, the function recursively checks the result of
```

src/Chainweb/Pact/PactService.hs
larskuhtz: yes, it is true -- but maybe the comment isn't clear. In both places it is called to resolve possible forks. It should only be called when it is acceptable (as during initialization) or desirable (during rewindTo) to discard the latest validated block. Calling this function when '_cpLatestBlock' is not a fork/orphan is inefficient. In particular when there are many local calls, calling this function would result in many redundant re-evaluations.

src/Chainweb/Pact/PactService.hs
larskuhtz: Agreed. Moving this into something like `PactService.Checkpointer` would make sense.

I had the code already moved into this sections to avoid intense scrolling during editing them. I only moved the code back in place to get nice diffs and make the review easier.

I propose that we do a follow up PR, that only moves the code but doesn't change any behavior.

src/Chainweb/Pact/PactService.hs
larskuhtz: I propose the following:

1. Implement your proposal and move '_psParentHeader' into the environment. That would give us the benefits that you describe.

2. solve the issue of the pending recently validated block header by using the existing '_psStateValidated' component in pact service state. It seems that it gets already initialized with the correct block header. It would be used for the single purpose of providing the block header of the latest block for initializing `_psParentHeader` in `withCurrentCheckpointer`.

Given that this PR is already pretty big and also has already been tested extensively, I would propose to implement that in a follow up PR.

slpopejoy: I'm a little apprehensive about resurrecting `_psStateValidated` as it's currently obsolete, so we don't know how correct its maintenance is (which is the problem we're going to get rid of with `_psParentHeader`). 

Also it seems like `withCurrentCheckpointer` could simply grab the latest block in checkpointer (via `_cpGetLatestBlock`) and restore with that, it's a miniscule db (and probably in sqlite RAM anyway) hit given that `local` and pre-insert have to hit the db anyway.

I would much prefer less state + more correctness, which would mean deleting `_psStateValidated` and proceeding with #1 above.

PR #1127, opened 2020-07-29
install coin v2 upgrade on new testnet chains

larskuhtz:


src/Chainweb/Version.hs
```diff
@@ -778,6 +778,9 @@ coinV2Upgrade Mainnet01 cid h
     | cid == unsafeChainId 8 = h == 140808
     | cid == unsafeChainId 9 = h == 140808
     | otherwise = error $ "invalid chain id " <> sshow cid
+coinV2Upgrade Testnet04 cid h
+    | chainIdInt @Int cid >= 10 = h == 337000
```
slpopejoy, line 782: Should we bound this with `< 20` as well? Future-proof as we will hopefully get the genesis blocks right next time.

larskuhtz:
```suggestion
    | chainIdInt @Int cid >= 10  && chainIdInt @Int cid < 20 = h == 337000
```

PR #1132, opened 2020-08-03
More efficient and correct database pruning

larskuhtz:
* [x] new more efficient block header pruning algorithm
* [ ] Mark and sweep garbage collection for payloads

### DIFF
--- a/src/Chainweb/BlockHeaderDB/PruneForks.hs
+++ b/src/Chainweb/BlockHeaderDB/PruneForks.hs
@@ -1,7 +1,14 @@
-{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 
 -- |
 -- Module: Chainweb.BlockHeaderDB.PruneForks
@@ -13,20 +20,31 @@
 -- Prune old forks from BlockHeader DB
 --
 module Chainweb.BlockHeaderDB.PruneForks
-( pruneForks
+( pruneForksLogg
+, pruneForks
 , pruneForks_
+
+-- * Mark and sweep GC for payloads
+, mkFilter
+, markPayload
 ) where
 
-import Control.Monad
-import Control.Monad.ST
+import Chainweb.MerkleLogHash
 
-import qualified Data.BloomFilter.Easy as BF (suggestSizing)
-import qualified Data.BloomFilter.Hash as BF
-import qualified Data.BloomFilter.Mutable as BF
+import Control.Monad
+import Control.Monad.Catch
+import Control.Monad.IO.Class
+
+import Data.Aeson
+import qualified Data.ByteArray as BA
+import Data.Coerce
+import Data.Cuckoo
 import Data.Function
-import qualified Data.HashSet as HS
+import qualified Data.List as L
 import Data.Maybe
 import Data.Semigroup
+
+import Foreign.Ptr
 
 import Numeric.Natural
 
@@ -41,7 +59,11 @@ import System.LogLevel
 import Chainweb.BlockHash
 import Chainweb.BlockHeader
 import Chainweb.BlockHeaderDB.Internal
+import Chainweb.BlockHeight
+import Chainweb.ChainId
 import Chainweb.Logger
+import Chainweb.Payload
+import Chainweb.Payload.PayloadStore
 import Chainweb.TreeDB
 import Chainweb.Utils hiding (Codec)
 
@@ -50,7 +72,32 @@ import Data.CAS.RocksDB
 import Data.LogMessage
 
 -- -------------------------------------------------------------------------- --
--- Prune Old Forks
+-- Chain Database Pruning
+
+pruneForksLogg
+    :: Logger logger
+    => logger
+    -> BlockHeaderDb
+    -> Natural
+        -- ^ The depth at which deletion starts. Note, that the max rank isn't
+        -- necessarly included in the current best cut. So one, should choose a
+        -- depth for which one is confident that all forks are resolved.
+
+    -> (BlockHeader -> IO ())
+        -- ^ Deletion call back. This hook is called /after/ the entry is
+        -- deleted from the database. It's main purpose is to delete any
+        -- resources that were related to the deleted header and that are not
+        -- needed any more.
+
+    -> (BlockPayloadHash -> IO ())
+        -- ^ Deletion call back. This hook is called /after/ the entry is
+        -- deleted from the database. It's main purpose is to delete any
+        -- resources that were related to the deleted header and that are not
+        -- needed any more.
+    -> IO Int
+pruneForksLogg logger = pruneForks logg
+  where
+    logg = logFunctionText (setComponent "ChainDatabasePrunning" logger)
 
 -- | Prunes most block headers and block payloads from forks that are older than
 -- the given number of blocks.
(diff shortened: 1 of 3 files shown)


src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -349,7 +349,12 @@ rewindTo rewindLimit (Just (ParentHeader parent)) = do
         -- This does a restore, i.e. it rewinds the checkpointer back in
         -- history, if needed.
         withCheckpointerWithoutRewind (Just target) "fastForward" $ \pdbenv -> do
-            payload <- liftIO (payloadWithOutputsToPayloadData <$> casLookupM payloadDb bpHash)
+            payload <- liftIO $ casLookup payloadDb bpHash >>= \case
```
larskuhtz, line 352: With this PR `Chainweb.Test.Pact.RemotePactTest/remote pact tests/await network` to hang forever. I think, this is the only change that could explain that behavior.

Before this was throwing a low-level exception from the `Chainweb-storage` package. Now it is throwing an `InternalPactError`.

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -61,160 +99,162 @@ import Data.LogMessage
...
+        -- needed any more.
+
+    -> (Bool -> BlockPayloadHash -> IO ())
         -- ^ Deletion call back. This hook is called /after/ the entry is
         -- deleted from the database. It's main purpose is to delete any
         -- resources that were related to the deleted header and that are not
-        -- needed any more. The Boolean argument indicates whether the payload
-        -- of the block is shared with any block header that isn't marked for
-        -- deletion.
+        -- needed any more.
```
slpopejoy, line 124: "deleted header" in previous line should be "deleted payload" no?

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -61,160 +99,162 @@ import Data.LogMessage
...
-        -- entry of largest POW weight.
-        --
-        -- Usually this number would be defined in terms of the chainweb
-        -- diameter and/or the difficulty adjustment window.
-    -> (BlockHeader -> Bool -> IO ())
+        -- ^ The depth at which deletion starts. Note, that the max rank isn't
+        -- necessarly included in the current best cut. So one, should choose a
+        -- depth for which one is confident that all forks are resolved.
+
+    -> (Bool -> BlockHeader -> IO ())
```
slpopejoy, line 114: Please document `Bool` argument

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -61,160 +99,162 @@ import Data.LogMessage
...
+        -- ^ Deletion call back. This hook is called /after/ the entry is
+        -- deleted from the database. It's main purpose is to delete any
+        -- resources that were related to the deleted header and that are not
+        -- needed any more.
+
+    -> (Bool -> BlockPayloadHash -> IO ())
```
slpopejoy, line 120: Please document `Bool` argument

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -61,160 +99,162 @@ import Data.LogMessage
...
-        -- Usually this number would be defined in terms of the chainweb
-        -- diameter and/or the difficulty adjustment window.
-    -> (BlockHeader -> Bool -> IO ())
+        -- ^ The depth at which deletion starts. Note, that the max rank isn't
+        -- necessarly included in the current best cut. So one, should choose a
+        -- depth for which one is confident that all forks are resolved.
+
+    -> (Bool -> BlockHeader -> IO ())
+        -- ^ Deletion call back. This hook is called /after/ the entry is
+        -- deleted from the database. It's main purpose is to delete any
```
slpopejoy, line 116: It looks like the `Bool` argument is whether it is deleted from the database *or not*, and it is called on both "pivots" and deleted items, meaning this comment is inaccurate as well as the one for the following callback?

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -61,160 +99,162 @@ import Data.LogMessage
...
+            -- note that the use of `head` here is safe
+
+        forM_ toDelete $ \b -> do
+            deleteHdr b
+            hdrCallback True b
+            let payload = _blockPayloadHash b
+            payloadCallback (payload `notElem` pivotPayloads) payload
+
+        forM_ newPivots $ \b -> do
+            hdrCallback False b
```
slpopejoy, line 211: What is the purpose for calling back on non-deletion?

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -61,160 +99,162 @@ import Data.LogMessage
...
+-- TODO: provide this function in chainweb-storage:
+--
+-- Returns the stream of key-value pairs of an 'RocksDbTableIter' in reverse
+-- order.
+--
+-- The iterator must be released after the stream is consumed. Releasing the
+-- iterator to early while the stream is still in use results in a runtime
+-- error. Not releasing the iterator after the processing of the stream has
+-- finished results in a memory leak.
+--
```
slpopejoy, line 238: What is the use case for exposing this outside of `withReverseHeaderStream` when it seems like the latter is the preferred function for safe usage?

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -0,0 +1,394 @@
...
+            forM_ markedTrans $ \cf ->
+                tryInsert cf "transactions hash" (gcHash $ _blockPayloadTransactionsHash payload)
+  where
+    tryInsert cf k a = do
+        -- inserting a somewhat larger number (I think, it's actually 7) of
+        -- equal elements causes the filter to fail.
+        unlessM (member cf a) $
+            unlessM (insert cf a) $ error
+                $ "failed to insert item " <> k <> " in cuckoo filter"
+                <> ": while very rare this can happen. Usually it is resolve by retrying."
```
slpopejoy, line 332:
```suggestion
                <> ": while very rare this can happen. Usually it is resolved by retrying."
```

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -0,0 +1,394 @@
...
+
+    pdb = _transactionDbBlockPayloads $ _transactionDb db
+
+-- -------------------------------------------------------------------------- --
+-- Delete Payload
+--
+-- Payload Components are deleted in a way such that it is guaranteed that there
+-- are no dangling references.
+--
+-- /Tables:/
```
slpopejoy, line 342: Can you break up this list to comment `deleteBlockPayload` and `deleteBlockTransactions` separately?

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -0,0 +1,394 @@
...
+    -> ChainwebVersion
+    -> IO ()
+fullGc logger rdb v = do
+    logg Info $ "Starting chain database garbage collection"
+
+    -- 1. Concurrently prune chain dbs (TODO use a pool to limit concurrency)
+    (markedPayloads, markedTrans) <- unzip <$> forConcurrently (toList $ chainIds v) pruneChain
+    logg Info $ "Finished mark phase"
+
+    -- TODO: paralleize he the sweep phases to use all available cors.
```
slpopejoy, line 165:
```suggestion
    -- TODO: paralleize he the sweep phases to use all available cores.
```

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -0,0 +1,394 @@
...
+import Chainweb.Payload.PayloadStore.RocksDB
+import Chainweb.TreeDB
+import Chainweb.Utils
+import Chainweb.Version
+
+import Data.LogMessage
+
+-- | Prune all chains and cleanup BlockPayloads and BlockOutputs.
+--
+-- This doesn't cleanup BlockTransactions in the payload database.
```
slpopejoy, line 115: Why? Is this because you can't safely delete a payload without a `fullGc`?

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -0,0 +1,394 @@
...
+    loggHdrDelete l True h = l Info
+        $ "pruned header " <> toText (_blockHash h)
+        <> " at height " <> sshow (_blockHeight h)
+
+-- -------------------------------------------------------------------------- --
+-- Utils for Mark and sweep GC for Payloads
+--
+
+-- | Wraps any types wrapper around a MerkleLogHash for usage with a Cuckoo
+-- filter.
```
slpopejoy, line 263:
```suggestion
-- | Wraps a MerkleLogHash for usage with a Cuckoo filter.
```

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -0,0 +1,394 @@
...
+newtype GcHash a = GcHash a
+    deriving newtype (Show, ToJSON)
+
+instance BA.ByteArrayAccess a => CuckooFilterHash (GcHash a) where
+    cuckooHash (Salt s) (GcHash a) = fnv1a_bytes s $ BA.takeView a 8
+    cuckooFingerprint (Salt s) (GcHash a) = sip_bytes s $ BA.takeView a 8
+    {-# INLINE cuckooHash #-}
+    {-# INLINE cuckooFingerprint #-}
+
+gcHash :: a -> GcHash a
```
slpopejoy, line 274: Why not just an accessor?

src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -349,7 +349,12 @@ rewindTo rewindLimit (Just (ParentHeader parent)) = do
...
+                Nothing -> throwM $ PactInternalError
```
slpopejoy, line 353: This change is for a more localized error yes?

src/Chainweb/BlockHeaderDB/PruneForks.hs
slpopejoy: Nm, understood in later review

src/Chainweb/Pact/PactService/Checkpointer.hs
larskuhtz: Yes, it was helpful during debugging. Before it was throwing just an generic message coming from the Chainweb-storage package.

src/Chainweb/BlockHeaderDB/PruneForks.hs
larskuhtz: Copy and paste from the respective non-reversed functions in chainweb-storage :-)

larskuhtz:
We have to do something. Without this node won't start.

larskuhtz:
One option is to take it and to do any of

1. set default pruning to `none`,
2. completely disable `full`,
3. completely disable `full` and `forks` (so that `none` is the only option).

larskuhtz:
I added some code that does a structural consistency check of the database during fork pruning. That should add some additional peace of mind.

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -41,180 +55,199 @@ import System.LogLevel
...
+pruneForks logg cdb depth callback = do
+    hdr <- maxEntry cdb
+    if
+        | int (_blockHeight hdr) <= depth -> do
+            logg Info
+                $ "Skipping database prunning because the requested depth "
+                <> sshow depth <> " is not larger than the maximum block height "
+                <> sshow (_blockHeight hdr)
+            return 0
+        | int (_blockHeight hdr) <= int genHeight + depth -> do
```
slpopejoy, line 130: Can you document how this case gets tested (I couldn't find anything in the manual tests)?

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -41,180 +55,199 @@ import System.LogLevel
...
-    --
-    -- TODO: would it be better to use a single shared filter?
-    --
-    let (psize, phashNum) = BF.suggestSizing s 0.0001
-    !markedPayloads <- stToIO $ BF.new (BF.cheapHashes phashNum) psize
+  where
+    go :: ([BlockHash], BlockHeight, Int) -> BlockHeader -> IO ([BlockHash], BlockHeight, Int)
+    go ([], _, _) cur = throwM $ InternalInvariantViolation
+        $ "PrunForks.pruneForks_: no pivots left at block " <> encodeToText (ObjectEncoded cur)
+    go (!pivots, !prevHeight, !n) !cur
```
slpopejoy, line 191: What is the purpose of `n`? It is not read in `go`

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -41,180 +55,199 @@ import System.LogLevel
...
+                <> ". Current header: " <> encodeToText (ObjectEncoded cur)
+                <> ". Previous height: " <> sshow prevHeight
+        | _blockHash cur `elem` pivots = do
+            callback False cur
+            let !pivots' = force $ L.nub $ _blockParent cur : L.delete (_blockHash cur) pivots
+            return (pivots', curHeight, n)
+        | otherwise = do
+            deleteHdr cur
+            callback True cur
+            return (pivots, curHeight, n+1)
```
slpopejoy, line 206: What is the loop invariant here, since only `n` is changed, which is not read in `go`?

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -41,180 +55,199 @@ import System.LogLevel
...
+      where
+        curHeight = _blockHeight cur
+
+    deleteHdr k = do
+        -- TODO: make this atomic (create boilerplate to combine queries for
+        -- different tables)
```
slpopejoy, line 212: How would we do this, does our CAS solution allow for atomic multi-table deletes?

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -0,0 +1,438 @@
...
+-- we'll have to take a global approach at garabage collecting payloads.
+--
+-- There's also the advantage of guaranteeing that there are no concurrent
+-- writes during the garbage collections, because the chain databases aren't
+-- initialized yet.
+--
+-- 1. Concurrently for each chain:
+--     * initialize temporary block header db,
+--     * prune the block header db, and
+--     * use the callbacks to delete those parts of the payloads that are
```
slpopejoy, line 31: This needs an update re payload deletion no?

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -1,7 +1,16 @@
-{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
```
slpopejoy, line 6: I don't see a GADT declared, is this for usage of a GADT, or RankNTypes, or?

src/Chainweb/BlockHeaderDB/PruneForks.hs
larskuhtz: It counts the number of deleted blocks, which is reported in the logs.

src/Chainweb/BlockHeaderDB/PruneForks.hs
larskuhtz: `n` counts the number of deleted blocks. The pivots are unchanged.

The algorithm works as follows:

1. initialization: pivots are initialized with the parents hashs of all blocks that should be kept. All ancestors of these blocks are kept. All other blocks are deleted.

2. traverse all blocks in the db reverse ordered by block height. For each block check:

   1. If the block hash is in the pivot set it's an ancestor of a block that is kept and not deleted. The respective pivot is replaced with the the current block hash.

   2. If the block hash is not in the pivot set, it's not an ancestor of a block that is kept. The block is deleted and the count of deleted blocks `n` is increased by 1.

  There is an additional (not strictly needed) check, that ensures that actually a match for each pivot is found. This ensures that the database was consistent in first place.

src/Chainweb/BlockHeaderDB/PruneForks.hs
larskuhtz: No, the current CAS instance doesn't support this. RocksDb, however, does support it.

This `TODO` isn't related to this PR. The code existed since the very first version. It's not critical.

src/Chainweb/BlockHeaderDB/PruneForks.hs
larskuhtz: I think, it's not needed any more and can be removed. But not 100% sure.

src/Chainweb/BlockHeaderDB/PruneForks.hs
larskuhtz: This gets triggered for instance on the mainnet database for the new chains. I haven seen that showing up on the mainnet-simulation

src/Chainweb/BlockHeaderDB/PruneForks.hs
larskuhtz: `n` is returned by the fold in line 185: `(\(_,_,!n) -> evaluate n)`

src/Chainweb/Chainweb/PruneChainDatabase.hs
larskuhtz: yes, the comment is outdated.

larskuhtz: Actually, a lot of the module comment is outdated.

larskuhtz: There are two pruning modes:

1. `forks`. Only remove block headers that are not reachable from the most recent blocks of the chain. Doesn't touch payloads at all.

2. `full`. Does header plus mark and sweep GC for payloads. GC takes a conservative approach that guarantees that even if something goes wrong in between, there are no dangling references in the -- even for non-reachable entries.

   1. mark all used payload hashes during block header pruning
   2. traverse all `BlockPaylaods` and sweep all unmarked payload entries and mark all transactions hashes and outputs hashes that are used by marked `BlockPayload`s.
   3. concurrently sweep all unmarked `BlockTransactions` and `BlockOutputs`.

So, it is guaranteed that if if a BlockPayload is kept (is marked) also the respective transactions and outputs are marked and kept. The algorithm works top-down, so referenced entries are deleted only after the referencing entry is deleted. If an exception occurs during the GC, some garbage is remains but there are not dangling references, i.e. the database is consistent.

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -0,0 +1,438 @@
...
+  where
+    logg = logFunctionText logger
+    diam = diameter $ chainGraphAt v (maxBound @BlockHeight)
+    depth = diam * 3
+
+    db = newPayloadDb rdb
+
+
+    -- Prune a single chain and return the sets of marked payloads and
+    -- transactions.
```
larskuhtz, line 186:
```suggestion
    -- Prune a single chain and return the sets of marked payloads
```

src/Chainweb/BlockHeaderDB/PruneForks.hs
slpopejoy: Right, and I forgot that since it's a `foldM_` the loop invariant is guaranteed, sorry. Thanks for the explanation

PR #1136, opened 2020-08-14
Remove legacy tx timing validation code

larskuhtz:
This PR simplifies the tx timing validation code.

Early versions of chainweb used the creation time of the current block header for timing validation. Recent version use the parent creation time. Special logic was needed during the transition period, which now can be removed.

In addition to removing transition code, this PR also simplifies the handling of old blocks. Since, once a tx passes timing validation, it has no further effect on the pact validation of the tx, it is possible to just skip the timing validation for historic blocks.

### DIFF
--- a/src/Chainweb/Pact/PactService/ExecBlock.hs
+++ b/src/Chainweb/Pact/PactService/ExecBlock.hs
@@ -133,7 +133,7 @@ execBlock currHeader plData pdbenv = do
       then
         return (ParentCreationTime $ _blockCreationTime currHeader)
       else
-         (ParentCreationTime . _blockCreationTime . _parentHeader) <$> use psParentHeader
+         ParentCreationTime . _blockCreationTime . _parentHeader <$> use psParentHeader
 
     -- prop_tx_ttl_validate
     valids <- liftIO $ V.zip trans <$>
--- a/src/Chainweb/Pact/PactService/ExecBlock.hs
+++ b/src/Chainweb/Pact/PactService/ExecBlock.hs
@@ -71,7 +71,6 @@ import Pact.Types.RPC
 import qualified Pact.Types.Runtime as P
 import qualified Pact.Types.SPV as P
 
-import Chainweb.BlockCreationTime
 import Chainweb.BlockHeader
 import Chainweb.BlockHeight
 import Chainweb.Mempool.Mempool as Mempool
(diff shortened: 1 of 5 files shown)


src/Chainweb/Pact/PactService/ExecBlock.hs
```diff
@@ -130,18 +129,15 @@ execBlock currHeader plData pdbenv = do
...
-        useLegacyCreationTimeForTxValidation v h || isGenesisBlockHeader currHeader
+    txValidationTime <- if isGenesisBlockHeader currHeader
       then
-        return (_blockCreationTime currHeader, False)
-      else do
-        parent <- use psParentHeader
-        return (_blockCreationTime $ _parentHeader parent, True)
+        return (ParentCreationTime $ _blockCreationTime currHeader)
+      else
+         (ParentCreationTime . _blockCreationTime . _parentHeader) <$> use psParentHeader
```
emilypi, line 136: extraneous parens

src/Chainweb/Pact/Utils.hs
```diff
@@ -72,17 +73,17 @@ maxTTL = ParsedInteger $ 2 * 24 * 60 * 60
 -- This is probably going to be changed. Let us make it 2 days for now.
 
 -- prop_tx_ttl_newBlock/validateBlock
+--
+-- Timing checks used to be based on the creation time of the validated
+-- block. That changed on mainnet at block height 449940. Tx creation time
+-- and TTL don't affect the tx outputs and pact state and can thus be
```
emilypi, line 79: can thus be...?

src/Chainweb/Pact/PactService/ExecBlock.hs
larskuhtz, line 136:
```suggestion
         ParentCreationTime . _blockCreationTime . _parentHeader <$> use psParentHeader
```

src/Chainweb/Pact/Utils.hs
```diff
@@ -72,17 +73,17 @@ maxTTL = ParsedInteger $ 2 * 24 * 60 * 60
...
+-- when replaying old blocks.
```
larskuhtz, line 80:
```suggestion
-- skipped when replaying old blocks.
```

larskuhtz:
@emilypi I did an offline pact history replay on the the mainnet database.

PR #1138, opened 2020-08-21
Update mempool after execValidateBlock and keep track of committed txs

larskuhtz:
This PR addresses two issues with the current Mempool:

1.  The mempool currently only deletes transactions that are included in a new block from the list of pending transactions in calls of `execNewBlock`. That means that transactions are removed from the pending set only on nodes that are mining.

    The PR adds the respective calls of

    ```Haskell
    mpaProcessFork memPoolAccess p
    mpaSetLastHeader memPoolAccess p
    ```

    to `execValidateBlock`. `p` is the parent header of the currently validate header. This isn't optimal. Ideally, the transactions in the current header should be removed, too. However, the implementation of these functions of the mempool accesses the payload db in order to obtain the transactions of the block. At the time these calls are made the current block is generally not yet available in the chain database. 

This could be fixed by explicitly deleting the transactions of the currently validated block from the set of pending transactions. However, the overhead for keeping the transactions a little longer is small and the logic is simpler this way.

2.  Transactions that are included in a block are deleted from the mempool only 1. when included in a new block or 2. when they expire. The latter can take up to 2 days. In the former case the following can happen:

    1. Node `A` validates Block `b` with tx `t` and deletes `t` from the set of pending tx
    2. Node `B` hasn't yet validated `b` and synchronizes its mempool with `A`. `t` gets added to the mempool of `A`.
    3. Node `B` validates Block `b` and deletes `t` from the set of pending tx
    4. Node `A` synchronizes its mempool with `B`. `t` gets added to the mempool of `B`.
    5. At this point `A` and `B` both have validated `b` and have `t` in their mempools. It can take up to 2 days until `t` is removed.

    The PR introduces a probabilistic data structure that keeps track of unexpired transactions that have been included
    in blocks and prevents these from being added to the mempool.

TODO:

* [x] make `maxCurrentTxIdSize` configurable.
* [x] motivate PR and document changes
* [x] add code comment explaining why mempool update is done only for the parent and the current header
* [x] consider renaming `CurrentTxIdx` into something more meaningful
* [x] Provide telemetry and test results from mainnet tests.

### DIFF
--- a/src/Chainweb/Mempool/CurrentTxs.hs
+++ b/src/Chainweb/Mempool/CurrentTxs.hs
@@ -3,7 +3,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 
 -- |
--- Module: Chainweb.Mempool.CurrentTxIndex
+-- Module: Chainweb.Mempool.CurrentTxs
 -- Copyright: Copyright © 2020 Kadena LLC.
 -- License: MIT
 -- Maintainer: Lars Kuhtz <lars@kadena.io>
@@ -12,7 +12,7 @@
 -- A probabilistic index for non-expired transactions.
 --
 -- The false positive rate is very low. False negatives rate is 0 when the
--- number of entries is smaller than 'maxCurrentTxIdxSize' and grows with each
+-- number of entries is smaller than 'maxCurrentTxsSize' and grows with each
 -- entry beyond that size.
 --
 -- The purpose of this data structure is to allow the mempool to keep track of
@@ -23,14 +23,14 @@
 -- pact service during new block validation once it has been included in another
 -- block.
 --
-module Chainweb.Mempool.CurrentTxIndex
-( CurrentTxIdx
-, newCurrentTxIdx
-, currentTxIdxSize
-, currentTxIdxInsert
-, currentTxIdxInsertBatch
-, currentTxIdxMember
-, currentTxIdxRemove
+module Chainweb.Mempool.CurrentTxs
+( CurrentTxs
+, newCurrentTxs
+, currentTxsSize
+, currentTxsInsert
+, currentTxsInsertBatch
+, currentTxsMember
+, currentTxsRemove
 ) where
 
 import qualified Data.ByteString as B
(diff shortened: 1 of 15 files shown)


src/Chainweb/Mempool/CurrentTxs.hs
```diff
@@ -0,0 +1,155 @@
...
+
+-- import Chainweb.Mempool.InMemTypes
+import Chainweb.Mempool.Mempool
+import Chainweb.Time
+import Chainweb.Utils
+
+-- -------------------------------------------------------------------------- --
+-- Configuration
+
+-- | Limits the size of the commit log. If the size is reached elements are
```
larskuhtz, line 56:
```suggestion
-- | Limits the size of the current txs set. If the size is reached elements are
```

src/Chainweb/Mempool/CurrentTxs.hs
```diff
@@ -0,0 +1,155 @@
...
+-- deleted, and queries may return false negatives. Not guarantee is made about
```
larskuhtz, line 57:
```suggestion
-- deleted, and queries may return false negatives. No guarantee is made about
```

src/Chainweb/Mempool/CurrentTxs.hs
```diff
@@ -0,0 +1,155 @@
...
+-- what elements are deleted.
```
larskuhtz, line 58:
```suggestion
-- | Limits the size of the set of current txs. If the size is reached elements are
-- deleted, and queries may return false negatives. No guarantee is made about
-- which elements are deleted.
```

larskuhtz:
This PR has been tested for over a week on us-w2 and a common ubuntu node. Both nodes perform fine. Telemetry indicates that the PR works as expected.

src/Chainweb/Mempool/CurrentTxs.hs
```diff
@@ -0,0 +1,154 @@
...
+        -- @maxCurrentTxsSize@ should be much larger than 1000. The worst case
+        -- expected running time is for @1000 == maxCurrentTxsSize@, which still
+        -- is very fast.
+        --
+        let n = min l (max 1000 (l - int maxCurrentTxsSize))
+        in take n . L.nub . randomRs (0, l-1) <$> newStdGen
+      else return []
+
+    -- Not sure if foldl' would be faster here
+    return $ CurrentTxs $! foldr S.deleteAt s0 indexesToBeDeleted
```
emilypi, line 154: It's probably good to `foldl'` either way - less unnecessary memory pressure.

PR #1139, opened 2020-08-25
Validate Request Key Length at Handler

emilypi:
@larskuhtz


emilypi:
@larskuhtz We're covered in the positive case (the tests fail otherwise!), but the negative case would be good to test. I'll add one.

PR #1142, opened 2020-08-28
update package version to 2.1.1

larskuhtz:
* [x] add notes about performance of different database pruning modes to CHANGELOG
* [x] check if versions in rosetta code need to be updated.


test/Chainweb/Test/Rosetta/RestAPI.hs
```diff
@@ -389,7 +389,7 @@ genesisId = BlockId 0 "rdfJIktp_WL0oMr8Wr6lH49YkERAJ9MlFp0RPLMXPDE"
 rosettaVersion :: RosettaNodeVersion
 rosettaVersion = RosettaNodeVersion
     { _version_rosettaVersion = "1.3.1"
-    , _version_nodeVersion = "2.1"
+    , _version_nodeVersion = "2.1.1"
```
slpopejoy, line 392:
```suggestion
    , _version_nodeVersion = VERSION_chainweb
```
Requires CPP, but makes this self-updating

emilypi: 👍 on that change @larskuhtz

PR #1144, opened 2020-09-09
remove dependency on neat-interpolation

larskuhtz:
Neat-interpolation doesn't consider the formatting of output as part of the stable API. Version 0.4 introduced white space change that broke genesis header hashes.

This PR drops the dependency on neat-interpolation and replaces its use by `unlined`, `intercalate` and `<>` from the `text` package that can be assumed to be stable.


test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -663,62 +661,63 @@ toTerm' :: ToTerm a => a -> Term Name
...
+    , "      \"content\": content"
+    , "      })"
+    , "  )"
+    , "  (defun read-table ()"
+    , "    (select test-table (constantly true))"
+    , "  )"
+    , ")"
+    , ""
+    , "(create-table test-table)"
+    ]
```
emilypi, line 722: We'll need to confirm with a full replay that this works. Pact is white-space aware when it generates its hashes. Any substantive change will mess things up.

tools/ea/Ea.hs
```diff
@@ -170,38 +170,39 @@ genPayloadModule' v tag cwTxs =
                     execNewGenesisBlock noMiner (V.fromList cwTxs)
 
             let payloadYaml = TE.decodeUtf8 $ Yaml.encode payloadWO
-                modl = T.unlines $ startModule tag <> [payloadYaml] <> endModule
```
emilypi, line 173: why this change?

test/Chainweb/Test/Pact/Checkpointer.hs
larskuhtz: This is test code, and the test still passes.

The only non-test change is to the genesis payloads. There are golden tests for the genesis hashes. Any change to the payloads should trigger those tests to fail.

tools/ea/Ea.hs
larskuhtz: This is the change that causes the generated payload modules to not depend on neat-interpolation. The code replaces the use of the QuasiQuote with `T.unlines`, `T.intercalate`, and `<>` for `Text`, which are expected to be stable.

test/Chainweb/Test/Pact/Checkpointer.hs
emilypi: Ah right, the goldens! They're finally good for something 😄

emilypi (approved):

PR #1147, opened 2020-09-17
upper bound on base16-bytestring and fix mainnet validation trigger

larskuhtz:


.github/workflows/applications.yml
```diff
@@ -462,10 +464,10 @@ jobs:
 
     steps:
     - name: trigger mainnet history validation
-      if: env.GIT_REF_SHORT == 'master'
+      if: ${{ needs.config.output.git-short-ref == 'master' }}
```
larskuhtz, line 467:
```suggestion
      if: ${{ needs.config.outputs.git-short-ref == 'master' }}
```

mightybyte (approved):

PR #1148, opened 2020-09-30
Update for 2.2 release

mightybyte:


node/ChainwebNode.hs
```diff
@@ -496,7 +496,7 @@ pkgInfoScopes =
 -- KILLSWITCH for version 2.1
```
larskuhtz, line 496:
```suggestion
-- KILLSWITCH for version 2.2
```

larskuhtz (commented):
I think, there's also some place in the Rosetta tests that uses the current version as a constant.

mightybyte:
> I think, there's also some place in the Rosetta tests that uses the current version as a constant.

@LindaOrtega ?

LindaOrtega:
> > I think, there's also some place in the Rosetta tests that uses the current version as a constant.
> 
> @LindaOrtega ?

I think this is the place that Lars is referring to, but it seems to read the version from a variable called `VERSION_chainweb`: https://github.com/kadena-io/chainweb-node/blob/master/test/Chainweb/Test/Rosetta/RestAPI.hs#L394

This is the PR that first made use of that variable: https://github.com/kadena-io/chainweb-node/pull/1142/files

larskuhtz (approved):

PR #1150, opened 2020-10-03
Support Pact Events

slpopejoy:
Forking change to support Pact Events (https://github.com/kadena-io/pact/pull/819) in output for SPV proofs etc.

TODOs: 
- [ ] Test against mainnet history (current block numbers should be high enough for this)
- [x] Populate block numbers for switch
- [x] merge `kpkgs` pact 3.6 bump to master https://github.com/kadena-io/kpkgs/pull/28 (waiting for mainnet replay to succeed for this)

PR #1157, opened 2020-10-21
Pact Events redux

slpopejoy:
Retry of #1150 with better back-compat.

Picking block heights might need to wait for 2.3 release. @larskuhtz @mightybyte how should we track? Also we need a mainnet replay.

Only way to test with Chainweaver etc IMO is in devnet, however this build should deploy into testnet/prod without disturbance with the current blockheights, so can at least verify that `local` backcompat is not broken.

### DIFF
--- a/test/Chainweb/Test/Pact/TransactionTests.hs
+++ b/test/Chainweb/Test/Pact/TransactionTests.hs
@@ -237,7 +237,7 @@ testCoinbase797DateFix = testCaseSteps "testCoinbase791Fix" $ \step -> do
             noSPVSupport Nothing 0.0 (RequestKey h) 0 def
           txst = TransactionState mempty mempty 0 Nothing (_geGasModel freeGasEnv)
 
-      CommandResult _ _ (PactResult pr) _ _ _ _ <- evalTransactionM tenv txst $!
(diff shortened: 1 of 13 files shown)


test/Chainweb/Test/Pact/TransactionTests.hs
```diff
@@ -237,7 +237,7 @@ testCoinbase797DateFix = testCaseSteps "testCoinbase791Fix" $ \step -> do
             noSPVSupport Nothing 0.0 (RequestKey h) 0 def
           txst = TransactionState mempty mempty 0 Nothing (_geGasModel freeGasEnv)
 
-      CommandResult _ _ (PactResult pr) _ _ _ _ <- evalTransactionM tenv txst $!
+      CommandResult _ _ (PactResult pr) _ _ _ _ _ <- evalTransactionM tenv txst $!
```
emilypi, line 240: It might be useful to add `resultOf` (defined above) to the Pact utils so we can just keep reusing it across tests.

emilypi (approved):

LindaOrtega (approved):

test/Chainweb/Test/Pact/TransactionTests.hs
slpopejoy: Not sure ... `_pactResult . _crResult` is pretty terse?

PR #1167, opened 2020-12-05
improve tagging of pact service logs

larskuhtz:


src/Chainweb/Pact/PactService.hs
```diff
@@ -395,7 +398,7 @@ attemptBuyGas miner (PactDbEnv' dbEnv) txs = do
         -> P.Gas
         -> PactServiceM cas (TransactionEnv db)
     createGasEnv db cmd gp gl = do
-        l <- view $ psCheckpointEnv . cpeLogger
+        l <- P.newLogger <$> view psLoggers <*> pure "AtemptByGas"
```
slpopejoy, line 401:
```suggestion
        l <- P.newLogger <$> view psLoggers <*> pure "attemptBuyGas"
```

src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -377,23 +380,38 @@ rewindTo rewindLimit (Just (ParentHeader parent)) = do
...
+        -> P.Logger
+            -- ^ logger for loggingprogress
+        -> Int
+            -- ^ How often a progress message is issued
+        -> S.Stream (S.Of BlockHeader) IO r
+        -> S.Stream (S.Of BlockHeader) IO r
+    progress runPact logger (n :: Int) s = s
+        & S.zip (S.enumFrom 1)
+        & S.mapM
+            (\(!i, !b) -> do
```
slpopejoy, line 402: Sorry, I don't understand `Stream` constructions, that is, `S.Stream a IO r -> S.Stream a IO r` ...

- does `progress` act as an independent streaming consumer alongside the `S.scanM` call above?
- The `S.mapM` call here naively looks like "every time progress gets called it loops through stuff" but is it instead the incremental consumer?

src/Chainweb/Pact/Types.hs
```diff
@@ -503,15 +517,32 @@ pactLoggers logger = P.Loggers $ P.mkLogger (error "ignored") fun def
         let namedLogger = addLabel ("logger", pack n) logger
         logFunctionText namedLogger (pactLogLevel cat) $ pack msg
 
+-- | Write log message
+--
+logg_ :: MonadIO m => P.Logger -> String -> String -> m ()
```
slpopejoy, line 522: Instead of the alternates, why not just migrate to this `MonadIO` version?

src/Chainweb/Pact/PactService/Checkpointer.hs
larskuhtz: Composing streams fuses them into a single loop.

src/Chainweb/Pact/Types.hs
larskuhtz: Most uses are within `PactServiceM c a`. In that context, using the `MonadIO` version would require to first extract the logger from the pact environment via `view psLogger`, which seems inconvenient.

src/Chainweb/Pact/PactService/Checkpointer.hs
larskuhtz: That's a nice feature of the streaming library the stream are composed just by function composition.

slpopejoy (approved):

emilypi (approved):

PR #1168, opened 2020-12-05
Incremental pact history replay

larskuhtz:
[This branch includes #1167 and should be reviewed and merged after it]

### Summary

This PR adds a new form of pact service checkpointer rewind that is incremental. Blocks are replayed in chunks and the state is committed to the pact database after each chunk. When the process is interrupted or a failure occurs, a restart resumes after the latest successful chunk.

During normal node operation, pactValidateBlock` continues to use the transactional `rewindTo`. The new incremental `rewindTo` is used for the initial pact database synchronization during node startup.

### Details

On startup chainweb-nodes syncs the pact database to the current cut that is recorded in the chain database. This is done by calling `pactValidateBlock` on the for the respective block on each chain. `pactValidateBlock` first rewinds the checkpointer to the parent of the evaluated block. `pactValidateBlock` is transactional, i.e. in case of a failure the the rewind is rolled back and the checkpointer is left unchanged.  There are a few scenarios to consider:

1. Normal node operation.
2. Node startup, after a node was offline for some period of time.
3. Resetting and rebuilding the pact database (e.g. when the database was corrupted or after a db schema or format change).
4. Replaying the pact history re-validating the payload hashes for testing and debugging purposes.

In the first scenario it is desired that the rewind is transactional and since it only involves a small number of blocks it doesn't matter anyways. 

In the second and the third scenario transactions may get very large. SQLite itself can handle even very large transactions efficiently. A large number of rewinded blocks might be discarded if a failure occurs or the process is interrupted. Replaying a large portion of the history may take a long time to replay.

Finally, during testing and debugging, failures are expected to some degree and it is not acceptable if repeating a test takes a long time.

### DIFF
--- a/src/Chainweb/Chainweb.hs
+++ b/src/Chainweb/Chainweb.hs
@@ -812,7 +812,10 @@ withChainwebInternal conf logger peer rocksDb pactDbDir resetDb inner = do
         syncOne :: (BlockHeader, ChainResources logger) -> IO ()
         syncOne (bh, cr) = do
             let pact = _chainResPact cr
-            let logCr = logFunctionText $ _chainResLogger cr
+            let logCr = logFunctionText
+                    $ addLabel ("component", "pact")
+                    $ addLabel ("sub-component", "init")
+                    $ _chainResLogger cr
             let hsh = _blockHash bh
             let h = _blockHeight bh
             logCr Info $ "pact db synchronizing to block "
--- a/src/Chainweb/Chainweb.hs
+++ b/src/Chainweb/Chainweb.hs
@@ -134,7 +134,6 @@ import Control.Monad.Catch (MonadThrow, throwM)
 import Control.Monad.Writer
 
 import Data.Bifunctor (second)
-import Data.CAS (casLookupM)
 import Data.Foldable
 import Data.Function (on)
 import qualified Data.HashMap.Strict as HM
@@ -183,7 +182,6 @@ import Chainweb.Pact.RestAPI.Server (PactServerData)
 import Chainweb.Pact.Service.Types (PactServiceConfig(..))
 import Chainweb.Pact.Types (defaultReorgLimit)
 import Chainweb.Pact.Utils (fromPactChainId)
-import Chainweb.Payload
 import Chainweb.Payload.PayloadStore
 import Chainweb.Payload.PayloadStore.RocksDB
 import Chainweb.RestAPI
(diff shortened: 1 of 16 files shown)


src/Chainweb/Chainweb.hs
```diff
@@ -812,14 +810,15 @@ withChainwebInternal conf logger peer rocksDb pactDbDir resetDb inner = do
...
+                    $ addLabel ("sub-component", "init")
+                    $ _chainResLogger cr
             let hsh = _blockHash bh
             let h = _blockHeight bh
             logCr Info $ "pact db synchronizing to block "
                 <> T.pack (show (h, hsh))
-            payload <- payloadWithOutputsToPayloadData
-                <$> casLookupM payloadDb (_blockPayloadHash bh)
-            void $ _pactValidateBlock pact bh payload
+            void $ _pactSyncToBlock pact bh
```
slpopejoy, line 821: Did you consider simply adding a flag to `_pactValidateBlock` to do incremental replay? Adding a queue RPC is expensive when this seems to be essentially a piece of advice on how to replay.

larskuhtz: I did consider it. But the use case is very different. Using pactValidateBlock for synchronizing the pact db with the consensus db always felt a bit like a misuse of it.

An `incrementalRewindTo` flag on `pactValidateBlock` would be feel a bit far-fetched and would only really be used for the initial db synchronization. So, adding an new API request for the synchronization seemed more natural to me.

src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -503,3 +525,123 @@ lookupBlockHeader bhash ctx = do
...
+blockStreamProgress
+    :: Monad m
+    => Int
+        -- ^ Progress reporting callback
+    -> (String -> m ())
+        -- ^ progress callback
+    -> S.Stream (S.Of BlockHeader) m r
+    -> S.Stream (S.Of BlockHeader) m r
+blockStreamProgress n logFun = progress n $ \i b -> logFun
+    $ "processes blocks: " <> sshow i
```
larskuhtz, line 645:
```suggestion
    $ "processed blocks: " <> sshow i
```

src/Chainweb/Chainweb.hs
slpopejoy: > Using pactValidateBlock for synchronizing the pact db with the consensus db always felt a bit like a misuse of it.
👍

slpopejoy (approved):

PR #1169, opened 2020-12-08
fix applyUpgrades to correctly update psInitCache

larskuhtz:
This PR ensures that fungible-v2 is included in the module cache after calling `applyUpgrade`. 

Perviously fungible-v2 was missing from psInitCache and would be added only after the next restart of pactService. This could break history replays or catchups (that started replay from before the upgrade).

### DIFF
--- a/src/Chainweb/Pact/TransactionExec.hs
+++ b/src/Chainweb/Pact/TransactionExec.hs
@@ -440,11 +440,6 @@ applyUpgrades v cid height
       txs <- map (fmap payloadObj) <$> liftIO txsIO
 
       --
-      -- Note (emily): the historical use of 'mapM' here means that we are not
-      -- threading the updated module cache from tx to tx in our upgrades. This
-      -- means that the resulting module cache is the set of modules loaded in
-      -- the /last/ tx, and that result will go into the hash of the block.
-      --
       -- In order to prime the module cache with all new modules for subsequent
       -- blocks, the caches from each tx are collected and the union of all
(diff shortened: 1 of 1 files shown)


emilypi (approved):

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -440,20 +440,19 @@ applyUpgrades v cid height
       txs <- map (fmap payloadObj) <$> liftIO txsIO
 
       --
-      -- Note (emily): the historical use of 'mapM_' here means that we are not
-      -- threading the updated module cache from tx to tx in our upgrades.
-      -- This means that the outputed module cache is the set of modules
-      -- loaded in the /last/ tx, and that result will go into the hash.
+      -- Note (emily): the historical use of 'mapM' here means that we are not
```
slpopejoy, line 443: This first paragraph is no longer relevant, just keep your comment starting on L448

slpopejoy (approved):

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -440,20 +440,19 @@ applyUpgrades v cid height
...
+      -- threading the updated module cache from tx to tx in our upgrades. This
+      -- means that the resulting module cache is the set of modules loaded in
+      -- the /last/ tx, and that result will go into the hash of the block.
       --
```
larskuhtz, line 447:
```suggestion
```

PR #1172, opened 2020-12-23
wrap noop checkpointer use in withBatch

larskuhtz:
From time to time some newly started nodes got stuck with

```
Received exception running pact service (syncToBlockBlock): Code invariant violation: withCheckpointerRewind called by rewindTo outside of batch. Please report this as a bug.
  error, called at src/Chainweb/Pact/PactService/Checkpointer.hs:197:9 in chainweb-2.3-inplace:Chainweb.Pact.PactService.Checkpointer
 ```
 
 This is a regression from #1168. The change in #1168 splits up a single huge pact-db batch of unbounded size into chunks of small, constant sized, transactions. In order to do so it moves the `withBatch` bracket into the inner rewind loop. However, the implementation missed to add `withBatch` for the corner case of a noop rewind, which by-passes the loop.

### DIFF
--- a/src/Chainweb/Pact/PactService/Checkpointer.hs
+++ b/src/Chainweb/Pact/PactService/Checkpointer.hs
@@ -83,6 +83,8 @@ import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Tuple.Strict
 
+import GHC.Stack
+
 import qualified Pact.Types.Logger as P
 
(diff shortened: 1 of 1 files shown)


larskuhtz:
This has been tested to solve the issue on several mainnet instances.

slpopejoy (approved):

PR #1174, opened 2020-12-24
Add X-Peer-Addr header

larskuhtz:
The purpose of this is to make node operation easier and remove centralized single points of failure. Currently the docker image uses a free  third party service to discover the external IP address of a node. With this change chainweb nodes can use the any node in the chainweb network for this. It also supports the removal of complex bash startup scripts for checking connectivity from the docker image.

### Changes

* [x] Re-add builtin bootstrap nodes. This also means that `--ignore-boostrap-nodes` regains its meaning.
* [x] Add X-Peer-Addr header that allows nodes to auto-discover their external network configuration
* [x] Support nodes to auto-configure their hostname. This eliminates the need to use a (centralized) third party service for that.
* [x] Validate p2p configuration on startup
* [x] Validate peer configuration on startup


src/P2P/Node/Configuration.hs
```diff
@@ -98,8 +109,35 @@ defaultP2pConfiguration = P2pConfiguration
...
     , _p2pConfigPrivate = False
+    , _p2pConfigBootstrapReachability = 0.5
     }
 
+validateP2pConfiguration :: Applicative a => ConfigValidation P2pConfiguration a
+validateP2pConfiguration c = do
+    validatePeerConfig $ _p2pConfigPeer c
+
+    when (_p2pConfigIgnoreBootstrapNodes c && null (_p2pConfigKnownPeers c)) $ tell
+        $ pure "Default bootstrap nodes are ignored and no known peers are configured. This node won't be able to communicate with the network"
```
sirlensalot, line 120:
```suggestion
        $ pure "Default bootstrap nodes are ignored and no known peers are configured. This node won't be able to communicate with the network."
```

src/P2P/Node/Configuration.hs
```diff
@@ -98,8 +109,35 @@ defaultP2pConfiguration = P2pConfiguration
...
+
+    validateRange "sessionTimeout" (60 {- 1 min -}, 900 {- 15 min -}) (_p2pConfigSessionTimeout c)
+
+    when (_p2pConfigSessionTimeout c < 120) $ tell
+        $ pure "This node is configured with a p2p session timeout of less than 120. This causes network overhead for creating new sessions. A connection timeout between 180 and 600 seconds is recommended."
+
+    validateRange "maxSessionCount" (2, 30) (_p2pConfigMaxSessionCount c)
+
+    when (_p2pConfigMaxSessionCount c < 3) $ tell
+        $ pure "This node is configured to have a maximum session count of less than 5. This will limit the ability of this node to communicate with the rest of the network. A max session count between 5 and 15 is adviced"
```
sirlensalot, line 133:
```suggestion
        $ pure "This node is configured to have a maximum session count of less than 5. This will limit the ability of this node to communicate with the rest of the network. A max session count between 5 and 15 is advised."
```

src/P2P/Node/Configuration.hs
```diff
@@ -98,8 +109,35 @@ defaultP2pConfiguration = P2pConfiguration
...
+
+    when (_p2pConfigMaxSessionCount c > 30) $ throwError
+        "This node is configured with a maximum session count of more than 30. This may put a high load on the network stack of the node and may cause connectivity problems. A max session count between 5 and 15 is adviced"
```
sirlensalot, line 136:
```suggestion
        "This node is configured with a maximum session count of more than 30. This may put a high load on the network stack of the node and may cause connectivity problems. A max session count between 5 and 15 is advised."
```

src/P2P/Peer.hs
```diff
@@ -248,13 +254,17 @@ peerInfoClientEnv mgr = mkClientEnv mgr . peerBaseUrl . _peerAddr
...
+
 -- -------------------------------------------------------------------------- --
 -- Peer Configuration
 
 data PeerConfig = PeerConfig
     { _peerConfigAddr :: !HostAddress
         -- ^ The public host address of the peer.
+        --
         -- A port number of 0 means that a free port is assigned by the system.
+        -- An IP address of 0.0.0.0 means that the node discovers it's
```
sirlensalot, line 266:
```suggestion
        -- An IP address of 0.0.0.0 means that the node discovers its
```

sirlensalot (approved):
Approved, small language edits to messages

PR #691, opened 2019-11-12
validate P2P configuration

larskuhtz:

PR #1175, opened 2020-12-24
validate p2p configuration on startup

larskuhtz:
NOTE: the changes of this PR are now contained in #1174 

This is an updated version of #691 

* [x] Warnings about problematic P2P configuration settings
* [x] Errors on unacceptable P2P configuration settings
* [ ] Connectivity of bootstrap nodes and known peers at startup
    * [ ] new configuration option `requiredReachableBootstrapNodes` for the required number of bootstrap nodes that are reachable at startup
    * [ ] check available bootstrap nodes
    * [ ] check connectivity with available bootstrap nodes (depends on #1174)
 * [ ] validate that configured host IP/domain name and port matches actual external IP address (depends on #1174)
 * [ ] auto-configure host IP if not set (depends on #1174)
 * [ ] option to turn of p2p-validation (for easier bootstrapping of development networks)

### Builtin Bootstrap Nodes

This PR also re-adds the list of builtin bootstrap nodes to the code base, which was removed before the mainnet launch in order to not reveal the network ahead of the official announcement.Including a builtin list of bootstrap nodes simplifies packaging and operation of chainweb-node. 

A hard code list of bootstrap nodes could be seen as a form of centralization. However, most Bitcoin clients also include such a list in the code. The code is open source and anybody can make a PR to add their own node, if they think that is satisfies the requirements for bootstrap nodes. Actually, I think, we should welcome additional bootstrap nodes from a diverse set of different entities. Moreover, it is sufficient that a single node from the set of bootstrap nodes is trusted and it is possible to configure additional known peers in addition to the builtin set of bootstrap nodes. It is always possible to ignore the builtin bootstrap nodes via a command line flag or configuration file setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,27 @@ cabal v2-install
 
 ## Bootstrap Nodes
 
-### Testnet Nodes
+Bootstrap nodes are used by chainweb-nodes on startup in order to discover other
+nodes in the network. A least one of the bootstrap nodes must be trusted.
+
+Chainweb node operators can configure additional bootstrap nodes by using the
+`--known-peer-info` command line option or in a configuration file. It is also
+possible to ignore the builtin bootstrap nodes by using the
+`--enable-ignore-bootstrap-nodes` option or the respective configuration file
+setting.
+
+Bootstrap nodes must have public DNS names and a corresponding TLS certificate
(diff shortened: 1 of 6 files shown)


larskuhtz:
the changes of this PR are now contained in #1174

PR #1177, opened 2020-12-30
Remove unused mining code

larskuhtz:
* [x] removes (finally!) support for public mining.
* [x] removes some other unused mining code and cleans up the mining coordination code.
* [x] Improves test coverage, but using the same code path for in-node mining that is also used for the mining API. All test code uses in-node mining, while mainnet uses only external mining via the mining API. This PR adds test coverage for the latter.
    * [x] fix races that in the mining API, that surfaced because of the changes in this PR.


src/Chainweb/Miner/Coordinator.hs
```diff
@@ -232,21 +218,25 @@ chainChoice c choice = case choice of
...
+publish lf cdb pwVar miner pd s = do
     c <- _cut cdb
     now <- getCurrentTimeIntegral
     try (extend c pd s) >>= \case
 
         -- Publish CutHashes to CutDb and log success
         Right (bh, Just ch) -> do
+
+            -- reset the primed payload for this cut extension
+            atomically $ modifyTVar pwVar $ resetPrimed miner (_chainId bh)
```
larskuhtz, line 235: This fixes a race in the mining code where a newWork would retrieve an old payload on a new cut. This could happen when a miner would request a primed payload after a new cut arrived but before the primed payload got updated.

There is probably still a race with external cuts, which requires separate attention.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -134,43 +162,32 @@ newtype PrevTime = PrevTime BlockCreationTime
...
     logFun @T.Text Debug $ "newWork: picked chain " <> sshow cid
 
-    mr <- case eminer of
-        Primed m -> primed m cid <$> readTVarIO tpw
-        Plebian m -> do
-            logFun @T.Text Debug
-                $ "newWork: call getCutExtension for chain "
-                <> sshow cid <> " on cut " <> cutIdToTextShort (_cutId c)
-            public m cid
+    PrimedWork pw <- readTVarIO tpw
```
larskuhtz, line 186: We should probably double check, that the payload is valid for the cut, i.e. that it is mined on top of the most recent block for the respective chain. For that the parent hash (or just the block height might be enough?) should be included in the PrimedWork value.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -134,82 +163,44 @@ newtype PrevTime = PrevTime BlockCreationTime
...
-                $ "newWork.public: got extension."
-                <> " Parent: " <> encodeToText (ObjectEncoded $ _parentHeader $ _cutExtensionParent ext)
-                <> " Adjacents: " <> encodeToText (_cutExtensionAdjacentHashes ext)
-                <> " Cuthashes: " <> encodeToText (cutToCutHashes Nothing (_cutExtensionCut ext))
-            -- This is an expensive call --
-            payload <- trace logFun "Chainweb.Miner.Coordinator.newWork.newBlock" () 1
-                (_pactNewBlock pact miner $ _cutExtensionParent ext)
-            pure . Just $ T2 (payloadWithOutputsToPayloadData payload) ext
+        Just (T2 (payload, parent) extension)
+            | parent == _blockHash (_parentHeader (_cutExtensionParent extension)) -> do
```
larskuhtz, line 197: Adding this check fixes a race where a new cut is picked up before the primed work is updated. I this case it could happen that that a mining client would receive an outdated payload. The respective block would only fail later during validation in the cut pipeline.

test/Chainweb/Test/Mining.hs
```diff
@@ -0,0 +1,95 @@
...
+    var <- newEmptyMVar
+    x <- race (takeMVar var) $ do
+        withTestCutDb rdb v 0 (\_ _ -> return fakePact) (logFunction logger) $ \cdb ->
+            withMiningCoordination logger conf cdb $ \case
+                Nothing -> error "nonEmptyMiningAccount: Bug in the mining Code"
+                Just coord -> do
+                    a logger coord
+                    putMVar var ()
+    case x of
+        Left () -> return ()
```
sirlensalot, line 78: Assuming this is an error, it would be good to fail.

src/Chainweb/Miner/Miners.hs
```diff
@@ -78,36 +75,36 @@ import Data.LogMessage (LogFunction)
...
-    pact :: PactExecutionService
-    pact = _webPactExecutionService $ view cutDbPactService cdb
-
-    hdb :: WebBlockHeaderDb
-    hdb = view cutDbWebBlockHeaderDb cdb
+        go height wh >>= solve coord
+        -- void $ awaitNewCut cdb c
+        void $ awaitNewCutByChainId cdb (_workHeaderChainId wh) c
 
+        -- race (awaitNewCutByChainId cdb (_workHeaderChainId wh) c) (go height wh) >>= \case
```
sirlensalot, line 97: commented out?

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -230,43 +228,57 @@ withMiningCoordination logger conf cdb inner
...
+-- production setups.
 --
 data MinerResources logger cas = MinerResources
     { _minerResLogger :: !logger
     , _minerResCutDb :: !(CutDb cas)
     , _minerChainResources :: HashMap ChainId (ChainResources logger)
     , _minerResConfig :: !NodeMiningConfig
+    , _minerResCoordination :: !(Maybe (MiningCoordination logger cas))
+        -- ^ The primed work cache. This is Nothing when coordination is
+        -- disable. It is needed by the in-node test miner. The mempoolNoopMiner
```
sirlensalot, line 242:
```suggestion
        -- disabled. It is needed by the in-node test miner. The mempoolNoopMiner
```

test/Chainweb/Test/Mining.hs
```diff
@@ -0,0 +1,95 @@
...
+        & miningCoordination . coordinationEnabled .~ True
+
+-- -------------------------------------------------------------------------- --
+-- Tests
+
+nonEmptyMiningAccount :: HasCallStack => RocksDb -> Assertion
+nonEmptyMiningAccount rdb = withTestCoordiantor rdb Nothing $ \_logger coord -> do
+    PrimedWork w <- readTVarIO (_coordPrimedWork coord)
+    forM_ (HM.keys w) $ \(MinerId k) ->
+        assertBool "miner account name must not be the empty string" (not (T.null k))
```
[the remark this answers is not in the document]
larskuhtz: It is a regression test for a potential/former bug in line 118 of `src/Chainweb/Chainweb/MinerResources.hs`.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -124,8 +124,8 @@ gp = 0.1
 -- | Note: These tests are intermittently non-deterministic due to the way
 -- random chain sampling works with our test harnesses.
 --
-tests :: RocksDb -> ScheduledTest
-tests rdb = testGroupSch "Chainweb.Test.Pact.RemotePactTest"
+tests_ :: RocksDb -> ScheduledTest
```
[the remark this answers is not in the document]
larskuhtz: Oh, that's related to debugging code that I forgot to remove.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -22,7 +22,7 @@
 --
 module Chainweb.Test.Pact.RemotePactTest
 ( tests
-, withNodes
+, tests_
```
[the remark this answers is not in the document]
larskuhtz: should be removed.

PR #1178, opened 2020-12-31
Improve TXOUT SPV to include events

sirlensalot:


larskuhtz (approved):
I skimmed only quickly over the tests. It's looking good otherwise.

[Unrelated of this PR, I am a bit concerned about the use of aeson parses in this module because it seems that in case of failures the aeson failure messages would made it onto the chain, which means that any change in those messages would fork the chain. I don't think that aeson considers error messages as part of the stable API. We may consider wrapping those calls in inject our own stable failure messages.]

src/Chainweb/Pact/SPV.hs
```diff
@@ -224,3 +234,64 @@ getTxIdx bdb pdb bh th = do
...
+            , ("txid", tStr $ maybe "" asString _crTxId)
+            , ("gas", toTerm $ (fromIntegral _crGas :: Integer))
+            , ("meta", maybe empty metaField _crMetaData)
+            , ("logs", tStr $ asString $ _crLogs)
+            , ("continuation", maybe empty contField _crContinuation)
+            , ("events", toTList TyAny def $ map eventField _crEvents)
+            ])
+    TyAny Nothing def
+  where
+    metaField v = case fromJSON v of
```
sirlensalot, line 259: @larskuhtz so this field is a little annoying -- `_crMetaData` is `Maybe Value`, as it's supposed to be an entirely platform-dependent free field in the Pact API. Here, `fromJSON` is converting to a `PactValue`, and note that if it fails, it simply adds an empty object.

PR #1179, opened 2021-01-01
ethereum receipt SPV

larskuhtz:
This branch is based on #1178

PR #1181, opened 2021-01-03
add endpoint for creating eth-receipt-proofs

larskuhtz:
based on #1179


sirlensalot (approved):
Approved. I didn't inspect terribly closely except the RemotePactTest part which looks good.

PR #1183, opened 2021-01-08
Improve TXOUT proof format

sirlensalot:
Currently, `(verify-spv "TXOUT" (read-msg 'proof))` expects the data to be the full JSON of `TransactionOutputProof`. However this is not what the `pact/spv` endpoint returns -- it returns this JSON encoded in base64url-unpadded. 

This changes the "new" TXOUT (post-fork) to expect an object `{ "proof": [base64url-unpadded text] }`. See `createVerify` in the SPV test for the structure.

### DIFF
--- a/src/Chainweb/Pact/SPV.hs
+++ b/src/Chainweb/Pact/SPV.hs
@@ -124,7 +124,7 @@ verifySPV bdb bh typ proof = go typ proof
           Right result -> return $ Right $ ethResultToPactValue result
 
       -- Chainweb tx output proof
-      "TXOUT" -> case extractProof o of
+      "TXOUT" -> case extractProof enableBridge o of
         Left t -> return (Left t)
         Right u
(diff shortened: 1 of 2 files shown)


larskuhtz:
Does this require that clients that use the SPV endpoints for cross-chain transfers must be updated?

larskuhtz (approved):

sirlensalot:
@larskuhtz no, it only affects post-fork `verify-spv` calls. Cross chain you'll notice is handled in a completely different function (`verifyCont`) than ETH and TXOUT proofs (`verifySPV`), corresponding to two different calls in the `SPVSupport` Pact backend API https://github.com/kadena-io/chainweb-node/blob/6bcd680b98bb0b1cdd285f3e75d98b2982374052/src/Chainweb/Pact/SPV.hs#L88

PR #1189, opened 2021-01-25
remove these-skinny dependency

larskuhtz:
Chainweb depends on these-skinny and indirectly also on these. Both export the module `Data.These`, which sometimes caused issues within ghci session. This PR remove the dependency on skinny-these.


emilypi (changes requested):
Since `aeson` now depends on `these`, it makes no sense to have `these-skinny` anymore. We should just use `these`.

larskuhtz:
Close in favor of #1190

PR #1192, opened 2021-02-11
Add payload batch APIs

larskuhtz:


src/Chainweb/Payload/RestAPI/Server.hs
```diff
@@ -51,7 +57,25 @@ import Chainweb.Version
...
+
+-- | The maximum number of items that are returned in a batch
+--
+payloadBatchLimit :: Int
+payloadBatchLimit = 100
+
+err404Msg :: ToJSON msg  => msg -> ServerError
+err404Msg msg = err404 { errBody = encode msg }
+
+catMaybes :: V.Vector (Maybe a) -> V.Vector a
```
sirlensalot, line 70: Remove please. Our cabal is >= 0.12 so this is unnecessary https://github.com/kadena-io/chainweb-node/blob/master/chainweb.cabal#L408

src/Chainweb/Payload/RestAPI/Server.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
```
sirlensalot, line 1: Remove, see comment below

src/Chainweb/Payload/RestAPI/Server.hs
```diff
@@ -77,6 +101,29 @@ payloadHandler db k = run >>= \case
...
+    => PayloadDb cas
+    -> [BlockPayloadHash]
+    -> Handler [PayloadData]
+payloadBatchHandler db ks = liftIO $ do
+    payloads <- catMaybes
+        <$> casLookupBatch payloadsDb (V.fromList $ take payloadBatchLimit ks)
+    txs <- V.zipWith (\a b -> payloadData <$> a <*> pure b)
+        <$> casLookupBatch txsDb (_blockPayloadTransactionsHash <$> payloads)
+        <*> pure payloads
+    return $ V.toList $ catMaybes txs
```
sirlensalot, line 119: It's not a problem that some hashes may be bad?

sirlensalot (changes requested):
Looks good, except for CPP, plus API question: how would a replay handle interspersed missing entries? `Maybe xxx` makes sense for a single call but in a batch, seems like it will be a chore to figure out what was missing.

src/Chainweb/Payload/RestAPI/Server.hs
larskuhtz: It is added only in 0.12.2, which is not in our nix-pin.

larskuhtz:
> Looks good, except for CPP, plus API question: how would a replay handle interspersed missing entries? `Maybe xxx` makes sense for a single call but in a batch, seems like it will be a chore to figure out what was missing.

I think, it's fine here, since the result items include the `BlockPayloadHash` values from the request. So one can easily associate the results with the request keys.

We could return `null` in the result for missing entries, but that doesn't work well in cases where the server responds less than the requested number of items because it exceeds the maximum.

sirlensalot (approved):

PR #1193, opened 2021-02-16
remove unused deps and enabled optimization for tls

larskuhtz:


chainweb.cabal
```diff
@@ -370,7 +370,6 @@ library
         , process >= 1.5
         , random >= 1.1
         , random-bytestring >= 0.1
-        , resourcet >= 1.2
```
mightybyte, line 373: Woah, nice dependency pruning!  Were these all unused?

mightybyte (approved):

chainweb.cabal
larskuhtz: Yes. I just recently discovered the GHC flag: `-Wunused-packages`. Not sure in what GHC version it was introduced.

PR #1200, opened 2021-02-25
Additional verification for coin contract

sirlensalot:
This creates a copy of the coin contract that only differs in properties added or removed, with the goal of demonstrating the entire scope of coin contract verification that is currently available.

### DIFF
--- a/pact/coin-contract/v2/verified/coin.pact
+++ b/pact/coin-contract/v2/verified/coin.pact
@@ -1,0 +1,590 @@
+(module coin GOVERNANCE
+
+  @doc "'coin' represents the Kadena Coin Contract. This contract provides both the \
+  \buy/redeem gas support in the form of 'fund-tx', as well as transfer,       \
+  \credit, debit, coinbase, account creation and query, as well as SPV burn    \
+  \create. To access the coin contract, you may use its fully-qualified name,  \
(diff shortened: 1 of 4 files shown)


pact/coin-contract/v2/verified/coin.pact
```diff
@@ -0,0 +1,590 @@
...
+    [ (defproperty conserves-mass
+        (= (column-delta coin-table 'balance) 0.0))
+
+      (defproperty valid-account (account:string)
+        (and
+          (>= (length account) 3)
+          (<= (length account) 256)))
+
+      (defproperty enforce-custody (account:string)
+        (row-enforced coin-table 'guard account))
```
emilypi, line 19: Just to confirm, this adds an additional layer of security which reinforces that the tx sender is authorized to modify an account's row via `enforce-keyset`?  It looks like this property is enforced as part of the `gas-guard` and any debiting operations. This makes sense to add!

emilypi (approved):

pact/coin-contract/v2/verified/coin.pact
sirlensalot: Yes, it leverages `row-enforced` which is a really great property!

PR #1201, opened 2021-02-26
Disable hanging tests on macOS

mightybyte:


mightybyte:
Made unnecessary by #1202

PR #1208, opened 2021-03-08
increase default listen timeoutto 3 minutes

larskuhtz:
Increase the listen timeout in the chainweb-node from 2min to 3min.


larskuhtz:
For the current network (since  2021-02-28) a tx latency of 3 minutes is somewhere between the 80% and 90% percentile:

image

mightybyte (approved):

PR #1210, opened 2021-03-09
Event Proofs

larskuhtz:
This PR adds the features that allow to create proofs for binary encoded pact Events using Keccak_256 as hash function.

## Usage Example

Create proof for transfer on mainnet:

```sh
curl -s "https://us-east1.api.chainweb.com/chainweb/0.0/mainnet01/chain/1/pact/spv2" -XPOST -H 'content-type: application/json' -d '{ "subjectIdentifier" : { "type":"events", "chain":1, "requestKey": "QiX2nIzoq21HoA7Hiq2Rjkx5Nl7jD-l4nUGOjsz8aPA"}, "minimalProofDepth": 3, "algorithm": "Keccak_256"}'
```

Result:

```json
{
  "subject": {
    "input": "ADAgAAAAQiX2nIzoq21HoA7Hiq2Rjkx5Nl7jD-l4nUGOjsz8aPAEAAAACAAAAFRSQU5TRkVSDAAAAGtzd2FwLmtwZW5ueSAAAADZU-C1BQTw6ghU2A09r-wtWZZdphpzWlmsvB3i3VjBQwMAAAAAQAAAADZmNTJjZTBkMWRmYzJmNzNlY2M4ZjE2MmFkODE1ZWZjMmM5NzY2YWI3NmIxYTQ4YzlhNjc4ZDhjZDc5ZGQ0MDIAKwAAAFRvWXViX1ZoeldHdWVEWVVwODBnVmVWTnZiVGdoaDA4dU9kVUJtMDZkcFkCAABAe6XwY4GWCgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAVFJBTlNGRVIJAAAAa3N3YXAueHl6IAAAAP3AV3e1ornnygmVlZcEYY61dPGffrku_zZpklQUCx75AwAAAABAAAAANmY1MmNlMGQxZGZjMmY3M2VjYzhmMTYyYWQ4MTVlZmMyYzk3NjZhYjc2YjFhNDhjOWE2NzhkOGNkNzlkZDQwMgArAAAAVG9ZdWJfVmh6V0d1ZURZVXA4MGdWZVZOdmJUZ2hoMDh1T2RVQm0wNmRwWQKAZ5fVFxAj49sAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAABNSU5UDAAAAGtzd2FwLnRva2VucyAAAAB6vJFqfTyW0fbYYSu1VSkKPDE8a-VkGeN67sH_V9xlJwMAAAAAFgAAAGtzd2FwLmtwZW5ueTprc3dhcC54eXoAQAAAADZmNTJjZTBkMWRmYzJmNzNlY2M4ZjE2MmFkODE1ZWZjMmM5NzY2YWI3NmIxYTQ4YzlhNjc4ZDhjZDc5ZGQ0MDICgOV-ibaj994DAwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGAAAAVVBEQVRFDgAAAGtzd2FwLmV4Y2hhbmdlIAAAACvNLqkjDRI58v4xbqbQJN-YFuqwTim7Mq2x9LUm_ZJFAwAAAAAWAAAAa3N3YXAua3Blbm55Omtzd2FwLnh5egJw0jUOpI8jq0lWwksAAAAAAAAAAAAAAAAAAAAAAAAAAAKgAuLs2Q3nkQNZJQYAAAAAAAAAAAAAAAAAAAAAAAAAAA"
  },
  "algorithm": "Keccak_256",
  "object": "AAAAAAAAAAAAAAAA",
  "rootType": "blockEvents"
}
```

Parsing the proof in GHCI:

```Haskell
:set -XQuasiQuotes
:set -XScopedTypeVariables
import NeatInterpolation
import qualified Data.Text.Encoding as T
import Data.Aeson
import Chainweb.SPV.EventProof
import Chainweb.SPV.PayloadProof
a = [text|
{
  "subject": {
    "input": "ADAgAAAAQiX2nIzoq21HoA7Hiq2Rjkx5Nl7jD-l4nUGOjsz8aPAEAAAACAAAAFRSQU5TRkVSDAAAAGtzd2FwLmtwZW5ueSAAAADZU-C1BQTw6ghU2A09r-wtWZZdphpzWlmsvB3i3VjBQwMAAAAAQAAAADZmNTJjZTBkMWRmYzJmNzNlY2M4ZjE2MmFkODE1ZWZjMmM5NzY2YWI3NmIxYTQ4YzlhNjc4ZDhjZDc5ZGQ0MDIAKwAAAFRvWXViX1ZoeldHdWVEWVVwODBnVmVWTnZiVGdoaDA4dU9kVUJtMDZkcFkCAABAe6XwY4GWCgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAVFJBTlNGRVIJAAAAa3N3YXAueHl6IAAAAP3AV3e1ornnygmVlZcEYY61dPGffrku_zZpklQUCx75AwAAAABAAAAANmY1MmNlMGQxZGZjMmY3M2VjYzhmMTYyYWQ4MTVlZmMyYzk3NjZhYjc2YjFhNDhjOWE2NzhkOGNkNzlkZDQwMgArAAAAVG9ZdWJfVmh6V0d1ZURZVXA4MGdWZVZOdmJUZ2hoMDh1T2RVQm0wNmRwWQKAZ5fVFxAj49sAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAABNSU5UDAAAAGtzd2FwLnRva2VucyAAAAB6vJFqfTyW0fbYYSu1VSkKPDE8a-VkGeN67sH_V9xlJwMAAAAAFgAAAGtzd2FwLmtwZW5ueTprc3dhcC54eXoAQAAAADZmNTJjZTBkMWRmYzJmNzNlY2M4ZjE2MmFkODE1ZWZjMmM5NzY2YWI3NmIxYTQ4YzlhNjc4ZDhjZDc5ZGQ0MDICgOV-ibaj994DAwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGAAAAVVBEQVRFDgAAAGtzd2FwLmV4Y2hhbmdlIAAAACvNLqkjDRI58v4xbqbQJN-YFuqwTim7Mq2x9LUm_ZJFAwAAAAAWAAAAa3N3YXAua3Blbm55Omtzd2FwLnh5egJw0jUOpI8jq0lWwksAAAAAAAAAAAAAAAAAAAAAAAAAAAKgAuLs2Q3nkQNZJQYAAAAAAAAAAAAAAAAAAAAAAAAAAA"
  },
  "algorithm": "Keccak_256",
  "object": "AAAAAAAAAAAAAAAA",
  "rootType": "blockEvents"
}
|]
let Right (proof :: PayloadProof Keccak_256) = eitherDecodeStrict' (T.encodeUtf8 a)
(root, events) <- runEventsProof proof
print root
print events
```


src/Chainweb/Pact/RestAPI.hs
```diff
@@ -128,6 +136,26 @@ pactSpvApi
...
 
+-- -------------------------------------------------------------------------- --
+-- GET Pact SPV Proof
+
+--  | The chain endpoint is the target chain of the proof, i.e. the chain where
+--  the root of the proof is located.
+--
+type PactSpv2Api_
+    = "pact"
+    :> "spv2"
```
sirlensalot, line 147: Can we consider a different name here? It will be hard to justify to users why it's just called `spv2`, especially since that sounds like an update to the first version.

src/Chainweb/Utils.hs
```diff
@@ -407,6 +410,14 @@ alignWithV f a b = V.zipWith (\a' -> f . These a') a b <> case (V.length a,V.len
           | la > lb -> V.map (f . This) $ V.drop lb a
           | otherwise -> V.map (f . That) $ V.drop la b
 
+-- | Substraction that returns 0 when the second argument is larger than the
+-- first. This can be in particular usefull when substracting 'Natural' numbers.
+-- The operator '-' would throw an 'Underflow' exception in this situation.
+--
+minusOrNull :: Ord a => Num a => a -> a -> a
```
sirlensalot, line 417:
```suggestion
minusOrZero :: Ord a => Num a => a -> a -> a
```
?

src/Data/Singletons.hs
```diff
@@ -188,3 +211,69 @@ singThat p = withSing $ \x -> if p (fromSing x) then Just x else Nothing
...
+-- -------------------------------------------------------------------------- --
+-- Constraint Dictionaries
+
+data Dict :: Constraint -> Type -> Type where
+    Dict :: c => a -> Dict c a
+
+-- -------------------------------------------------------------------------- --
+-- Type Level Peano Numbers
+
+data N = Z | S N
```
sirlensalot, line 223: Surely we don't need to write this. Isn't there a Richard library somewhere for this, aka `singletons` or something? @emilypi

src/Data/Singletons.hs
```diff
@@ -188,3 +211,69 @@ singThat p = withSing $ \x -> if p (fromSing x) then Just x else Nothing
...
+type family Index (h :: Type) (l :: [Type]) :: N where
+    Index h (h ': _) = 'Z
+    Index h (_ ': t) = 'S (Index h t)
+    Index h '[] = TypeError
+        ( 'Text "Data.Singletons.Index: element of type "
+        ':<>: 'ShowType h
+        ':<>: 'Text " not found in list"
+        )
+
+-- data HList (l :: [Type]) where
```
sirlensalot, line 276: comments?

test/Chainweb/Test/Pact/PactInProcApi.hs
```diff
@@ -62,7 +62,7 @@ import Chainweb.Test.Pact.Utils
 import Chainweb.Test.Utils
 import Chainweb.Time
 import Chainweb.Transaction
-import Chainweb.Utils
+import Chainweb.Utils hiding (label)
```
sirlensalot, line 65: Hmm, where is the conflict coming from?

test/Chainweb/Test/Orphans/Pact.hs
```diff
@@ -0,0 +1,82 @@
...
+
+import Pact.Types.Command
+import Pact.Types.Runtime
+
+import Test.QuickCheck
+
+-- -------------------------------------------------------------------------- --
+--
+
+instance Arbitrary RequestKey where
```
sirlensalot, line 39: We should move these into Pact no?

src/Chainweb/Pact/RestAPI.hs
sirlensalot: Unless this _is_ actually the v2 (ie, v1 with more hash functions?)

src/Chainweb/Utils.hs
sirlensalot: Also "useful" not "usefull" in comment

src/Data/Singletons.hs
emilypi: I think these are already defined in `GHC.TypeNats` - is there a reason we redefined it here?

larskuhtz: The singletons package is highly experimental. It used to be bound to only the latest versions of GHC. The API broke regularly. It also used to be very big, since it included a lifted version of the complete base package.

I think, some of this may have changed with version 3.0 which was release only about a month ago (2021-03-12). I haven't yet gotten a chance to take a closer look. The code in this module comes from a draft for a micro-singletons package. I think singletons-3.0 may eliminate the need for such a package. For now, I think, it's too early to make a call on that. It would also be work for a separate PR.

larskuhtz: @emilypi GHC.TypeNats does not implement Peano numbers. It only provides type level literals. Unfortunately, the literals from that module can not be used for inductive proofs.

There are some packages on Hackage that work around that limitation by either providing an inductive representation on the fly or by directly injecting evidence for theorems over naturals into the type checker. I tried a few of those in the past, but some require type checker plugins, other didn't seem stable, and some were buggy.

In lack of a commonly used and stable library for type level Peano numbers, it seems the easiest to just add a minimal set of definitions that suffice for our use case here.

Specifically, there is the singletons-nat package. But that package used to come with its own singletons infrastructure hasn't been compatible with the definitions in this module. The most recent version of singletons-nat is base on the new minimal singletons-3.0 package, which is worth taking a closer look. But I propose that we first evaluate how stable singletons-3.0 is going to be and then possibly migrate to it in a separate PR.

src/Chainweb/Pact/RestAPI.hs
larskuhtz: This new SPV API should still be considered experimental. The goal is, that once it is finalized, we get a unified SPV proof format that can be used with bridges as well as internally for Chainweb.

The new SPV API is meant to be a generalization that subsumes and replaces the previous API. However, additional work has to be done such that output can be used with pact cross chain transactions.

larskuhtz (commented):
Not sure about the new of the new SPV API. I think, eventually, once we finalized the design we want it to become v2. Until, then we can keep the new and just not "publish" it officially or we could name it something like spvng or staging/spv and rename before we make it public. I tent towards the former.

src/Data/Singletons.hs
larskuhtz: The code in this module is mostly a minimal core of definitions from the singletons package at the time this module was written.

Yes, singletons-3.0 looks much leaner and mature than previous versions and I agree that we should consider taking it as a new dependency instead of this code.

The code we are talking here has been in chainweb since the very first version. So, let's keep it for a separate PR to update it. This PR is already too large.

src/Data/Singletons.hs
```diff
@@ -188,3 +211,65 @@ singThat p = withSing $ \x -> if p (fromSing x) then Just x else Nothing
...
+
+type family Index (h :: Type) (l :: [Type]) :: N where
+    Index h (h ': _) = 'Z
+    Index h (_ ': t) = 'S (Index h t)
+    Index h '[] = TypeError
+        ( 'Text "Data.Singletons.Index: element of type "
+        ':<>: 'ShowType h
+        ':<>: 'Text " not found in list"
+        )
+
```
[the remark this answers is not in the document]
larskuhtz: I like the proof erasure. But this code also scares me a bit. So far I haven't been very successful in using constraints package style abstractions productively.

Actually, I tried to get this PR working without Peano numbers using the `constraints` approach, but didn't succeed. Using the classic unary N approach was the most straight forward to make it work for me.

Does your code compile with all the compiler versions that we currently support? Also does it get the job done to resolve the MerkleLog Indexes?

Assuming that the current code is correct, I propose that we keep it for this PR and that we schedule some time and come up with a separate PR to improve the type machinery around MerkleLogs after the 2.7 release?

src/Chainweb/Pact/RestAPI.hs
```diff
@@ -128,6 +136,26 @@ pactSpvApi
     . Proxy (PactSpvApi v c)
 pactSpvApi = Proxy
 
+-- -------------------------------------------------------------------------- --
+-- GET Pact SPV Proof
```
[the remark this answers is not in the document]
larskuhtz: oh, that's true. The HTTP standard is a little unclear with respect to it, but generally GET should not have a request body (even though most implementations support it). I'll change the comment.

```suggestion
-- POST Query for Pact SPV Proof
```

src/Data/Singletons.hs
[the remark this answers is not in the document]
emilypi: I think you'll probably need `-XStandaloneKindSignatures` at the least, so 8.10 and above (are we there yet, or still on 8.8?)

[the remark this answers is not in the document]
larskuhtz: That's what we currently have:

```
tested-with:
    GHC == 8.10.4
    GHC == 8.8.4
    GHC == 8.6.5
```

nix-builds: 8.6.5
stack builds: 8.8.4
cabal builds: 8.10.4

Last time I checked, cryptonite was the only remaining blocker for ghc-9, if I remember correctly. Might be fixed by now. Not sure.

sirlensalot (approved):

PR #1214, opened 2021-03-30
drop deprecated random-bytestring dependency

larskuhtz:


src/Chainweb/Utils.hs
```diff
@@ -927,6 +930,30 @@ leadingZeros b =
...
+-- and life locks.
+--
+-- However, use of the generator is still sequentialized. Thus, for long
+-- 'ByteString's it can be more efficient to split the generator to speed up
+-- concurrent access.
+
+randomShortByteString :: MonadIO m => Natural -> m BS.ShortByteString
+randomShortByteString n
+    -- don't split the generators for less than 64 words.
+    | n < 8 * 64 = getStdRandom $ genShortByteString (int n)
```
emilypi, line 948: May as well just inline this and save a calculation. 

```suggestion
    -- 512 = 8 * 64
    | n < 512 = getStdRandom $ genShortByteString (int n)
```

src/Chainweb/Utils.hs
```diff
@@ -927,6 +930,30 @@ leadingZeros b =
...
+    | otherwise = fst . genShortByteString (int n) <$> newStdGen
+
+randomByteString :: MonadIO m => Natural -> m B.ByteString
+randomByteString n
+    -- don't split the generators for less than 64 words.
+    | n < 8 * 64 = getStdRandom $ genByteString (int n)
```
emilypi, line 954:
```suggestion
    -- 512 = 8 * 64
    | n < 512 = getStdRandom $ genByteString (int n)
```

emilypi (approved):
Optional suggestions. Otherwise 👍

src/Chainweb/Utils.hs
```diff
@@ -927,6 +930,30 @@ leadingZeros b =
     in out
 {-# INLINE leadingZeros #-}
 
+-- -------------------------------------------------------------------------- --
+-- Random ByteString
+--
+-- 'getStdRandom' provides an generator that is stored in an 'IORef' and updated
+-- via an optimistic atomci swap. 'atomicModifyIORef'' is implemented such that
+-- the swapped pointer is updated lazily, which minimizes the chance of retries
+-- and life locks.
```
larskuhtz, line 939:
```suggestion
-- 'getStdRandom' provides a generator that is stored in an 'IORef' and updated
-- via an optimistic atomic swap. 'atomicModifyIORef'' is implemented such that
-- the swapped pointer is updated lazily, which minimizes the chance of retries
-- and life locks.
```

PR #1217, opened 2021-04-16
Generate coin.TRANSFER events for gas, coinbase, crosschain

sirlensalot:
With this change all changes to `coin` accounts will be visible via `coin.TRANSFER` events. This is also the model for a new KIP to require `TRANSFER` events for burn and create by using an empty account `""` to indicate:

- Gas payment: transfer from `sender` to miner
- Coinbase: transfer from `""` to miner
- Crosschain step 1: transfer from `sender` to `""`
- Crosschain step 2: transfer from `""` to `receiver`

TODO investigate if allocation release can be detected for generating events.

FORKING CHANGE.

### DIFF
--- a/src/Chainweb/Version.hs
+++ b/src/Chainweb/Version.hs
@@ -864,7 +864,7 @@ pactGenerateEvents :: ChainwebVersion -> BlockHeight -> Bool
 pactGenerateEvents Mainnet01 = (>= 1_600_000) -- 2021-05-07T14:14:46
 pactGenerateEvents Testnet04 = (>= 1_140_000) -- 2021-05-06T13:47:27
 pactGenerateEvents Development = (>= 140)
-pactGenerateEvents (FastTimedCPM g) = const $ g == pairChainGraph || g == petersonChainGraph
+pactGenerateEvents (FastTimedCPM g) = const $ g == pairChainGraph
 pactGenerateEvents _ = const True
 
 -- -------------------------------------------------------------------------- --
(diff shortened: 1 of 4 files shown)


stack.yaml
```diff
@@ -61,11 +61,10 @@ extra-deps:
 
   # --- Custom Pins --- #
   - github: kadena-io/pact
-    commit: a9c9b92e313a680574479ffda4700533f19d16d5
+    commit: ccce78004d567966f5cfff131287fbdb3bf63121
```
larskuhtz, line 64: This pin must also be updated in `cabal.project` and `project.nix`.

src/Chainweb/Version.hs
```diff
@@ -858,6 +859,14 @@ enableSPVBridge Development = (>= 130)
 enableSPVBridge (FastTimedCPM g) = const $ g == pairChainGraph || g == petersonChainGraph
 enableSPVBridge _ = const True
 
+-- | Generate events for gas, coinbase, crosschain
+pactGenerateEvents :: ChainwebVersion -> BlockHeight -> Bool
+pactGenerateEvents Mainnet01 = (>= 1_600_000) -- 2021-05-07T14:14:46
+pactGenerateEvents Testnet04 = (>= 1_140_000) -- 2021-05-06T13:47:27
```
larskuhtz, line 865: 👍 Looks good. These there the estimates that I get base on current block heights:

mainnet: 2021-05-07 14:37:06
testnet: 2021-05-06 13:54:06

larskuhtz (changes requested):
The changes look good to me. I would have expected that some golden test results would need to be updated. Also, are there any unit tests for events? If not, it might be good to add those.

sirlensalot:
We will do this with #1218 instead.

PR #1219, opened 2021-04-19
Single-key account protocol

sirlensalot:


sirlensalot:
Merged into #1218 which is going to DRAFT for this release.

PR #1222, opened 2021-04-23
Avoid injecting the local peer into the peer db

larskuhtz:
This PR involves some long pending code cleanup in order to avoid cyclic dependencies. While touches a lot of code the changes are mostly straight forward.

* [x] cleanup use of `*Resources` datastructures. These data structure are meant as temporary collections of different resources during node initialization. They shouldn't be used after the services are started. Usage outside of initialization code can lead to dependency cycles and can also keep data alive that otherwise could be GCed.
* [x] start temporary P2P peer endpoints during initial reachability test of node.
* [x] avoid injection of the local peer into the peer db

### DIFF
--- a/src/Chainweb/RestAPI.hs
+++ b/src/Chainweb/RestAPI.hs
@@ -26,16 +26,20 @@
 --
 module Chainweb.RestAPI
 (
-  ChainwebServerDbs(..)
+-- * Utils
+  serveSocketTls
+
+-- * Chainweb Server DBs
+, ChainwebServerDbs(..)
 , emptyChainwebServerDbs
 
--- * Chainweb API
-, someChainwebApi
-, someChainwebPeerApi
-, someChainwebServiceApi
-, prettyShowChainwebApi
-, apiVersion
-, prettyApiVersion
+-- -- * Full Chainweb API (only used for generating swagger spec)
+-- , someChainwebApi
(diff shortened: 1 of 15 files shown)


src/Chainweb/RestAPI.hs
```diff
@@ -26,16 +26,20 @@
...
-, prettyShowChainwebApi
-, apiVersion
-, prettyApiVersion
+-- -- * Full Chainweb API (only used for generating swagger spec)
+-- , someChainwebApi
+-- , someChainwebPeerApi
+-- , someChainwebServiceApi
+-- , prettyShowChainwebApi
+-- , apiVersion
+-- , prettyApiVersion
```
larskuhtz, line 42:
```suggestion
```

src/Chainweb/RestAPI.hs
```diff
@@ -26,16 +26,20 @@
...
 , emptyChainwebServerDbs
 
--- * Chainweb API
-, someChainwebApi
-, someChainwebPeerApi
-, someChainwebServiceApi
-, prettyShowChainwebApi
-, apiVersion
-, prettyApiVersion
+-- -- * Full Chainweb API (only used for generating swagger spec)
```
sirlensalot: Commented out?

sirlensalot (approved):
Approved, small comment

PR #1223, opened 2021-04-24
fix mining newWork function

larskuhtz:
This fixes a condition in the newWork method that cause the mining coordinator to get stuck on an outdated primed work item.

### DIFF
--- a/src/Chainweb/Miner/Coordinator.hs
+++ b/src/Chainweb/Miner/Coordinator.hs
@@ -39,7 +39,6 @@ module Chainweb.Miner.Coordinator
 , newWork
 , publish
 , removePrimed
-, removeOutdatedPrimed
 ) where
 
 import Control.Concurrent.STM (atomically)
@@ -144,20 +143,6 @@ removePrimed
     -> IO ()
 removePrimed pwVar mid cid =
     atomically $ modifyTVar pwVar $ resetPrimed mid cid
-
-removeOutdatedPrimed
-    :: TVar PrimedWork
-    -> MinerId
-    -> ChainId
-    -> BlockPayloadHash
-    -> IO ()
-removeOutdatedPrimed pwVar mid cid phash = atomically $ do
-    PrimedWork pw <- readTVar pwVar
-    case join (HM.lookup mid pw >>= HM.lookup cid) of
-        Nothing -> return ()
-        Just (payload, _)
-            | _payloadDataPayloadHash payload == phash -> writeTVar pwVar $ resetPrimed mid cid (PrimedWork pw)
-            | otherwise -> return ()
 
 -- | Data shared between the mining threads represented by `newWork` and
 -- `publish`.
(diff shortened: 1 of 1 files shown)


src/Chainweb/Miner/Coordinator.hs
```diff
@@ -135,6 +137,28 @@ resetPrimed :: MinerId -> ChainId -> PrimedWork -> PrimedWork
...
+    -> BlockPayloadHash
+    -> IO ()
+removeOutdatedPrimed pwVar mid cid phash = atomically $ do
+    PrimedWork pw <- readTVar pwVar
+    case join (HM.lookup mid pw >>= HM.lookup cid) of
+        Nothing -> return ()
+        Just (payload, _)
+            | _payloadDataPayloadHash payload == phash -> writeTVar pwVar $ resetPrimed mid cid (PrimedWork pw)
+            | otherwise -> return ()
+
```
larskuhtz, line 161:
```suggestion
```

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -38,6 +38,8 @@ module Chainweb.Miner.Coordinator
 -- ** Internal Functions
 , newWork
 , publish
+, removePrimed
+, removeOutdatedPrimed
```
larskuhtz, line 42:
```suggestion
```

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -38,6 +38,7 @@ module Chainweb.Miner.Coordinator
 -- ** Internal Functions
 , newWork
 , publish
+, removePrimed
```
sirlensalot, line 41: Why export?

sirlensalot: Also, why export `newWork`? Clearly nothing broke by changing its signature.

src/Chainweb/Miner/Coordinator.hs
```diff
@@ -135,6 +136,14 @@ resetPrimed :: MinerId -> ChainId -> PrimedWork -> PrimedWork
 resetPrimed mid cid (PrimedWork pw) = PrimedWork
     $! HM.update (Just . HM.insert cid Nothing) mid pw
 
+removePrimed
+    :: TVar PrimedWork
+    -> MinerId
+    -> ChainId
+    -> IO ()
+removePrimed pwVar mid cid =
```
sirlensalot, line 144: Why is this top-level

src/Chainweb/Miner/Coordinator.hs
larskuhtz: Oh, I used `removePrimed` for some ad-hoc testing/debugging. Forgot to remove it.

```suggestion
```

larskuhtz: Not sure about `newWork`. It might be used in test code. I'll double check.

src/Chainweb/Miner/Coordinator.hs
larskuhtz: From an intermediate version and some testing. I'll move it back down.

sirlensalot (approved):

PR #1225, opened 2021-04-29
Do not log InsecureConnectionDenied Exceptions

larskuhtz:


emmanueldenloye (approved):
Unless there is some extra context that I'm missing, I'll say this is good!

PR #1226, opened 2021-04-30
P2P API endpoint to get node config

larskuhtz:


src/Chainweb/RestAPI/Config.hs
```diff
@@ -0,0 +1,64 @@
...
+-- GET config endpoing
+
+type GetConfigApi = "config" :> Get '[JSON] ChainwebConfiguration
+
+someGetConfigApi :: SomeApi
+someGetConfigApi = SomeApi (Proxy @GetConfigApi)
+
+someGetConfigServer :: ChainwebConfiguration -> SomeServer
+someGetConfigServer config = SomeServer (Proxy @GetConfigApi) $ return
+    -- hide sensible information
```
gregorycollins, line 48: My only comment here: it is safer to create a new object that exports a set of fields that you whitelist, rather than redacting a set of fields that are sensitive -- even at the cost of a lot of boilerplate in this section. If a new sensitive field is added elsewhere later, it might be leaked if the redaction in this section is overlooked.

gregorycollins (approved):

src/Chainweb/RestAPI/Config.hs
larskuhtz: True, but I think, the only truly sensitive field are the SSL keys. Everything else may, at worst, be exploited for more indirect attacks, like DOS attacks. So, I think, it doesn't matter that much in this case.

PR #1229, opened 2021-05-13
[mempool] remember high water marks across sync sessions

gregorycollins:
Currently, we forget a peer's mempool high water mark after a sync session ends. Keeping this info around will reduce round-trip traffic on subsequent sync sessions, because we can continue from the last sync point rather than retrieving everything again.


src/Chainweb/Mempool/Mempool.hs
```diff
@@ -428,62 +522,51 @@ syncMempools' log0 us localMempool remoteMempool = sync
     deb :: Text -> IO ()
     deb = log0 Debug
 
-    sync = finally (initialSync >>= subsequentSync) (deb "sync exiting")
+    sync = finally initialSync (deb "sync exiting")
 
     initialSync = do
-        deb "Get full list of pending hashes from remote"
+        mHistory <- fetchGlobalSyncHistory peer
```
emmanueldenloye, line 528: Given that you fetch peer sync history between the current mempool and a peer, is it possible for the actual highwater mark of the mempool to be different than `ourHwMark` that you get from the past? Is this a problem?

gregorycollins: No, because it's stored in RAM and the server nonce is changed at restart. Even if server nonces were different, the default behaviour is "return everything" on nonce mismatch.

PR #1230, opened 2021-05-14
[mempool] add hop counts to API calls.

gregorycollins:
1. add a hop count to internal mempool data structures and api calls (lookup and insert)
2. extend insert and lookup json to include # hops
3. unfortunately, the json for the insert call cannot be easily extended -- we need to maintain outbound call compatibility with older clients for a while, so the easiest fix is to add a new "gossip" call, which is insert + hop counts. Mempool p2p sync will continue to use the old insert call for now, although the plan shortly is to lookup the client version during sync and use `/gossip` where appropriate.


src/Chainweb/Mempool/Mempool.hs
```diff
@@ -145,23 +148,42 @@ import P2P.Peer (PeerInfo)
 
 ------------------------------------------------------------------------------
 data LookupResult t = Missing
-                    | Pending t
+                    | Pending { _lookupResultTx :: !t
```
emmanueldenloye, line 157: Maybe this record should just become another data type (perhaps `PendingResult`)

PR #1231, opened 2021-05-17
fix warp-tls on premature close

larskuhtz:


larskuhtz:
the issue is fixed in warp-tls-3.3.1

PR #1232, opened 2021-05-17
[WIP] Apply flag to disable pact 4.0

emmanueldenloye:


emmanueldenloye:
This is not the right way to approach this problem... I need to review another PR before going any further.

PR #1236, opened 2021-05-26
Index module init cache by block height

sirlensalot:
Implement a module init cache that avoids issues with globally-updated singleton cache.

Init cache is now an ordered map of (parent block height -> module cache). An invariant is that each cache is the union of the previous cache with any new definitions introduced at that height.

Caches are stored at the "adjusted parent block height" (APBH). At genesis, this is 0, but after, this is (parent block height + 1).

Caches are retrieved by finding the largest-height cache less than or equal to the parent block height at read time. Thus at block 5, an upgrade stored at 4 is visible but one stored at 5 is not.

This ensures that cache corruption during new block -> validate block is avoided. An upgrade at block 5 will, in new block, read an old cache and store the new one at 5. In validate block, the newly-stored value at 5 will not be retrieved ensuring that the old cache is again read, ensuring identical application.


src/Chainweb/Pact/Types.hs
```diff
@@ -370,15 +377,45 @@ defaultOnFatalError lf pex t = do
...
+    case M.lookupLE (pbh _psParentHeader) _psInitCache of
+      Just (_,mc) -> return mc
+      Nothing -> return mempty
+  where
+    pbh = _blockHeight . _parentHeader
+
+-- | Update init cache at adjusted parent block height (APBH), unioning contents with
+-- any cache found at or before APBH. APBH is modulated by first argument, in practice
+-- it is 'id' for genesis and 'succ' for thereafter.
+updateInitCache :: (BlockHeight -> BlockHeight) -> ModuleCache -> PactServiceM cas ()
```
emmanueldenloye, line 410: Do you think that it perhaps could be better for the first argument here to be a boolean (or some ADT) that makes its purpose easier to investigate?

src/Chainweb/Pact/Types.hs
```diff
@@ -370,15 +377,45 @@ defaultOnFatalError lf pex t = do
...
 
 
+_debugMC :: Text -> PactServiceM cas ()
+_debugMC t = do
+  mc <- fmap (fmap instr) <$> use psInitCache
+  liftIO $ print (t,mc)
+  where
+    instr (ModuleData{..},_) = preview (_MDModule . mHash) _mdModule
+
+-- | Look up an init cache that is stored before the height of the current parent header.
```
emmanueldenloye, line 398: Should this comment instead read as `before or at the height of the current parent header.`? (Since you're using `lookupLE`)

src/Chainweb/Pact/Types.hs
```diff
@@ -370,15 +377,45 @@ defaultOnFatalError lf pex t = do
...
+      Nothing -> return mempty
+  where
+    pbh = _blockHeight . _parentHeader
+
+-- | Update init cache at adjusted parent block height (APBH), unioning contents with
+-- any cache found at or before APBH. APBH is modulated by first argument, in practice
+-- it is 'id' for genesis and 'succ' for thereafter.
+updateInitCache :: (BlockHeight -> BlockHeight) -> ModuleCache -> PactServiceM cas ()
+updateInitCache bf mc = get >>= \PactServiceState{..} -> do
+    let pbh = bf . _blockHeight . _parentHeader $ _psParentHeader
```
larskuhtz, line 412: Can `bf` parameter be removed?

```suggestion
    let cur =  _blockHeight . _parentHeader $ _psParentHeader
        pbh = if (cur == 0) then 0 else cur + 1
```

emmanueldenloye (approved):
LGTM... There are just cosmetic changes that I'm suggesting. Nothing that should prevent a merge IMO.

test/Chainweb/Test/Pact/ModuleCacheOnRestart.hs
```diff
@@ -112,16 +113,17 @@ testCoinbase iobdb = (initPayloadState >> doCoinbase,snapshotCache)
       void $ execValidateBlock mempty nextH (payloadWithOutputsToPayloadData pwo)
 
 -- | Interfaces can't be upgraded, but modules can, so verify hash in that case.
-justModuleHashes :: ModuleCache -> HM.HashMap ModuleName (Maybe ModuleHash)
-justModuleHashes = HM.map $ \v -> preview (_1 . mdModule . _MDModule . mHash) v
+justModuleHashes :: ModuleInitCache -> HM.HashMap ModuleName (Maybe ModuleHash)
+justModuleHashes = upd. snd . last . M.toList where
```
larskuhtz, line 117: I think, future versions of GHC will require white space around function composition `.` (not sure if on both sides, though).

```suggestion
justModuleHashes = upd . snd . last . M.toList where
```

larskuhtz (approved):
Just two suggestions.

src/Chainweb/Pact/Types.hs
larskuhtz: Why not just drop it? It seems that it's always `id` when the parent block height is 0 and `succ` otherwise.

PR #1244, opened 2021-06-18
Use exponential distribution for P2P session timeouts

larskuhtz:
Along the way, fix implementation of `timeSpanToMicros` and add roundtrip tests for time span conversions.

### DIFF
--- a/src/Chainweb/Time.hs
+++ b/src/Chainweb/Time.hs
@@ -326,10 +326,8 @@ secondsToTimeSpan :: Num a => Seconds -> TimeSpan a
 secondsToTimeSpan (Seconds s) = scaleTimeSpan s second
 {-# INLINE secondsToTimeSpan #-}
 
--- | Assumes that the `TimeSpan` contains milliseconds.
---
 timeSpanToSeconds :: Integral a => TimeSpan a -> Seconds
-timeSpanToSeconds (TimeSpan ms) = Seconds . int $ ms `div` 1000000
+timeSpanToSeconds (TimeSpan us) = Seconds . int $ us `div` 1000000
 {-# INLINE timeSpanToSeconds #-}
 
 secondsToText :: Seconds -> T.Text
@@ -361,10 +359,8 @@ microsToTimeSpan :: Num a => Micros -> TimeSpan a
 microsToTimeSpan (Micros us) = scaleTimeSpan us microsecond
 {-# INLINE microsToTimeSpan #-}
 
--- | Assumes that the `TimeSpan` contains milliseconds.
---
 timeSpanToMicros :: Integral a => TimeSpan a -> Micros
-timeSpanToMicros (TimeSpan ms) = Micros . int $ ms * 1000
+timeSpanToMicros (TimeSpan us) = Micros . int $ us
 {-# INLINE timeSpanToMicros #-}
 
 microsToText :: Micros -> T.Text
--- a/src/P2P/Node.hs
+++ b/src/P2P/Node.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiWayIf #-}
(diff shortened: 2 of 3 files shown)


src/P2P/Node.hs
```diff
@@ -585,26 +588,40 @@ newSession conf node = do
...
+-- within the range of a a tenth of the expectation and ten times the
+-- expectation.
+--
+-- The expected value of the actual distribution gets increasinly imprecise
+-- as the input value gets smaller, because a minimum result of 5 seconds
+-- is implemented. Input values of less than 30s should be avoided.
+--
+sessionTimeout :: TimeSpan Int -> IO (TimeSpan Int)
+sessionTimeout expected = do
+    g <- MWC.createSystemRandom
```
larskuhtz, line 617: Seeding the generator on each call from system entropy is inefficient.

src/Chainweb/Time.hs
```diff
@@ -326,10 +326,8 @@ secondsToTimeSpan :: Num a => Seconds -> TimeSpan a
 secondsToTimeSpan (Seconds s) = scaleTimeSpan s second
 {-# INLINE secondsToTimeSpan #-}
 
--- | Assumes that the `TimeSpan` contains milliseconds.
---
 timeSpanToSeconds :: Integral a => TimeSpan a -> Seconds
-timeSpanToSeconds (TimeSpan ms) = Seconds . int $ ms `div` 1000000
+timeSpanToSeconds (TimeSpan us) = Seconds . int $ us `div` 1000000
```
gregorycollins, line 330: While you're here -- maybe make these functions strict?

src/P2P/Node.hs
```diff
@@ -585,26 +587,43 @@ newSession conf node = do
...
+-- within the range of a a tenth of the expectation and ten times the
+-- expectation.
+--
+-- The expected value of the actual distribution gets increasinly imprecise
+-- as the input value gets smaller, because a minimum result of 5 seconds
+-- is implemented. Input values of less than 30s should be avoided.
+--
+sessionTimeout :: TimeSpan Int -> IO (TimeSpan Int)
+sessionTimeout expected = do
+    x <- exponential (1 / us) :: IO Double
```
gregorycollins, line 616: Use `System.Random.MWC` instead, it's better in basically every way -- including already shipping with a truncated exponential distribution. I used it for exactly this in https://github.com/kadena-io/chainweb-node/pull/1254/files, you could crib the code from there

gregorycollins (approved):
In general I obviously approve -- see my suggestion re swapping System.Random out in preference for mwc-random (way faster)

src/Chainweb/Time.hs
larskuhtz, line 330:
```suggestion
timeSpanToSeconds (TimeSpan us) = Seconds $! int $ us `div` 1000000
```

src/Chainweb/Time.hs
```diff
@@ -361,10 +359,8 @@ microsToTimeSpan :: Num a => Micros -> TimeSpan a
 microsToTimeSpan (Micros us) = scaleTimeSpan us microsecond
 {-# INLINE microsToTimeSpan #-}
 
--- | Assumes that the `TimeSpan` contains milliseconds.
---
 timeSpanToMicros :: Integral a => TimeSpan a -> Micros
-timeSpanToMicros (TimeSpan ms) = Micros . int $ ms * 1000
+timeSpanToMicros (TimeSpan us) = Micros . int $ us
```
larskuhtz, line 363:
```suggestion
timeSpanToMicros (TimeSpan us) = Micros $! int $ us
```

src/P2P/Node.hs
larskuhtz: I'd like to use it. However, I run into compatibility issues with our rather old nix snapshot. I don't remember the details. However, this implementation should be the same than what is used in mac-random. Also, I think, modern versions of the random package have much better performance than previous versions.

gregorycollins: Lars -- mwc-random was already in the deps list?

PR #1252, opened 2021-06-24
drop dependency on streaming-concurrency

larskuhtz:
streaming-concurrency depends on streaming-with. Both packages are not actively maintained and the latter is broken for GHC-9.

chainweb-node makes use of streaming-concurrency in a single location in cwtool and it is not clear if its use provides any benefit over and sequential stream. Local tests suggests that there is no performance improvement. So, it makes sense to drop this dependency.


tools/header-dump/HeaderDump.hs
```diff
@@ -647,22 +630,47 @@ withChainDbsConcurrent
...
+--     go (cid:t) !s = withBlockHeaderDb rdb v cid $ \cdb ->
+--         entries cdb Nothing Nothing start end $ \x ->
+--             S.withBuffer buffer (S.writeStreamBasket x) $ \out ->
+--                 S.withStreamBasket out $ \y ->
+--                     go t (() <$ S.mergeOn _blockHeight s (y & val))
+--
+--     buffer = S.bounded 4000
+--
+--     val = if doValidation then validate else id
+
```
larskuhtz, line 673:
```suggestion
```

tools/header-dump/HeaderDump.hs
```diff
@@ -374,7 +372,7 @@ withBlockHeaders logger config inner = do
         initializePayloadDb v pdb
         liftIO $ logg Info "start traversing block headers"
         liftIO $ logg Info $ "header validation: " <> sshow (_configValidate config)
-        withChainDbsConcurrent rdb v cids (_configValidate config) start end $ inner pdb . void
+        withChainDbs rdb v cids (_configValidate config) start end $ inner pdb . void
```
sirlensalot, line 375: So is this single threaded now?

sirlensalot (approved):

tools/header-dump/HeaderDump.hs
larskuhtz: Yes, but performance seemed still fine. Well, actually performance never really was that great :-) I think, this has bottlenecks elsewhere.

Also, I think, I am the only one using this code in personal scripts.

PR #1258, opened 2021-07-26
Increase batch size of payloads to 1000

larskuhtz:
* [x] Increases batch size for payloads to 1000
* [x] limit batch size for payloads with outputs to 1000


emmanueldenloye (approved):
LGTM

PR #1264, opened 2021-08-11
avoid rounding errors when reading miner rewards file

larskuhtz:
**IMPORTANT: Only merge after completing full mainnet history validation**

* [x] Mainnet history validation

### DIFF
--- a/src/Chainweb/Miner/Pact.hs
+++ b/src/Chainweb/Miner/Pact.hs
@@ -43,7 +43,6 @@ import Control.Monad.Catch (MonadThrow)
 
 import Data.Aeson hiding (decode)
 import Data.ByteString (ByteString)
-import qualified Data.ByteString.Char8 as B8
 import qualified Data.ByteString.Lazy as BL
 import qualified Data.Csv as CSV
 import Data.Decimal (Decimal)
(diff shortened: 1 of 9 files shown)


src/Chainweb/Miner/Pact.hs
```diff
@@ -159,8 +162,14 @@ readRewards =
...
-    formatRow :: (Word64, Double) -> (BlockHeight, Decimal)
-    formatRow (!a,!b) = (BlockHeight $ int a,  fromRational $ toRational b)
+    formatRow :: (Word64, CsvDecimal) -> (BlockHeight, Decimal)
+    formatRow (!a,!b) = (BlockHeight $ int a, _csvDecimal b)
+
+newtype CsvDecimal = CsvDecimal { _csvDecimal :: Decimal }
+    deriving newtype (Show, Read)
+
+instance CSV.FromField CsvDecimal where
+    parseField = either fail pure . readEither . B8.unpack
```
mightybyte, line 172: In my experience it's usually a bad idea to use `Data.ByteString.Char8` due to incorrect edge cases with Unicode.

larskuhtz: I think, it should be fine here. The input is ASCII and `readEither` should fail if it isn't.

We could interpret the input as `Text` and do an Utf8 encoding roundtrip to filter out any nasty code points or deal with non-ASCII decimal digit representation, but that seems overkill to me here.

mightybyte (approved):
Assuming this doesn't break any block serializations, looks good to me.

PR #1265, opened 2021-08-11
attempt to make remote pact tests less flaky

larskuhtz:


mightybyte (approved):

sirlensalot (approved):

larskuhtz:
hmm, so it seems that the test failure is not just a timeout issue:

```
Chainweb Tests
  Chainweb-Pact Tests
    Chainweb.Test.Pact.RemotePactTest
      remote pact tests
        local continuation test:                                                                             FAIL (55.73s)
          execute /send with initial pact continuation tx           (0.09s)
          check /poll responses to extract pact id for continuation (55.64s)
            PollingFailure "polling check failed after 200 attempts: {}"
        /poll rejects keys of incorrect length:                                                              SKIP
```

larskuhtz:
I have been running this test for over an hour in a loop on my laptop and I didn't fail a single time. So, the issue is probably some race that is related to the restricted resources on the Github runner environment.

larskuhtz:
Another datapoint is that in all cases where the test fails, there is appears a warning similar to the following right before the failure:

```
"[2021-08-11 20:10:43.859414733 UTC][Warn][node:0,peerId:9LkpIG,port:38797,host:localhost,component:cut] Timeout while processing cut BFOsag at height 190 from unknown origin"
"[2021-08-11 20:10:43.868570747 UTC][Warn][node:0,peerId:9LkpIG,port:38797,host:localhost,component:cut] cut BFOsag: failed to get prerequesites for some blocks. Missing: {}"
```

larskuhtz:
It seems that the issue occurs only with GHC-8.8.4. Since we are moving away from the version, I put this PR on hold until we we are able to observe the issue with a more recent version of GHC.

PR #1282, opened 2021-10-15
[mempool] configurable min gas price

emmanueldenloye:


larskuhtz (approved):
Looks good to me. Just to double check: pact/send would reject a submitted tx right away if it doesn't meet the min gas price limit?

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -363,6 +365,7 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configThrottling = defaultThrottlingConfig
     , _configMempoolP2p = defaultEnableConfig defaultMempoolP2pConfig
     , _configBlockGasLimit = 150000
+    , _configMinGasPrice = 1e-12
```
larskuhtz, line 368:
```suggestion
    , _configMinGasPrice = 1e-10
```

PR #1284, opened 2021-10-26
[PactService] add telemetry to PactQueue

emmanueldenloye:


src/Chainweb/Pact/Service/PactQueue.hs
```diff
@@ -12,37 +18,83 @@
...
 newPactQueue sz = do
-  _pactQueueValidateBlock <- newTBQueue sz
-  _pactQueueNewBlock <- newTBQueue sz
-  _pactQueueOtherMsg <- newTBQueue sz
+  (_pactQueueValidateBlock, _pactQueueNewBlock, _pactQueueOtherMsg) <-
+    atomically $ do
+      v <- newTBQueue sz
+      n <- newTBQueue sz
+      o <- newTBQueue sz
+      return (v,n,o)
```
edmundnoble, line 69: Use `newTBQueueIO` instead?

src/Chainweb/Pact/Service/PactQueue.hs
```diff
@@ -51,11 +103,102 @@ addRequest q msg =  atomically $ writeTBQueue priority msg
...
+      , ())
+
+resetPactQueueStats :: PactQueue -> IO ()
+resetPactQueueStats q = do
+  resetPactQueueStats' (_pactQueuePactQueueValidateBlockMsgStats q)
+  resetPactQueueStats' (_pactQueuePactQueueNewBlockMsgStats q)
+  resetPactQueueStats' (_pactQueuePactQueueOtherMsgStats q)
+
+resetPactQueueStats' :: PactQueueStats -> IO ()
+resetPactQueueStats' stats = atomicModifyIORef' (_pactQueueStatsCounters stats) (const (initPactQueueCounters, ()))
```
edmundnoble, line 175: `atomicModifyIORef r (const (x, ()))` is the same as `atomicWriteIORef r x`.

edmundnoble (approved):

src/Chainweb/Pact/Service/PactQueue.hs
edmundnoble: As discussed offline, it's the same except that it's not strict in `x`, which is fine here.

src/Chainweb/Pact/Service/PactInProcApi.hs
```diff
@@ -89,15 +89,34 @@ withPactService'
...
+        (validateblock_stats, newblock_stats, other_stats) <- getPactQueueStats pq
+        logFunctionText l Debug "got validateBlock stats"
+        logFunctionJson l Info validateblock_stats
+        logFunctionText l Debug "logged validateBlock stats"
+        logFunctionText l Debug "got newBlock stats"
+        logFunctionJson l Info newblock_stats
+        logFunctionText l Debug "logged newBlock stats"
+        logFunctionText l Debug "got otherMsg stats"
+        logFunctionJson l Info other_stats
+        logFunctionText l Debug "logged otherMsg stats"
```
larskuhtz, line 117: I wonder if those `Debug` logs are a bit too excessive. Maybe just remove those? Or just condense all of them into a single `Debug` message?

src/Chainweb/Pact/Service/PactQueue.hs
```diff
@@ -12,37 +18,79 @@
...
+        , _pactQueueStatsCounters = counters
+        }
+  _pactQueuePactQueueOtherMsgStats <- do
+      counters <- newIORef initPactQueueCounters
+      return PactQueueStats
+        {
+          _pactQueueStatsQueueName = "OtherMsg"
+        , _pactQueueStatsCounters = counters
+        }
   return PactQueue {..}
```
larskuhtz, line 88: Please avoid `RecordWildCards`. If you want to avoid an helper variable, you may consider applicative functor notation (`<$>` and `<*>`). Otherwise just being explicit is fine, too.

larskuhtz, line 88: What about the following?

```suggestion
    PactQueue
        <$> newTBQueueIO sz -- _pactQueueValidateBlock
        <*> newTBQueueIO sz -- _pactQueueNewBlock
        <*> newTBQueueIO sz -- _pactQueueOtherMsg
        <*> PactQueueStats "ValidateBlockMsg" <$> newIORef initPactQueueCounters
        <*> PactQueueStats "NewBlockMsg" <$> newIORef initPactQueueCounters
        <*> PactQueueStats "OtherMsg" <$> newIORef initPactQueueCounters
```

Applicative functor syntax has the disadvantage that it doesn't include the record filed names (thus the comments), but overall it's more readable and maintainable, I think.

I've been burned by `RecordWildCards` more than once...

src/Chainweb/Pact/Service/PactQueue.hs
```diff
@@ -51,11 +99,102 @@ addRequest q msg =  atomically $ writeTBQueue priority msg
...
+    _pactQueueCountersCount :: {-# UNPACK #-} !Int
+  , _pactQueueCountersSum   :: {-# UNPACK #-} !Micros
+  , _pactQueueCountersMin   :: {-# UNPACK #-} !Micros
+  , _pactQueueCountersMax   :: {-# UNPACK #-} !Micros
+  } deriving (Show, Generic)
+    deriving anyclass NFData
+
+
+instance ToJSON PactQueueCounters where
+  toJSON (PactQueueCounters {..}) =
```
larskuhtz, line 147: Could you also add an `toEncoding` implementation? You may use the "properties-pattern" that is used for other `ToJSON` instances, too. (This is a new thing and wasn't around when this PR was created)

src/Chainweb/Pact/Service/PactQueue.hs
```diff
@@ -51,11 +99,102 @@ addRequest q msg =  atomically $ writeTBQueue priority msg
...
+  resetPactQueueStats' (_pactQueuePactQueueValidateBlockMsgStats q)
+  resetPactQueueStats' (_pactQueuePactQueueNewBlockMsgStats q)
+  resetPactQueueStats' (_pactQueuePactQueueOtherMsgStats q)
+
+resetPactQueueStats' :: PactQueueStats -> IO ()
+resetPactQueueStats' stats = atomicWriteIORef (_pactQueueStatsCounters stats) initPactQueueCounters
+
+getPactQueueStats :: PactQueue -> IO (Value, Value, Value)
+getPactQueueStats = getPactQueueStats' >=> \case
+    (vstats',nbstats',ostats') -> do
```
larskuhtz, line 184: Isn't this `case` redundant?

```suggestion
getPactQueueStats = getPactQueueStats' >=> \(vstats',nbstats',ostats') -> do
```

src/Chainweb/Pact/Service/PactQueue.hs
```diff
@@ -51,11 +99,102 @@ addRequest q msg =  atomically $ writeTBQueue priority msg
...
+      Time timestamp  <- getCurrentTimeIntegral @Micros
+      let
+        vstats  = withObject' (toJSON vstats') ("msg_type" .= ("validate" :: T.Text) <> "timestamp" .= timestamp)
+        nbstats = withObject' (toJSON nbstats') ("msg_type" .= ("newblock" :: T.Text) <> "timestamp" .= timestamp)
+        ostats  = withObject' (toJSON ostats') ("msg_type" .= ("other" :: T.Text) <> "timestamp" .= timestamp)
+      return (vstats, nbstats, ostats)
+  where
+    withObject' v o' = case v of
+      Object o -> Object $ o' <> o
+      _ -> error "getPactQueueStats: impossible"
```
larskuhtz, line 198: You may add a `HasCallStack` constraint to any function that calls `error`.

src/Chainweb/Pact/Service/PactQueue.hs
```diff
@@ -51,11 +99,102 @@ addRequest q msg =  atomically $ writeTBQueue priority msg
...
+
+resetPactQueueStats' :: PactQueueStats -> IO ()
+resetPactQueueStats' stats = atomicWriteIORef (_pactQueueStatsCounters stats) initPactQueueCounters
+
+getPactQueueStats :: PactQueue -> IO (Value, Value, Value)
+getPactQueueStats = getPactQueueStats' >=> \case
+    (vstats',nbstats',ostats') -> do
+      Time timestamp  <- getCurrentTimeIntegral @Micros
+      let
+        vstats  = withObject' (toJSON vstats') ("msg_type" .= ("validate" :: T.Text) <> "timestamp" .= timestamp)
```
larskuhtz, line 187: Why not using an `ToJSON` instance for `PactQueueStats`?

Also, I am not sure whether the timestamp is needed, since each log message receives a timestamp when it is created.

src/Chainweb/Pact/Service/PactInProcApi.hs
```diff
@@ -89,15 +89,34 @@ withPactService'
     -> (PactQueue -> IO a)
     -> IO a
 withPactService' ver cid logger memPoolAccess bhDb pdb sqlenv config action = do
-    reqQ <- atomically $ newPactQueue (_pactQueueSize config)
-    race (server reqQ) (client reqQ) >>= \case
+    reqQ <- newPactQueue (_pactQueueSize config)
+    race (concurrently_ (monitor reqQ) (server reqQ)) (client reqQ) >>= \case
         Left () -> error "pact service terminated unexpectedly"
```
larskuhtz, line 94: Please add a `HasCallStack` constraint to this function, which would help to trace the error in case it triggers. Also, please prefix the error message with the Module and function name.

larskuhtz (changes requested):
The chainweb codebase uses 4 characters for indentation. We should keep that consistent, because otherwise editors may get confused. Inconsistent indenting can also make the code harder to read lead to merge conflicts.

We can talk about change the indentation (I think the pact code base uses 2). But if we do it, we should do it globally (in a separate PR, that would touch almost any single line of code in the project...)

larskuhtz (approved):

PR #1287, opened 2021-11-02
Enforce keyset formats

sirlensalot:
Enables pact flag to enforce valid keys in `read-keyset`. Before the switch any key value used in `read-keyset` will succeed; after, only valid Pact public key values will work.

Development block height is 100.


larskuhtz (approved):
Looks good. Haven't yet checked the block heights, which I'll do tomorrow.

emmanueldenloye (approved):
LGTM

PR #1299, opened 2021-11-14
Remove stack support

larskuhtz:


mightybyte (commented):
LGTM, but @sirlensalot should also take a look.

PR #1300, opened 2021-11-14
logging and throttling

larskuhtz:
* [x] remove throttling from service API endpoints (those should be made public only via a reverse proxy)
* [x] Add support for rate based log message filtering

An alternative implementation of could use a proper token-bucket based rate limiter, but that seems overkill for the logging system. The approach in this PR is much simpler.


src/Utils/Logging.hs
```diff
@@ -252,90 +263,196 @@ genericLogHandle f b msg = case fromBackendLogMessage msg of
...
+--             Just level -> Just (All $ level >= msgLevel)
+--             Nothing -> Nothing
+--     case mconcat $ apply <$> L._logMsgScope msg of
+--         Nothing
+--             | _logFilterDefaultLevel sf >= msgLevel -> return (Just msg)
+--             | otherwise -> return Nothing
+--         Just (All True) -> return (Just msg)
+--         Just (All False) -> return Nothing
+-- {-# INLINEABLE logFilterHandle_ #-}
+
```
larskuhtz, line 399:
```suggestion
```

src/Utils/Logging.hs
```diff
@@ -252,90 +263,183 @@ genericLogHandle f b msg = case fromBackendLogMessage msg of
...
-      where
-        f = withObject "LogRule" $ \o -> (\x y z -> ((x,y),z))
-            <$> o .: "key"
-            <*> o .: "value"
-            <*> o .: "level"
+        <*> o .:? "default-rate" .!= Probability 1
     {-# INLINE parseJSON #-}
 
+logRuleRng :: IORef Prob.Seed
+logRuleRng = unsafePerformIO (Prob.createSystemSeed >>= newIORef)
```
sirlensalot, line 387: Why can't this simply be in `IO`, since it is only used in one place, in IO?

larskuhtz: `Prob.createSystemSeed` is an expensive operation, in particular when performed on high rate, because it draws entropy from the systems random source.

`logRuleRng` initializes a global `RNG` with a random seed from the system entropy and reuses it after that.

I guess, the RNG could also be initialized higher up in the code and passed down to the use site. It seemed easier to keep things local.

sirlensalot: How does this code prevent duplicate calls to `createSystemSeed`?

larskuhtz: Since it is a pure top-level function, it is evaluated only once into a globally shared heap reference (CAF). The `NOINLINE` pragma prevents the code from being inlined and evaluated multiple times.

larskuhtz: `logRuleRng` is a pure top-level function (CAF), which get evaluated only once into a heap reference. `NOINLINE` prevents the code from getting evaluated repeatedly through inlining.

sirlensalot: Sounds fine then. Please document the function to this effect, including the importance of NOINLINE

sirlensalot (approved):

PR #1302, opened 2021-11-16
implement toEncoding in JSON instances

larskuhtz:
* [ ] run mainnet db validation


sirlensalot (commented):
I think this needs some kind of solution. Creating duplicate encoders and bespoke `xxxProperties` is too hard to justify and maintain, as it all seems to come from warts in `aeson` that we shouldn't require maintainers to have expert-level knowledge of.

I propose we design a solution and adapt it to `aeson`. An example is https://github.com/kadena-io/pact/blob/3b221d5a9efb0f61fa2e63a88a766258991b2f29/src/Pact/Types/Codec.hs#L59-L62 , which has the advantage of closing over local definitions for field names to be reused in decoders. I've never liked the decoupling of `ToJSON`/`FromJSON` ...

This would forge a path to eventually drop `aeson`.

larskuhtz:
This PR isn't about the separation between encoders and decoders, but only about adding `toEncoding` implementations to the existing encoders. If we want to do something about the separation of encoder and decoders, I propose that we tackle that in a different PR.

I think, there are currently not good practical alternatives to writing encoders and decoders separately.

There are grammar frameworks on Hackage for generating encoders and decoders from the same code. However, I don't think there's anything that has broader adaption.

An alternative is code generation (either TH or generic) directly from the types. However, that effectively require to duplicate all types by adding a "representation" version of the type. That would allow to cover most types but would fail in situations when there is not trivial mapping between structure of the Haskell AST and the encoded type. This would probably include all types with higher order parameters.

larskuhtz:
While it is possible that the code for `toEncoding` and `toJSON` could get of of sync, I found that this is rarely happens in practice for the following reasons:

1. These instances have very good test coverage. There are QuickCheck based generic roundtrip tests that are very sensitive to inconsistencies.

2. `ToJSON` and `FromJSON` instances are usually highly idiomatic and follow obvious patterns. They are verbose but a the same time easy to write and review.

sirlensalot:
> These instances have very good test coverage.

Do the quickchecks test both `toJSON` and `toEncoding`?

sirlensalot:
> I think, there are currently not good practical alternatives to writing encoders and decoders separately.

OK we can leave that discussion and focus on the "ToXXX" side for this PR.

sirlensalot:
> There are grammar frameworks on Hackage for generating encoders and decoders from the same code. However, I don't think there's anything that has broader adaption.

> An alternative is code generation (either TH or generic) directly from the types. 

Agreed and not looking for a templated/generic solution here, explicit code is better for codecs.

sirlensalot:
I'm looking for something principled. How about a simple ADT + an `EncodeJSON` typeclass?

```haskell
data Encoded = 
  EObject :: [(Text,Encoded)] -- can define our own '.=' analogue
  EText :: Text
  ENumber :: Scientific
  EBool :: Bool
  EArray :: Vector Encoded
  ENull :: () -- or just omit Null if we can
  
class EncodeJSON a where
  encodeJSON :: a -> Encoded

instance ToJSON Encoded where
  toJSON = \case
    EObject kvs = object kvs
    EText t = String t
    ...
  toEncoding = \case
    EObject kvs = pairs $ mconcat kvs
    EText t = toEncoding t
```

This has the big advantage of ensuring all contained objects have an `EncodeJSON` instance; in the current state only vigilance can ensure a given `ToJSON` implements `toEncoding` (and that it does so according to our standards)

I don't think the extra wrapper will impact GC significantly.

larskuhtz:
I think, this adds just another layer of an generic intermediate representation on top of Aeson's `Value` type.

The difference is that it puts encoder and decoder into the same class, which enforces that one has to provide both or none. But this is effectively the same as introducing a new constraint `type ToFromJSON a = (ToJSON a, FromJSON a)`. The other benefit is that it enforces the existence of `toEncoding`, but the my concern is that it eliminates the benefits by still building an intermediate AST.

Beside that it has the same drawbacks of the current use of an intermediate data. In particular, I think, that it would eliminate the performance benefits for `toEncoding`.

larskuhtz:
I am not completely sure how big the problem is that we are trying to solve with a new wrapper around Aeson instances.

I don't think those instances have caused issues in the past except for 1. usage of generated generic instances (which wouldn't be captures by a new wrapper) and 2. use of JSON in for hashed on-chain data, which wouldn't be solved neither.

Aeson instances are a bit verbose, but they are also very idiomatic and thus relatively easy to maintain.

larskuhtz:
I think the `*Properties` functions serve a similar rule like the `Encoded` class, without introducing a data type and tagging intermediate values with constructors. While I am not 100% sure,  I think, that GHC would still be possible to inline and fully eliminate it, since it is just a list fold.

sirlensalot:
> I think, this adds just another layer of an generic intermediate representation on top of Aeson's `Value` type.

It is not generic and there is a specific change which you seem to have missed. Please study the example further.

> 
> The difference is that it puts encoder and decoder into the same class, which enforces that one has to provide both or none. But this is effectively the same as introducing a new constraint `type ToFromJSON a = (ToJSON a, FromJSON a)`. The other benefit is that it enforces the existence of `toEncoding` ...

Again please actually look at the code. This does nothing on the decoding side and I explicitly conceded that argument earlier. 

> , but the my concern is that it eliminates the benefits by still building an intermediate AST.
> Beside that it has the same drawbacks of the current use of an intermediate data. In particular, I think, that it would eliminate the performance benefits for `toEncoding`.

Please explain how a ad-hoc "properties" function that returns a new list is so dramatically worse than the exact same list in a constructor.

sirlensalot:
> I think the `*Properties` functions serve a similar rule like the `Encoded` class, without introducing a data type and tagging intermediate values with constructors. While I am not 100% sure, I think, that GHC would be able to inline and fully eliminate it, since it is just a list fold.

Presumably the same optimization can be applied to the datatype. 

One thing you missed about this design is it elides `Value` for `toEncoding`, so any concerns about it "duplicating" `Value` are kind of not the point. `Value` has an obvious failing which this fixes so it is never referenced.

larskuhtz:
Sorry, I missed that is proposal would only cover the encoding side.

larskuhtz:
> Please explain how a ad-hoc "properties" function that returns a new list is so dramatically worse than the exact same list in a constructor.

Encoding of a Haskell value `a` to JSON via `Encoded` would be done as `encode (encodeJSON a)`. So, `a` would be first recursively converted to an `Encoded` value and after that `toEncoding` (or `toJSON`) from Aeson would be applied to serialize the value to a `Builder` (or `Value` in case of `toJSON`).

Since all constructors of `Encoded` are applied lazily, I think, there is some chance that GHC is smart enough optimize away the constructor application and apply the builder functions directly without creation of heap objects. But I am not sure about that. I am also not sure if there would be any benefit from inlining `encodeJSON`. There might be if the optimizer can resolve the pattern match on the constructor tags statically.  

I think, most Haskell frameworks for efficient streaming, rely sophisticated `RULE` pragmas to eliminate the constructors and the related heap objects (and those `RULE`s are known for being unreliable). `Encoding` and `Builder` are  just newtype wrappers around the continuation that serializes the data to memory.

In case of `*Properties` I am relatively certain that GHC eliminates the lazy list constructor and turns it into a generator function. There is also a good chance that inlining of the function together with `mconcat` enables further optimizations from the `ByteString.Builder`.

larskuhtz:
I think, the best way to find out the performance of the approaches would be to either profile or to build a small example and compile it to core.

sirlensalot:
Why can't we just make `toJSON` `error`? That would solve the duplication problem.

larskuhtz:
We use toJSON in some situation when converting to Value. So, we would have to try out when that's the case.

It could also trigger if someone in the future implements a `ToJSON` instance for a type that has a type that errors on `toJSON` somewhere as inner type. When calling `toJSON` on the outer type the `error` would trigger. Though, I guess that the error might be intended.

Currently, I didn't implement `toEncoding` for those large configuration types that are only parsed at startup. Those would fail because they would call toJSON on inner types. But, it wouldn't be too much work to add `toEncoding` to those, too.

From a practical point and maintenance point of view it wouldn't make a big difference. `toJSON = object . ...Properties` vs. `toJSON = error "..."`. What we really want to enforce is that either of the two is present and the generic instance isn't used. I guess, the only really reliable way to enforce that are unit tests.

larskuhtz:
For moving forward I propose the following:

1. I'll try to benchmark / analyze core for your proposal. I am curious how smart GHC is about it. (Maybe I can also pitch it to someone else with some more GHC core experience, @emilypi, @jmcardon? :-)). 
  
2. I am about to push a commit that greatly extends unit tests coverage to (almost) all existing JSON instances.

3. We may add a workflow to Github, that would check that all JSON instances (and possibly also a few other encoding types) are covered by unit tests. So if someone adds a new instance and forgets to add a unit test, that workflow would fail and produce a warning in the PR. The idea would be to grep through the codebase for instance and match those against the list of existing jsonTestCases.

sirlensalot:
>  1. I'll try to benchmark / analyze core for your proposal. I am curious how smart GHC is about it. 

Maybe before that, benchmark this PR as-is against master. I can see how it might improve but numbers talk louder than smart programmers. 

> [unit test coverage]

How do unit tests validate that BOTH `toEncoding` and `toJSON` are implemented, and that those implementations are equivalent?

larskuhtz:
> How do unit tests validate that BOTH toEncoding and toJSON are implemented, and that those implementations are equivalent?

In the test roundtrips module there is a list of all type with json instances. We have QuickCheck instance for those types and perform all kinds of encoding decoding roundtrip and check that those yield equivalent results. (We do that also for other encodings; binary, read-show, text-representation).

However, that list got outdated regularly in the past, every few month I have been curating that list (a big chunk of updates is coming). So, I am thinking about ways to tickle code contributes to add their new types to this list as needed and provide Arbitrary instances.

larskuhtz:
> Maybe before that, benchmark this PR as-is against master. I can see how it might improve but numbers talk louder than smart programmers.

I think, at this point micro-benchmarking is probably the shortest route to get an idea about the benefits. Benchmarking chainweb-node as a whole is tricky. All our nodes have very different load profiles, depending an hardware and request loads. The closes we have to a formal benchmark is the multimode test in slow-test test-suite. Beside that the best we could do in the past was:

1. deploy to a node (or a subset of nodes), wait a few hours/days, and use our performance telemetry to compare the behaviors of the new and the old version. This has been the most useful way to assess effectiveness of changes.

2. Use linux perf tools to collect perf traces (this gave many hints that JSON processing is currently consuming considerable amounts of resources)

3. Profiling builds and running nodes with profiling enabled (however that was of limited use because the amounts of generated data that are needed for meaningful results are difficult to handle)

4. Use gdb, strace, and similar tools to explore the low-level behavior of nodes (system calls, memory allocations, etc.)

The latest GHC-9.2 comes with several features that will hopefully improve our ability to get insights into the performance behavior of full nodes.

sirlensalot:
I would think a dedicated serialization benchmark PR would be welcome at this point. JSON serialization of these datatypes (not necessarily config ones, who cares) ... we already have ChainwebPactDb dedicated benchmarks which have proven very useful.

larskuhtz:
Some preliminary numbers: For pages of payloads and object encoded block Header this PR results in performance gains between 20% - 30% in micro benchmarks. Those datastructure make up for a significant part of API load on some nodes. ( It's possible that actual gains are larger because, if I remember correctly, criterion microbenchmarks tent to hide GC overhead on small benchmark instances, but I still have to confirm that).

I am going to do more benchmarking. I'll also include the `Encode` proposal.

larskuhtz:
Performance gains depend on the number of objects in the JSON structure and are around 50% minus constant overheads.

 For `PayloadWithOutputs`  benchmarks show:
 
 * page size of 10: 30% performance gain (90us vs 130us)
 * page size > 50: 50% performance gain  (440us vs 900)
 
 For instance, chainweb-data refill queries pages of 300 `PayloadWithOutputs` structures at high rate.

larskuhtz:
image
image

larskuhtz:
The following results include the approach via an `Encoded` types as implemented in [this gist](https://gist.github.com/larskuhtz/6b17cd339e3fc278a440fab66fa77b91). It shows the the approach via `Encoded` performs equivalent to `toJSON`. (Note that the x axis uses a logarithmic scale.)

image

(Benchmarks where compiled with `-O1 -threaded`.)

I think, replacing the strict `V.Vector` in `EArray` by a lazy list may provide the compiler a chance to optimize away the extra heap allocations for the intermediate representation. Using a different representation for the inner types (e.g. replacing `ByteString` by `ShortByteString` may also make a difference, although, I think, in this case it may actually decrease performance, but not sure).

larskuhtz:
Replacing the strict `Vector` argument to `EArray` by a lazy list constructor, and ensuring that the implementation of all instances is maximally lazy improves performance and puts the approach half way between `toJSON` and `toEncoding` and larger instances (and small instances there is no improvement). On the larger instances the approach is about 10-20% slower than direct `toEncoding`.

image

Interestingly, the difference between `Encoded` and and direct `toEncoding` becomes smaller (about 10%) when compiling without any optimization. This means that introduction of the `Encoded` constructor isn't fully eliminated prevents GHC from applying some optimizations.

All these results are with GHC-9.0.1.

larskuhtz:
Summing up, I think, the direct approach via `toEncoding` gives the most robust results. (This probably is the reason for the existence of the numerous builder frameworks in Haskell, and why streaming / fusion frameworks rely on fragile `RULE` pragmas.)

Overall, the introduction of the `EncodeJSON` type class results in similar code as the use of `...Property` functions, since the body of `encodeJSON` and `...Property` are almost identical.

I think that the robust performance gains of the `toEncoding` approach justify its use. Performance of `Encoded` is at least 10% worse in the benchmarked examples, and 15%-20% with `-O1`. Those numbers rely on lazy evaluation, which makes this approach less robust. Without laziness the approach performs on par with `toJSON`, which is about 100% slower than `toEncoding` on moderately large instances.

While working on the benchmarks I noticed that many of the `ToJSON` and `FromJSON` instances for newtypes can be eliminated through use of `deriving via` (which should be done in another PR).

Here is a side by side comparison of both patterns

```haskell
data Example = Example { _exampleA :: Int, _exampleB :: Text }

type JSONProperties a = KeyValue kv => a -> [kv]
```

*Direct `toEncoding` Approach:*

```haskell
instance ToJSON Example where
    toJSON = object . exampleProperties
    toEncoding = pairs . mconcat . exampleProperties
 
 exampleProperties :: Properties Example
 exampleProperties e = [ "a" .= _exampleA e, "b" .= _exampleB e ]
 ```
 
 *`Encoded` Approach:*
 
 ```haskell
 instance JSONEncoded Example where
     encodeJSON = [ "a" .== _exampleA e, "b" .== _exampleB e ]
 ```
 
 There is also a one time overhead for defining  the `Encoded` type and the `JSONEncoded` class as well as instances for all "third-party" types used within the code base (full details are [in this gist](https://gist.github.com/larskuhtz/6b17cd339e3fc278a440fab66fa77b91#file-bench-hs-L140)): 
 
 ```haskell
(.==) :: EncodeJSON a => T.Text -> a -> (T.Text, Encoded)
(.==) k v = (k, encodeJSON v)

 instance JSONEncoded Int where
     encodeJSON = ENumber . fromIntegral
     
 instance JSONEncoded Text where
     encodeJSON = EText
 ```

**Comment by sirlensalot** [comment]
As an alternate suggestion, is there some way to make `toEncoding` canonical such that `toJSON` uses it? 

I'm sure it'd be terrible and slow ... basically `fromJSON` to `Value` ... but the idea would be that any uses of `toJSON` on our protocol datatypes are undesirable and not performance-critical. 

Another idea is to have `toJSON` always return a signal value like `Null` or `"---INVALID JSON ENCODER--"` instead of `error` . But insofar as there may be valid reasons for it (like config objects), this seems to be unworkable.

Thanks for doing the benchmarks. It does seem regrettable to bother to go to `toEncoding` and not realize the full perf gains. But duplicate encoders is still bad and I am very motivated to try to find a way out.

**Comment by larskuhtz** [comment]
Here is another approach that eliminate some of the noise. It would have the same performance as plain `toEncoding`. The newtype would be erased at compile time. It is only used to resolve the `ToJSON` dictionary at compile time.

Define the following help type once:

```haskell
class IsJsonObject a where
    jsonProperties :: KeyValue kv => a -> [k]

newtype JsonObject a = JsonObject a

instance IsJsonObject a => ToJSON (JsonObject a) where
    toJSON (JsonObject a) = object $ jsonProperties a
    toEncoding (JsonObject a) = pairs . mconcat $ jsonProperties a
```

Then use it as follows:

```haskell
data Example = Example { _exampleA :: Int, _exampleB :: Text }
    deriving (ToJSON) via (JsonObject Example)

instance IsJsonObject Example where
    jsonProperties e = [ "a" .= _exampleA e, "b" .= _exampleB e ]
    {-# INLINE jsonProperties #-}
```

It would be possible to extend the approach to other JSON encoding types, either via adding additional newtypes or via adding a phantom type to the newtype. There would be one class per JSON type and deriving via would be used to select the correct `ToJSON` dictionary for the respective type.

Not sure, though, if the saving in lines of code justify the extra layer of abstraction and custom coding pattern that wouldn't be familiar to outsiders.

larskuhtz:
I think that the `...Properties` pattern resolves the "duplicate encoders" issue similarly well as other approaches. With any approach a programmer has to be aware of the respective idiom (and the current state of Aeson in general), to not screw things up. Luckily unit tests should reliably catch inconsistencies (assuming the type is added to the json roundtrip test list).

Generally, I prefer a simple and local pattern, that is easily recognized by a programer who is new to the codebase or contributes only infrequently. My feeling is that the `...Properties` approach does better with this regard. It uses a plain Haskell function and doesn't rely on a custom setup, that is hidden somewhere in an `Encoding.Utils` module.

The `newtype` wrapper + `deriving via` approach has the advantage of being elegant and concise. But it has the disadvantage of being less obvious and more difficult to understand.

I am pretty certain that performance is exactly the same (I still have to confirm via benchmarks to be 100% sure). 

I tent towards the straight forward `...Properties` function pattern, because of its straight-forwardness. But I do not have a very strong opinion as long as it delivers good performance.

sirlensalot:
I think your design is interesting and worth considering. However at this point I agree that it's a formalization on top of the code you have in this PR so I'm ready to go with `xxxProperties` to achieve the encoding stability and performance improvements.

I do think that this was a productive exercise informing our future approaches to "fixing" our JSON with a homegrown lib:

- It should have a story for working with `aeson`/`toEncoding` as it will probably be hard/impossible to disconnect our ecosystem entirely from it.
- I had been thinking we would simply use `Map` as the backing store of object and call it a day, whereas properties canonicalizes an arbitrary ordering which has advantages too.
- It's nice to be getting rid of `Hashable` on the encoding side and maybe never having to deal with that issue directly.

Anyway I will proceed with a review of the code as-is. Thanks for working so hard on this!

larskuhtz:
Yeah, this was a helpful exercise. It is interesting to see how lazy constructor application in `Encoded` lets `toEncoding . encodeJSON` outperform `toEncoding . toJSON`.

> I had been thinking we would simply use Map as the backing store of object and call it a day

The most recent version of aeson allows to choose between `HashMap` and `Map`, with `Map` being the default. I didn't yet get a chance to try it with chainweb, due to dependency issues. Now that nix builds have been updated to ghc-8.10, I may give it another try.

larskuhtz:
I am going to remove the experimental JSON encoding approach from the PR and store them for reference and future use [in this gist](https://gist.github.com/larskuhtz/6b17cd339e3fc278a440fab66fa77b91)

src/Chainweb/Cut/CutHashes.hs
```diff
@@ -305,32 +328,46 @@ instance ToJSON CutHashes where
         ifNotEmpty s l
             | x <- view l c, not (HM.null x) = [ s .= x ]
             | otherwise = mempty
+    {-# INLINE toJSON #-}
+
+    toEncoding c = pairs
```
sirlensalot, line 333: properties?

src/Chainweb/MerkleLogHash.hs
```diff
@@ -136,7 +136,7 @@ randomMerkleLogHash :: MerkleHashAlgorithm a => MonadIO m => m (MerkleLogHash a)
 randomMerkleLogHash = unsafeMerkleLogHash <$> randomByteString merkleLogHashBytesCount
 
 merkleLogHashToText :: MerkleHashAlgorithm a => MerkleLogHash a -> T.Text
-merkleLogHashToText = encodeB64UrlNoPaddingText . runPutS . encodeMerkleLogHash
+merkleLogHashToText (MerkleLogHash bytes) = encodeB64UrlNoPaddingText (encodeMerkleRoot bytes)
```
sirlensalot, line 139: This is equivalent?

src/Chainweb/Miner/Config.hs
```diff
@@ -185,19 +190,20 @@ nodeTestMiners = lens _nodeTestMiners (\m c -> m { _nodeTestMiners = c })
...
         [ "enabled" .= _nodeMiningEnabled o
-        , "miner" .= _nodeMiner o ]
+        , "miner" .= _nodeMiner o
+        ]
 
 instance FromJSON (NodeMiningConfig -> NodeMiningConfig) where
     parseJSON = withObject "NodeMiningConfig" $ \o -> id
         <$< nodeMiningEnabled ..: "enabled" % o
         <*< nodeMiner ..: "miner" % o
-        <*< nodeTestMiners ..: "testMiners" % o
```
sirlensalot, line 194: ?

src/Chainweb/SPV.hs
```diff
@@ -107,6 +108,26 @@ proofToJson cid p = object
         [ "input" .= encodeB64UrlNoPaddingText bytes
         ]
 
+proofToEncoding
```
sirlensalot, line 111: properties?

src/Chainweb/SPV/PayloadProof.hs
```diff
@@ -104,6 +108,22 @@ instance MerkleHashAlgorithmName a => ToJSON (PayloadProof a) where
         subj (InputNode bytes) = object
             [ "input" .= encodeB64UrlNoPaddingText bytes
             ]
+    {-# INLINE toJSON #-}
+
+    toEncoding p = pairs
```
sirlensalot, line 113: properties?

src/Chainweb/SPV/PayloadProof.hs
```diff
@@ -139,8 +159,20 @@ instance (MerkleHashAlgorithm a, MerkleHashAlgorithmName a) => FromJSON (Payload
 data SomePayloadProof where
     SomePayloadProof :: (MerkleHashAlgorithm a, MerkleHashAlgorithmName a) => !(PayloadProof a) -> SomePayloadProof
 
+deriving instance Show SomePayloadProof
+
+instance Eq SomePayloadProof where
+    (SomePayloadProof (a :: PayloadProof aalg)) == (SomePayloadProof (b :: PayloadProof balg))
+        | merkleHashAlgorithmName @aalg == merkleHashAlgorithmName @balg =
+            unsafeCoerce b == a
```
sirlensalot, line 167: isn't there a better way to do this given the witnesses above?

src/Chainweb/Utils.hs
```diff
@@ -651,6 +652,10 @@ encodeB64UrlNoPaddingText :: B.ByteString -> T.Text
 encodeB64UrlNoPaddingText = T.dropWhileEnd (== '=') . T.decodeUtf8 . B64U.encode
 {-# INLINE encodeB64UrlNoPaddingText #-}
 
+encodeB64UrlNoPadding :: B.ByteString -> B.ByteString
```
sirlensalot, line 655: Do we have a roundtrip for this?

src/Chainweb/Utils/RequestLog.hs
```diff
@@ -99,6 +101,11 @@ instance ToJSON RequestLog where
     toJSON = genericToJSON defaultOptions
         { fieldLabelModifier = over _head toLower . drop 11
         }
+    toEncoding = genericToEncoding defaultOptions
```
sirlensalot, line 104: Why do we still need generics?

src/Chainweb/Utils/RequestLog.hs
```diff
@@ -140,6 +147,11 @@ instance ToJSON RequestResponseLog where
     toJSON = genericToJSON defaultOptions
         { fieldLabelModifier = over _head toLower . drop 19
         }
+    toEncoding = genericToEncoding defaultOptions
```
sirlensalot, line 150: See line 104

src/Chainweb/MerkleLogHash.hs
larskuhtz: The second is potentially more efficient (since it avoid a trivial roundtrip through a builder). However, the original code is more idiomatic and friendlier to maintain, because all other encodings are derived from `encodeMerkleHashLog`. It is also unrelated to this PR.

I'll to revert this changes.

src/Chainweb/Miner/Config.hs
larskuhtz: the `testMiners` setting is used in some tests to adjust the hash power when simulating PoW. It used to be a "hidden" configuration setting. But nowadays it is only used programmatically in tests. So, the PR removes it from the config parser.

src/Chainweb/SPV/PayloadProof.hs
larskuhtz: The alternative is to use `Data.Type.Equality.testEquality` which would require either 

1. add a proper `Singleton` for `PayloadProof` and define an `TestEquality` instance based on that,
2. move the `unsafeCoerce` into a `TestEquality` instance for `PayloadProof`, or
3. use `typeOf` from `Data.Typeable`.

It's not really important, since the instance is only used for testing (in order to ensure that roundtrips yield same results). I think, I try (3.). If that doesn't work I'll just move the instance out of production code to the test orphans module.

larskuhtz: The `Data.Typeable` approach kinda leads into a rabbit hole (and generally I try to avoid generic reflection). I think, it is to disruptive for an instance that is only used for testing. I am going to just move it to the test orphans.

src/Chainweb/Utils.hs
larskuhtz: I ended up not using it in this PR. Going to remove it.

src/Chainweb/Utils/RequestLog.hs
larskuhtz: A lazy programer :-) The type is only used for telemetry logging.

Anyways, I'll add an explicit encoding for consistency.

sirlensalot (approved):

PR #1303, opened 2021-11-16
GHC 8.10 support for Nix builds

mightybyte:


jmininger (commented):
Did you accidentally move `deps/kpkgs` into `deps/chainweb-storage` ?

PR #1309, opened 2021-11-18
include cut, header, and payload endpoints into service api

larskuhtz:
* [x] exclude `PUT /cut` from service API


larskuhtz:
Those are endpoints from the P2P API. But in practice, many service API clients use these endpoints, too. (I guess, most information would be available at the Rosetta endpoints, which are on the Service API, but Rosetta isn't really used by anybody).

So, all deployments use reverse proxies to map these endpoints from the P2P API to the service API. This change would eliminate the need to use a reverse proxy for this and also would simplify the setup of the proxy where it is needed for other reasons.

sirlensalot (approved):

PR #1312, opened 2021-11-19
option to only sync pact databases

larskuhtz:


src/Chainweb/Chainweb.hs
```diff
@@ -447,36 +448,37 @@ withChainwebInternal conf logger peer serviceSock rocksDb pactDbDir resetDb inne
...
-                            , _chainwebManager = mgr
-                            , _chainwebPactData = pactData
-                            , _chainwebThrottler = throttler
-                            , _chainwebMiningThrottler = miningThrottler
-                            , _chainwebPutPeerThrottler = putPeerThrottler
-                            , _chainwebLocalThrottler = localThrottler
-                            , _chainwebConfig = conf
-                            , _chainwebServiceSocket = serviceSock
-                            }
+            unless (_configOnlySyncPact conf) $
```
sirlensalot, line 451: Going forward, please don't re-indent so much code to add a single line, it makes the diff hard to review.

sirlensalot (approved):

src/Chainweb/Chainweb.hs
larskuhtz: In this case, what would have been the alternative to reindenting?

PR #1315, opened 2021-11-26
report pruning progress

larskuhtz:


src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -189,22 +189,44 @@ pruneForks_ logg cdb mar mir callback = do
...
         -- check. Comprehensive checks on various levels are availabe through
         -- callbacks that are offered in the module "Chainweb.Chainweb.PruneChainDatabase"
         | prevHeight /= curHeight && prevHeight /= curHeight + 1 =
             throwM $ InternalInvariantViolation
-                $ "PrunForks.pruneForks_: detected a corrupted database. Some block headers are missing"
+                $ "PruneForks.pruneForks_: detected a corrupted database. Some block headers are missing"
+                <> ". Current pivots: " <> encodeToText pivots
+                <> ". Current header: " <> encodeToText (ObjectEncoded cur)
+                <> ". Previous height: " <> sshow prevHeight
+        | _blockChainwebVersion  cur /= ver =
```
larskuhtz, line 216:
```suggestion
        | _blockChainwebVersion cur /= ver =
```

src/Chainweb/BlockHeaderDB/PruneForks.hs
```diff
@@ -189,22 +189,44 @@ pruneForks_ logg cdb mar mir callback = do
...
+            throwM $ InternalInvariantViolation
+                $ "PruneForks.pruneForks_: detected a corrupted database. A block header in the database did not match the chainweb version of the node"
+                <> ". Current pivots: " <> encodeToText pivots
+                <> ". Current header: " <> encodeToText (ObjectEncoded cur)
+                <> ". Previous height: " <> sshow prevHeight
+                <> ". Expected chainweb version: " <> sshow ver
+        | _blockChainId  cur /= cid =
```
larskuhtz, line 223:
```suggestion
        | _blockChainId cur /= cid =
```

src/Chainweb/BlockHeaderDB/PruneForks.hs
larskuhtz, line 216: Unfortunately, this may actually trigger on all dbs that originated from a bootstrap node. It seems that for some reason those dbs contain development genesis headers -- which is something that we should fix. (I think, this can happen if a node is started with the wrong chainweb version and a fix should prevent that a non-empty db is initialized with genesis headers that do not match the existing headers in the database)

larskuhtz (commented):
The new check for the correct chainweb version must be tested

edmundnoble (approved):

PR #1319, opened 2021-11-30
remove legacy CPP macros and raise lower bounds

larskuhtz:
Remove some legacy CPP macros and lower bounds that are now obsolete since the nix builds use GHC-8.10.

### DIFF
--- a/src/Chainweb/Utils.hs
+++ b/src/Chainweb/Utils.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE DefaultSignatures #-}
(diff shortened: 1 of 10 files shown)


src/Chainweb/Utils.hs
```diff
@@ -1405,14 +1373,6 @@ showIpv6 ha = T.intercalate ":"
   where
     (a0,a1,a2,a3,a4,a5,a6,a7) = hostAddress6ToTuple ha
 
-#if !MIN_VERSION_network(3,0,0)
```
emmanueldenloye, line 1408: I noticed that NFData instance in network, version 3.1.1.1, is slightly different than what you had previously. Is this desired?

emmanueldenloye (approved):
Aside from my one comment... LGTM

src/Chainweb/Utils.hs
larskuhtz: hmm, in any case whatever is in the network package is authoritative.

PR #1322, opened 2021-12-21
Enable eventlog (do not merge)

edmundnoble:


edmundnoble:
Why doesn't this build?

PR #1328, opened 2021-12-29
RocksDB checkpoints with SIGUSR1, now SIGUSR2 does not terminate the node

edmundnoble:
Requires upstream PR https://github.com/kadena-io/chainweb-storage/pull/10. Makes checkpoints in `$DBDIR/0/rocksDbCheckpoints`.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -40,8 +40,8 @@ source-repository-package
 
 source-repository-package
     type: git
-    location: https://github.com/edmundnoble/chainweb-storage.git
-    tag: d648827619be2d4545387f4d85d9c55f7ba46ad4
+    location: https://github.com/kadena-io/chainweb-storage.git
+    tag: 1ee64539f9b6e27372caa14a1b4cbdf68736ed43
 
 source-repository-package
     type: git
--- a/cabal.project
+++ b/cabal.project
@@ -40,8 +40,8 @@ source-repository-package
 
 source-repository-package
     type: git
-    location: https://github.com/kadena-io/chainweb-storage.git
-    tag: bf37c394de9a0c680a7a09bbfb82eb255519b3fb
+    location: https://github.com/edmundnoble/chainweb-storage.git
+    tag: d648827619be2d4545387f4d85d9c55f7ba46ad4
 
 source-repository-package
     type: git
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -602,6 +602,7 @@ executable chainweb-node
         , configuration-tools >= 0.6
         , deepseq >= 1.4
         , directory >= 1.3
+        , filepath >= 1.4
         , http-client >= 0.5
         , http-client-tls >=0.3
         , lens >= 4.17
--- a/node/ChainwebNode.hs
+++ b/node/ChainwebNode.hs
@@ -56,6 +56,7 @@ import Control.Monad.Managed
 
 import Data.CAS
 import Data.CAS.RocksDB
+import Data.IORef
 import qualified Data.Text as T
 import Data.Time
 import Data.Typeable
@@ -70,6 +71,7 @@ import qualified Network.HTTP.Client.TLS as HTTPS
 import qualified Streaming.Prelude as S
 
 import System.Directory
+import System.FilePath
 import System.IO (BufferMode(LineBuffering), hSetBuffering, stderr)
 import qualified System.Logger as L
 import System.LogLevel
@@ -92,6 +94,7 @@ import Chainweb.Miner.Coordinator (MiningStats)
 import Chainweb.Pact.RestAPI.Server (PactCmdLog(..))
 import Chainweb.Payload
 import Chainweb.Payload.PayloadStore
+import Chainweb.Time
 import Chainweb.Utils
 import Chainweb.Utils.RequestLog
 import Chainweb.Version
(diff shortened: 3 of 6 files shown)


node/ChainwebNode.hs
```diff
@@ -299,20 +305,32 @@ node conf logger = do
...
         logFunctionText logger Info $ "opened rocksdb in directory " <> sshow rocksDbDir
-        withChainweb cwConf logger rocksDb pactDbDir (_nodeConfigResetChainDbs conf) $ \cw -> mapConcurrently_ id
-            [ runChainweb cw
-              -- we should probably push 'onReady' deeper here but this should be ok
-            , runCutMonitor (_chainwebLogger cw) (_cutResCutDb $ _chainwebCutResources cw)
-            , runQueueMonitor (_chainwebLogger cw) (_cutResCutDb $ _chainwebCutResources cw)
-            , runRtsMonitor (_chainwebLogger cw)
-            , runBlockUpdateMonitor (_chainwebLogger cw) (_cutResCutDb $ _chainwebCutResources cw)
-            ]
+        bracket (newIORef (Just rocksDb)) (flip writeIORef Nothing) $ \ref -> do
```
edmundnoble, line 312: This is extremely paranoid, perhaps too paranoid, about not letting us access a `RocksDb` when it's out of scope.

node/ChainwebNode.hs
```diff
@@ -470,7 +488,7 @@ mainInfo = programInfoValidate
 
 main :: IO ()
 main = do
-    installSignalHandlers
+    installFatalSignalHandlers [ sigHUP, sigTERM, sigXCPU, sigXFSZ ]
```
edmundnoble, line 491: I made `sigUSR2` not fatal anymore in case we want to use it later, so that we don't accidentally kill older versions of the node with it.

node/ChainwebNode.hs
```diff
@@ -299,20 +305,32 @@ node conf logger = do
...
+                , runBlockUpdateMonitor (_chainwebLogger cw) (_cutResCutDb $ _chainwebCutResources cw)
+                ]
   where
     cwConf = _nodeConfigChainweb conf
 
+makeCheckpoint :: FilePath -> IORef (Maybe RocksDb) -> IO ()
+makeCheckpoint checkpointDir rdbRef = 
+    readIORef rdbRef >>= \case
+        Nothing -> pure ()
+        -- 0 ~ never flush WAL log before checkpoint, to avoid making extra work
```
larskuhtz, line 329: Do you know how often the WAL log is flushed normally? Assuming that it's done often, this should be fine. Otherwise, we may consider flushing.

Writing the snapshot can only be triggered by a node admin and probably won't happen too often, so some performance overhead is probably fine.

node/ChainwebNode.hs
larskuhtz: Is `USER2` the only change compared to the previous behavior?

node/Utils/InstallSignalHandlers.hs
```diff
@@ -77,6 +79,15 @@ installHandler signal handler = do
...
 
+installHandlerCross :: Signal -> (Signal -> IO ()) -> IO ()
+installHandlerCross s h = 
+#ifdef mingw32_HOST_OS
+    installHandler s h
+#else
+    void $ installHandler s (Catch (h s)) Nothing
+#endif
+
+
```
larskuhtz, line 90:
```suggestion
```

node/ChainwebNode.hs
```diff
@@ -299,20 +305,32 @@ node conf logger = do
...
+        Just rdb -> do 
+            now <- getCurrentTimeIntegral @Integer
+            checkpointRocksDb rdb 0 (checkpointDir </> formatTimeMicros now)
```
larskuhtz, line 332: I assume that the resulting file names lexicographically ordered, so that using `ls | tail -n 1` will provide the latest snapshot? Also, are all characters portable and compatible with all major file systems?

chainweb.cabal
```diff
@@ -602,6 +602,7 @@ executable chainweb-node
         , configuration-tools >= 0.6
         , deepseq >= 1.4
         , directory >= 1.3
+        , filepath >= 1.4
```
larskuhtz, line 605: I am not sure if `</>` is worth the new dependency. We used to be more liberal about dependencies while we were building up the codebase, but have recently transitioning to a mode where we try to reduce the number of dependencies in order to reduce attack surface.

What about adding `</>` to `Chainweb.Utils` or, if we need more path specific functions, adding a`Chainweb.Utils.Filepath` module?

cabal.project
```diff
@@ -40,8 +40,8 @@ source-repository-package
 
 source-repository-package
     type: git
-    location: https://github.com/kadena-io/chainweb-storage.git
-    tag: bf37c394de9a0c680a7a09bbfb82eb255519b3fb
+    location: https://github.com/edmundnoble/chainweb-storage.git
+    tag: d648827619be2d4545387f4d85d9c55f7ba46ad4
```
larskuhtz, line 44: Can you update this to the latest master commit of chainweb-storage?

larskuhtz: The nix pin probably needs to be updated as well.

larskuhtz (changes requested):
I think, it would be nice to add a test for this. I think, the test suite already initializes global rocksdb (and individual tests create a "sub-db" in separate namespaces). Maybe that could be used to test snapshot creation.

node/ChainwebNode.hs
edmundnoble: sigUSR1 and sigUSR2 not being fatal is the only change.

node/ChainwebNode.hs
edmundnoble: It's flushed when it gets to 4MB IIRC. This makes sense to me, avoiding extra catchup is probably more important than a momentary hiccup.

node/ChainwebNode.hs
edmundnoble: Seems that it's ISO8601, for which lexicographic and chronologic orderings coincide. But... it's not compatible with NTFS or exFAT, which reserve `:` apparently. So I guess this should just be unix timestamps.

node/ChainwebNode.hs
edmundnoble: Specifically, this is for not letting the signal handler access a `RocksDb` when it's out of scope, but that assumes that we can exit this `bracket` *and then* trigger the signal handler without the program having terminated.

chainweb.cabal
edmundnoble: `</>` is surprisingly large for cross-platform stuff. `directory` transitively depends on `filepath`, does that factor in at all?

node/ChainwebNode.hs
larskuhtz: 4MB is probably around a few thousand blocks. I think, it's fine for now to not flush. Maybe we could do a follow up PR that addresses it. We could use another signal for flushing or add an (optional) endpoint on the service API that would take parameters (e.g. snapshot name and flushing)

node/ChainwebNode.hs
larskuhtz: hmm, not sure if we care about NTFS and FAT. Readable dates make snapshot maintenance easier. But scripts could still use the file creation time. So POSIX timestamps would be fine.

Not sure, I don't have a strong opinion here.

chainweb.cabal
larskuhtz: If directory already depends on it, let's take the dependency.

node/ChainwebNode.hs
larskuhtz: Oh, just noticed that you changed it. It's fine either way.

larskuhtz (approved):

larskuhtz (approved):
Actually, I think a test case would be nice -- assuming that it's relatively easy to implement. 

But it's not critical for this PR. Maybe in a follow up PR?

PR #1330, opened 2021-12-30
db size telemetry

edmundnoble:
For pactdb and rocksdb. Requires https://github.com/kadena-io/chainweb-storage/pull/11/files.


edmundnoble:
I'm changing this to just report overall rocksdb size, to avoid rocksdb-specific features and the `tables` table.

larskuhtz (approved):

PR #1340, opened 2022-01-08
More pact queue stats

edmundnoble:


src/Chainweb/Pact/Service/PactQueue.hs
```diff
@@ -131,36 +123,31 @@ instance ToJSON PactQueueCounters where
 pactQueueCountersProperties :: KeyValue kv => PactQueueCounters -> [kv]
 pactQueueCountersProperties pqc =
     [ "count" .= _pactQueueCountersCount pqc
-    , "sum" .= _pactQueueCountersSum pqc
+    , "sum" .= getSum (_pactQueueCountersSum pqc)
     , "min" .= _pactQueueCountersMin pqc
```
emmanueldenloye, line 127: Why  don't you use `getMin` (and correspondingly `getMax` below) here?

edmundnoble: There's instances of `ToJSON a => ToJSON (Min a)` and `ToJSON a => ToJSON (Max a)`, but for some reason not `Sum`.

emmanueldenloye (approved):

PR #1341, opened 2022-01-10
Enforce miner keyset formats

sirlensalot:


src/Chainweb/Version.hs
```diff
@@ -913,6 +914,14 @@ doCheckTxHash Development = (>= 110)
 doCheckTxHash (FastTimedCPM g) | g == petersonChainGraph = (>= 7)
 doCheckTxHash _ = const False
 
+-- | Placeholder name for 2.13 Pact changes
+chainweb213Pact :: ChainwebVersion -> BlockHeight -> Bool
+chainweb213Pact Mainnet01 = const False -- TODO
+chainweb213Pact Testnet04 = const False -- TODO
+chainweb213Pact Development = (>= 95)
+chainweb213Pact (FastTimedCPM g) | g == petersonChainGraph = (>= 5)
+chainweb213Pact _ = const False
```
larskuhtz, line 923: Why not make the  new behavior the default for all tests that don't explicitly require the old behavior?

larskuhtz (approved):
I have one question regarding the default value of `chainweb213Pact`. Looks good otherwise.

src/Chainweb/Version.hs
```diff
@@ -913,6 +914,14 @@ doCheckTxHash Development = (>= 110)
 doCheckTxHash (FastTimedCPM g) | g == petersonChainGraph = (>= 7)
 doCheckTxHash _ = const False
 
+-- | Placeholder name for 2.13 Pact changes
+chainweb213Pact :: ChainwebVersion -> BlockHeight -> Bool
+chainweb213Pact Mainnet01 = const False -- TODO
+chainweb213Pact Testnet04 = const False -- TODO
```
larskuhtz, line 920: This is what is used in #1358

```Haskell
pactForceEof Mainnet01 = (>= 2_447_358) -- 2022-02-26 00:00:18
pactForceEof Testnet04 = (>= 1_977_816) -- 2022-02-25 00:00:18
```

emilypi (commented):
LGTM, pending the TODOs

src/Chainweb/Version.hs
sirlensalot: Hmm ... we've been doing `const False` up until now, probably because we're worried about genesis. Maybe it's best to change these across the board. 

How about leaving it `const False` for now and doing a dedicated PR for that?

src/Chainweb/Version.hs
sirlensalot: To clarify -- I'm proposing `chainweb213Pact` to be the only fork and used for force-eof, module gas etc.

PR #1337, opened 2022-01-06
Prune cut hashes on CutDB startup

edmundnoble:

PR #1342, opened 2022-01-11
Cut hashes space efficiency

edmundnoble:
Based on #1337, prunes periodically and writes less cuts to the database in the first place.


test/Chainweb/Test/CutDB.hs
```diff
@@ -38,6 +40,7 @@ import Data.Foldable
 import Data.Function
 import qualified Data.HashMap.Strict as HM
 import qualified Data.Vector as V
+-- import qualified Data.Text as T
```
larskuhtz, line 43: remove?

```suggestion
```

test/Chainweb/Test/CutDB.hs
```diff
@@ -486,3 +497,31 @@ fakePact = WebPactExecutionService $ PactExecutionService
     getFakeOutput (Transaction txBytes) = TransactionOutput txBytes
     coinbase = noCoinbaseOutput
     fakeMiner = MinerData "fakeMiner"
+
+tests :: RocksDb -> TestTree
+tests rdb = testGroup "CutDB"
+    [ testCutPruning rdb version
+    ]
+    where
```
larskuhtz, line 505:
```suggestion
  where
```

test/Chainweb/Test/CutDB.hs
```diff
@@ -486,3 +497,31 @@ fakePact = WebPactExecutionService $ PactExecutionService
...
+
+testCutPruning :: RocksDb -> ChainwebVersion -> TestTree
+testCutPruning rdb v = testCase "cut pruning" $ do
+    let pruningThreshold = 100
+    -- initialize cut DB and mine enough to trigger pruning
+    let alterPruningSettings = 
+            set cutDbParamsAvgBlockHeightPruningDepth 50 .
+            set cutDbParamsAvgBlockHeightPruningThreshold pruningThreshold .
+            set cutDbParamsAvgBlockHeightWritingGap 10
+        minedBlockHeight = 201
```
larskuhtz, line 516: move to `where`-clause for better readability?

test/Chainweb/Test/CutDB.hs
```diff
@@ -486,3 +497,31 @@ fakePact = WebPactExecutionService $ PactExecutionService
...
+    (cutHashesStore, ()) <- withTestCutDbWithoutPact rdb v alterPruningSettings (int $ avgCutHeightAt v minedBlockHeight) (\_ _ -> return ()) (const $ return ())
```
larskuhtz, line 517: Shorten line

```suggestion
    (cutHashesStore, ()) <- withTestCutDbWithoutPact rdb v alterPruningSettings (int $ avgCutHeightAt v minedBlockHeight) 
        (\_ _ -> return ()) 
        (const $ return ())
```

larskuhtz, line 517: Why not move the assertions into the scope of the `withTestCutDbWithoutPact` bracket? That would also make above changes to the return value of `withTestCutDb` redundant.

test/Chainweb/Test/CutDB.hs
```diff
@@ -111,8 +120,8 @@ withTestCutDb
     -> LogFunction
         -- ^ a logg function (use @\_ _ -> return ()@ turn of logging)
     -> (forall cas . PayloadCasLookup cas => CutDb cas -> IO a)
-    -> IO a
-withTestCutDb rdb v n pactIO logfun f = do
+    -> IO (RocksDbCas CutHashes, a)
```
larskuhtz, line 123: Why is this needed. It is probably safe to let the cut-hashes table escape the scope of `withTestCutDb`, but is a bit unintuitive. For instance, some future code refactoring may implement some sort of cleanup for the temporary test cut db, which would cause code using the result db to break.

test/Chainweb/Test/Mining.hs
```diff
@@ -67,8 +67,8 @@ withTestCoordiantor
     -> IO ()
 withTestCoordiantor rdb maybeConf a = do
     var <- newEmptyMVar
-    x <- race (takeMVar var) $ do
-        withTestCutDb rdb v 0 (\_ _ -> return fakePact) (logFunction logger) $ \cdb ->
+    x <- race (takeMVar var) $ 
+        fmap snd $ withTestCutDb rdb v id 0 (\_ _ -> return fakePact) (logFunction logger) $ \cdb ->
```
larskuhtz, line 71: see comment above on `test/Chaiwneb/Test/CutDb.hs` regarding the change of the return type of `withTestCutDb`.

test/SlowTests.hs
```diff
@@ -16,24 +16,28 @@ module SlowTests ( main ) where
...
 
-suite :: TestTree
-suite = testGroup "ChainwebSlowTests"
+suite :: RocksDb -> TestTree
+suite rdb = testGroup "ChainwebSlowTests"
     [ Chainweb.Test.MultiNode.test loglevel (TimedConsensus petersonChainGraph twentyChainGraph) 10 120
     , testGroup "Network.X05.SelfSigned.Test"
         [ Network.X509.SelfSigned.Test.tests
         ]
+    , Chainweb.Test.CutDB.tests rdb
```
larskuhtz, line 42: These tests already got added to `testsChainwebTests`. No need to add them here again.

(The tests in ` Network.X509.SelfSigned.Test.tests` are here not because they are slow, but because of nix historically had trouble with executing certain networking tests)

larskuhtz (changes requested):
I wonder if the logic for computing the delay between writing of two cuts could be simpler. 

I think we didn't use direct sampling with memoryless Bernoulli trials, because we were concerned of the long tail of geometric distribution. I now wonder if that concern was overly conservative. 

We could also consider a distribution that based on the current cut height modulo the expected gap plus some random seed and some random jitter that would approximate the current normal distribution of the gap, but without the need of the mutable state.

Similarly we may replace pruning by cut-height by a geometric distribution or an memoryless approximation of the current distribution.

... Just a though; the current solution would be fine, too.

src/Chainweb/CutDB.hs
```diff
@@ -457,32 +501,50 @@ processCuts
...
+        & S.mapM_ (\newCut -> do
+            curCut <- readTVarIO cutVar
+            !resultCut <- trace logFun "Chainweb.CutDB.processCuts._joinIntoHeavier" () 1
+                $ joinIntoHeavier_ hdrStore (_cutMap curCut) newCut
+            let curCutAvgBlockHeight = BlockHeight $ round $ avgBlockHeightAtCutHeight (_chainwebVersion headerStore) (_cutHeight curCut)
+            lastPrunedAvgBlockHeight <- readTVarIO lastPrunedAvgBlockHeightVar
+            let sinceLastPrune = curCutAvgBlockHeight - lastPrunedAvgBlockHeight
+            when (sinceLastPrune > _cutDbParamsAvgBlockHeightPruningThreshold conf) $ do
+                atomically $ writeTVar lastPrunedAvgBlockHeightVar curCutAvgBlockHeight
+                pruneCuts logFun (_chainwebVersion headerStore) conf curCut cutHashesStore
```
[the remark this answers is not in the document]
edmundnoble: Pruning is asynchronous yes, so they can definitely overlap. RocksDB has a queue of pending compaction operations so it seems fine.

test/Chainweb/Test/CutDB.hs
edmundnoble: I'm not sure I follow; we can't access the `cutHashesStore` from the `CutDb` record.

larskuhtz (approved):

chainweb.cabal
```diff
@@ -675,6 +675,7 @@ executable cwtool
         , exceptions >= 0.8
         , extra >= 1.6
         , file-embed >= 0.0
+        , hashable >= 1.3 && < 1.3.1
```
larskuhtz, line 678:
```suggestion
        , hashable >= 1.3
```

test/Chainweb/Test/CutDB.hs
larskuhtz: I don't understand. Can't the whole test, including the assertions be moved into the `withTestCutDb` bracket?

If `cutHashesStore` isn't available from `CutDb`, the type of the inner action could become:

```Haskell
(forall cas . PayloadCasLookup cas => CutDb cas -> RocksDbCas CutHashes -> IO a)
```

That way any access to the cutDb and the cutHashesStore would occur within the bracket.

test/Chainweb/Test/CutDB.hs
```diff
@@ -111,8 +120,8 @@ withTestCutDb
     -> LogFunction
         -- ^ a logg function (use @\_ _ -> return ()@ turn of logging)
     -> (forall cas . PayloadCasLookup cas => CutDb cas -> IO a)
```
larskuhtz, line 122:
```suggestion
    -> (forall cas . PayloadCasLookup cas => CutDb cas -> RocksDbCas CutHashes -> IO a)
```

test/Chainweb/Test/CutDB.hs
```diff
@@ -486,3 +497,33 @@ fakePact = WebPactExecutionService $ PactExecutionService
...
+        (const $ return ())
+    -- peek inside the cut DB's store to find the oldest and newest cuts
+    let table = _getRocksDbCas cutHashesStore
+    Just (leastCutHeight, _, _) <- tableMinKey table
+    Just (mostCutHeight, _, _) <- tableMaxKey table
+    -- we must have pruned the older cuts
+    assertBool "oldest cuts are too old" 
+        (round (avgBlockHeightAtCutHeight v leastCutHeight) >= minedBlockHeight - pruningFrequency - int (degreeAt v minedBlockHeight) - 1)
+    -- we must keep the latest cut
+    avgBlockHeightAtCutHeight v mostCutHeight @?= 201
```
larskuhtz, line 528: After changing the type of `withTestCutDbWithoutPact` above:

```suggestion
    (cutHashesStore, ()) <- withTestCutDbWithoutPact rdb v alterPruningSettings (int $ avgCutHeightAt v minedBlockHeight)  (\_ _ -> return ()) $ \_ cutHashesStore -> do 
        -- peek inside the cut DB's store to find the oldest and newest cuts
        let table = _getRocksDbCas cutHashesStore
        Just (leastCutHeight, _, _) <- tableMinKey table
        Just (mostCutHeight, _, _) <- tableMaxKey table
        -- we must have pruned the older cuts
        assertBool "oldest cuts are too old" 
            (round (avgBlockHeightAtCutHeight v leastCutHeight) >= minedBlockHeight - pruningFrequency - int (degreeAt v minedBlockHeight) - 1)
        -- we must keep the latest cut
        avgBlockHeightAtCutHeight v mostCutHeight @?= 201
```

larskuhtz (changes requested):

test/Chainweb/Test/CutDB.hs
edmundnoble: Done.

test/Chainweb/Test/CutDB.hs
edmundnoble: I had the same thought this morning, implemented it, and just read this now :D

cabal.project
```diff
@@ -40,8 +40,8 @@ source-repository-package
 
 source-repository-package
     type: git
-    location: https://github.com/kadena-io/chainweb-storage.git
-    tag: 1ee64539f9b6e27372caa14a1b4cbdf68736ed43
+    location: https://github.com/edmundnoble/chainweb-storage.git
+    tag: 464f75e80e72f8051e4aa32f7072ceac6f493d86
```
larskuhtz, line 44:
```suggestion
    tag: 5bfe043ab11512b6eb4d625deaee0a79edc595df
```

larskuhtz (changes requested):
I think this is almost ready. There are just three small suggestions and redundant test in `test/SlowTests.hs`

test/SlowTests.hs
```diff
@@ -16,22 +16,25 @@ module SlowTests ( main ) where
...
 loglevel :: LogLevel
 loglevel = Warn
 
-suite :: TestTree
-suite = testGroup "ChainwebSlowTests"
+suite :: RocksDb -> TestTree
+suite rdb = testGroup "ChainwebSlowTests"
     [ Chainweb.Test.MultiNode.test loglevel (TimedConsensus petersonChainGraph twentyChainGraph) 10 120
     , testGroup "Network.X05.SelfSigned.Test"
         [ Network.X509.SelfSigned.Test.tests
```
larskuhtz, line 40: Let's just revert these changes. They are redundant now.

larskuhtz (approved):

PR #1344, opened 2022-01-13
Make cut from blockheight specified on cmdline

emmanueldenloye:
This PR should probably be renamed to something more descriptive.


src/Chainweb/CutDB.hs
```diff
@@ -420,6 +430,16 @@ startCutDb config logfun headerStore payloadStore cutHashesStore = mask_ $ do
                             -- FIXME: this requires that the pact db is deleted because it
                             -- interfers with rewind limit. As a fix temporarily disable
                             -- rewind limit during initialization.
+    readCutForHeight h wbhdb' = do
+      unsafeMkCut v . HM.fromList <$> (flip traverse (HS.toList $ chainIdsAt v h) $ \cid -> do
+            db <- getWebBlockHeaderDb wbhdb' cid
+            let randomStrElem s = do
+                  l <- S.length_ s
+                  i <- createSystemRandom >>= asGenIO (uniformR (0,l-1))
```
emmanueldenloye, line 438: This line should be 
```
i <- createSystemRandom >>= asGenIO (uniformR (0,l-2))
```
But I know someone else is going to take this over. I'm just putting this here as an FYI.

edmundnoble:
Tested by rewinding by around 60000 block heights on ap2.testnet.chainweb.com; system seemed to catch up healthily.
Rewinding 1_000000 block heights killed the Pact checkpointer as I think is expected.

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -333,7 +337,7 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configMempoolP2p = defaultEnableConfig defaultMempoolP2pConfig
     , _configBlockGasLimit = 150000
     , _configMinGasPrice = 1e-8
-    , _configPactQueueSize = 2000
+    , _configPactQueueSize = 10000
```
larskuhtz, line 340: What is the motivation for this. For all chains that would 20*10000 pact requests. Also if each pact requests takes a few ms (as typical), 10000 items would be a backlog of something like 10-20seconds at least.

src/Chainweb/CutDB.hs
```diff
@@ -159,7 +160,7 @@ data CutDbParams = CutDbParams
     , _cutDbParamsTelemetryLevel :: !LogLevel
     , _cutDbParamsUseOrigin :: !Bool
     , _cutDbParamsFetchTimeout :: !Int
-    , _cutDbParamsInitialHeightLimit :: !(Maybe CutHeight)
+    , _cutDbParamsInitialHeightLimit :: !(Maybe BlockHeight)
```
larskuhtz, line 163: 👍

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -124,7 +124,7 @@ doRestore v cid dbenv (Just (bh, hash)) = runBlockEnv dbenv $ do
     setSortedKeys = bsSortedKeys .= pact420Upgrade v bh
 doRestore _ _ dbenv Nothing = runBlockEnv dbenv $ do
     clearPendingTxState
-    withSavepoint DbTransaction $
+    withSavepoint ResetCheckpointer $
```
larskuhtz, line 127: We should double check, that no code elsewhere makes assumptions about this checkpoint name.

src/Chainweb/Pact/Types.hs
```diff
@@ -353,9 +353,10 @@ defaultPactServiceConfig :: PactServiceConfig
 defaultPactServiceConfig = PactServiceConfig
       { _pactReorgLimit = fromIntegral $ defaultReorgLimit
       , _pactRevalidate = True
-      , _pactQueueSize = 1000
+      , _pactQueueSize = 10000
```
larskuhtz, line 356: What is the motivation for this change? (also see comment above)

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -829,7 +829,7 @@ deleteHistory bh = do
 
 initSchema :: BlockHandler SQLiteEnv ()
 initSchema = do
-    withSavepoint DbTransaction $ do
+    withSavepoint InitSchema $ do
```
larskuhtz, line 832: We have to double check that no code elsewhere makes assumptions about this savepoint name

src/Chainweb/Chainweb.hs
```diff
@@ -487,15 +490,15 @@ withChainwebInternal conf logger peer serviceSock rocksDb pactDbDir resetDb inne
...
+        , _cutDbParamsInitialHeightLimit = _cutInitialBlockHeightLimit cutConf
         }
       where
         cutConf = _configCuts conf
 
     synchronizePactDb :: HM.HashMap ChainId (ChainResources logger) -> CutDb cas -> IO ()
     synchronizePactDb cs cutDb = do
         currentCut <- _cut cutDb
-        mapConcurrently_ syncOne $ mergeCutResources $ _cutMap currentCut
+        traverse_ syncOne $ mergeCutResources $ _cutMap currentCut
```
larskuhtz, line 501: Does this synchronize pact dbs for chains sequentially? That would be very slow. What is the reason for this change?

src/Chainweb/CutDB.hs
```diff
@@ -416,10 +417,7 @@ startCutDb config logfun headerStore payloadStore cutHashesStore = mask_ $ do
                 Right hm -> unsafeMkCut v
                     <$> case _cutDbParamsInitialHeightLimit config of
                         Nothing -> return hm
```
larskuhtz, line 419: I think, that even this case could interfere with the pact rewind limit (although it's very unlikely). I assume with this PR we always disable the rewind limit during pact db initialization. Is that correct?

src/Chainweb/CutDB.hs
```diff
@@ -416,10 +417,7 @@ startCutDb config logfun headerStore payloadStore cutHashesStore = mask_ $ do
...
-                        Just h -> limitCutHeaders wbhdb h hm
-                            -- FIXME: this requires that the pact db is deleted because it
-                            -- interfers with rewind limit. As a fix temporarily disable
-                            -- rewind limit during initialization.
+                        Just h -> limitCutHeaders wbhdb (avgCutHeightAt v h) hm
```
larskuhtz, line 420: Maybe we should change the argument of `limitCutHeaders` from cut height to block height?

cabal.project
```diff
@@ -40,8 +40,8 @@ source-repository-package
 
 source-repository-package
     type: git
-    location: https://github.com/kadena-io/chainweb-storage.git
-    tag: 1ee64539f9b6e27372caa14a1b4cbdf68736ed43
+    location: https://github.com/edmundnoble/chainweb-storage.git
+    tag: eb3c0477aa728526b18b4013560b88488bd49fd0
```
larskuhtz, line 44: Is this change required for this PR?

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -235,7 +238,8 @@ pCutConfig = id
     <*< cutFetchTimeout .:: option auto
         % long "cut-fetch-timeout"
         <> help "The timeout for processing new cuts in microseconds"
-    -- cutInitialCutHeightLimit isn't supported on the command line
+    -- cutInitialBlockHeightLimit isn't supported on the command line
+    -- cutResetToCut isn't supported on the command line
```
larskuhtz, line 242: I think, it would be fine to expose this on the command line

src/Chainweb/CutDB.hs
```diff
@@ -395,8 +395,10 @@ startCutDb
     -> RocksDbCas CutHashes
     -> IO (CutDb cas)
 startCutDb config logfun headerStore payloadStore cutHashesStore = mask_ $ do
-    logg Debug "obtain initial cut"
+    logg Info "obtain initial cut"
     initialCut <- readInitialCut
+    -- TODO before merge: is this a good idea or not?
+    -- deleteRangeRocksDb (_getRocksDbCas cutHashesStore) (Just $ over _1 succ $ casKey $ cutToCutHashes Nothing initialCut, Nothing)
```
larskuhtz, line 401: What are we planning to do here? I think, deleting would make sense, but not 100% sure.

src/Chainweb/CutDB.hs
```diff
@@ -455,10 +457,12 @@ startCutDb config logfun headerStore payloadStore cutHashesStore = mask_ $ do
...
                         Nothing -> return hm
-                        Just h -> limitCutHeaders wbhdb h hm
-                            -- FIXME: this requires that the pact db is deleted because it
-                            -- interfers with rewind limit. As a fix temporarily disable
-                            -- rewind limit during initialization.
+                        Just h -> do
+                            let
+                                cutHeightTarget =
+                                    avgCutHeightAt v h -
+                                        CutHeight (int $ diameterAtCutHeight v (maxBound :: CutHeight) * chainCountAt v (maxBound :: BlockHeight))
```
larskuhtz, line 464: Why is this term needed? 

Can `cutHeightTarget` become negative?

src/Chainweb/Cut.hs
```diff
@@ -240,6 +242,35 @@ unsafeMkCut v hdrs = Cut'
     , _cutChainwebVersion = v
     }
 
+-- In some scenarios it's useful to reset a node's latest cut to some
+-- past cut, which we can specify either as the ancestor of the latest
+-- cut at some block height or as the cut sitting in a file at the indicated path.
+data CutResetTarget
```
larskuhtz, line 248: Where is this used? Are we planning to use this in a future iteration of the reset logic?

Maybe we should come up with a concrete list of scenarios for cut initialization and reset logic.

src/Chainweb/CutDB.hs
edmundnoble: I think deleting makes sense, because in practice this feature exists to recover from broken forks, and in that case we certainly don't want the old cut hanging around until the new fork becomes heavier. I'll uncomment this.

src/Chainweb/CutDB.hs
edmundnoble: This term compensates for the `avgCutHeightAt v h` potentially having some chains which are past the block height in question. In other words, with this term in here, we make sure that all of the chains are actually at or before the target block height at `cutHeightTarget`. Technically `cutHeightTarget` can become negative because of this but for the current chain graph we'd need to reset to a block height lower than 3.

src/Chainweb/CutDB.hs
```diff
@@ -505,16 +509,16 @@ processCuts
 processCuts conf logFun headerStore payloadStore cutHashesStore queue cutVar = do
     rng <- Prob.createSystemRandom
     queueToStream
-        & S.chain (\c -> loggc Debug c "start processing")
+        & S.chain (\c -> loggc Info c "start processing")
```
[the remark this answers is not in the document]
edmundnoble: We still have "got all prerequisites" below as `Info`, so we're already logging at least once per valid cut. This logging happens even for invalid cuts though, is that too much? I thought it was an okay amount of logs.

Ditto for below "fetch all prerequisites"

src/Chainweb/CutDB.hs
larskuhtz: Let's add `min 0 ...` to make sure that we don't go past genesis?

src/Chainweb/Cut.hs
larskuhtz: It depends if we plan to use that in the future. I think, the ability to reset to a particular cut could be useful to resolve forks or also for testing. But I am not sure.

edmundnoble: Deleted for now.

larskuhtz (approved):

PR #1354, opened 2022-02-01
NumCast instance only if platform allows

edmundnoble:


larskuhtz (approved):
Looks good to me, assuming that `MachDeps.h` is available with all GHC version that chainweb-node supports.

PR #1355, opened 2022-02-02
set locale to en_US.utf-8 in container images

larskuhtz:
Fixes the output of `chainweb-node --long-info`, which contains unicode characters.


edmundnoble:
I don't think this is necessarily trivial because it will affect all of our textual IO in chainweb-node, but it's probably an improvement regardless.

edmundnoble (approved):

PR #1361, opened 2022-02-11
Fork latest memory upgrades

jmcardon:


cabal.project
```diff
@@ -36,7 +36,7 @@ debug-info: True
 source-repository-package
     type: git
     location: https://github.com/kadena-io/pact.git
-    tag: 0ea3e6f63a2f677520fa425fa5f01984a7345160
+    tag: 151c45846c6cea0d31f875c9e77619ce9538810d
```
jmcardon, line 39: TODO: fix pact pin once https://github.com/kadena-io/pact/pull/964 is merged.

cabal.project
```diff
@@ -59,7 +59,7 @@ package vault
 package pact
     ghc-options: -Wwarn
     -- avoid conflict with cryptonite during linking
-    flags: +cryptonite-ed25519 -build-tool
+    flags: +cryptonite-ed25519 -build-tool +no-advice
```
jmcardon, line 62: This is the new flag to CPP advice out of pact.

cabal.project.freeze
```diff
@@ -209,7 +209,7 @@ constraints: any.Boolean ==0.2.4,
              any.old-time ==1.1.0.3,
              any.optparse-applicative ==0.16.1.0,
              optparse-applicative +process,
-             any.pact ==4.2.0,
+             any.pact ==4.2.0.1,
```
jmcardon, line 212: Not sure if this version needs to go

sirlensalot: No, it's basically a cw-only release which is fine

test/Chainweb/Test/Pact/PactInProcApi.hs
```diff
@@ -284,7 +286,7 @@ minerKeysetTest :: TestBlockDb -> IO (IORef MemPoolAccess) -> WebPactExecutionSe
 minerKeysetTest bdb _mpRefIO pact = do
 
   -- run past genesis, upgrades
-  forM_ [(1::Int)..3] $ \_i -> runCut' noMiner
+  forM_ [(1::Int)..24] $ \_i -> runCut' noMiner
```
sirlensalot, line 289: why so high? this makes the test slower (applies to change in Version too obviously)

sirlensalot (approved):
Approving since test issue is acceptable for merging

src/Chainweb/Version.hs
```diff
@@ -924,8 +924,8 @@ chainweb213Pact :: ChainwebVersion -> BlockHeight -> Bool
 chainweb213Pact Mainnet01 = (>= 2_447_358) -- 2022-02-26 00:00:18
 chainweb213Pact Testnet04 = (>= 1_977_816) -- 2022-02-25 00:00:18
 chainweb213Pact Development = (>= 95)
-chainweb213Pact (FastTimedCPM g) | g == petersonChainGraph = (>= 5)
-chainweb213Pact _ = const True
+chainweb213Pact (FastTimedCPM g) | g == petersonChainGraph = (> 25)
+chainweb213Pact _ = const False
```
larskuhtz, line 928: Is this needed? It means that in general tests only cover pre chainweb213 legacy behavior.

test/Chainweb/Test/Pact/SPV.hs
```diff
@@ -262,7 +262,7 @@ roundtrip'
     -> IO (CutOutputs, CutOutputs)
 roundtrip' v sid0 tid0 burn create step = withTestBlockDb v $ \bdb -> do
   tg <- newMVar mempty
-  withWebPactExecutionService v bdb (chainToMPA' tg) $ \pact -> do
+  withWebPactExecutionService v bdb (chainToMPA' tg) freeGasModel $ \pact -> do
```
larskuhtz, line 265: How much test coverage of gas computation code do we loose by using `freeGasModel` in so many places in the tests?

larskuhtz (commented):
I am a bit concerned that this PR may reduce test coverage in pact service:

* By setting the default value `chainweb213Pact` to `False` for most test chainweb versions, tests that use those versions only cover pre chainweb-2.13 legacy behavior.

* By using `freeGasModel` in many tests we exclude previously covered gas related code paths and logic from the tests.

test/Chainweb/Test/Pact/PactInProcApi.hs
jmcardon: It was making a test fail. 

The same thing that was causing the intermittent test failure for the pact 4.2.0 release wrt gas that I found with @LindaOrtega , there's a set of tests that rely on block height and the fork being applied too early (similar to when coin contract fork happened @ a low block height) makes some tests fail.

test/Chainweb/Test/Pact/SPV.hs
jmcardon: We don't lose any test coverage, we were already using `freeGasmodel`.

We can make a case for refactoring the tests that do this to use the proper gas model later on.

src/Chainweb/Version.hs
jmcardon: Changed, all tests pass with it.

larskuhtz (approved):

PR #1367, opened 2022-02-16
Add Rosetta README

LindaOrtega:


mightybyte (changes requested):
Can we move these files out of the `src` directory?  `src` is usually reserved for Haskell source. Maybe make a top level `rosetta` directory?

mightybyte (approved):

PR #1383, opened 2022-02-25
Add a savepoint to rewindBlock

edmundnoble:


emmanueldenloye (approved):
LGTM

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -758,7 +758,7 @@ handlePossibleRewind v cid bRestore hsh = do
         return $ fromIntegral txid
       where msg = "handlePossibleRewind: newChildBlock: error finding txid"
 
-    rewindBlock bh = do
+    rewindBlock bh = withSavepoint Rewind $ do
```
larskuhtz, line 761: How does this affect possible rollbacks within this safepoint? Is it guaranteed that those still have same effect as before?

larskuhtz (commented):
What is the motivation for this change?

Overall it looks good, but I try to better understand the effects.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
edmundnoble: Actually I'm not sure now that we need this PR, because `handlePossibleRewind` is always called within `withSavepoint PreBlock`. @emmanueldenloye thoughts?

emmanueldenloye: I've given some more thought to this. I don't think we need this PR.

PR #1385, opened 2022-03-02
Request size limit middleware

edmundnoble:


src/Chainweb/Chainweb.hs
```diff
@@ -572,6 +573,7 @@ runChainweb cw = do
                 $ httpLog
                 . throttle (_chainwebPutPeerThrottler cw)
                 . throttle (_chainwebThrottler cw)
+                . requestSizeLimitMiddleware defaultRequestSizeLimitSettings
```
larskuhtz, line 576: I think, we may also use this for the service api in line 581 below

larskuhtz (approved):

PR #1386, opened 2022-03-04
remove unused InMemoryCheckpointer

larskuhtz:


edmundnoble (commented):
Not sure why we had this in the first place, but clearly deleting it has no effect on the codebase, so that's fine with me.

edmundnoble (approved):
Emmanuel told me about the history, LGTM

PR #1387, opened 2022-03-07
Disable exclusive locking mode and log when savepoint escapes

edmundnoble:


larskuhtz (approved):

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -110,20 +110,22 @@ initRelationalCheckpointer' bstate sqlenv loggr v cid = do
...
     beginSavepoint Block
     return $! PactDbEnv' $! PactDbEnv chainwebPactDb dbenv
   where
     -- Module name fix follows the restore call to checkpointer.
     setModuleNameFix = bsModuleNameFix .= enableModuleNameFix v bh
     setSortedKeys = bsSortedKeys .= pact420Upgrade v bh
-doRestore _ _ dbenv Nothing = runBlockEnv dbenv $ do
+doRestore logger _ _ dbenv Nothing = runBlockEnv dbenv $ do
     clearPendingTxState
+    checkNotInTx logger
```
larskuhtz, line 128: Cf. comment above. I think, this can cause falls positives.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -152,6 +154,7 @@ doSave dbenv hash = runBlockEnv dbenv $ do
     -- FIXME: if any of the above fails with an exception the following isn't
     -- executed and a pending SAVEPOINT is left on the stack.
     commitSavepoint Block
+    checkNotInTx logger
```
larskuhtz, line 157: I think, this can trigger false positives as well. Although not every call of `withCheckpointer` calls `_cpSave`, I think, the decision is independent from wether the call is nested or not:

```.haskell
        try (restore (act cenv)) >>= \case
            Left !e -> discardTx checkPointer >> throwM @_ @SomeException e
            Right (Discard !result) -> discardTx checkPointer >> return result
            Right (Save header !result) -> saveTx checkPointer header >> return result
```

larskuhtz (changes requested):
In think all invocations of `checkNotInTx` can result in false error messages, because they don't take into account the nesting depth of checkpointer save points.

The values of `_psCheckpointerDepth` and `_psIsBatch` keep track of nesting of save points, and could e consulted to determine if invocations of `doRestore` and `doSave` are on the top-level tx save point.

I think, the introduction of `checkNotInTx` should be moved into a separate PR, since it is an unrelated (and less urgent) feature from the use of normal tx mode.

src/Chainweb/Pact/Backend/Utils.hs
```diff
@@ -212,14 +212,14 @@ chainwebPragmas :: [Pragma]
 chainwebPragmas =
   [ "synchronous = NORMAL"
   , "journal_mode = WAL"
-  , "locking_mode = EXCLUSIVE"
+  -- , "locking_mode = EXCLUSIVE"
+  , "locking_mode = NORMAL"
```
larskuhtz, line 216: Add a comment about the effects of the setting and the reason why it was changed

larskuhtz (approved):

PR #1388, opened 2022-03-09
Fix math so that cut pruning happens less often

edmundnoble:


src/Chainweb/CutDB.hs
```diff
@@ -537,7 +537,7 @@ processCuts conf logFun headerStore payloadStore cutHashesStore queue cutVar = d
 
     maybePrune rng curCutAvgBlockHeight = do
         r :: Double <- Prob.uniform rng
-        when (r > 1 / int (int (_cutDbParamsPruningFrequency conf) * chainCountAt v maxBound)) $
+        when (r < 1 / int (int (_cutDbParamsPruningFrequency conf) * chainCountAt v maxBound)) $
```
larskuhtz, line 540: I wonder if we have to document better the meaning of the `_cutDbParamsPruningFrequency` value. Right now it states:

```haskell
    , _cutDbParamsPruningFrequency :: BlockHeight
    -- ^ How often do we prune cuts (approximately)?
```

But a pruning is triggered with probability `1 / (frequency * chainCount)`. The default value for the frequency is 10000. So, on average every 200000 cuts we trigger cut pruning. I guess that's about once a week. The documentation should at least mention that the value is normalized by chain.

larskuhtz (approved):

PR #1396, opened 2022-03-22
Add principals to Coin Contract

emilypi:


pact/coin-contract/v4/coin-v4.pact
```diff
@@ -469,9 +469,7 @@
       (if (= "" r) true
```
sirlensalot, line 469: hmm ... the point is to stop parsing in pact altogether, really even the `check-reserved` call is slow. Suggest instead that we test `validate-principal` in a `let`, and if `true` return fast. if `false`, _then_ `check-reserved` and only fail if reserved (otherwise it's vanity so pass).

sirlensalot (commented):
One comment, plus needs repl coverage for at least one other principal case.

pact/coin-contract/v4/coin-v4.pact
emilypi: Works for me - I figured we'd want to optimize for the non-protocol account branch, but if we want to optimize for protocols, that works.

pact/coin-contract/v4/coin-v4.pact
```diff
@@ -465,16 +465,18 @@
...
-      (if (= "" r) true
-        (if (= "k" r)
-          (enforce
-            (= (format "{}" [guard])
-               (format "KeySet {keys: [{}],pred: keys-all}"
-                       [(drop 2 account)]))
-            "Single-key account protocol violation")
-          (enforce false
-            (format "Unrecognized reserved protocol: {}" [r]))))))
+    (let ((valid (validate-principal guard account)))
```
sirlensalot, line 468: looks like we don't need the `let`, just do 
```
    (if (validate-principal guard account)
      true 
      ...
```

sirlensalot (commented):
LG!! However tests are failing. One comment to save us a `let`

sirlensalot (approved):
Looks good! Just needs the final pin from pact and GTG

PR #1420, opened 2022-04-14
log legacy tx-hash and signatures issues at debug level

larskuhtz:


edmundnoble (approved):
Yeah I saw a crapton of these messages doing a replay, good change

PR #1430, opened 2022-05-09
Resolve module cache consistency issue on reset (take 2)

emmanueldenloye:


sirlensalot (approved):
Approved. @emilypi was going to take a look too. Also we need to have @edmundnoble attempt to reproduce his rewind issue, but we can do that after merge if that makes things easier.

edmundnoble:
Rewinding to 2134320 (the testnet fork happens at 2134331) works with this patch added.

PR #1434, opened 2022-05-11
OpenApi validation 

edmundnoble:
WIP because the code in other projects has to be reviewed first.


src/Chainweb/Utils.hs
```diff
@@ -407,6 +408,11 @@ minusOrZero :: Ord a => Num a => a -> a -> a
 minusOrZero a b = a - min a b
 {-# INLINE minusOrZero #-}
 
+-- | Analogous to `unsafeInterleaveIO` but doesn't hide the effect behind evaluation.
+interleaveIO :: IO a -> IO (IO a)
+interleaveIO act = evaluate <$> unsafeInterleaveIO act
+{-# INLINE interleaveIO #-}
```
enobayram, line 414: This is an interesting implementation of [`once`](https://hackage.haskell.org/package/io-memoize-1.1.1.0/docs/System-IO-Memoize.html#:~:text=once%20%3A%3A%20IO%20a%20%2D%3E%20IO%20(IO%20a)), (see also this [`once`](https://hackage.haskell.org/package/once-0.4/docs/Control-Once.html#v:once)) without introducing any dependency. So I'd recommend renaming this function to `once` since that's what it's often called.

Once caveat of this implementation is what happens in the presence of an async exception:
``` haskell
ghci> d <- evaluate <$> unsafeInterleaveIO (threadDelay 10000000)
ghci> d
^CInterrupted.
ghci> d
Interrupted.
```

I.e. the async exception gets trapped inside `a`, so I'd say this is a semantically slippery ground since an arbitrary `timeout` somewhere in the code has a chance of being the first one to ask for `a` and poisoning it for everyone else.

In comparison, [`io-memoize`](https://hackage.haskell.org/package/io-memoize-1.1.1.0/docs/System-IO-Memoize.html) retries the action in that case:
``` haskell
ghci> d <- once $ putStrLn "Starting" >> threadDelay 10000000
ghci> d
Starting
^CInterrupted.
ghci> d
Starting
```
Though to be fair, `io-memoize` always retries if it encounters an exception:
``` haskell
ghci> d <- once $ putStrLn "Starting" >> threadDelay 10000000 >> error "FAIL!"
ghci> d
Starting
*** Exception: FAIL!
  error, called at <interactive>:9:60 in interactive:Ghci5
ghci> d
Starting
*** Exception: FAIL!
  error, called at <interactive>:9:60 in interactive:Ghci5
```

I don't think there's any "right" behavior here, retrying on async exceptions might also invalidate the caller's assumption that the IO inside can be executed at most once, so I guess the least we can do is to mention this behavior in the haddock comment.

src/Chainweb/OpenAPIValidation.hs
```diff
@@ -0,0 +1,70 @@
...
+    where
+    findPact pactSpec rawVersion rawChainId rest = do
+        reqVersion <- chainwebVersionFromText (T.decodeUtf8 rawVersion)
+        guard (reqVersion == v)
+        reqChainId <- chainIdFromText (T.decodeUtf8 rawChainId)
+        guard (HS.member reqChainId (chainIds v))
+        return (BS8.intercalate "/" ("":rest), pactSpec)
+
+    fetchOpenApiSpecs = do
+        let chainwebUri = "https://raw.githubusercontent.com/kadena-io/chainweb-openapi/validation-fixes-3/chainweb.openapi.yaml"
```
enobayram, line 53: It seems a little iffy to me to point at a non-permanent link here. What exactly is the intention here? If I use a chainweb binary that was compiled before the latest update to the `validation-fixes-3` branch, why is it an error that my binary doesn't conform to the spec there? Why is it the `validation-fixes-3` branch anyway?

In any case, the reasoning here needs explanation.

src/Chainweb/OpenAPIValidation.hs
```diff
@@ -0,0 +1,70 @@
...
+        WV.mkValidator apiCoverageRef (WV.Log logValidationFailure (logApiCoverage apiCoverageLogTimeRef)) $ \path -> asum
+            [ case BS8.split '/' path of
+                ("" : "chainweb" : "0.0" : rawVersion : "chain" : rawChainId : "pact" : "api" : "v1" : rest) -> do
+                    findPact pactSpec rawVersion rawChainId rest
+                ("" : "chainweb" : "0.0" : rawVersion : "chain" : rawChainId : "pact" : rest) -> do
+                    findPact pactSpec rawVersion rawChainId rest
+                _ -> Nothing
+            , (,chainwebSpec) <$> BS8.stripPrefix (T.encodeUtf8 $ "/chainweb/0.0/" <> chainwebVersionToText v) path
+            , Just (path,chainwebSpec)
+            ]
```
enobayram, line 43: Why do we have to do this kind of analysis for determining which endpoint the request has came to? Why not make two independent validator middlewares, one for the p2p and another for the service endpoints and inject them separately into the corresponding servers?

enobayram (commented):
@edmundnoble Thanks for this PR, seems very useful. I've asked for some clarifications and made some style recommendations in my review. Would you mind updating the PR description to indicate what we're trying to achieve here and where is it intended to be used. Is there any issue related to this PR?

Also, you've mentioned in the existing PR description that this code depends on the review of code in other projects, which projects are they?

test/Chainweb/Test/Utils.hs
```diff
@@ -647,20 +657,67 @@ withTestAppServer tls v appIO envIO userFunc = bracket start stop go
...
+    , vResp :: (ResponseHeaders, Status, BL.ByteString)
+    , vErr :: WV.TopLevelError
+    }
+    deriving (Show, Typeable)
+
+instance Exception ValidationException
+
+{-# NOINLINE chainwebOpenApiSpec #-}
+chainwebOpenApiSpec :: OpenApi
+chainwebOpenApiSpec = unsafePerformIO $ do
```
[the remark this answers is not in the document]
edmundnoble: It is actually helping us a lot, because it is never evaluated more than once, despite the underneath `evaluate` being executed more than once. Without this, we quickly ran into github's rate limiting.

src/Chainweb/Utils.hs
edmundnoble: This is not the behavior that I expected, so I'll leave a comment and rename.

src/Chainweb/OpenAPIValidation.hs
edmundnoble: My explanation is, we need review on the validation-fixes-3 branch to merge it into master :D

src/Chainweb/OpenAPIValidation.hs
edmundnoble: > Why not make two independent validator middlewares, one for the p2p and another for the service endpoints and inject them separately into the corresponding servers?
That isn't quite what this does. We have two OpenAPI specs, one for the Pact API, and one for the chainweb-node API. The service API includes both the chainweb-node API and the Pact API, the P2P API includes only part of the chainweb-node API. So while the Pact stuff could be excluded from the P2P API, this code would still be necessary for the service API endpoint at least.

enobayram: I see, that makes sense, thanks! 

There's some non-trivial coupling here. As far as I can tell, this mixes the routing of the request to the correct API spec with some pre-processing of the `path` before it's passed to the validator. Why do we need this kind of path pre-processing? I guess this means there's a gap between the API spec and the actual API, so where does this gap come from, is it something like a middleware defined by the `chainweb-node` codebase? Or does it come from the behavior of some external middleware like `nginx`? Is this special path pre-processing explained inside the spec itself, or maybe it's supposed to be legacy?  

Anyway, I guess these questions are mostly outside the scope of this PR, but I think it could be nice to decouple the spec-selection from the spec-specific path pre-processing. For example how about a refactor like this to make it more clear: https://github.com/kadena-io/chainweb-node/commit/0522c11c25ace2eaa0dc435499e30b4d7ecc30ba I usually do this kind of decoupling so that I have natural places in the code to explain why "surprising" logic like this exists, maybe write a comment to point to another place in the code that we're implementing the inverse of?

src/Chainweb/Miner/RestAPI/Server.hs
```diff
@@ -99,16 +105,28 @@ solvedHandler
...
-                    return $ Left err404 { errBody = "No associated Payload" }
+                    return $ Left err404
+                        { errBody = "No associated Payload"
+                        , errHeaders = [("Content-Type", "text/plain;charset=utf-8")]
+                        }
                 , E.Handler $ \(InvalidSolvedHeader _ msg) ->
-                    return $ Left err400 { errBody = "Invalid solved work: " <> toErrText msg}
+                    return $ Left err400
+                        { errBody = "Invalid solved work: " <> toErrText msg
+                        , errHeaders = [("Content-Type", "text/plain;charset=utf-8")]
```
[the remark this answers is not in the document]
enobayram: Yeah, maybe a better abstraction is `errPlain :: Int -> String -> ServantErr`, but the moral of my suggestion here is that capturing the connection between the body being a `String` and the corresponding `Content-Type` being a `text/plain;charset=utf-8` is a minor improvement.

src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -233,7 +233,10 @@ sendHandler logger mempool (SubmitBatch cmds) = Handler $ do
        Left err -> failWith $ "Validation failed: " <> err
   where
     failWith :: String -> ExceptT ServerError IO a
-    failWith err = throwError $ err400 { errBody = BSL8.pack err }
+    failWith err = throwError $ err400
+        { errBody = BSL8.pack err
```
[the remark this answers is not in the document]
enobayram: Yeah, I think so.

test/Chainweb/Test/Utils.hs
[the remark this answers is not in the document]
enobayram: Right! For some reason, I got the impression that the `withChainwebTestServer` below was something we were calling only once as part of the test suite setup, but I now see that's wrong.

src/Chainweb/OpenAPIValidation.hs
enobayram: :D that's definitely an improvement! Still though, when we merge new changes into `chainweb-openapi`'s `master` branch,  existing node binaries will start validating differently. Does that make sense? Especially in the context of tests, is it OK that if I go back and execute the tests of a `chainweb-node` commit from a year ago, they will fail since the API spec living in the `master` branch will have changed?

edmundnoble: It's not great, I would prefer that the spec be inside of this repo.

cabal.project
```diff
@@ -52,6 +52,11 @@ package yet-another-logger
 -- -------------------------------------------------------------------------- --
 -- Source Repository Packages
 
+source-repository-package
+    type: git
+    location: https://github.com/edmundnoble/wai-middleware-validation.git
+    tag: ea2f735d7e15f72b572182153bbd9b0a7ae57afe
+
```
larskuhtz, line 59: maybe publish on Hackage?

edmundnoble:
@chessai and I were successful running the integration tests with this branch.

src/Chainweb/Chainweb.hs
```diff
@@ -674,7 +690,7 @@ runChainweb cw = do
 
     -- collect server resources
     proj :: forall a . (ChainResources logger -> a) -> [(ChainId, a)]
-    proj f = map (second f) chains
+    proj f = chains & mapped . _2 %~ f
```
chessai, line 693: Nit:

this change seems extraneous, is lens the preferred style?
personally I find the old version more readable

src/Chainweb/OpenAPIValidation.hs
```diff
@@ -0,0 +1,70 @@
...
+    where
+    findPact pactSpec rawVersion rawChainId rest = do
+        reqVersion <- chainwebVersionFromText (T.decodeUtf8 rawVersion)
+        guard (reqVersion == v)
+        reqChainId <- chainIdFromText (T.decodeUtf8 rawChainId)
+        guard (HS.member reqChainId (chainIds v))
+        return (BS8.intercalate "/" ("":rest), pactSpec)
+
+    fetchOpenApiSpecs = do
+        let chainwebUri = "https://raw.githubusercontent.com/kadena-io/chainweb-openapi/main/chainweb.openapi.yaml"
```
chessai, line 53: are these going to change at all? or are these the standard locations for these openapi definitions?

src/Chainweb/OpenAPIValidation.hs
```diff
@@ -0,0 +1,70 @@
...
+        chainwebSpec <- Yaml.decodeThrow . BL.toStrict . HTTP.responseBody =<< HTTP.httpLbs (HTTP.parseRequest_ chainwebUri) mgr
```
chessai, line 54: Nit:

Consider making a helper function
```haskell
getSpec :: (FromJSON a) => String -> IO a
getSpec uri = Yaml.decodeThrow . BL.toStrict . HTTP.requestBody =<< HTTP.httpLbs (HTTP.parseRequest_ uri) mgr
```

src/Chainweb/RestAPI/Utils.hs
```diff
@@ -137,6 +142,27 @@ type family ReassocBranch (a :: s) (b :: [Type]) :: Type where
...
     ReassocBranch a (b ': rest) = a :> ReassocBranch b rest
 
+newtype DuplicateHeader = DuplicateHeader HTTP.HeaderName
+instance Show DuplicateHeader where
+    show (DuplicateHeader hn) = "Attempted to set a duplicate header with this name: " <> show hn
+instance Exception DuplicateHeader
+
+addHeader :: HTTP.Header -> [HTTP.Header] -> [HTTP.Header]
+addHeader (hn, hv) hs =
+    (hn, hv) : [ if n == hn then throw (DuplicateHeader hn) else h | h@(n, _) <- hs ]
```
chessai, line 152: I think `find` is more readable/clear than this list comprehension
```suggestion
    case find ((== hn) . fst) hs of
      Just _ -> throw (DuplicateHeader hn)
      Nothing -> (hn, hv) : hs
```

chessai: (Alternatively with `isJust`)

src/Chainweb/Utils.hs
chessai: +1 to renaming and commenting about the behaviour.

test/Chainweb/Test/RestAPI.hs
```diff
@@ -122,7 +122,7 @@ noMempool = []
 simpleSessionTests :: RocksDb -> Bool -> ChainwebVersion -> TestTree
 simpleSessionTests rdb tls version =
     withBlockHeaderDbsResource rdb version $ \dbs ->
-        withBlockHeaderDbsServer tls version dbs (return noMempool)
+        withBlockHeaderDbsServer True tls version dbs (return noMempool)
```
chessai, line 125: can this be a variable instead of a raw Bool? Since this is just `True`, I can't immediately tell what the context of the argument is, without going to the function. for `tls`, `version`, and `dbs`, I can immediately infer what they are based on their names.

test/Chainweb/Test/RestAPI.hs
```diff
@@ -328,7 +328,7 @@ simpleClientSession envIO cid =
 
 pagingTests :: RocksDb -> Bool -> ChainwebVersion -> TestTree
 pagingTests rdb tls version =
-    withBlockHeaderDbsServer tls version
+    withBlockHeaderDbsServer True tls version
```
chessai, line 331: same here

test/Chainweb/Test/Utils.hs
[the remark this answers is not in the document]
chessai: +1 to the use of unsafePerformIO with NOINLINE here

test/Chainweb/Test/Utils.hs
```diff
@@ -649,20 +659,68 @@ withTestAppServer tls v appIO envIO userFunc = bracket start stop go
...
+        coverageRef <- newIORef $ WV.CoverageMap Map.empty
+        _ <- evaluate chainwebOpenApiSpec
+        _ <- evaluate pactOpenApiSpec
+        let
+            lg (_, req) (respBody, resp) err = do
+                let ex = ValidationException req (W.responseHeaders resp, W.responseStatus resp, respBody) err
+                print $ ppShow ex
+                error "validation error"
+            findPath path = asum
+                [ case B8.split '/' path of
```
chessai, line 708: This code feels duplicated from the code in the other module. Perhaps it should be extracted to some kind of `Util` function?

src/Chainweb/Chainweb.hs
jwiegley: The old version is actually the preferred style in our code. We don't tend to use the Lens operators as much.

chessai (approved):
Remaining change requests are mostly nits, I don't think they are entirely necessary.

test/Chainweb/Test/RestAPI.hs
edmundnoble: Made a one-off ADT.

src/Chainweb/RestAPI/Utils.hs
edmundnoble: That would be clearer, I did this for extra laziness but it doesn't seem worth it now.

src/Chainweb/OpenAPIValidation.hs
edmundnoble: This is where we can find the latest spec, ideally we could configure it somehow, maybe with an envvar.

chessai (approved):

src/P2P/Node/Configuration.hs
```diff
@@ -160,8 +164,9 @@ instance ToJSON P2pConfiguration where
         , "private" .= _p2pConfigPrivate o
         , "bootstrapReachability" .= _p2pConfigBootstrapReachability o
         ]
-        -- hidden: Do not print the default value. Included only if explicitely set to False
```
jwiegley, line 163: Or to True?

larskuhtz (commented):
Only reviewed production relevant code paths, which looks good to me.

I wonder if this could be moved into a speparate binary that functions as reverse proxy in order to even more separate it from prod code.

Or alternatively include only via CPP and cabal flag in order to not link the code into the production binary at all?

PR #1438, opened 2022-05-14
freeze file for version 2.14.1

larskuhtz:


sirlensalot (commented):
What is the process for review? I checked CI and there are still differences on GHC 8.10 builds on ubuntu boxen which I guess is the target OS/GHC for the freeze file? 

https://github.com/kadena-io/chainweb-node/runs/6436496838?check_suite_focus=true#step:13:14

```

Run diff -w <(git show HEAD:cabal.project.freeze) cabal.project.freeze || true
351c351
<              any.wai-extra ==3.1.12,
---
>              any.wai-extra ==3.1.12.1,
370c370
< index-state: hackage.haskell.org 2022-05-14T00:56:28Z
---
> index-state: hackage.haskell.org 2022-05-14T18:06:58Z
```

larskuhtz:
Yes, different builds can result in different freeze files. The main reason for differences is usually the GHC version. But it can also depend on the time of the build (Hackage snapshot), state of dist-cache, and/or platform.

In the freeze file we track dependency changes for the GHC and os version that we use to create the docker images.

Diffs are recorded and logged in each CI build and a manual review of changes is part of the release check list.

larskuhtz:
For this release interesting changes were updates to `megaparsec` and `parsers`. I checked the changelogs for both upgrades on Hackage and they are fine. I put notes about that in the release checklist for 2.14.1 in the infra repo.

There's also the complete mainnet replay to check for possible incompatibilities.

sirlensalot (approved):

PR #1470, opened 2022-06-27
Fix our ixset-typed work on the PeerEntry population

jwiegley:
The basic idea of this PR is as follows:

We do a lot of work using `ixset-typed` to determine a collection of peers, and then randomly walk through the peers in `findNextPeer`. The problem is that much of the ixset work happens within an STM block, prior to a `retry` if we don't find the peer we're looking for. Debug prints confirm that this `retry` can happen 12-20 times before a node is chosen -- *even though the set of peers and sessions is unchanged from the previous attempt*. This means redoing the ixset work over and over until the desired node is found.

This PR changes the code to cache the set of peer entries until the set of peers or sessions actually changes. That way, the node can just randomly walk the same set of entries over and over until there is a change in the P2P situation. This eliminates a ton of work and memory consumption. @larskuhtz said this is similar to an idea he implemented a while ago, so it's a nice confirmation that this change is just rediscovering a previous idea.

With the PR, `ixset-typed` drops out of the top profiling results entirely; this is what catchup now looks like after a few minutes:
```
        Mon Jun 27 16:53 2022 Time and Allocation Profiling Report  (Final)

           chainweb-node +RTS -N -H1G -A64M -p -RTS --config-file chainweb-node.config --disable-node-mining

        total time  =      126.03 secs   (355420 ticks @ 1000 us, 20 processors)
        total alloc = 478,527,150,120 bytes  (excludes profiling overheads)

COST CENTRE                 MODULE                          SRC                                                   %time %alloc

allocAndFreeze              Data.ByteArray.Methods          Data/ByteArray/Methods.hs:74:1-45                      18.0    6.4
jstring                     Data.Aeson.Parser.Internal      src/Data/Aeson/Parser/Internal.hs:320:1-32              9.6    7.2
runPutS                     Data.Bytes.Put                  src/Data/Bytes/Put.hs:253:1-18                          6.1   42.2
hashMutableUpdate           Crypto.Hash.IO                  Crypto/Hash/IO.hs:(52,1)-(57,70)                        5.4    2.2
newMerkleLog                Chainweb.Crypto.MerkleLog       src/Chainweb/Crypto/MerkleLog.hs:(527,1)-(533,78)       3.6    4.0
decodeWithTable/fill        Data.ByteString.Base64.Internal Data/ByteString/Base64/Internal.hs:(174,62)-(197,68)    3.6    4.5
array_                      Data.Aeson.Parser.Internal      src/Data/Aeson/Parser/Internal.hs:172:35-59             3.4    2.5
genericParseJSON            Data.Aeson.Types.FromJSON       src/Data/Aeson/Types/FromJSON.hs:264:1-60               2.3    1.8
peek8                       Data.ByteString.Base64.Internal Data/ByteString/Base64/Internal.hs:38:1-12              1.7    1.7
fromList                    Data.HashMap.Internal.Strict    Data/HashMap/Internal/Strict.hs:631:1-64                1.5    0.1
parseJSON                   Pact.Types.Orphans              src/Pact/Types/Orphans.hs:95:15-36                      1.4    1.1
object_                     Data.Aeson.Parser.Internal      src/Data/Aeson/Parser/Internal.hs:135:46-89             1.4    1.2
termCodec.dec.parseWithInfo Pact.Types.Term                 src/Pact/Types/Term.hs:1301:62-77                       1.4    0.9
encodeMerkleInputNode       Chainweb.Crypto.MerkleLog       src/Chainweb/Crypto/MerkleLog.hs:739:1-59               1.2    1.1
genesisBlockHeader'         Chainweb.BlockHeader.Genesis    src/Chainweb/BlockHeader/Genesis.hs:(286,1)-(306,84)    1.2    0.8
unsafeInsert                Data.HashMap.Internal           Data/HashMap/Internal.hs:(909,1)-(939,76)               1.1    0.8
unescapeText                Data.Aeson.Parser.UnescapePure  src-pure/Data/Aeson/Parser/UnescapePure.hs:254:1-70     0.8    1.2
```
Still a lot of memory passing through the allocator, but at least the worst offender has been dealt with.

- Catchup speed is MUCH faster, now showing 50 blocks per sec consistently.
- Resident memory (heap) is down from 3.6 GB to 2.35 GB on macOS.
- CPU utilization is more consistent, 180-200% on average, whereas before this PR it was very volatile and could peak near 2000% (20 virtual core machine).

### DIFF
--- a/src/P2P/Node.hs
+++ b/src/P2P/Node.hs
@@ -193,6 +193,8 @@ makeLenses ''P2pSessionInfo
 -- -------------------------------------------------------------------------- --
 -- P2P Node State
 
+type SessionsMap = M.Map PeerInfo (P2pSessionInfo, Async (Maybe Bool))
+
 -- | P2P Node State
 --
 data P2pNode = P2pNode
@@ -200,7 +202,7 @@ data P2pNode = P2pNode
     , _p2pNodeChainwebVersion :: !ChainwebVersion
     , _p2pNodePeerInfo :: !PeerInfo
     , _p2pNodePeerDb :: !PeerDb
-    , _p2pNodeSessions :: !(TVar (M.Map PeerInfo (P2pSessionInfo, Async (Maybe Bool))))
+    , _p2pNodeSessions :: !(TVar SessionsMap)
     , _p2pNodeManager :: !HTTP.Manager
     , _p2pNodeLogFunction :: !LogFunction
     , _p2pNodeStats :: !(TVar P2pNodeStats)
@@ -470,16 +472,11 @@ syncFromPeer node info = do
 -- -------------------------------------------------------------------------- --
 -- Sample Peer from PeerDb
 
--- | Sample next active peer. Blocks until a suitable peer is available
---
--- @O(_p2pConfigActivePeerCount conf)@
---
-findNextPeer
+getPeersAndSessions
     :: P2pConfiguration
     -> P2pNode
-    -> STM PeerEntry
-findNextPeer conf node = do
-
+    -> STM (PeerSet, SessionsMap)
+getPeersAndSessions conf node = do
     -- check if this node is active. If not, don't create new sessions,
     -- but retry until it becomes active.
     --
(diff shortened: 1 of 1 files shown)


src/P2P/Node.hs
```diff
@@ -546,31 +541,35 @@ findNextPeer conf node = do
             i <- randomR node (0, max 1 (length s) - 1)
```
larskuhtz, line 541: The use of `randomR` has no semantic effect on success or failure of the transaction. However, we expect a lot of contention on the PRNG variable because all threads make use of it.

Therefor the use of randomR should be isolated in a separate STM transaction, or, probably even better, be replaced by a concurrency mechanism that performance better under high contention.

PR #1472, opened 2022-06-27
add version property for block gas limit

larskuhtz:


edmundnoble:
@larskuhtz we discussed having a two-stage approach to this PR: first hardcoding an upper limit to the configured limit, then forking and making that the only limit. This is the first stage.

src/Chainweb/Version.hs
```diff
@@ -751,6 +753,24 @@ workSizeBytes
...
+--
+blockGasLimit
+    :: ChainwebVersion
+    -> ChainId
+    -> BlockHeight
+    -> Natural
+blockGasLimit Mainnet01 _ _ = 180000
+blockGasLimit Testnet04 _ _ = 180000
+blockGasLimit Development _ _ = 180000
+blockGasLimit _ _ _ = 180000
```
larskuhtz, line 772: Maybe rename to `maxBlockGasLimit` as used above, to avoid confusion with the user-configured `blockGasLimit`?

src/Chainweb/Pact/PactService.hs
```diff
@@ -497,11 +497,11 @@ execNewBlock mpAccess parent miner = do
                 <> " (parent height = " <> sshow pHeight <> ")"
                 <> " (parent hash = " <> sshow pHash <> ")"
 
-        blockGasLimit <- view psBlockGasLimit
-        let initState = BlockFill blockGasLimit mempty 0
+        blockGasLim <- view psBlockGasLimit
+        let initState = BlockFill blockGasLim mempty 0
```
larskuhtz, line 501: What is the reason for this change? Avoiding a name clash or just a redundant change?

edmundnoble (approved):

src/Chainweb/Version.hs
```diff
@@ -751,6 +753,24 @@ workSizeBytes
...
 {-# INLINE workSizeBytes #-}
 
+-- -------------------------------------------------------------------------- --
+-- Pact Validation Parameters
+
+-- | This the hard upper limit of the gas within a block. Blocks that use more
+-- gas are invalid and rejected. This limit is needed as a DOS protection.
+--
+-- Smaller limits can be configured for creating new blocks.
+--
```
larskuhtz, line 763:
```suggestion
--
-- WARNING: this isn't yet enforced as block validation property. The current use of this ignores the
-- block height and use the value only during new block creation. Future versions of chainweb-node
-- will enforce this limit during block validation.
--
```

larskuhtz (commented):
One suggestion. Also, can we add a unit test for this? E.g. initializing pact service and a test mempool access, and checking that new blocks respect the limit?

PR #1473, opened 2022-06-28
fail early during p2p peer selection

larskuhtz:
This PR is greatly motivated by the ideas and insights from #1470:

*   Moving the `_p2pConfigMaxSessionCount` check out of the fold will cause the transaction to fail early and prevent retrying the transaction for each single peer in the peer database.

    Ideally this reduces the overhead of P2P session creation by a factor of that is equal to the number of peers in the peer database. The current median number of peers on mainnet nodes is 360.

*   The latest iteration of this PR uses STM only for awaiting a non-empty set of candidate peers. The actual peer selection is not transactional, which is supposed to greatly reduce contention.

*   Expensive transactional computations that are also less likely to fail and cause retries are moved towards the end of the STM transaction.

*   The PRNG is stored in an `IORef` and accessed via `atomicModifyIORef'`. This should further reduce contention, because the PRNG if accessed and mutated by all all peer selection calls but shouldn't actually cause a transaction to retry.

### DIFF
--- a/src/P2P/Node.hs
+++ b/src/P2P/Node.hs
@@ -78,7 +78,6 @@ import Control.Monad.STM
 import Control.Scheduler (Comp(..), traverseConcurrently)
 
 import Data.Aeson hiding (Error)
-import Data.Foldable
 import Data.Function
 import Data.Hashable
 import qualified Data.HashSet as HS
--- a/src/P2P/Node.hs
+++ b/src/P2P/Node.hs
@@ -497,13 +497,15 @@ findNextPeer conf node = do
     --
     check (sessionCount < peerCount)
 
+    -- Retry if there are more active sessions than the maximum number
+    -- of sessions
+    check (int sessionCount < _p2pConfigMaxSessionCount conf)
+
     -- Create a new sessions with a random peer for which there is no active
     -- sessions:
 
     let checkPeer (n :: PeerEntry) = do
             let !pid = _peerId $ _peerEntryInfo n
-            -- can this check be moved out of the fold?
-            check (int sessionCount < _p2pConfigMaxSessionCount conf)
             check (M.notMember (_peerEntryInfo n) sessions)
             check (pid /= myPid)
             return n
@@ -537,14 +539,13 @@ findNextPeer conf node = do
 #else
 
     -- random circular shift of a set
-    let shift i s = uncurry (++)
-            $ swap
-            $ splitAt (fromIntegral i)
-            $ s
+    let shift i = uncurry (++)
+            . swap
+            . splitAt (fromIntegral i)
 
         shiftR s = do
             i <- randomR node (0, max 1 (length s) - 1)
-            return $! shift i s
+            return $ shift i s
 
     let p0 = toList
             $ IXS.getGT (ActiveSessionCount 0)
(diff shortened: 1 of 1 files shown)


jwiegley:
Hi Lars, I think this isn’t doing quite the same thing as the larger PR. All of the IxSet work will still be being repeated in this version with each `retry`, because it has to evaluate all of the query instructions on the same dataset.

larskuhtz:
yeah, it would still have to evaluate those. But, it would be a chance that is guaranteed to be safe and may already capture the vast majority of the perf improvement of #1470.

#1470 would require a more thorough review with respect to corner cases -- which would still be worth doing. This one would be an insta-merge -- assuming it actually makes a perf difference.

jwiegley:
I much prefer the way the code is organized here, but there are still too many IxSet manipulation happening within a retried block. Below is the new profile. Running this, I only see 50 blocks progressing every 5 seconds or so.

A question I have is: Is it bad if we pick a "stale peer"? I.e., if the peer list changes and we use a value that might be slightly out of date?

```
	Tue Jun 28 08:21 2022 Time and Allocation Profiling Report  (Final)

	   chainweb-node +RTS -N -H1G -A64M -p -RTS --config-file chainweb-node.config --disable-node-mining

	total time  =       72.34 secs   (203995 ticks @ 1000 us, 20 processors)
	total alloc = 124,553,299,304 bytes  (excludes profiling overheads)

COST CENTRE                 MODULE                          SRC                                                   %time %alloc

runForever                  Chainweb.Utils                  src/Chainweb/Utils.hs:(926,1)-(932,54)                 23.7    2.7
delete                      Data.IxSet.Typed.Ix             src/Data/IxSet/Typed/Ix.hs:(102,1)-(105,64)             9.4   14.8
fromList                    Data.IxSet.Typed.Ix             src/Data/IxSet/Typed/Ix.hs:(95,1)-(96,77)               7.8   10.8
p2pStartNode                P2P.Node                        src/P2P/Node.hs:(741,1)-(743,74)                        7.2    1.6
insert                      Data.IxSet.Typed.Ix             src/Data/IxSet/Typed/Ix.hs:86:1-76                      3.4    3.8
jstring                     Data.Aeson.Parser.Internal      src/Data/Aeson/Parser/Internal.hs:320:1-32              3.1    1.9
allocAndFreeze              Data.ByteArray.Methods          Data/ByteArray/Methods.hs:74:1-45                       2.6    2.3
getEQ                       Data.IxSet.Typed                src/Data/IxSet/Typed.hs:910:1-17                        2.4    5.0
MAIN                        MAIN                            <built-in>                                              2.4    0.3
runPutS                     Data.Bytes.Put                  src/Data/Bytes/Put.hs:253:1-18                          1.3   15.9
getObject                   Data.ASN1.Parse                 Data/ASN1/Parse.hs:(95,1)-(99,42)                       1.1    2.2
incrementActiveSessionCount P2P.Node.PeerDB                 src/P2P/Node/PeerDB.hs:(434,1)-(435,75)                 1.0    0.0
decodeWithTable/fill        Data.ByteString.Base64.Internal Data/ByteString/Base64/Internal.hs:(174,62)-(197,68)    1.0    2.1
newMerkleLog                Chainweb.Crypto.MerkleLog       src/Chainweb/Crypto/MerkleLog.hs:(527,1)-(533,78)       0.5    1.3
uintOfBytes                 Data.ASN1.Internal              Data/ASN1/Internal.hs:23:1-87                           0.2    1.1
```

jwiegley (approved):

PR #1475, opened 2022-06-28
Remove uses of unsafeThaw in the mempool and checkpointer

edmundnoble:
I am pretty sure writing to an `unsafeThaw`ed vector is never safe, and there is no way for it to improve our performance here anyway... more likely it worsens our performance. I get not having `mutableVectorFromList` is annoying though, so I added that.

(see https://stackoverflow.com/questions/54041813/just-how-unsafe-are-data-vectors-unsafefreeze-unsafethaw)

### DIFF
--- a/src/Chainweb/Utils.hs
+++ b/src/Chainweb/Utils.hs
@@ -62,6 +62,7 @@ module Chainweb.Utils
 , (&)
 , IxedGet(..)
 , minusOrZero
+, mutableVectorFromList
 
 -- * Encoding and Serialization
 , EncodingException(..)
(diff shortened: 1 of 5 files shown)


src/Chainweb/Utils.hs
```diff
@@ -404,6 +407,17 @@ minusOrZero :: Ord a => Num a => a -> a -> a
...
 {-# INLINE minusOrZero #-}
 
+-- | Equivalent to V.thaw . V.fromList but by inspection probably faster.
+mutableVectorFromList
+    :: PrimMonad m
+    => [a]
+    -> m (MV.MVector (PrimState m) a)
+mutableVectorFromList as = do
+    vec <- MV.unsafeNew (length as)
+    forM_ (zip [0..] as) $ uncurry (MV.unsafeWrite vec)
```
larskuhtz, line 417: This traverses the list twice, which could be an issue for very large lists. For moderately sized lists it probably is fine.

larskuhtz (approved):

src/Chainweb/Utils.hs
edmundnoble: I'm fairly sure (though not familiar enough with how Vector does its fusion) that `V.fromList` does it as well, so at least it isn't a loss.

PR #1477, opened 2022-06-29
Always finalize SQLite statements and make ForkingBench more efficient

larskuhtz:


jwiegley:
`bracket` is goodness

PR #1479, opened 2022-06-29
Debug/benchmark tests

DevopsGoth:
Limit heap and stack size to prevent Github Runner memory issues from affecting benchmarks.

There is possibly a race condition in one of the benchmark tests as well.  We will see if over time the memory limits here help with the intermittent benchmark failures.


edmundnoble (changes requested):
Actually while we have this PR, can you try deleting `-N`? In my testing, higher `-N` values were more likely to trigger the benchmark failure.

.github/workflows/applications.yml
```diff
@@ -355,7 +355,7 @@ jobs:
         ghc: ${{ matrix.ghc }}
     - name: Run all benchmarks
       timeout-minutes: 12
-      run: bench --json $BENCH_FILE +RTS -N
+      run: bench --json $BENCH_FILE +RTS -H4G -K512m -RTS
```
larskuhtz, line 358: I think, we should not remove it. In production we run with `-N`, and even though this is "just" a benchmark it would reduce coverage or my skew the results. It should not fail with `-N`, so removing it, potentially hide a bug. 

We could, however, consider hardcoding it to a reasonably small value, say `-N4` or `-N2`.

larskuhtz (changes requested):

.github/workflows/applications.yml
larskuhtz: @edmundnoble what are your thoughts on this? What about using `-N2` or `-N4`?

emilypi: While i don't think we want to run these benchmarks in serial, github runners have 2 cores dedicated for linux/windows and 3 for macos. Seems like `-N4` would make the most sense to me.

larskuhtz:
The benchmark failures were actually caused by a bug in the benchmarking code, which was fixed in  #1483.

PR #1480, opened 2022-06-30
Update pact pin

jwiegley:


emilypi (approved):

larskuhtz:
@emilypi is this pact version backwards compatible with the current pact pin?

jwiegley:
The only change between this pin and the previous is that the `direct-sqlite` dependency has bumped by a mini sub-version. I think that should be compatible, unless another project has a fixed dependency on `direct-sqlite` at the previous version.

PR #1483, opened 2022-07-01
remove pow mining from forking bench

larskuhtz:


bench/Chainweb/Pact/Backend/ForkingBench.hs
```diff
@@ -104,13 +99,20 @@ import Chainweb.Version.Utils
...
     withArgs args $ C.defaultMain [bench rdb]
 
+-- -------------------------------------------------------------------------- --
+-- Benchmarks
+
 bench :: RocksDb -> C.Benchmark
 bench rdb = C.bgroup "PactService"
-    [ withResources rdb 10 Quiet forkingBench
+    [ forkingBench
+    , nonForkingBench
```
sirlensalot, line 115: Is this a new test?

sirlensalot (approved):

bench/Chainweb/Pact/Backend/ForkingBench.hs
larskuhtz: yes, but it can probably be removed now. It was a previous attempt  to scope the issue that we were investigating.

jwiegley (approved):

PR #1485, opened 2022-07-01
Use singletonChainGraph in ForkingBench

edmundnoble:
Now that we know what the issue was with ForkingBench, we know that the headers are being created with no adjacent hashes anyway, so we have to use a `singletonChainGraph`.

### DIFF
--- a/bench/Chainweb/Pact/Backend/ForkingBench.hs
+++ b/bench/Chainweb/Pact/Backend/ForkingBench.hs
@@ -380,7 +380,7 @@ cid :: ChainId
 cid = someChainId testVer
 
 testVer :: ChainwebVersion
-testVer = FastTimedCPM petersonChainGraph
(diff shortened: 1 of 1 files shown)


larskuhtz (approved):

larskuhtz (changes requested):
ForkingBench is failing with pact validation errors with this change.

One thing to check is whether there is a mismatching Chainweb-version hardcoded somewhere in the benchmark code.

edmundnoble:
I've looked for a mismatching version hardcoded anywhere in this benchmark and haven't been able to find it... seems the problem is more complex than that. The validation issue is only for the payload and only for block 1. If I make the blocks all empty of course that fixes the issue. I've also removed every hardcoded `petersonChainGraph` from Chainweb.Version, no dice there either.

PR #1486, opened 2022-07-02
Build a proper default.nix with a modern nixpkgs pin

jwiegley:
This default.nix file uses Nix release 22.05 (May 2022) as its basis, with GHC 8.10.7 as the compiler, and setting version pins where necessary for compatibility with chainweb.cabal. It strives to keep overrides to a minimum, and to avoid disabling as many build-time checks as possible.

Running nix-build in the chainweb-node directory should produce a build that's very close to our Linux-based production build.

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -372,7 +372,7 @@ library
         , optparse-applicative >= 0.14
         , pact >= 4.2.0.1
         , pem >=0.2
-        , primitive >= 0.7.4.0
+        , primitive >= 0.7.1.0
         , random >= 1.2
         , rosetta >= 1.0
         , safe-exceptions >= 0.1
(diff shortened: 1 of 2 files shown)


jwiegley:
It looks like CI is failing with the bug that @edmundnoble and @larskuhtz and I tracked down, so once that I fix is pushed it should pass.

chainweb.cabal
```diff
@@ -372,7 +372,7 @@ library
         , optparse-applicative >= 0.14
         , pact >= 4.2.0.1
         , pem >=0.2
-        , primitive >= 0.7.4.0
+        , primitive >= 0.7.1.0
```
edmundnoble, line 375: :+1:

edmundnoble (approved):

emmanueldenloye (approved):
LGTM

PR #1494, opened 2022-08-02
Our own serialization newtypes, using binary

edmundnoble:
Monomorphizes our binary codecs to use `binary` instead of the `bytes` type classes, but behind a newtype so that we know that no questionable `Serialize` or `Binary` instances are used.


edmundnoble:
This branch has run on a node for >12 hours with no errors logged or downtime.

src/Chainweb/ChainId.hs
```diff
@@ -177,20 +175,18 @@ instance HasTextRepresentation ChainId where
 -- -------------------------------------------------------------------------- --
 -- Serialization
 
-encodeChainId :: MonadPut m => ChainId -> m ()
-encodeChainId (ChainId i32) = putWord32le $ unsigned i32
+encodeChainId :: ChainId -> Put
+encodeChainId (ChainId w32) = putWord32le w32
```
larskuhtz, line 179: Is it ok to drop the `unsigned`?

larskuhtz (approved):
Nice PR! Just on question.

PR #1502, opened 2022-08-11
Ensure that p2p topology is sufficiently dynamic and random

larskuhtz:
TODO

* [x] add test case for sampling the 'geometric' distribution
* [ ] add test case for peer selection

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -446,6 +446,7 @@ test-suite chainweb-tests
         Chainweb.Test.Mempool.InMem
         Chainweb.Test.Mempool.RestAPI
         Chainweb.Test.Mempool.Sync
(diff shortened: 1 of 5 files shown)


larskuhtz:
Distribution of connections to peers for 1200 peers after 3600 iterations:

1. All nodes grouped into bootstrap and non-bootstrap peers
2. Only bootstrap peers
3. Like (1.) but focused on peers with at least one active connection

image

edmundnoble:
Though I don't know how to be sure that the geometric distribution is exactly right for us here, 95% seems conservative enough that we can try this out.

edmundnoble (approved):

enobayram:
Can we just merge `master` here so that I can `nix-build` this branch locally?

PR #1504, opened 2022-08-16
Try fprof-late plugin

edmundnoble:


edmundnoble:
Branch is useful, PR is not.

PR #1507, opened 2022-08-17
Principal namespaces

emilypi:
This PR modifies the namespace contract `ns` to allow for autonomously-generated namespaces using [principal account protocols](https://github.com/kadena-io/KIPs/pull/22). The schema is as follows: 

1. Users may create a principal namespace name using the new `create-principal-namespace` function added to the contract in this PR for the admin guard they might care to use to govern their namespace. 
2. Autonomously generated namespace names are of the format `n_ + (160 bit truncations of Blake2b_256 hash of guard)`. 
3. When issuing `define-namespaace` and `namespace` in a Pact script, if the namespace name is the same  as the created name (checked against the admin guard), then usage will succeed and otherwise fail.


jwiegley (approved):

pact/namespaces/ns.pact
```diff
@@ -31,25 +27,48 @@
...
+        , 'active : false }
+        { 'admin-guard := ag
+        , 'active := is-active }
 
         (enforce is-active "Inactive or unregistered namespace")
         (enforce (= ns-admin ag) "Admin guard must match guard in registry")
 
-        true))
+        true)
+      ))
```
sirlensalot, line 71: Do we want to prevent principal namespaces from being registered via `write-registry`? Per `validate` they "do no harm" as the principal would not hit the record, so it's more correctness for it's own sake.

pact/namespaces/ns.repl
```diff
@@ -5,6 +5,7 @@
  , 'ns-operate-keyset: ["operate"]
  , 'ns-genesis-keyset: { "keys": [], "pred": "="} })
 
+(load "v1/ns.pact")
```
sirlensalot, line 8: This is a little confusing, as presumably it's to show that v1 can be upgraded, but it doesn't cover governance/upgrade for v2. Suggest we 
- document the upgrade test (just a comment is fine)
- exercise v2 governance: fail to upgrade without admin sigs, succeed upgrade with. I think just loading v2 over itself should be fine.

pact/namespaces/ns.pact
emilypi: So, we did this in [previous commits](https://github.com/kadena-io/chainweb-node/blob/e21f1d3620efc00b67563143928b52401991615b/pact/namespaces/v2/ns.pact#L75) but it was removed on the basis of being unnecessary. I'm fine with adding it back using the new-and-improved principal ns format. It's no sweat over here, we just need to commit to something.

pact/namespaces/ns.repl
emilypi: Sounds good - I'll add.

pact/namespaces/ns.pact
sirlensalot: No, if we removed it let's stick with that. Resolving

pact/namespaces/ns.repl
```diff
@@ -1,16 +1,34 @@
...
+
+;; Test that we can upgrade ns.pact only when appropriate keys
+;; are in scope.
+(env-data {})
+(expect-failure
+  "Cannot upgrade the namespace contract due to governance failure"
+  "tx failure on upgrade"
+  (try (enforce false "tx failure on upgrade") (load "ns.pact")))
+
+(env-data
```
sirlensalot, line 22: I think we don't need this?

emilypi: Removed the `env-data` calls here - they were unnecessary

sirlensalot (approved):

PR #1513, opened 2022-08-24
update pact pin + tests for legacy keyset rotation

jmcardon:


emilypi (approved):
Approved pending up to date pin

PR #1521, opened 2022-08-27
Overhaul pact multichain regression, add musl regression

sirlensalot:


jmcardon (approved):

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -518,7 +519,7 @@ mkCmd nonce rpc = defaultCmd
 
 -- | Build parsed + verified Pact command
 --
-buildCwCmd :: CmdBuilder -> IO ChainwebTransaction
+buildCwCmd :: MonadThrow m => MonadIO m => CmdBuilder -> m ChainwebTransaction
```
jmcardon, line 522: Tuple constraints?

test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -532,11 +533,11 @@ buildTextCmd = fmap (fmap T.decodeUtf8) . buildRawCmd
 
 -- | Build a raw bytestring command
 --
-buildRawCmd :: CmdBuilder -> IO (Command ByteString)
+buildRawCmd :: MonadThrow m => MonadIO m => CmdBuilder -> m (Command ByteString)
```
jmcardon, line 536: Same here

test/Chainweb/Test/Pact/PactInProcApi.hs
```diff
@@ -427,115 +442,98 @@ chainweb213Test bdb mpRefIO pact = do
         , "  (defun fselect () (select tbl (constantly true))))"
         , "(create-table tbl)"
         ]
-    runCut' = runCut testVersion bdb pact (offsetBlockTime second) zeroNoncer noMiner
 
-pact43UpgradeTest :: TestBlockDb -> IO (IORef MemPoolAccess) -> WebPactExecutionService -> IO ()
-pact43UpgradeTest bdb mpRefIO pact = do
+buildVector :: MonadIO m => [m a] -> m (V.Vector a)
```
jmcardon, line 446: doesn't need `MonadIO` constraint I believe. Just `Monad` will do

test/Chainweb/Test/Pact/PactInProcApi.hs
```diff
@@ -296,128 +334,105 @@ newBlockRewindValidate mpRefIO reqIO = testCase "newBlockRewindValidate" $ do
...
       }
 
-minerKeysetTest :: TestBlockDb -> IO (IORef MemPoolAccess) -> WebPactExecutionService -> IO ()
-minerKeysetTest bdb _mpRefIO pact = do
+minerKeysetTest :: MultichainTest
+minerKeysetTest _mpRefIO = do
 
   -- run past genesis, upgrades
-  forM_ [(1::Int)..24] $ \_i -> runCut' noMiner
+  forM_ [(1::Int)..24] $ \_i -> runCut'
```
emilypi, line 344: I do think Jose's hiding of this indexing is cheap and easy, and more legible.

emilypi (approved):

test/Chainweb/Test/Pact/PactInProcApi.hs
sirlensalot: ? no indexes are used here

sirlensalot: Assuming you mean tupling tx builders with tests that could be elegant. The base case is very common so that's probably worth having an abstraction. 

However it's not always so straightforward since some tests are dispatching to different chains, or refilling etc, and getting a fixture to handle those seamlessly is pretty hard. I think indexed access is still necessary for certain tests but they should be able to coexist.

sirlensalot: I do want some kind of mempool abstraction that can make that behavior easier to model. Probably getting away from concrete `MemPoolAccess` implementations would be an improvement.

test/Chainweb/Test/Pact/PactInProcApi.hs
sirlensalot: this will probably go away

emilypi (approved):

jwiegley (approved):
It all seems straightforward enough. If it still passes, then LGTM.

jmcardon (approved):

sirlensalot:
Mac builds failing are not part of this PR, so merging

PR #1526, opened 2022-09-02
Less logs

edmundnoble:
Stuart added on this because some of the logs are for block fill.

These messages are by far the most common log messages, thousands are emitted per minute. And it seems we don't really use them in prod.


sirlensalot (approved):
LGTM

src/Chainweb/Chainweb.hs
```diff
@@ -646,43 +647,64 @@ runChainweb cw = do
...
-        shouldDisplayException e
-            | Just InsecureConnectionDenied <- fromException e = False
-            | otherwise = defaultShouldDisplayException e
+    serverSettings :: IORef Int -> Settings
+    serverSettings closedConnectionsRef = setOnException
+        (\r e -> if
+            | Just InsecureConnectionDenied <- fromException e ->
+                return ()
+            | Just ClientClosedConnectionPrematurely <- fromException e ->
+                atomicModifyIORef' closedConnectionsRef (\n -> (n + 1, ()))
```
larskuhtz, line 656: It might be overkill here, but you could consider using https://github.com/kadena-io/chainweb-node/blob/2.16/src/Chainweb/Counter.hs. One (small) advantage would be the that log message format would be consistent with other counters.

src/Chainweb/Chainweb.hs
```diff
@@ -646,43 +647,64 @@ runChainweb cw = do
...
+                when (defaultShouldDisplayException e) $
+                    logg Warn $ loggServerError r e
+        ) $ peerServerSettings (_peerResPeer $ _chainwebPeer cw)
+
+    monitorConnectionsClosedByClient :: IORef Int -> IO ()
+    monitorConnectionsClosedByClient closedConnectionsRef =
+        runForever logg "ConnectionClosedByClient.counter" $ do
+            closedConnections <- atomicModifyIORef' closedConnectionsRef (\n -> (0, n))
+            when (closedConnections /= 0) $
+                logg Info $ "Clients closed connections prematurely " <> sshow closedConnections <> " times"
```
larskuhtz, line 667: A disadvantage of this textual encoding is that it is hard to process with elasticsearch (and similar) tools. For instance, we have a panel on the dashboard that counts connections failures of different types, which would require to write an extra ingress rule to parse the message and extract the value.

One solution would be to use the predefined counter type from `Chainweb.Counters` along with a dedicated telemetry logger. This would also allow to add more counters (e.g. for other failure types).

Though I am not sure. It might be overkill to create a dedicated telemetry logger and elasticsearch index just for a single counter value.

larskuhtz (approved):
Approved with three suggestions

src/Chainweb/Chainweb.hs
edmundnoble: How about now, using the existing data stream (telemetry logger) just with a different string attached to the counter?

src/Chainweb/Chainweb.hs
```diff
@@ -588,26 +589,26 @@ runChainweb
     -> IO ()
 runChainweb cw = do
     logg Info "start chainweb node"
-    runConcurrently $ ()
+    foldr1 concurrently_
```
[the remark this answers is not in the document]
edmundnoble: It's too short to be a utility.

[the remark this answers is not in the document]
larskuhtz: It's not completely obvious that `foldr1 concurrently_` has the same performance as `runConcurrently`. It depends on how async manages threads internally. It actually is equivalent with what `runConcurrently` does. But that could justify to make it an utility function.

PR #1528, opened 2022-09-03
Strict logging

edmundnoble:
Just checking out what happens if we encode a log message prior to enqueueing it.


edmundnoble:
This broke the telemetry loggers, but I'm not sure why. Leaving this for now. I had hoped to make it less problematic for the node if the log queue filled.

larskuhtz:
I am pretty sure that the logging code calls `deepseq` on the message (maybe not on the full message but only on the payload?) before putting in the queue, but you may double check that.

PR #1530, opened 2022-09-08
Principal namespaces slp

sirlensalot:


pact/namespaces/ns.pact
```diff
@@ -35,21 +31,42 @@
...
+         (lambda (n)
+           (let
+             ((take-two (take 2 n))
+              (drop-two (drop 2 n)))
+
+             (cond
+              ((= "k#" take-two)
+                (+ "k:" drop-two))
+              ((= "w#" take-two)
+                (+ "w:" (+ (take 43 drop-two) (+ ":" (drop 46 n)))))
```
jwiegley, line 51: Since this is validation, should there be an enforce that the address is indeed long enough for this take and drop to be meaningful here?

jwiegley (approved):

sirlensalot:
Redoing as merge into #1507

PR #1532, opened 2022-09-13
Check file descriptor rlimit on startup

edmundnoble:


node/Utils/CheckRLimits.hs
```diff
@@ -0,0 +1,36 @@
...
+        openFileLimit =
+            [ n
+            | l <- lines limits
+            , Just rest <- [stripPrefix "Max open files" l]
+            , Just (n :: Int) <- [readMaybe (takeWhile isDigit $ dropWhile isSpace rest)]
+            ]
+    case openFileLimit of
+        [n] ->
+            when (n < 32768) $
+                error $
```
enobayram, line 30: I wonder if we should just print this message to stdout and exit instead of `error`. Since this is a user facing error from a very specific check, the source information and the indentation from the`error` display is probably unnecessary noise.

edmundnoble: :+1:

edmundnoble:
> Should we provide the user with a CLI option to bypass this check? I.e. when file descriptor limit is smaller than 32768 is chainweb-node guaranteed to crash, could there be specific use cases that will work with lower limits so a user could plausibly want to bypass the check?

It may not be guaranteed to crash (immediately), for example if the user disables P2P and is doing a replay or some kind of testing, or are running devnet. Still, I think it's about as easy in these cases to raise the limit as it is to set a flag disabling the check.

enobayram:
Seems like this PR still needs a fix for `getCurrentPid` on MacOS, right? Would also be nice to have some manual test steps here, both for Linux and Mac. For example the terminal output from a session that starts the node with a low ulimit sees it fail, starts it with a high ulimit and sees it not fail.

edmundnoble:
Sample terminal output:
```
edmundnoble mx:~/kadena/chainweb-node
$ cabal run chainweb-node
Up to date
This process is only able to open 1024 file descriptors at once, which is not enough to run chainweb-node.
Set the limit higher than 32768 using the ulimit command or contact an administrator.
 ^ Tue Oct 04 08:51:09
edmundnoble mx:~/kadena/chainweb-node
$ ulimit -n
1024
 ^ Tue Oct 04 08:51:15
edmundnoble mx:~/kadena/chainweb-node
$ ulimit -n 32768
 ^ Tue Oct 04 08:51:20
edmundnoble mx:~/kadena/chainweb-node
$ cabal run chainweb-node
Up to date
2022-10-04T12:51:22.318Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] start ServiceDate
2022-10-04T12:51:22.320Z [Warn] [chainwebVersion=mainnet01|type=ChainwebApp] This version of chainweb node will stop to work at 2022-12-01 00:00:00 UTC. Please upgrade to a new version before that date.
2022-10-04T12:51:22.352Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] opened rocksdb in directory "/home/edmundnoble/.local/share/chainweb-node/mainnet01/0/rocksDb"
2022-10-04T12:51:22.352Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] backup config: BackupConfig {_configBackupApi = EnableConfig {_enableConfigEnabled = False, _enableConfigConfig = BackupApiConfig}, _configBackupDirectory = Nothing}
2022-10-04T12:51:22.685Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from us-e3.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":42690}}
2022-10-04T12:51:22.689Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from us-e2.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":44938}}
2022-10-04T12:51:22.689Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from us-e1.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":53222}}
2022-10-04T12:51:22.731Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from us-w3.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":55830}}
2022-10-04T12:51:22.733Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from us-w2.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":35426}}
2022-10-04T12:51:22.754Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from us-w1.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":34642}}
2022-10-04T12:51:22.891Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from fr2.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":39776}}
2022-10-04T12:51:22.892Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from fr1.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":60270}}
2022-10-04T12:51:22.900Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] got remote info from fr3.chainweb.com:443: {"version":"2.16","timestamp":1664887882,"hostaddress":{"hostname":"64.229.87.120","port":34248}}
^C2022-10-04T12:51:23.052Z [Info] [chainwebVersion=mainnet01|type=ChainwebApp] ServiceDate stopped

 ^ Tue Oct 04 08:51:23
edmundnoble mx:~/kadena/chainweb-node
$ 
```

enobayram (approved):
LGTM!

edmundnoble:
Just realized the error says to increase the limit *above* 32768, when in fact it is fine with a limit of 32768. Fixed now, error asks for above 32767.

node/Utils/CheckRLimits.hs
```diff
@@ -0,0 +1,39 @@
...
+import Data.List(stripPrefix)
+import System.Exit
+import System.IO
+import System.Posix.Process (getProcessID)
+import Text.Read
+
+checkRLimits :: IO ()
+checkRLimits = do
+    pid <- getProcessID
+    limits <- readFile $ "/proc/" <> show pid <> "/limits"
```
larskuhtz, line 16: This won't work on mac-os and windows. Maybe use `getrlimit` and `setrlimit` with FFI, I think, I think are both POSIX.

larskuhtz: Also, if this fails, the error message would probably be confusing for most users.

larskuhtz (changes requested):

node/Utils/CheckRLimits.hsc
```diff
@@ -0,0 +1,58 @@
...
+checkRLimits = void $ try @IOException $ do
+    (soft, hard) <- getOpenFileLimits
+    when (hard < 32768) $ do
+        hPutStrLn stderr $
+            "This process is only able to open " <> show hard <> " file descriptors at once, "
+            <> "which is not enough to run chainweb-node.\n" 
+            <> "Set the open file limit higher than 32767 using the ulimit command or contact an administrator."
+        exitFailure
+    when (soft < 32768) $ 
+        setOpenFileLimits (hard, hard)
```
enobayram, line 57: Wouldn't it be better to set the file limits to `32768` in this case? Seems odd to me that if the limit was already `32768` we keep it at that, but if it happens to be `32767`, then we silently set it to the hard limit. There's definitely an element of surprise there.

node/Utils/CheckRLimits.hsc
```diff
@@ -0,0 +1,58 @@
...
+
+foreign import ccall "sys/resource.h setrlimit" 
+    c_setrlimit :: CInt -> Ptr () -> IO CInt
+
+getOpenFileLimits :: HasCallStack => IO ((#type rlim_t), (#type rlim_t))
+getOpenFileLimits = allocaBytes (#size struct rlimit) $ \ptr -> do
+    err <- c_getrlimit (#const RLIMIT_NOFILE) ptr
+    if err /= 0 then do
+        Errno errno <- getErrno
+        error $ "getOpenFileLimits: errno not equal to 0: " <> show errno
```
enobayram, line 35: This error condition makes me worried that we'll encounter an environment that, for some reason, errors out on the `getrlimit` call, but actually still has a high enough file handle limit. In that case this mandatory check will be a big inconvenience to the user who could otherwise run the node without any issues.

enobayram (commented):
Awesome, this looks much more robust now. 

We need to run this PR on Linux and MacOS for both `limit < 32768` and `limit >= 32768` cases and post the terminal dump as a comment for all cases.

edmundnoble:
New version uses setrlimit/getrlimit. Testing instructions:
First, check your hard and soft open file limits in your session. `ulimit -Hn` will print the hard limit, `ulimit -n` will print the soft limit.
You can also set the hard and soft limits to an integer `x` with `ulimit -Hn x` and `ulimit -n x`. You can only ever increase the hard limit if you're root or have a special capability.
Then, run the node in these configurations:
- with the hard limit < 32768, and the soft limit < 32768 - expect an error from the node regarding the limit being too low.
- with the hard limit >= 32768, and the soft limit < 32768 - expect no errors from the node and check that there is no message like this: "Failed to set open file limit"
- with the hard limit >= 32768, and the soft limit >= 32768 - expect no errors from the node.

larskuhtz (approved):

PR #1534, opened 2022-09-20
Principal namespaces: use hashes, rework upgrade methodology

sirlensalot:


pact/namespaces/ns.pact
```diff
@@ -31,25 +27,38 @@
     (enforce (is-charset CHARSET_LATIN1 name)
              "Name must be in latin1 charset"))
 
+  (defun create-principal-namespace:string
+      ( g:guard
+        )
+    " Format principal namespace as Pact hash (BLAKE2b256) of principal \
+    \ in hex truncated to 160 bits (40 characters), prepended with 'n'. "
+    (+ "n" (take 40 (int-to-str 16 (str-to-int 64 (hash g)))))
```
emilypi, line 35: Small nit here: I'd like this to be more distinct than `n` + `40 hash chars` which looks fairly ambiguous. Can we at least call this `ns` or `kns` or smth to make it more scriable?

emilypi (commented):
Code LGTM - I like the idea of exposing namespace formatting as a function of the namespace module. Approved pending small nit.

emilypi (approved):

pact/namespaces/ns.pact
sirlensalot: As discussed, moved to `n_` prefix

PR #1538, opened 2022-09-24
Fix default.nix for building chainweb-node on macOS

### LINKED ISSUE #1537: New rocksdb-haskell-kadena dependency does not build on macOS
Building the latest chainweb-node on Mac does not seem to work. This is due to a compilation failure in rocksdb by way of the new rocksdb-haskell-kadena dependency introduced in 6e5d32bc11dc21e8d3ad6e70d188d902a26c560d. The exact error looks like:
```
rocksdb-6.29.3/util/crc32c.cc:757:13: error:
     error: always_inline function '_mm_crc32_u64' requires target feature 'sse4.2', but would be inlined into function 'crc32c_3way' that is compiled without support for 'sse4.2'
                CRCtriplet(crc, next, -128);
                ^
    |
757 |             CRCtriplet(crc, next, -128);
    |             ^
```
Since we've decided to support Mac as a production deployment platform, we’ll need a way for CI to catch discrepancies like this. Or is it just happening because of the Nix environment that we’ve pinned in default.nix? I’m still gathering information.

A possibly relevant note, from https://github.com/chakra-core/ChakraCore/issues/3551:
>Unfortunately if -msse4.2 is enabled then SSE4.2 instructions will be enabled globally throughout the codebase. This means that Clang is free to emit SSE4.2 code wherever it pleases (not just where you have used an intrinsic such as _mm_crc32_u32). This means it will generate code that will not work correctly on processors that only have support for SSE2. I think this is an example of how Clang differs in behaviour to MSVC. I think MSVC will only emit SSE4.2 instructions specifically where you have used SSE4.2 intrinsics.

jwiegley:
Fixes #1537

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -47,11 +47,17 @@ pkgs.haskell.packages.${compiler}.developPackage {
           # date = "2022-08-10T09:03:56-04:00";
         }) {})
         (attrs: {
+          preConfigure = (attrs.preConfigure or "") +
+            pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
+              perl -i -ne 'print unless /HAVE_SSE42/;' rocksdb-haskell-kadena.cabal
+            '';
           librarySystemDepends = (attrs.librarySystemDepends or []) ++ [
+            pkgs.rocksdb
             pkgs.snappy.dev
             pkgs.zlib.dev
             pkgs.zstd.dev
             pkgs.lz4.dev
+            pkgs.perl
           ];
         });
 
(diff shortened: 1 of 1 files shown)


jwiegley:
I can confirm that this PR builds and passes tests on my Mac.

mightybyte (approved):
This fixed the SSE build error I was getting on Linux.

jwiegley:
It seems that M1 Mac has problems building `tbb` that `rocksdb-haskell` depends on. It is addressed in this issue, so working on the patch now: https://github.com/oneapi-src/oneTBB/issues/186

emmanueldenloye (approved):
LGTM

mightybyte (approved):
Works for me on multiple Linux machines.

jwiegley:
Just to note, there is some patching being done in the Nix file that should be applied to the Cabal file and the source directly. Since that is a more invasive approach that affects all users, I will take that up in a separate PR.

PR #1542, opened 2022-10-03
Use new table abstraction from chainweb-storage

edmundnoble:
This is blocked on https://github.com/kadena-io/chainweb-storage/pull/16. There are some interesting things going on here (and in the chainweb-storage PR) at the intersection of type families, functional dependencies and constraint kinds, which make the instances we define a bit unintuitive. See my comments for details.


src/Chainweb/Payload/PayloadStore.hs
```diff
@@ -343,29 +291,29 @@ instance PayloadCasLookup_ a cas => HasCasLookup (PayloadDb_ a cas) where
...
+                tableDelete
                     (_transactionDbBlockTransactions $ _transactionDb db)
                     (_blockPayloadTransactionsHash pd)
-                casDelete
+                tableDelete
                     (_payloadCacheBlockOutputs $ _payloadCache db)
                     (_blockPayloadOutputsHash pd)
             Nothing -> return ()
-    {-# INLINE casDelete #-}
+    {-# INLINE tableDelete #-}
```
edmundnoble, line 319: You'd think we could just write
```hs
instance (MerkleHashAlgorithm a, CanPayloadCas_ a tbl) => 
  Cas (PayloadDb_ a tbl) (PayloadWithOutputs_ a)
```
since `type Cas t v = Table t (CasKeyType v) v`. Unfortunately, that instance is not possible to write due to https://gitlab.haskell.org/ghc/ghc/-/issues/13267.

You might think that instead we can write
```hs
instance (MerkleHashAlgorithm a, CanPayloadCas_ a tbl) => 
  Table (PayloadDb_ a tbl) (CasKeyType (PayloadWithOutputs_ a)) (PayloadWithOutputs_ a)
```
but this is illegal in the first place, this time deliberately, see https://gitlab.haskell.org/ghc/ghc/-/issues/13262 or https://gitlab.haskell.org/ghc/ghc/-/issues/14046. We can't have type family applications in our instance heads at all. So instead, we need to ask for `k ~ CasKeyType (PayloadOutputs_ a)`, and in the instance head we ask for a `Table` with key type `k`. This pattern repeats in other places in this file.

enobayram (approved):
LGTM! All the changes look safe to me. And thanks for renaming all those `cas` variables to `tbl`.

src/Chainweb/Payload/PayloadStore.hs
enobayram: So, the definitions of `ReadableTable` and `Table` are:
``` haskell
class ReadableTable t k v | t -> k v where
  ...
class ReadableTable t k v => Table t k v | t -> k v where
  ...
```
If we instead had:
``` haskell
class ReadableTable t where
  type TableKey t
  type TableValue t 
  ...
class ReadableTable t => Table t where  
  ...
```

Then these instance heads wouldn't be nearly as gnarly:
```
instance (MerkleHashAlgorithm a, CanReadablePayloadCas_ a tbl) => ReadableTable (PayloadDb_ a tbl) 
  type TableKey (PayloadDb_ a tbl) = CasKeyType (PayloadWithOutputs_ a)
  type TableValue (PayloadDb_ a tbl) = PayloadWithOutputs_ a
  ...
instance (MerkleHashAlgorithm a, CanPayloadCas_ a tbl) => Table (PayloadDb_ a tbl) 
  -- No need to repeat the `TableKey` & `TableValue` types
  ...
```

I'm not sure if it would be useful here, but using associated types of a single class parameter instead of encoding them as functional dependencies between multiple parameters also enables us to use `DerivingVia`.

So, is there any reason why we're using functional dependencies for `ReadableTable` and `Table`?

edmundnoble: I've forgotten whether I tried that, so I'm going to give it a try now.

edmundnoble: Ah so after trying this, the places that don't need to talk about (constrain) the `TableKey` and `TableValue` types are shorter, but most places that *do* need to constrain them is way more verbose, with a bunch of equality constraints.

More generally the issue is, say we have `class C x y | x -> y`. Whenever we don't need to constrain `y`, the mention of `y` in constraints is useless verbosity. But when we do need to constrain it, say that we want it to equal a ground type `T`, we can just write `C x T` as a constraint, whereas with tyfams we have to write `(C x, CTyFam x ~ T)`, and this is only more verbose with more tyfams/dependent type parameters. And that ends up really annoying in practice because almost *everyone* who wants to use a table is going to care about the key and value types being exactly equal to some ground type.

enobayram: Oh, that makes sense. Thanks for trying this out! BTW, this observation makes me wonder how much it would help to have the following type alias (constraint alias, really :) ):

``` haskell
type C' x y = (C x, CTyFam x ~ y)
```

edmundnoble: Assuming it works it'd help a bit but it doesn't make instances much less verbose, because the type families still have to be defined for the instance itself, and then we end up with both `C` and `C'` which is a bit much. I'm not sure it's worse, but it doesn't seem clearly better.

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -228,15 +229,15 @@ checkPayloads pdb False h = casLookup pdb (_blockPayloadHash h) >>= \case
...
 --
--- TODO: casMember for rocksdb internally uses the default implementation that
--- uses casLookup. The Haskell rocksdb bindings don't provide a member check
+-- TODO: tableMember for rocksdb internally uses the default implementation that
+-- uses tableLookup. The Haskell rocksdb bindings don't provide a member check
 -- without getting the value. It could be done via the iterator API, which has
 -- different cost overheads. One optimization could be an implementation that
--- avoids decoding of the value.
+-- avoids decoding of the value (now implemented)
 --
```
larskuhtz, line 237: This comment is redundant now

larskuhtz (approved):

PR #1543, opened 2022-10-27
Jose/no more errors

jmcardon:


emilypi (approved):

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -215,6 +216,9 @@ applyCmd v logger gasLogger pdbenv miner gasModel txCtx spv cmd initialGas mcach
       cr <- catchesPactError $! runPayload cmd managedNamespacePolicy
       case cr of
         Left e
+          | chainweb217Pact' -> do
+            r <- jsonErrorResult (PactError ArgsError def [] "") "tx failure for request key when running cmd"
```
larskuhtz, line 220: Would this also swallow error messages from local results?

If that's the case, why not leave the message untouched here and only drop it when assembling the final result for PayloadWithOutputs and computing the PayloadHash?

emilypi: @larskuhtz this is a different code path than local cmd execution, which is done in `applyLocal`.

larskuhtz: What about newBlock? I assume it is using this code path?

emilypi: I believe so yeah

larskuhtz (changes requested):
There are some failing test cases, for which it is not immediately clear how they relate to the changes in this PR.

I would also have expected some golden tests failing with this PR. The fact that this isn't the case may somehow indicate that pact failures were not covered before very well or that those tests aren't executed for the new (default) code path. 

(We generally need to make sure that when we add features flags and forks that existing tests suites cover the new code. We had issues before where most tests where only executed on legacy code was not receiving a lot of coverage.)

Also, do we need some more high-level tests, that ensure that we get the correct behavior for the top-level pact-service requests, in particular `execValidateBlock`, `execNewBlock`, and `execLocal`?

src/Chainweb/Pact/TransactionExec.hs
jmcardon: > If that's the case, why not leave the message untouched here and only drop it when assembling the final result for PayloadWithOutputs and computing the PayloadHash?

`jsonErrorResult` `show`s the error, which fill force it and cause the long computation.

jmcardon:
Why would golden tests fail for this PR? As long as the goldens are using something like devnet or FastTimedCpm, they shouldn't fail at all, since the new fork height would be past all previous ones.

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -209,6 +210,20 @@ chainweb213Test = do
         , "(create-table tbl)"
         ]
 
+pact45UpgradeTest :: PactTestM ()
+pact45UpgradeTest = do
+  runToHeight 54
```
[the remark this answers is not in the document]
sirlensalot: This is more an issue with `ChainwebVersion` than with the forks per se.  Ideally, tests would be able to mock their own `ChainwebVersion` and set forks for its own requirements (perhaps by implementing a typeclass that specifies all the fork points). 

As it is, our approach suffers from a great deal of coupling and complexity, including abusing different version graphs to get "sub-versions" so that tests don't clobber each other.

Finally, it is why it is hard to do something like "just get the fork block height", as we are instead usually wanting some kind of particular test (see `AtOrAfter` in some of the fork functions). 

Agree with @jmcardon that this is out of scope for this PR.

sirlensalot (commented):
Following up on @larskuhtz's review -- `PactMultiChainTest` tests both new block and validate block -- but it does _not_ test local. I guess we can add this in the multi chain test though (verify that local still has errors after the fork)?

test/Chainweb/Test/Pact/PactMultiChainTest.hs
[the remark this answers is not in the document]
edmundnoble: Coming soon (TM)

larskuhtz:
> Why would golden tests fail for this PR? As long as the goldens are using something like devnet or FastTimedCpm, they shouldn't fail at all, since the new fork height would be past all previous ones.

yeah, that's my concern, that existing golden tests (and we have many of them) cover mostly legacy code paths but not the actual production code.

I think, we should either add new golden tests or port existing golden tests to include new default code paths.

src/Chainweb/Pact/TransactionExec.hs
larskuhtz, line 220: I don't know how fine-grained the information in `PactError` is. Would it make sense to include more high-level some info about the failure in the message? Is there some classification of failures? Anything that doesn't require fragile or unstable formatting.

jmcardon:
> yeah, that's my concern, that existing golden tests (and we have many of them) cover mostly legacy code paths but not the actual production code.

> I think, we should either add new golden tests or port existing golden tests to include new default code paths.

Goldens belong in the pact repo. We already do this. I don't think this is necessary in chainweb.

I've added tests across the fork boundary for new behaviors and gas changes.

jwiegley (approved):

larskuhtz (approved):
Devnet validation now fails in CI. Do we have to adjust the devnet fork points and/or the test suite for the changes in this PR?

PR #1545, opened 2022-10-31
ci updates

larskuhtz:


.github/workflows/applications.yml
```diff
@@ -353,20 +357,10 @@ jobs:
...
-        else
-          cp dist-newstyle/build/*/ghc-${{ matrix.ghc }}/chainweb-${VER}/build/*/chainweb-node artifacts/chainweb
-          cp dist-newstyle/build/*/ghc-${{ matrix.ghc }}/chainweb-${VER}/build/*/chainweb-tests artifacts/chainweb
-          cp dist-newstyle/build/*/ghc-${{ matrix.ghc }}/chainweb-${VER}/build/*/cwtool artifacts/chainweb
-          cp dist-newstyle/build/*/ghc-${{ matrix.ghc }}/chainweb-${VER}/build/*/bench artifacts/chainweb
-        fi
+        cp $(cabal list-bin chainweb-node) artifacts/chainweb
+        cp $(cabal list-bin chainweb-tests) artifacts/chainweb
+        cp $(cabal list-bin cwtool) artifacts/chainweb
+        cp $(cabal list-bin bench) artifacts/chainweb
```
edmundnoble, line 363: A thousand :+1:s

.github/workflows/applications.yml
```diff
@@ -287,18 +287,22 @@ jobs:
         EOF
 
     # dist cache
+    # the cache-key forces uploading of cache at least once a day, which ensures that
+    # upstream depenency changes are captured regularly.
```
edmundnoble, line 291: s/depenency/dependency

.github/actions/install-chainweb-from-artifacts/action.yml
```diff
@@ -16,8 +16,11 @@ runs:
       shell: bash
       run: |
         case "${{ inputs.os }}" in
-        ubuntu*)
-          sudo apt-get install -y build-essential libgflags-dev libtbb-dev
+        ubuntu-20.04)
+          sudo apt-get install -u libtbb2 libgflags2.2
+          ;;
+        ubuntu-22.04)
```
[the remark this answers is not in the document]
larskuhtz: Not sure. The libtbb version is different: libtbb2 vs. libtbb12. We could also use the libtbb-dev package (as we did before) which has the same name on both versions. But I thought, it may be better to to precise about the devs.

PR #1558, opened 2022-11-13
Transaction simulator

sirlensalot:


src/Chainweb/Pact/TxSimulator.hs
```diff
@@ -0,0 +1,127 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+module Chainweb.Pact.TxSimulator
```
larskuhtz, line 6: If this is used only in `cwtool`, can we move the module there, in order to keep the prod code base small?

src/Chainweb/Pact/TxSimulator.hs
```diff
@@ -0,0 +1,127 @@
...
+              <> help "Pact database directory")
+        <*> strOption
+             (short 'h'
+              <> metavar "PARENT_HEADER"
+              <> help "Parent header, example: curl 'https://api.chainweb.com/chainweb/0.0/mainnet01/chain/2/header?maxheight=3195192&minheight=3195192' -s | jq '.items[0]'")
+        <*> strOption
+             (short 'p'
+              <> metavar "PAYLOAD"
+              <> help "Block payload, example: curl https://api.chainweb.com/chainweb/0.0/mainnet01/chain/2/payload/3uU_CZVy3ZynNOniLbSDykNdzO5pFYuKftuynax9eUo")
+        <*> (fromIntegral @Int <$> option auto
```
larskuhtz, line 115:
```suggestion
        <*> (option auto
```

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -161,11 +163,21 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
         -- First, check: did we create this table during this block? If so,
         -- there's no point in looking up the key.
         checkDbTableExists tableName
-        result <- lift $ callDb "doReadRow"
-                       $ \db -> qry db queryStmt [SText rowkey] [RBlob]
+        (result,l) <- lift $ callDb "doReadRow"
+                       $ \db -> do
+          (t0 :: Time Micros) <- getCurrentTimeIntegral
```
larskuhtz, line 168: There exist a utility to generate traces telemetry message. However, we use that only conservatively and probably don't want to merge these logs here to master.

We could make them debug logs or hide them behind CPP. Or maybe a cost center pragma would be sufficient?

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -161,11 +163,21 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
...
+          r <- qry db queryStmt [SText rowkey] [RBlob]
+          t1 <- getCurrentTimeIntegral
+          return (r,("read"::String,tableName,rowkey,diff t1 t0))
+        lift $ logDebug $ show l
         case result of
             [] -> mzero
-            [[SBlob a]] -> MaybeT $ return $! decode $ fromStrict a
+            [[SBlob a]] -> do
+              (t0 :: Time Micros) <- liftIO getCurrentTimeIntegral
```
larskuhtz, line 176: see comment above

src/Chainweb/Pact/PactService.hs
```diff
@@ -34,6 +34,7 @@ module Chainweb.Pact.PactService
     , initPactService'
     , execNewGenesisBlock
     , getGasModel
+    , chainweb213GasModel
```
jwiegley, line 37: Why does the simulator need this gas model? Is it special like "the latest gas model" or something? If so, would it make sense to have an alias for it and export that?

tools/cwtool/TxSimulator.hs
```diff
@@ -0,0 +1,117 @@
...
+  cmdTx <- decodeStrictOrThrow tx
+  miner <- decodeStrictOrThrow $ _minerData md
+  case validateCommand cmdTx of
+    Left _ -> error "bad cmd"
+    Right cmdPwt ->
+      withSqliteDb cid cwLogger dbDir False $ \sqlenv -> do
+        (CheckpointEnv cp _) <- initRelationalCheckpointer (initBlockState 0) sqlenv logger ver cid
+        bracket
+          (_cpBeginCheckpointerBatch cp)
+          (\_ -> _cpDiscardCheckpointerBatch cp) $ \_ -> do
```
jwiegley, line 63: You can use `bracket_` instead of ignoring the bracketed value.

tools/cwtool/TxSimulator.hs
```diff
@@ -0,0 +1,117 @@
...
+            let cmd = payloadObj <$> cmdPwt
+                txc = txContext parent cmd
+            (PactDbEnv' pde) <- _cpRestore cp $ Just (succ (_blockHeight parent), _blockHash parent)
```
jwiegley, line 66: Ok, going full nerd now: Why the parens on the left side?

src/Chainweb/Pact/PactService.hs
sirlensalot: It would require functionality to actually compute the gas model at a given height, so I opted with the least complicated approach which is to use the gas model we're using currently and address when that's no longer the case (which is probably soon no?)

sirlensalot: Actually upon review there's `getGasModel` LOL so I'll use that

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
sirlensalot: I left this out for a later PR

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
sirlensalot: I left this out for a later PR

tools/cwtool/TxSimulator.hs
sirlensalot: I de-parenthesized all such expressions :)

jwiegley (approved):
This looks great!

larskuhtz (approved):

PR #1559, opened 2022-11-15
module cache debug logging

edmundnoble:


src/Chainweb/Pact/PactService/ExecBlock.hs
```diff
@@ -329,7 +330,8 @@ execTransactions isGenesis miner ctxs enfCBFail usePrecomp (PactDbEnv' pactdbenv
             l <- asks _psLogger
             pd <- getTxContext def
             mc <- liftIO (readInitModules l pactdbenv pd)
-            updateInitCache mc
```
emilypi, line 332: deleting this line is going to make things hard - are we logging *instead* of updating?

PR #1560, opened 2022-11-15
Re-enable cabal cache

edmundnoble:
Seems like installing deps takes like an hour now so the github cache isn't doing its job properly, I think this will re-enable the cabal cache but I'll check how long build takes.

### DIFF
--- a/.github/workflows/applications.yml
+++ b/.github/workflows/applications.yml
@@ -327,7 +327,7 @@ jobs:
       with:
         bucket: "kadena-cabal-cache"
         region: "us-east-1"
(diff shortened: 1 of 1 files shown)


jwiegley:
Is there any reason not to do this?

edmundnoble:
Lars had hoped that the github cache would actually do its job when the cabal cache was turned off, but it didn't (at least for dependencies).

edmundnoble:
This seems to work fine, the latest build installed dependencies in 8 seconds.

larskuhtz (approved):

PR #1561, opened 2022-11-16
Add telemetry logging for node top-level status

edmundnoble:
Added bonus, we can look for the `PactReplaySuccessful` message to check that a replay was successful instead of relying on exit code.


edmundnoble (commented):
Lars: Add cases for termination of the node?

edmundnoble:
What happens when we get killed at the top level?

larskuhtz (approved):
Maybe at top level handler for logging the reason for terminating the node.

edmundnoble:
Tested this by starting in development mode, then `chown -R root ~/.local/share/chainweb-node/development`, then restarting in development mode. Got this only on stderr without this PR:
```
chainweb-node: user error (open: IO error: while open a file for lock: /home/edmundnoble/.local/share/chainweb-node/development/0/rocksDb/LOCK: Permission denied)
```
And with this PR:
```
2022-12-14T15:35:53.477Z [Error] [chainwebVersion=development|type=ChainwebStatus] {"tag":"ProcessDied","contents":"user error (open: IO error: while open a file for lock: /home/edmundnoble/.local/share/chainweb-node/development/0/rocksDb/LOCK: Permission denied)"}

```

PR #1566, opened 2022-11-22
Fix replaying via config file

edmundnoble:
Setting `initialBlockHeightLimit` in the config file was broken, this fixes it. A subtle cause; 
`optional $ BlockHeight <$> option Auto (long "initial-block-height-limit")` will *always succeed*, so if the CLI option is not provided, `Nothing` will override the existing config file flag.


larskuhtz (approved):
Nice catch!

PR #1573, opened 2022-12-03
Simulate whole block(s)

sirlensalot:


jwiegley (approved):

tools/cwtool/TxSimulator.hs
```diff
@@ -2,113 +2,242 @@
...
+    go mv cp = modifyMVar mv $ searchOuts cp
+    searchOuts _ [] = return ([],Left "spv: proof not found")
+    searchOuts cp@(ContProof pf) ((Transaction ti,TransactionOutput o):txs) =
+      case codecDecode (chainwebPayloadCodec (Just (scVersion sc,_blockHeight bh))) ti of
+        Left {} -> internalError "input decode failed"
+        Right cmd -> case _pPayload $ payloadObj $ _cmdPayload cmd of
+          Continuation cm | _cmProof cm == Just cp -> do
+            cr :: CommandResult Hash <- decodeStrictOrThrow o
+            case _pactResult (_crResult cr) of
+              _ -> do -- TODO failed SPV error heuristic
```
emilypi, line 167: on `Left` here we could make it less opaque and error out iwth something like "tx failure for spv proofs unsupported at this time" rather than comment.

tools/cwtool/TxSimulator.hs
```diff
@@ -2,113 +2,242 @@
...
+    mgrSettings = mkManagerSettings
+        (TLSSettingsSimple True False False)
+        Nothing
+
+-- | note, fetches [low - 1, hi] to have parent headers
+fetchHeaders :: SimConfig -> ClientEnv -> IO [BlockHeader]
+fetchHeaders sc cenv = do
+  r <- (`runClientM` cenv) $
+      headersClient (scVersion sc) (scChain sc) Nothing Nothing
+      (Just $ fromIntegral $ pred $ fst $ scRange sc)
```
emilypi, line 195: why `pred` here ... is it to make sure we always fetch the parent header? If so, why not make that part of the range specification (or just comment it)?

tools/cwtool/TxSimulator.hs
sirlensalot: I'll just remove it, we're covering every error seen so far with just the ones thrown here, and it will be faster not to decode outputs.

tools/cwtool/TxSimulator.hs
sirlensalot: it was already commented :)

emilypi (approved):

PR #1576, opened 2022-12-05
withSavepoint now catches SomeAsyncException

edmundnoble:
`withSavepoint` here is trying to rethrow asynchronous exceptions but fails because it catches `AsyncException` instead of `SomeAsyncException`. I noticed this looking at the logs which show `non-pact exception: AsyncCancelled`.

### DIFF
--- a/src/Chainweb/Pact/Backend/Utils.hs
+++ b/src/Chainweb/Pact/Backend/Utils.hs
@@ -57,7 +57,7 @@ module Chainweb.Pact.Backend.Utils
   , chainwebPragmas
   ) where
 
-import Control.Exception (AsyncException, evaluate)
+import Control.Exception (SomeAsyncException, evaluate)
 import Control.Exception.Safe (tryAny)
 import Control.Lens
 import Control.Monad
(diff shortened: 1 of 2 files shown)


src/Chainweb/Pact/Backend/Utils.hs
```diff
@@ -166,7 +166,7 @@ withSavepoint name action = mask $ \resetMask -> do
         liftIO $ evaluate r
     throwErr s = internalError $ "withSavepoint (" <> toText name <> "): " <> s
     handlers = [ Handler $ \(e :: PactException) -> throwErr (sshow e)
-               , Handler $ \(e :: AsyncException) -> throwM e
+               , Handler $ \(e :: SomeAsyncException) -> throwM e
```
larskuhtz, line 169: Oh, good catch. Is this change safe to apply or can it break anything? I guess it should be safe, but why may want quickly grep over all use sites.

larskuhtz (approved):
Approved with one comment

src/Chainweb/Pact/Backend/Utils.hs
larskuhtz: It's used only in two places and it seems fine. Also checked the respective functions in `Chainweb.Utils`, which have correct behavior.

edmundnoble: This should be safe I think, any place which is fine with `ThreadKilled` should also be fine with `AsyncCancelled`.

PR #1577, opened 2022-12-06
Checkpointer Module Cache

larskuhtz:


src/Chainweb/Pact/Backend/ModuleCache.hs
```diff
@@ -0,0 +1,184 @@
...
+checkModuleCache key rowdata txid = runState $ do
+    readModuleCache txid addy >>= \case
+
+        -- Cache hit
+        Just !x -> return $ Just x
+
+        -- Cache miss: decode module and insert into cache
+        Nothing -> case decodeStrict' rowdata of
+            Nothing -> return Nothing
+            Just !m -> do
```
larskuhtz, line 127: Should this raise an error instead of returning `Nothing`?

```suggestion
        Nothing -> case eitherDecodeStrict' rowdata of
            e@Left{} -> return e
            Right !m -> do
```

src/Chainweb/Pact/Backend/ModuleCache.hs
```diff
@@ -0,0 +1,184 @@
...
+checkModuleCache
+    :: Utf8
+        -- ^ Row Key for the module
+    -> ByteString
+        -- ^ row data that contains the encoded module
+    -> TxId
+        -- ^ TxId. This is used as priority for cache eviction. Smaller values
+        -- are evicted first.
+    -> ModuleCache
+    -> (Maybe PersistModuleData, ModuleCache)
```
larskuhtz, line 117: Return `Either` instead of `Maybe` in case decoding fails?

src/Chainweb/Pact/Backend/ModuleCache.hs
```diff
@@ -0,0 +1,184 @@
...
+-- | Prune Cache to meet size limit
+--
+-- Complexity: \(O(n \log(n))\) in the number of entries
+--
+maintainCache :: State ModuleCache ()
+maintainCache = do
+    sz <- use mcSize
+    lim <- use mcLimit
+    e <- gets isEmpty
+    unless (e || sz < lim) $ do
```
larskuhtz, line 173: For `lim > 0`, `e` implies `sz < lim`. Therefor the check of emptiness seem redundant here.

src/Chainweb/Pact/Backend/ModuleCache.hs
```diff
@@ -0,0 +1,184 @@
...
+  mcStore %= HM.insert ca (CacheEntry txid ca sz v)
+  mcSize %= (+ sz)
+  maintainCache
+
+-- | Prune Cache to meet size limit
+--
+-- Complexity: \(O(n \log(n))\) in the number of entries
+--
+maintainCache :: State ModuleCache ()
+maintainCache = do
```
larskuhtz, line 169: For caches with many items it may make sense to trigger this function only when the cache size is, say, twice the limit. That could amortize for the (somewhat) expensive call of `sort`. On the down-side it would make latencies of cache usage less consistent. Most likely it doesn't pay off.

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -240,6 +243,7 @@ initBlockState initialBlockHeight = BlockState
     , _bsModuleNameFix = False
     , _bsSortedKeys = False
     , _bsLowerCaseTables = False
+    , _bsModuleCache = emptyCache (1024 * 1024 * 1024) -- FIXME: what is a reasonable value here?
```
larskuhtz, line 246: This should be configurable.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -156,20 +168,33 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
...
         -- there's no point in looking up the key.
         checkDbTableExists tableName
         result <- lift $ callDb "doReadRow"
                        $ \db -> qry db queryStmt [SText rowkey] [RBlob]
         case result of
             [] -> mzero
-            [[SBlob a]] -> MaybeT $ return $! decode $ fromStrict a
+            [[SBlob a]] ->
+              checkCache rowkey a <|>
+              (MaybeT $ return $! decode $ fromStrict a)
```
jwiegley, line 186: Instead of `decode . fromStrict`, you can use `decodeStrict`.

src/Chainweb/Pact/Backend/ModuleCache.hs
```diff
@@ -0,0 +1,184 @@
...
+--
+writeModuleCache
+    :: TxId
+    -> CacheAddress
+    -> PersistModuleData
+    -> State ModuleCache ()
+writeModuleCache txid ca v = do
+  let sz = sizeOf SizeOfV1 $ void (_mdModule v)
+  mcStore %= HM.insert ca (CacheEntry txid ca sz v)
+  mcSize %= (+ sz)
```
jwiegley, line 161: I think `+=` would be clearer here.

src/Chainweb/Pact/Backend/ModuleCache.hs
sirlensalot: See response above

src/Chainweb/Pact/Backend/ModuleCache.hs
sirlensalot: agree, remove

src/Chainweb/Pact/Backend/ModuleCache.hs
sirlensalot: In the specific case of modules, writes happen pretty rarely so it's probably fine to maintain each time.

bench/Chainweb/Pact/Backend/Bench.hs
```diff
@@ -103,7 +103,7 @@ cpWithBench torun =
     name = "batchedCheckpointer"
     cid = unsafeChainId 0
 
-    initialBlockState = initBlockState $ genesisHeight v cid
+    initialBlockState = initBlockState (1024^(2::Int) {- 1MiB -}) $ genesisHeight v cid
```
sirlensalot, line 106: I was going to suggest using `mega` from `Chainweb.Utils` but those are SI. Maybe we should make powers-of-2 versions for use cases like this instead of having to comment the size.

sirlensalot: Also, any reason for this size for the benchmark?

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -439,6 +441,7 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configOnlySyncPact = False
     , _configSyncPactChains = Nothing
     , _configBackup = defaultBackupConfig
+    , _configModuleCacheLimit = 10 * 1024^(2::Int) -- 10MiB (per chain)
```
sirlensalot, line 444: Seems like a small default. I guess we will revisit this later?

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -545,4 +550,8 @@ pChainwebConfiguration = id
         <> help "The only Pact databases to synchronize. If empty or unset, all chains will be synchronized."
         <> metavar "JSON list of chain ids"
     <*< configBackup %:: pBackupConfig
+    <*< configModuleCacheLimit .:: option auto
+        % long "module-cache-limit"
+        <> help "Maximum size of the module cache in bytes in terms of the raw module data"
```
sirlensalot, line 555:
```suggestion
        <> help "Maximum size of the per-chain checkpointer module cache in bytes"
```

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -120,25 +120,36 @@ doReadRow
...
+        Pacts -> lookupWithKey (convPactId k) noCache
   where
     tableName = domainTableName d
     (Utf8 tableNameBS) = tableName
 
+    checkModuleCache u b = MaybeT $ do
+        txid <- use bsTxId -- cache priority
+        mc <- use bsModuleCache
+        (r, mc') <- liftIO $ checkDbCache u b txid mc
+        modify' (bsModuleCache .~ mc')
```
sirlensalot, line 138: I think lines 136-138 could be simply `modify' $ liftIO . checkDbCache u b txid`?

sirlensalot: Also this was originally down by `lookupInDb` and `noCache`, is there a reason it's up here so far away from those?

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -224,14 +226,18 @@ data BlockState = BlockState
...
-    deriving Show
 
 emptySQLitePendingData :: SQLitePendingData
 emptySQLitePendingData = SQLitePendingData mempty mempty mempty mempty
 
-initBlockState :: BlockHeight -> BlockState
-initBlockState initialBlockHeight = BlockState
+initBlockState
+    :: Int
+        -- ^ Module Cache Limit (in bytes of corresponding rowdata)
```
sirlensalot, line 237: Hmm should we newtype `Natural` as `ModuleCacheLimitBytes` or something since it shows up here needing a comment and in config, document that it's bytes, etc?

src/Chainweb/Pact/Types.hs
```diff
@@ -389,6 +389,7 @@ defaultPactServiceConfig = PactServiceConfig
       , _pactUnlimitedInitialRewind = False
       , _pactBlockGasLimit = defaultBlockGasLimit
       , _pactLogGas = False
+      , _pactModuleCacheLimit = 1024^(3::Int) -- 1GiB
```
sirlensalot, line 392: Why is this different than the default for the config? Maybe we should define a default limit somewhere and use here, config, and for `TxSimulator` below (since it generally wants to mimic "production"

bench/Chainweb/Pact/Backend/Bench.hs
larskuhtz: No particular reason, I just picked something. We could just use the prod defaults. I can think of pros and cons for both larger or smaller values. So not sure.

src/Chainweb/Chainweb/Configuration.hs
larskuhtz: Yes, it's a bit too small, in particular for catchup and replay, based on a number of experiments that I performed. In the most recent versions I increased it to 60MiB per chain, which seems fine for prod and also gives decent results for replay. Although for replay it's probably better to configure something larger -- depending on available memory.

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -392,6 +392,8 @@ data ChainwebConfiguration = ChainwebConfiguration
     , _configSyncPactChains :: !(Maybe [ChainId])
         -- ^ the only chains to be synchronized on startup to the latest cut.
         --   if unset, all chains will be synchronized.
+    , _configModuleCacheLimit :: !Natural
+        -- ^ module cache size limit in bytes of raw module data.
```
larskuhtz, line 396:
```suggestion
        -- ^ module cache size limit in bytes
```

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
larskuhtz: > I think lines 136-138 could be simply modify' $ liftIO . checkDbCache u b txid?

Unfortunately `modify'` returns just `()`. There is probably a way to do it by combining `MaybeT`, and `StateT` with some lensy tricks, but that's a bit above my production type proficiency.

src/Chainweb/Pact/Backend/DbCache.hs
```diff
@@ -0,0 +1,236 @@
...
+
+data CacheEntry a = CacheEntry
+  { _ceTxId :: !TxId
+    -- ^ Priority for cache eviction
+  , _ceAddy :: !CacheAddress
+    -- ^ Key. Uniquely identifies entries in the cache
+  , _ceSize :: !Int
+    -- ^ The size of the entry. Used to keep track of the cache limit
+  , _ceData :: !a
+    -- ^ Cached data.
```
larskuhtz, line 65:
```suggestion
    -- ^ Cached data. This is stored in a compact region.
```

src/Chainweb/Pact/PactService.hs
```diff
@@ -128,46 +128,50 @@ initPactService'
...
-
-        -- If the latest header that is stored in the checkpointer was on an
-        -- orphaned fork, there is no way to recover it in the call of
-        -- 'initalPayloadState.readContracts'. We therefore rewind to the latest
-        -- avaliable header in the block header database.
-        --
-        exitOnRewindLimitExceeded $ initializeLatestBlock (_pactUnlimitedInitialRewind config)
-        act
+initPactService' ver cid chainwebLogger bhDb pdb sqlenv config act =
+    -- checkpointEnv <- initRelationalCheckpointer initialBlockState sqlenv cplogger ver cid
```
larskuhtz, line 132:
```suggestion
```

src/Chainweb/Pact/PactService.hs
```diff
@@ -128,46 +128,50 @@ initPactService'
...
-    initialBlockState = initBlockState $ genesisHeight ver cid
+    initialBlockState = initBlockState (int $ _pactModuleCacheLimit config) $ genesisHeight ver cid
     loggers = pactLoggers chainwebLogger
     cplogger = P.newLogger loggers $ P.LogName "Checkpointer"
     pactLogger = P.newLogger loggers $ P.LogName "PactService"
     gasLogger = P.newLogger loggers $ P.LogName "GasLogs"
 
+    checkpointerLogger = addLabel ("sub-component", "checkpointer") chainwebLogger
+
+
```
larskuhtz, line 174:
```suggestion
```

tools/cwtool/TxSimulator.hs
```diff
@@ -82,7 +82,7 @@ simulate sc@(SimConfig dbDir txIdx' _ _ cid ver) = do
   pwos <- fetchOutputs sc cenv hdrs
   withSqliteDb cid cwLogger dbDir False $ \sqlenv -> do
     cpe@(CheckpointEnv cp _) <-
-      initRelationalCheckpointer (initBlockState 0) sqlenv logger ver cid
+      initRelationalCheckpointer (initBlockState (1024^(3::Int)) 0) sqlenv logger ver cid
```
larskuhtz, line 85: Use default prod value

src/Chainweb/Pact/Backend/DbCache.hs
```diff
@@ -0,0 +1,236 @@
...
+  , _ceSize :: !Int
+    -- ^ The size of the entry. Used to keep track of the cache limit
+  , _ceData :: !a
+    -- ^ Cached data. This is stored in a compact region.
+  , _ceHits :: Int
+    -- ^ Count of cache hits for this entry
+  }
+
+ceTxId :: Lens' (CacheEntry a) TxId
+ceTxId = lens _ceTxId (\a b -> a { _ceTxId = b })
```
jwiegley, line 71: Is there a reason we don't want to use `makeLenses ''CacheEntry` here?

src/Chainweb/Pact/Backend/DbCache.hs
```diff
@@ -0,0 +1,236 @@
...
+ceHits = lens _ceHits (\a b -> a { _ceHits = b })
+
+instance Eq (CacheEntry a) where
+  a == b = _ceAddy a == _ceAddy b && _ceTxId a == _ceTxId b
+
+-- | Sort by '_ceTxId' so that smaller tx ids are evicted first from the
+-- cache. '_ceAddy' is just for stable sort.
+--
+instance Ord (CacheEntry a) where
+  a `compare` b = cmp _ceTxId (cmp _ceAddy EQ)
```
jwiegley, line 83: I assume you mean here `comparing _ceTxId a b <> comparing _ceAddy a b`?

src/Chainweb/Pact/Backend/DbCache.hs
```diff
@@ -0,0 +1,236 @@
...
+    -> TxId
+       -- ^ Current TxId from checkpointer, used as priority for cache eviction.
+       -- Smaller values are evicted first.
+    -> DbCache a
+    -> IO (Maybe a, DbCache a)
+checkDbCache key rowdata txid = runStateT $ do
+    readCache txid addy >>= \case
+
+        -- Cache hit
+        Just !x -> do
```
jwiegley, line 150: Why do you need strictness on `x` here?

src/Chainweb/Pact/Backend/DbCache.hs
```diff
@@ -0,0 +1,236 @@
...
+        -- GCed only when no pointer to any data in the region is live any more.
+        --
+        Nothing -> do
+            c <- lift $ compact (decodeStrict {- lazy, forced by compact -} rowdata)
+            case getCompact c of
+                Nothing -> return Nothing
+                Just m -> do
+                    s <- lift $ compactSize c
+                    writeCache txid addy (fromIntegral s) m
+                    modify' (dcMisses +~ 1)
```
jwiegley, line 166: I'd kind of prefer `dcMisses +=' 1`, provided we define a set of lens operators based on `modify'`. But this is fine too.

src/Chainweb/Pact/Backend/DbCache.hs
```diff
@@ -0,0 +1,236 @@
...
+    :: TxId
+        -- ^ Current TxId from checkpointer
+    -> CacheAddress
+    -> StateT (DbCache a) IO (Maybe a)
+readCache txid ca = do
+    mc <- use dcStore
+    forM (HM.lookup ca mc) $ \e -> do
+        modify'
+            $ (dcStore . ix ca . ceTxId .~ txid)
+            . (dcStore . ix ca . ceHits +~ 1)
```
jwiegley, line 199: You could replace this with:
```
        modify' $ \s ->
          s . dcStore . ix ca & ceTxId .~ txid
                              & ceHits +~ 1
```

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -603,7 +603,7 @@ runSQLite' runTest sqlEnvIO = runTest $ do
     cp <- initRelationalCheckpointer initialBlockState sqlenv logger testVer testChainId
     return (cp, sqlenv)
   where
-    initialBlockState = set bsModuleNameFix True $ initBlockState $ genesisHeight testVer testChainId
+    initialBlockState = set bsModuleNameFix True $ initBlockState (1024^(2::Int)) $ genesisHeight testVer testChainId
```
jwiegley, line 606: Could we make this a symbolic constant somewhere? We seem to refer to this 1 megabyte constant in a few places, but it's not clear from reading the test what it implies.

test/Chainweb/Test/Pact/DbCacheTest.hs
```diff
@@ -0,0 +1,98 @@
...
+
+import Test.Tasty
+import Test.Tasty.HUnit
+
+tests :: TestTree
+tests = testGroup "Chainweb.Test.Pact.DbCacheTest"
+    [ testCache ]
+
+defaultChunkSize :: Int
+defaultChunkSize = 2^(15::Int)
```
jwiegley, line 25: Maybe we just say 32_768 here?

larskuhtz:
Commit [9c4634f](https://github.com/kadena-io/chainweb-node/pull/1577/commits/9c4634fb21b72600d6b21c6930d9dfbb23a3b1c1) requires a mainnet replay.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -154,22 +165,35 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
...
         -- there's no point in looking up the key.
         checkDbTableExists tableName
         result <- lift $ callDb "doReadRow"
                        $ \db -> qry db queryStmt [SText rowkey] [RBlob]
         case result of
             [] -> mzero
-            [[SBlob a]] -> MaybeT $ return $! decode $ fromStrict a
+            [[SBlob a]] ->
+              checkCache rowkey a <|>
+              (MaybeT $ return $! decodeStrict' a)
```
edmundnoble, line 185: Should we be reading from sqlite before checking the cache?

src/Chainweb/Pact/Backend/DbCache.hs
edmundnoble: We definitely don't, seeing as `readCache` returns `_ceData e` and `_ceData` is a strict field.

src/Chainweb/Pact/PactService.hs
```diff
@@ -128,46 +128,48 @@ initPactService'
...
-    runPactServiceM pst pse $ do
-
-        -- If the latest header that is stored in the checkpointer was on an
-        -- orphaned fork, there is no way to recover it in the call of
-        -- 'initalPayloadState.readContracts'. We therefore rewind to the latest
-        -- avaliable header in the block header database.
-        --
-        exitOnRewindLimitExceeded $ initializeLatestBlock (_pactUnlimitedInitialRewind config)
-        act
+initPactService' ver cid chainwebLogger bhDb pdb sqlenv config act =
```
edmundnoble, line 131: It's a bit disturbing that we have an `initX` function calling a `withX` function. This PR isn't responsible, but perhaps at some point we can consider renaming `initPactService'` to `withPactService`.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
larskuhtz: The cache is not caching the result of the db query, but the result of the result of decoding the row data.

Line 164 reads the row data from the db. Line 184

src/Chainweb/Pact/Backend/DbCache.hs
larskuhtz: We would have to export it from the module (which we don't want) or GHC would warn about unused symbols. I guess there's a way to only generate some lenses. But, I don't think it's worth adding TH overhead for those four lines of code.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
larskuhtz: Oh, line 185 is needed for the `noCache`. I guess, it should be moved there.

src/Chainweb/Pact/Backend/DbCache.hs
larskuhtz: Also, `_ceData` is stored in a compact regions so it is fully evaluated.

src/Chainweb/Pact/Backend/DbCache.hs
larskuhtz: Agreed. Though I propose that for a separate PR that would introduce those operators and apply them pervasively across the code base. (Actually, I am pretty sure that we worked on that before in pact, but it seems it never got finalized).

src/Chainweb/Pact/Backend/DbCache.hs
larskuhtz: I tried, but it didn't type check.

src/Chainweb/Pact/PactService.hs
larskuhtz: I renamed `initPactService` into `runPactService`, I think, that should make it clear that it is a long-running service process.

test/Chainweb/Test/Pact/DbCacheTest.hs
```diff
@@ -0,0 +1,98 @@
...
+
+tests :: TestTree
+tests = testGroup "Chainweb.Test.Pact.DbCacheTest"
+    [ testCache ]
+
+defaultChunkSize :: Int
+defaultChunkSize = 32768
+
+-- | Create entries with a size the corresponds to the default
+-- chunk size of the cache.
```
sirlensalot, line 28: There's a lot of internals here that are exposed to the test making it difficult to understand.

1) What is the origin of `defaultChunkSize`

2) Why do we care about chunk size when creating test data? Modules are arbitrarily large/small so it seems like we're crafting the test to the implementation here.

test/Chainweb/Test/Pact/DbCacheTest.hs
```diff
@@ -0,0 +1,98 @@
...
+--
+entry :: MonadIO m => Int -> Char -> m ([String], Int)
+entry i c = do
+    let e = [replicate (i * defaultChunkSize `div` 256) c]
+    s <- liftIO $ compactSize =<< compact e
+    return (e, fromIntegral s)
+
+combineEntries :: MonadIO m => Monoid a => a -> a -> m (a, Int)
```
sirlensalot, line 36: I don't think we need this. The original version was using string literals so `<>` was useful, but now `entry` is explicitly sized in integers so let's just call it directly, such that if `a` is `1`, `b` is `2` and `c` is `4`, etc.

src/Chainweb/Pact/Backend/DbCache.hs
```diff
@@ -0,0 +1,245 @@
...
+            modify' (dcHits +~ 1)
+            return $ Just x
+
+        -- Cache miss: decode module and insert into cache
+        --
+        -- The cache entry is stored within a compact region. The whole region is
+        -- GCed only when no pointer to any data in the region is live any more.
+        --
+        Nothing -> do
+            c <- lift $ compact (decodeStrict {- lazy, forced by compact -} rowdata)
```
sirlensalot, line 169: It'd be nice if we could encapsulate the compact region functionality to use in the test instead of having to replicate it there (could even save a dep in the test stanza). What about

```haskell
compactCacheEntry :: FromJSON a => ByteString -> Maybe (T2 a Word)
compactCacheEntry rowdata = do
    c <- lift $ compact (decodeStrict rowdata)
    case getCompact c of
      Nothing -> return Nothing
      Just m -> do
        s <- lift $ compactSize c
        return $ T2 m c
```

This way the test can simply use `compactCacheEntry`, and the prod code could just traverse over the result to update the cache.

test/Chainweb/Test/Pact/DbCacheTest.hs
larskuhtz: 1. Origin of default chunk size is in `ghc-compact`. It's an implementation constant for compact regions.

2. We care about the chunk size in order to test cache eviction. The the size of cache entries depends on this value. The size of the input is crafted so that it triggers eviction. The precise values don't matter, just the rough order.

We are crafting the test inputs such that the cache exhibits the behavior that we want to test.

test/Chainweb/Test/Pact/DbCacheTest.hs
```diff
@@ -0,0 +1,88 @@
...
+
+import Test.Tasty
+import Test.Tasty.HUnit
+
+tests :: TestTree
+tests = testGroup "Chainweb.Test.Pact.DbCacheTest"
+    [ testCache ]
+
+-- | Create entries with a size the corresponds to the default
+-- chunk size of the cache.
```
sirlensalot, line 25: this comment is out of date now no? chunk size is unused.

test/Chainweb/Test/Pact/DbCacheTest.hs
```diff
@@ -0,0 +1,88 @@
...
+--
+entry :: MonadIO m => String -> m ([String], Int)
+entry c = do
+    s <- liftIO $ compactSize =<< compact [c]
```
sirlensalot, line 26: Can we use `compactCacheEntry` here? It would get rid of the `ghc-compact` dep and not have us replicating compaction code in the test.

larskuhtz: well, it may pay off if we refactor the tests to move the encode step out of `doCheck` below.

larskuhtz: Hmm, I think, it may not be the best idea, because the test wouldn't cover creation of the compact region any more, because we would give it already a compact region as input. The test would still work, but it would reduce coverage.

jwiegley (approved):

sirlensalot (approved):

Issue #1575, opened 2022-12-05
On-chain error reporting should be restored


CryptoPascal31:
Since this PR: https://github.com/kadena-io/chainweb-node/pull/1543
on-chain errors are not reported anymore.

I ignore the rational behind, maybe to reduce the block size ? **But IMHO this is a big regression.**

Using the `/local` l endpoint can't be an acceptable workaround, because conditions on chain could change between the /local call and the moment where is the transaction is being mined. In case of a `/local` success and on-chain failure, user must be able to troubleshoot.

Moreover, with some other people as well I give support on Discord. 
I often explain to people why their transaction failed, giving guidelines to try their transactions again. But now, this becomes impossible. 

Post-replaying the failed transaction in  `/local` for troubleshooting is not possible:
- Most of the time, I don't have the required keys
- Conditions on-chain have changed => replaying the transaction in the same conditions is not useful.

Please restore the feature, as soon as possible. It's a barrier to adoption. What should think a new user, when this transaction fails, having no way to understand why ?

masch1na:
I 100% agree it should be restored for all of the above mentioned reasons. Imagine being a new user on Kadena and your transaction fails, but you have no idea why. Additionally, this basically doesn't even work as explained in my example scenario.

Example scenario:
If the original transaction failed because of "insufficient output amount" for example, where there wasn't enough liquidity let's say, but currently, at this point of time there is enough liquidity, how is sending the transaction again using /local helpful in finding out what the original error was? 

The transaction I would send now using /local would pass, because there is enough liquidity now and so it wouldn't return the original error.

tuncatunc:
I’m new to Kadena and every other blockchain has error messages and even logs. Kadena should have it too.

jwiegley:
Allowing errors to appear on chain creates an unfortunate sort of dependency, since it makes any change to error messages a “forking change” — due to the fact that all chain data is hashed. There was also a mismatch in certain places between the gas cost of an operation, and the compute time required to generate error messages when that operation failed. For these reasons, errors have been reduced for the time being to a simple “Error” note on-chain. We will gradually be bringing back more concise and useful on-chain error messages.

CryptoPascal31:
IMHO At least, the error message emitted by a failed `(enforce ..)` should be returned. Call stack is less important.

Since `(enforce ..)` is now lazily evaluated, devs are free to build informative and explicit error messages.

masch1na:
[...]

PR #1585, opened 2022-12-22
Improved /local preflight simulation

emilypi:
In Chainweb 2.17, we removed error reporting from `/send`. However, we didn't have "best simulations" for users to try their code out in as close to real world context as possible. `/local` used to fill this gap, but it was not robust enough. This PR enhances the `/local` endpoint to allow the user toggle on or off "preflight" (as close as we can get to mainnet) simulations, as well as the ability for the user to specify rewinds and simulation at a particular block height, and the ability to bypass user signature verification. 

The usage is thus: 

- The behavior of the bare `/local` endpoint has not changed, allowing people to keep their work flows.
- The `/local?preflight=<bool>` query flag allows people to run their commands against the full mainnet workflow. If `false` or left unspecified, reverts to the legacy behavior (running a tx through the interpreter to get outputs)
- The `/local?rewindDepth=<word (non-negative integer)>` parameter allows people to rewind to a particular block height `rewindDepth`-many block heights in the past and run their commands (think: regressing against forks). If left unspecified, defaults to running at the most current block height on that chain.
- The `/local?signatureVerification=<bool>` query flag allows people to run their commands with or without user signature verification (useful in the case of detached signature signing). If unspecified, defaults to verifying the tx.

An example full query would be `/local?preflight=true&rewindDepth=100&signatureVerification=true` to run a full command workflow at a blockheight 100 heights in the past, with full signature verification. 

See: https://github.com/kadena-io/chainweb-node/issues/1575


src/Chainweb/Pact/PactService.hs
```diff
@@ -651,7 +651,7 @@ execLocal cwtx preflight = withDiscardedBatch $ do
         --
         r <- liftIO $ if preflight
           then do
-            T2 cr _mc' <-
+            T2 cr _mc' <- do
```
jwiegley, line 664: Unneeded `do`.

src/Chainweb/Pact/RestAPI/Server.hs
```diff
@@ -349,7 +349,7 @@ localHandler logger pact preflight cmd = do
     case r of
       Left err ->
         throwError $ err400 { errBody = "Execution failed: " <> BSL8.pack (show err) }
-      (Right !r') -> return r'
+      (Right !r') -> error "branch on metadata error formatting or command"
```
jwiegley: We should never call `error`; I assume this is just a temporary hole to be filled with real code?

emilypi: This was a todo to stub out how we were going to handle metadata validations. It's fixed now

src/Chainweb/Pact/PactService.hs
emilypi: Fixed in recent iterations

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -279,21 +278,94 @@ localChainDataTest iot nio = do
...
+      "Gas price decimal precision too high"
+
+    step "Execute preflight /local tx - network id mismatch"
+    SubmitBatch (cmd5 NEL.:| []) <- mkTx mv =<<
+      mkCmdBuilder sigs Mainnet01 sid 1000 gp
+    runClientFailureAssertion sid cenv cmd5 "Network id mismatch"
+
+    step "Execute preflight /local tx - invalid signers/usersigs"
+
+    step "Execute preflight /local tx - too many sigs"
```
jwiegley, line 319: I'm guessing this is still to be done?

emilypi: This needs to be removed because we cannot actually hit the mentioned `invalid signers/usersigs` problem via the /local endpoint (it calls `verifyCommand` which validates this)

jwiegley: Got it, thanks!

emilypi: Btw, are you on an old commit? I don't see this in the diff.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -273,6 +277,93 @@ localChainDataTest iot nio = do
...
+    runClientFailureAssertion sid cenv cmd4
+      "Gas price decimal precision too high"
+
+    step "Execute preflight /local tx - network id mismatch"
+    cmd5 <- mkTx mv =<< mkCmdBuilder sigs Mainnet01 sid 1000 gp
+    runClientFailureAssertion sid cenv cmd5 "Network id mismatch"
+
+    step "Execute preflight /local tx - too many sigs"
+    let pcid = Pact.ChainId $ chainIdToText sid
+    sigs1' <- (>>= replicate 101) <$> testKeyPairs sender00 Nothing
```
jwiegley, line 321: Comprehension-wise, this one is a bit nasty. I think you're using `<$>` to fmap into something that returns a list, and then you're using the list monad's bind operator to generate a list of lists? Is there some way to use a point to make this a bit more transparent to future editors?

test/Chainweb/Test/RestAPI/Utils.hs
```diff
@@ -119,28 +120,42 @@ repeatUntil test action = retrying testRetryPolicy
...
       --
-      runClientM (pactLocalApiClient v sid cmd) cenv >>= \case
+      runClientM (pactLocalWithQueryApiClient v sid pf rd cmd) cenv >>= \case
         Left e -> throwM $ LocalFailure (show e)
         Right t -> return t
   where
     h _ = Handler $ \case
-      LocalFailure _ -> return True
-      _ -> return False
+      LocalFailure _ -> pure False
```
jwiegley, line 146: Why the boolean flip? Was this a bug?

src/Chainweb/Pact/PactService.hs
```diff
@@ -622,21 +625,62 @@ execNewGenesisBlock miner newTrans = withDiscardedBatch $
...
+                pure $ Right cr
+              Left e -> pure $ Left e
+          else liftIO $ do
+            cr <- applyLocal
+              logger _psGasLogger pdbenv
+              chainweb213GasModel ctx spv
+              cwtx mc execConfig
+            pure $ Right cr
+
+        return $! Discard (toHashCommandResult <$> r)
```
jwiegley, line 683: This will have no effect. The construction is already in WHNF form: We know it's the `Discard` constructor!

emilypi: This was as it was before - i can remove the bang if we care to

test/Chainweb/Test/Pact/RemotePactTest.hs
emilypi: It's actually a bit of a terse hack - `testKeyPairs` always produces a singleton list, so i'm using the list monad instance to join a singleton list of lists, which is... a list. I can make this less opaque if you want.

test/Chainweb/Test/RestAPI/Utils.hs
emilypi: This was a stale change - nice catch. Turns out we never trip it because we don't hit any retry cases in the local tests anymore.

test/Chainweb/Test/Pact/RemotePactTest.hs
emilypi: Updated to be more readable

src/Chainweb/Pact/PactService.hs
```diff
@@ -627,10 +627,12 @@ execLocal
     => ChainwebTransaction
     -> Bool
       -- ^ preflight flag
+    -> Bool
```
imalsogreg, line 630: nit: A bool arg with a negative meaning might get confusing. Worth adding an enum `data SigVerification = On | Off`?

emilypi: Yeah... i've been thinking about whether a negative param makes sense here and I tend to agree. It makes the API a little awkward. I'll do a pass to refactor today.

emilypi: Addressed in later commits

jwiegley: @imalsogreg We're on the same page! I suggested this to Emily yesterday:
```
data SigVerification = NoVerifySigs | VerifySigs
```
to put some glasses on all of this boolean blindness floating around.

src/Chainweb/Mempool/InMem.hs
```diff
@@ -265,7 +265,7 @@ addToBadListInMem lock txs = withMVarMasked lock $ \mdata -> do
     let !pnd' = foldl' (flip HashMap.delete) pnd txs
     -- we don't have the expiry time here, so just use maxTTL
     now <- getCurrentTimeIntegral
-    let (ParsedInteger mt) = maxTTL
+    let (ParsedInteger mt) = defaultMaxTTL
```
imalsogreg, line 268: Sorry for the lazy question: does the existence of a `defaultMaxTTL` mean that there is some machinery for overriding that default? If so, should that configurable value be used here?

Or, if there are ideas to make it configurable later, and that task will involve grepping for all uses of the default, replacing them with the configured value?

src/Chainweb/Pact/PactService.hs
```diff
@@ -622,21 +624,64 @@ execNewGenesisBlock miner newTrans = withDiscardedBatch $
 execLocal
     :: CanReadablePayloadCas tbl
     => ChainwebTransaction
-    -> PactServiceM tbl (P.CommandResult P.Hash)
-execLocal cmd = withDiscardedBatch $ do
+    -> Maybe LocalPreflightSimulation
+      -- ^ preflight flag
+    -> Maybe LocalSignatureVerification
+      -- ^ turn off signature verification checks?
+    -> Maybe BlockHeight
```
imalsogreg, line 631: It's slightly confusing to have `BlockHeight` (which has absolute vibes) used as a delta (which has relative vibes). Maybe just another couple words of documentation could help.

Or more drastically, changing the parameter to be a function `(BlockHeight -> BlockHeight)`, and the user would pass `id` to indicate they want no rewinding, and `const (BlockHeight 10)` to indicate they want to play from block height 10.

i guess the case I'm worrying about is someone seeing the `BlockHeight` type and passing `Just (BlockHeight 1000)` to try to play from block height 1000. Confusing a relative value for an absolute one.

src/Chainweb/Pact/RestAPI.hs
```diff
@@ -119,6 +127,22 @@ pactPollApi
     . Proxy (PactPollApi v c)
 pactPollApi = Proxy
 
+-- -------------------------------------------------------------------------- --
+-- POST Queries for Pact Local Pre-flight
```
imalsogreg, line 131: Is the comment too specific? (Is this POST query also for /local endpoint with no pre-flight?)

src/Chainweb/Pact/PactService.hs
emilypi: That's a good point. We're not rewinding to a block height, but to a block height delta. When we configure the checkpointer to execute for a rewind, it cares about a blockheight delta defined to be blockheights to be subtracted from the current height. We don't really have a good language for expressing this. I would probably reject the idea that it should be a function, since it's always going to be an override if it's a positive height (it's either `id` or `\_ -> rewindDepth`, with no in between). So perhaps this warrants a newtype a la 

```haskell
newtype RewindDepth = RewindDepth BlockHeight
  deriving ... via BlockHeight
```

It can just exist with the other local data types. However, at this point i think we're kind of getting into defining a universe of types specific to the theory of mind for what a particular endpoint intends to do with existing datatypes rather than defining concretely different things. If we're okay with doing that, I'd want to follow up this PR with a PR that changes it for the Checkpointer as well, since the `BlockHeight` representation here percolates up from the API defined for the Checkpointer.

src/Chainweb/Pact/RestAPI.hs
emilypi: It's only for the preflight api - the original local api definition comes from Pact proper

src/Chainweb/Mempool/InMem.hs
emilypi: I grepped around and actually, i think `InMem.hs` is the only place that actually uses it, so perhaps we might want to just inline it there

emilypi: So to answer your question - there's no means of overriding it once it's in, but it does also apply to transactions we use to test, since we specify a max admissible ttl there too and those can be overwritten.

src/Chainweb/Pact/PactService.hs
emilypi: Addressed in DM

src/Chainweb/Pact/RestAPI.hs
imalsogreg: Ah. I think I was conflating "preflight API" with a particular preflight query parame.

imalsogreg (approved):
Looks great!

jwiegley (approved):

PR #1589, opened 2023-01-04
Use new rocksdb-haskell without tbb

edmundnoble:


larskuhtz (approved):

jwiegley:
@edmundnoble What does this mean then for ARM64 architectures, which don't have SSE instructions?

PR #1591, opened 2023-01-05
Try building in parallel in CI

edmundnoble:


.github/workflows/applications.yml
```diff
@@ -333,9 +333,9 @@ jobs:
     - name: Install build dependencies
       run: cabal build chainweb --only-dependencies
     - name: Build chainweb library
-      run: cabal build lib:chainweb
+      run: cabal build --ghc-options=-j2 lib:chainweb
```
larskuhtz, line 336: Why not `cabal build -j2` here, too?

.github/workflows/applications.yml
```diff
@@ -333,9 +333,9 @@ jobs:
...
     - name: Build chainweb applications
-      run: cabal build exe:chainweb-node test:chainweb-tests exe:cwtool chainweb:bench:bench
+      run: cabal build -j2 --ghc-options=-j2 exe:chainweb-node test:chainweb-tests exe:cwtool chainweb:bench:bench
```
larskuhtz, line 338: Does't `cabal build -j2` imply `ghc -j2?

larskuhtz (approved):

.github/workflows/applications.yml
edmundnoble: If only cabal and GHC coordinated their parallelism.

PR #1592, opened 2023-01-11
Update link to whitepapers

### LINKED ISSUE #1590: Incorrect Whitepaper link in "About" Section
The "About" section of this directory links to the Whitepapers, but when I click on the link there (https://docs.kadena.io/basics/whitepapers/overview), I get a "Page Not Found" error.

The new link to the Whitepapers should be: https://docs.kadena.io/kadena-docs/whitepapers/overview

jwiegley:
Fixes #1590


DevopsGoth:
The docs are being reverted to the previous system, so the old link should be good.  I would not merge this request, as it should not be needed after the docs are restored.

jwiegley:
Understood, thank you!

PR #1593, opened 2023-01-11
Payload block height index

edmundnoble:


bench/Chainweb/Pact/Backend/ForkingBench.hs
```diff
@@ -186,7 +186,7 @@ mineBlock
     -> IO (T3 ParentHeader BlockHeader PayloadWithOutputs)
 mineBlock parent nonce pdb bhdb pact = do
     !r@(T3 _ newHeader payload) <- createBlock True parent nonce pact
-    addNewPayload pdb payload
+    addNewPayload pdb (succ $ _blockHeight $ _parentHeader parent) payload
```
jwiegley, line 189: This is a hair split, but using `$` here isn't too helpful is it? I'd prefer just regular parens for argument passing to pure functions. `succ (_blockHeight (_parentHeader parent))` uses exactly the same number of characters as the dollar version, but is instantly friendly to readers of any other programming language.

src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
```diff
@@ -303,7 +301,7 @@ headerStreamHandler db = Tagged $ \req resp -> do
 
     g :: BlockHeader -> IO HeaderUpdate
     g bh = do
-        x <- casLookupM cas $ _blockPayloadHash bh
+        Just x <- lookupPayloadWithHeight cas (_blockHeight bh) $ _blockPayloadHash bh
```
jwiegley, line 304: If this partial match fails, we'll just abort. I'd like to see better error reporting here.

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -371,7 +371,7 @@ sweepPayloads
     -> [Filter BlockPayloadHash]
     -> IO (Filter BlockTransactionsHash, Filter BlockOutputsHash)
 sweepPayloads logg db markedPayloads = do
-    logg Info $ "Sweeping BlockPayloads"
+    logg Info "Sweeping BlockPayloads"
```
jwiegley, line 374: Music to my soul.

src/Chainweb/Chainweb/PruneChainDatabase.hs
```diff
@@ -528,7 +526,8 @@ deleteBlockPayload
     -> IO ()
 deleteBlockPayload logg db p = do
     logg Debug $ "Delete PayloadHash for " <> encodeToText (_blockPayloadPayloadHash p)
-    tableDelete pdb (_blockPayloadPayloadHash p)
+    Just h <- tableLookup (_transactionDbBlockPayloadHeights $ _transactionDb db) (casKey p)
```
jwiegley, line 529: Same here, need a descriptive error if the lookup fails.

src/Chainweb/Pact/SPV.hs
```diff
@@ -6,6 +6,7 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE ViewPatterns #-}
```
jwiegley, line 9: I don't see `ViewPatterns` being used in the changes to this module...

src/Chainweb/Payload/RestAPI/Server.hs
```diff
@@ -116,7 +125,7 @@ outputsHandler
     => PayloadDb tbl
     -> BlockPayloadHash
     -> Handler PayloadWithOutputs
-outputsHandler db k = liftIO (tableLookup db k) >>= \case
+outputsHandler db k = liftIO (fmap snd <$> tableLookup db k) >>= \case
```
jwiegley, line 128: Better than `fmap snd <$>`, why not just pattern match on `Just (_, e)` below?

jwiegley (changes requested):

src/Chainweb/Payload/RestAPI.hs
```diff
@@ -142,6 +148,25 @@ payloadPostApi
...
+    :> Post '[JSON] [PayloadData]
+
+type HeightedPayloadPostApi (v :: ChainwebVersionT) (c :: ChainIdT)
+    = 'ChainwebEndpoint v :> ChainEndpoint c :> HeightedPayloadPostApi_
+
+heightedPayloadPostApi
+    :: forall (v :: ChainwebVersionT) (c :: ChainIdT)
+    . Proxy (HeightedPayloadPostApi v c)
+heightedPayloadPostApi = Proxy
+
```
larskuhtz, line 169: What about just using the normal payload batch endpoint with optional block height property in the request body on a per payload basis? The implementation could do something like:

1. Collect and sort all request items that have a (non-null) value for block height
2. group collected items in continuous range.
3. Query each range group using an rocksdb iterator
4. Collect all request items that don't have a block height
5. (optional) order response items according to order of request items

For backward compatibility, the response would included in the block height (see comment above) if it was present in the request.

This approach would result in heterogenous arrays, which is fine in JSON. Alternatively, we could require that the block height is specified for either all or none of the items, which would avoid heterogenous arrays.

On the P2P API we should probably enforce the block height property on request items. On the service API it would be optional. Maybe even there we could deprecate the query by payload hash, because payload hashes are obtained from the block header which also includes the height.

src/Chainweb/Payload/RestAPI.hs
```diff
@@ -142,6 +148,25 @@ payloadPostApi
...
 
+-- | @POST \/chainweb\/\<ApiVersion\>\/\<InstanceId\>\/chain\/\<ChainId\>\/heightedpayload\/batch@
+--
+-- The query may return any number (including none) of the requested payload
+-- data. Results are returned in any order.
+--
+type HeightedPayloadPostApi_
+    = "payload"
+    :> "batch"
+    :> ReqBody '[JSON] [(BlockHeight, BlockPayloadHash)]
```
larskuhtz, line 159: I think tuples are serialized as lists, Is that what we want, or would we rather prefer an object with meaningful property name?

src/Chainweb/Payload/RestAPI.hs
```diff
@@ -142,6 +148,25 @@ payloadPostApi
...
+    :> Post '[JSON] [PayloadData]
```
larskuhtz, line 160: I wonder if we should include the block height also in the response. E.g. analogous to the request as`[(Blockheight, PayloadData)]` (or whatever we end up using, cf. comment above).

src/Chainweb/Payload/RestAPI.hs
```diff
@@ -110,6 +115,7 @@ somePayloadDbVal v cid db = runIdentity $ do
 type PayloadGetApi_
     = "payload"
     :> Capture "BlockPayloadHash" BlockPayloadHash
+    :> QueryParam "blockHeight" BlockHeight
```
larskuhtz, line 118: Making this a path parameter (and possibly deprecating the query without it) may be a strong incentive for the use of the block height.

larskuhtz (commented):
Commented on the Payload API changes

src/Chainweb/Payload/RestAPI.hs
```diff
@@ -144,6 +148,25 @@ payloadPostApi
     . Proxy (PayloadPostApi v c)
 payloadPostApi = Proxy
 
+-- | @POST \/chainweb\/\<ApiVersion\>\/\<InstanceId\>\/chain\/\<ChainId\>\/heightedpayload\/batch@
+--
+-- The query may return any number (including none) of the requested payload
+-- data. Results are returned in any order.
+--
+type HeightedPayloadPostApi_
+    = "payload"
```
larskuhtz, line 157: this conflicts with 'PayloadPostApi_'. Should this be 'heightedpayload'?

PR #1595, opened 2023-01-17
verify coin

rsoeldner:


pact/coin-contract/v5/coin-v5.pact
```diff
@@ -545,6 +547,7 @@
 
         (validate-account sender)
         (validate-account receiver)
+        (enforce (!= sender receiver) "sender and receiver should differ")
```
emilypi, line 550: aha but they _can_ be the same! Just, they should exist on different chains, which we have no means of expressing in the current system.

pact/coin-contract/v5/coin-v5.pact
```diff
@@ -437,8 +438,9 @@
              (property (valid-account miner))
              ;(property conserves-mass) not supported yet
            ]
-
-    (step (buy-gas sender total))
+    (step (let ((unused 0))
+            (validate-account miner)
+            (buy-gas sender total)))
```
emilypi, line 443: what does `unused` get us here?

emilypi (changes requested):

pact/coin-contract/v5/coin-v5.pact
rsoeldner: the `let` is used to group the two expressions. We have to `validate-account miner` on the first step too.

pact/coin-contract/v5/coin-v5.pact
rsoeldner: this is enforced by the `crosschain-transfer` of the `fungible-v2` interface, see https://github.com/kadena-io/chainweb-node/blob/master/pact/coin-contract/v2/fungible-v2.pact#L94

emilypi: `fungible-v2` has an invalid property then :disappointed:

pact/coin-contract/v5/coin-v5.pact
emilypi: but do we need to `let` it? Can we not write `

```
(step 
  (validate-account miner)
  (buy-gas sender total))
```
  ???

rsoeldner: Unfortunately not, `step` takes a single expression ...

emilypi: we really need `progn` then.

jwiegley: Indeed, or better named would be `do`.

emilypi: Indeed.

jwiegley (approved):

emilypi (commented):
As discussed off-band, we'll not want a `fungible-v3`, but rather we'll want to modify `fungible-v2` to replace the offending models with the "correct" ones so that, while on-chain, the old `fungible-v2` will remain, but off-chain, people will be able to cargo-cult the fixed interface. Concretely, my proposal is this: 

1. `fungible-v3` should be renamed to `fungible-v2`, and replace the current `fungible-v2` i the main repo. 
2. The on-chain (i.e. old) `fungible-v2` should be moved to its own directory, including a note (say, CAVEAT.md), which details the discrepancy and why there are two, plus how to distinguish between them. 
3. We scope any potential changes to `coin` for the next release, which will include more than just what we spoke about. It will include changes such as shutting off account creation for non-principals, and removing `rotate`.

PR #1596, opened 2023-01-24
resource pool 0.4.0.0

larskuhtz:


emilypi (approved):

test/Chainweb/Test/Mempool/RestAPI.hs
```diff
@@ -98,12 +98,11 @@ destroyTestServer :: TestServer -> IO ()
 destroyTestServer = killThread . _tsServerThread
 
 newPool :: IO (Pool.Pool TestServer)
-newPool = Pool.newPool Pool.PoolConfig
-    { Pool.createResource = newTestServer
-    , Pool.freeResource = destroyTestServer
-    , Pool.poolCacheTTL = 10
-    , Pool.poolMaxResources = 20
```
emilypi, line 105: alternatively, it looks like the `Pool.poolNumStripes = Nothing` is the only addition here that materially changes. We could just make this a 1 line addition

PR #1550, opened 2022-11-08
Chainweb.Version recordification and overhaul

edmundnoble:
Integration test replay is expected to fail, because devnet parameters change. I will follow up by running the integration-tests before merge. EDIT: that is no longer true, the devnet parameters are now the same and devnet integration tests replay fine.

PR #1598, opened 2023-01-30
Make devnet fork heights earlier

edmundnoble:


jwiegley:
@edmundnoble Is there any downside to doing this? Why hasn't it happened before?

edmundnoble:
I wasn't planning to touch this PR until the automated integration tests can confirm that it works. Here's my understanding of the background:

I didn't know this until recently, but @larskuhtz mentioned that there is something special about the first hundred and twenty blocks or so on devnet with regard to difficulty adjustment. Accordingly something funky may happen when we do the forks, especially the 20 chain adjustment, so much earlier than that. That needs to be tested still. I suspect an issue would manifest as some difficulty weirdness making blocks either way too fast or way too slow. 

With that in mind as a possibility, using devnet with real miners is no longer a common usecase for devnet - usually we disable POW validation and use a simulation miner. Without real POW, quirks of difficulty adjustment don't matter to us. So we can likely get away with much faster forks on devnet and make "fresh" integration test runs about 4x as fast. Again though, testing needed.

edmundnoble:
In favor of #1550

PR #1599, opened 2023-02-02
Add Warning System to `/local` preflight

emilypi:
TODO: 

- [x] Hookup warning system
- [x] Decide on final datatypes for /local output
  - Should we have enriched local output that is `CommandResult` when legacy is called, with warnings defined when preflight?
- [ ] ~~Fold pact 4.6 prep into this pr?~~

### DIFF
--- a/src/Chainweb/Pact/TransactionExec.hs
+++ b/src/Chainweb/Pact/TransactionExec.hs
@@ -105,7 +105,7 @@ import Chainweb.Pact.Templates
 import Chainweb.Pact.Transactions.UpgradeTransactions
 import Chainweb.Pact.Types hiding (logError)
 import Chainweb.Transaction
-import Chainweb.Utils (encodeToByteString, sshow, tryAllSynchronous, T2(..))
+import Chainweb.Utils (encodeToByteString, sshow, tryAllSynchronous, T2(..), T3(..))
 import Chainweb.Version as V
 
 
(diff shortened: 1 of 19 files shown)


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -700,6 +708,10 @@ applyExec initialGas interp em senderSigs hsh nsp = do
     for_ _erLogGas $ \gl -> gasLog $ "gas logs: " <> sshow gl
     logs <- use txLogs
     rk <- view txRequestKey
+
+    -- set tx warnings to eval warnings
+    txWarnings .= _erWarnings
```
jwiegley, line 713: `txWarnings` will never have an existing value here?

src/Chainweb/Pact/Validations.hs
```diff
@@ -64,7 +65,7 @@ assertLocalMetadata
     :: P.Command (P.Payload P.PublicMeta c)
     -> TxContext
     -> Maybe LocalSignatureVerification
-    -> PactServiceM tbl (Either MetadataValidationFailure ())
+    -> PactServiceM tbl (Either Text ())
```
jwiegley, line 68: Why aren't we classifying this text value anymore?

emilypi: `MetadataValidationFailure` was a newtype over `Text` originally, but I've lifted it into the ADT `LocalResult` here, which gets built directly in `execLocal` in `PactService.hs` to avoid committing this code to just being "local-specific", if that makes sense. It's a general series of validations, so it should deal with generally producing messages that other places in the code can use, and we can separate concerns a bit by lifting the result into whatever datatype we care to at the appropriate place.

src/Chainweb/Pact/TransactionExec.hs
jwiegley: Perhaps then you could just add an assert that txWarnings is always empty here? It would be helpful to know that this is an assumption, and that nothing will be thrown away. Alternatively, we could use `<>=`.

jmcardon (approved):

emilypi:
Merging on Green

PR #1603, opened 2023-02-08
update setup-haskell action to latest version

larskuhtz:


lbpage (approved):
This looks good and it might be good to update other sections to include "startsWith" at some point.

PR #1610, opened 2023-02-15
Add node upgrade process to the readme

Luzzotica:
Not sure if this is the best place to put changes like this, but having it here feels like a good start.


jwiegley:
Hi @Luzzotica! Thanks for the list, that's very thoughtful. We're actually working on just such a list right now, something that will integrate with an Asana checklist so that we ensure a full workthrough each cycle, but I'll suggest reflecting some of that information here as well. It adds a few more items to what you have, and a bit more depth on some of the items, but the spirit of the idea is quite similar!

PR #1611, opened 2023-02-18
Optimize JSON+base64 encoding

edmundnoble:
base64 text is already both a) valid UTF8 and b) escaped with respect to JSON. So why spend time encoding to UTF8, or escaping for JSON? This PR stops us doing that.

### DIFF
--- a/src/Chainweb/Utils.hs
+++ b/src/Chainweb/Utils.hs
@@ -103,6 +103,7 @@ module Chainweb.Utils
 , encodeB64UrlText
 , decodeB64UrlText
 , encodeB64UrlNoPaddingText
+, b64UrlNoPaddingTextEncoding
 , decodeB64UrlNoPaddingText
 
 -- ** JSON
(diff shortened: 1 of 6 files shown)


src/Chainweb/Utils.hs
```diff
@@ -650,6 +653,12 @@ encodeB64UrlNoPaddingText :: B.ByteString -> T.Text
 encodeB64UrlNoPaddingText = T.dropWhileEnd (== '=') . T.decodeUtf8 . B64U.encode
 {-# INLINE encodeB64UrlNoPaddingText #-}
 
+-- | Encode a binary value to a base64-url (without padding) JSON encoding.
+--
+b64UrlNoPaddingTextEncoding :: B.ByteString -> Encoding
+b64UrlNoPaddingTextEncoding t =
+    Aeson.unsafeToEncoding $ BB.char8 '\"' <> BB.byteString (B8.dropWhileEnd (== '=') $ B64U.encode t) <> BB.char8 '\"'
```
larskuhtz, line 660: It would be nice to use a version of `B64U.encode` (without padding) that would directly encode (lazily) into a Builder. But I think the bytestring-base64 package doesn't offer that.

larskuhtz (approved):

PR #1612, opened 2023-02-19
Preflight /local fixes: return metadata in results

emilypi:
As discussed: 

- `?preflight=true` should return metadata (requiring enrichment in `applyCmd` - it does not by default). 
- Add Hee Kyun's rewind regression. 
- Add tests for metadata regressions


jwiegley (approved):

ak3n:
@edmundnoble I have added a check, please take a look.

emilypi:
It's been a while at this point, so I'm not actually sure what Hee Kyun's rewind regression means at this point. My best guess would be that metadata exists across fork points, but since this is preflight, it's using `applyCmd` and fork flags should be sound, and this makes it so that metadata always exists regardless of whether errors were removed (or added back).

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -331,10 +331,14 @@ localPreflightSimTest iot nio = testCaseSteps "local preflight sim test" $ \step
         assertFailure "Preflight /local call produced legacy result"
       Right MetadataValidationFailure{} ->
         assertFailure "Preflight produced an impossible result"
-      Right (LocalResultWithWarns _ ws) -> case ws of
-        [w] | "decimal/integer operator overload" `T.isInfixOf` w ->
-          pure ()
-        ws' -> assertFailure $ "Incorrect warns: " ++ show ws'
+      Right (LocalResultWithWarns cr' ws) -> do
+        -- check the presence of metadata
+        assertBool "Preflight result should have metadata" $ isJust $ _crMetaData cr'
```
edmundnoble, line 336: Should we test exactly what metadata we get here?

emilypi: Existence here matters, otherwise we don't test specific metadata. We _could_ expand it, but i'm satisfied with existence.

edmundnoble: I'm happier with more coverage, if that's practical. Why not test metadata?

edmundnoble: We'll do that later, follow-up is on Asana.

edmundnoble (approved):

PR #1616, opened 2023-03-02
Housekeeping for p2p API

larskuhtz:


src/Chainweb/Chainweb.hs
```diff
@@ -587,12 +589,16 @@ withChainwebInternal conf logger peer serviceSock rocksDb pactDbDir backupDir re
 -- Throttling
 
 mkGenericThrottler :: Double -> IO (Throttle Address)
-mkGenericThrottler rate = mkThrottler 5 rate (const True)
+mkGenericThrottler rate = mkThrottler 30 rate (const True)
 
 mkPutPeerThrottler :: Double -> IO (Throttle Address)
-mkPutPeerThrottler rate = mkThrottler 5 rate $ \r ->
+mkPutPeerThrottler rate = mkThrottler 30 rate $ \r ->
```
jwiegley, line 595: Are you allowing more traffic than before here?

larskuhtz: This number is the how long a stale rate limiting bucket (for an address) is cached before being GCed. The value in in seconds. IIRC buckets are initialized full, so  increasing it restricts the number of requests that are available to a client.

edmundnoble (approved):

jwiegley (approved):

PR #1617, opened 2023-03-05
Force less

edmundnoble:
We do `Control.DeepSeq.force` a few things which don't need to be forced, which is expensive, this deletes those.


jwiegley (approved):

jwiegley:
Seems pretty straightforward, thank you. Why was `catchAllSynchronous` being used before?

edmundnoble:
Not really sure, I think the big reason was to make sure we only caught synchronous exceptions. The `force` was just some defensive programming I think. safe-exceptions should give us all we need.

larskuhtz:
I think, `catchAny` and `catchAllSynchronous` are the same. `catchAny` is the version from safe-exception and `catchAllSynchronous` is our own version from `Chainweb.Utils`. 

Originally, we didn't use safe-exceptions (IIRC the package was still in flux and also we are only using a fraction of it).

I think, we should try to be consistent and either go full safe-exceptions or stick to our own solution. I have a slight preference for shrinking the set of dependencies by avoiding packages from which we use only very little code. But I am fine with safe-exception, assuming that it is a de-facto standard that most Haskell devs are familiar with.

edmundnoble:
> I think, catchAny and catchAllSynchronous are the same. catchAny is the version from safe-exception and catchAllSynchronous is our own version from Chainweb.Utils.

Almost the same. `catchAllSynchronous` also forces its argument with `DeepSeq`.

larskuhtz:
> Almost the same. catchAllSynchronous also forces its argument with DeepSeq.

Oh, good point. Yeah, we should be careful about that. In particular pact service, but also the cut validation pipeline is sensitive with respect where exceptions surface. I don't think that it could cause correctness issues, but may have perf impacts or may even results in DoS vectors.

src/P2P/TaskQueue.hs
```diff
@@ -171,7 +171,7 @@ session_ limit q logFun env = mask $ \restore -> do
     let go = tryReadIVar (_taskResult task) >>= \case
             Nothing -> do
                 logg task Debug "run task"
-                flip catchAllSynchronous (retry task) $ restore $ do
+                handle (retry task) $ restore $ do
```
larskuhtz, line 174: This is only safe if we assume that all relevant exceptions arise from computations in `IO`, which probably is the case.

src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -530,7 +530,7 @@ lookupBlockHeader bhash ctx = do
       then return cur
       else do
         bhdb <- asks _psBlockHeaderDb
-        liftIO $! lookupM bhdb bhash `catchAllSynchronous` \e ->
+        liftIO $! lookupM bhdb bhash `catchAny` \e ->
```
larskuhtz, line 533: Are we sure that `BlockHeader` deserialization is forced nf in the result of `lookupM`. Realistically, we only expect failures from `IO` actions, like broken db connections or missing keys. But I wonder if corrupted db data could hide in an unevaluated thunk in the result.

larskuhtz (approved):
### Regarding `catchAny` vs. `catchAllSynchronous`:

`catchAllSynchronous` comes at a cost, but it defends against upstream changes (possibly in other packages) that could result in bad bugs. So, we should replace it only where we think it is absolutely safe.

We may want to rename `catchAllSynchronous` to be more explicit about deep forcing the result of the inner computation. Maybe `catchAnyForced`? If `catchAny` is all that we use form `safe-exception` we may consider implementing it in `Chainweb.Utils` and drop the dependency, in order to enforce a uniform exception handling style across the code base.

We should ensure (and possibly document) that `lookupM` returns a fully evaluated result.

Actually, it would be nice if that would just be the default and every Haskell programmer would explicitly mention for API methods when there's a chance that the result could contain unevaluated thunks. But realistically that will never happen.

I guess, the best option is to extensively use the nothunks package. I have been using it with pact and started using it with chainweb some time ago. It never got merged, I think. Maybe we should resume working on it.

src/Chainweb/Pact/PactService/Checkpointer.hs
edmundnoble: Good point, I have added some extra strictness to `decodeBlockHashRecord` for this but yes there are no other errors that are not thrown during deserialization.

PR #1618, opened 2023-03-06
Fix master

edmundnoble:
master is broken because new sqlite direct has an `open2` function which conflicts with ours. I've added a TODO to consider using theirs instead, but until then this fixes master.


jwiegley (approved):
Aren't these the same changes @larskuhtz made in his P2P cleanup PR?

edmundnoble:
Yeah I took them from that PR, will close seeing as it's in

PR #1619, opened 2023-03-06
Query headers as binary and disable object-encoded headers

edmundnoble:
* [x] disable `JsonBlockHeaderObject` on P2P API
* [x] use `Binary` encoding by default in the header client

### DIFF
--- a/src/Chainweb/BlockHeaderDB/RestAPI.hs
+++ b/src/Chainweb/BlockHeaderDB/RestAPI.hs
@@ -91,7 +91,6 @@ import Data.Bifunctor
 import Data.Foldable
 import Data.Proxy
 import Data.Text (Text)
-import Debug.Trace
 
 import Network.HTTP.Media ((//), (/:))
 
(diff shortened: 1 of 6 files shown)


src/Chainweb/BlockHeaderDB/RestAPI.hs
```diff
@@ -278,7 +299,13 @@ type HeadersApi_
     = "header"
     :> PageParams (NextItem BlockHash)
     :> FilterParams
-    :> Get '[JSON, JsonBlockHeaderObject] BlockHeaderPage
+    :> Get '[JSON, JsonBlockHeaderObject, OctetStream] BlockHeaderPage
```
larskuhtz, line 302: I think `OctetStream` isn't yet implemented.

src/Chainweb/BlockHeaderDB/RestAPI.hs
```diff
@@ -278,7 +299,13 @@ type HeadersApi_
...
+
+type P2pHeadersApi_
+    = "header"
+    :> PageParams (NextItem BlockHash)
+    :> FilterParams
+    :> Get '[JSON, OctetStream] BlockHeaderPage
```
larskuhtz, line 308: I think `OctetStream` isn't yet implemented

larskuhtz (approved):

PR #1621, opened 2023-03-10
branch bound limits

larskuhtz:
---
  - https://app.asana.com/0/0/1204717400575174


src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
```diff
@@ -137,14 +156,19 @@ branchHashesHandler
...
-    liftIO
-        $ branchKeys db nextChecked (succ <$> effectiveLimit) minr maxr
-            (_branchBoundsLower checkedBounds)
-            (_branchBoundsUpper checkedBounds)
-        $ finiteStreamToPage id effectiveLimit . void
+branchHashesHandler db limit next minr maxr bounds
+    | length (_branchBoundsUpper bounds) > defaultBoundsLimit = throwError $ err400Msg $
+        "upper branch bound limit exceeded. Only " <> show defaultBoundsLimit <> " values are supported."
+    | length (_branchBoundsLower bounds) > defaultBoundsLimit = throwError $ err400Msg $
+        "upper branch bound limit exceeded. Only " <> show defaultBoundsLimit <> " values are supported."
```
jwiegley, line 163: Should this say "lower branch bound limit exceeded"?

src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
```diff
@@ -156,6 +180,8 @@ branchHeadersHandler
     :: TreeDb db
     => ToJSON (DbKey db)
     => db
+    -> Int
```
jwiegley, line 183: How about we start using new types for these?

src/Chainweb/BlockHeaderDB/RestAPI/Server.hs
```diff
@@ -164,14 +190,19 @@ branchHeadersHandler
...
-    liftIO
-        $ branchEntries db nextChecked (succ <$> effectiveLimit) minr maxr
-            (_branchBoundsLower checkedBounds)
-            (_branchBoundsUpper checkedBounds)
-        $ finiteStreamToPage key effectiveLimit . void
+branchHeadersHandler db boundsLimit maxLimit limit next minr maxr bounds
+    | length (_branchBoundsUpper bounds) > boundsLimit = throwError $ err400Msg $
+        "upper branch bound limit exceeded. Only " <> show boundsLimit <> " values are supported."
+    | length (_branchBoundsLower bounds) > boundsLimit = throwError $ err400Msg $
+        "upper branch bound limit exceeded. Only " <> show boundsLimit <> " values are supported."
```
jwiegley, line 197: Same question as above here.

src/Chainweb/Chainweb.hs
```diff
@@ -753,11 +753,26 @@ runChainweb cw = do
...
+    -- NOTE: this may need to have to be adjusted if the p2p limits for batch
+    -- sizes or number of branch bound change. It may also need adjustment for
+    -- other protocol changes, like additional HTTP request headers or changes
+    -- in the mempool protocol.
+    --
+    -- FIXME: is 8K enough for the mempool?
+    --
+    p2pRequestSizeLimit :: Middleware
+    p2pRequestSizeLimit = requestSizeLimitMiddleware $
+        setMaxLengthForRequest (\_req -> pure $ Just $ 8 * 1024) -- 8KB
```
jwiegley, line 774: Would it be better to make these tunable values via the config file? Or do we want to specifically discourage that?

edmundnoble (approved):

larskuhtz:
Possible unit tests could check the limits of the API, although, again I think, it would be good to do that more systematically instead of just adding individual tests on an ad-hoc basis. 

We used to have comprehensive tests of REST parameters. But I don't think that has been maintained over time. But the test tests and the test setup should still be around -- so maybe we could update that to the latest APIs?

chessai:
Currently depends on https://github.com/kadena-io/chainweb-openapi/pull/20/files

edmundnoble (approved):

chessai:
> Currently depends on https://github.com/kadena-io/chainweb-openapi/pull/20/files

This is now merged, updated here as well

PR #1624, opened 2023-03-20
Explicit numeric conversions

edmundnoble:


imalsogreg:
It's a big improvement, making all these casts explicit.

I bet there are also some readability wins to be had by scanning the `int @_ @_`'s for cases where a purpose-named function can take the place of a generic conversion function. E.g.

```
case window ver of
     Just (WindowWidth w)
         | int @BlockHeight @Natural (_blockHeight h) <= w -> foo
```

->

```
case window ver of
  Just win | windowContainsBlock win h -> foo
```

jwiegley:
@edmundnoble What prompted this work?

jwiegley:
Would it be better to have separate newtypes for block heights and cut heights?

edmundnoble:
We do, they just... Both have Num instances.

jwiegley:
How often do we need to transparently use them as "numbers"? What if we removed those instances and made the conversions explicit?

edmundnoble:
That is a possibility too. But it will be more verbose. For example we frequently convert though between (`MaxRank`, `MinRank`) (TreeDB types) and `BlockHeight`, but `MaxRank` and `MinRank` wrap `Natural`, so we convert `MaxRank -> Natural -> Word64 -> BlockHeight`, three conversions.

PR #1626, opened 2023-03-30
remove redundant file rocksdb-arm64.patch

larskuhtz:
I assume this file is now part of the kadena-rocksdb-haskell package and not needed here any more. @jwiegley Is that correct?


edmundnoble:
This file was used on this line
https://github.com/kadena-io/chainweb-node/commit/b64db9c419f43c628a1fedb8bc9a37daa9a103bb#diff-245563af9e34754739b12b294f3324c1d42c116ace7ce8735e21dfbc82d8e064R68
which was subsequently deleted, so this is not used anymore.

edmundnoble (approved):

PR #1627, opened 2023-03-30
Fast devnet

edmundnoble:
Upcoming L2 needs custom genesis blocks, this is a step in that direction: a version of devnet, called "fast-devnet", that starts with all forks enabled, with 20 chains, and importantly with coin V5 and fungible V2. This can be used for integration tests without disturbing the nice properties of current devnet for testing PoW.

This PR also cleans up genesis block payload generation, including deleting some special cases for the 20 chain transition and adding a sanity check that payloads which are destined for more than one chain have identical outputs on all of those chains.

WIP pending a fix to the SPV tests.

---
  - https://app.asana.com/0/0/1204139973163258
  - https://app.asana.com/0/0/1204876391903203


edmundnoble:
As-is, on my machine, this branch can run the integration tests on the `fast-development` chainweb version in 300 seconds (5 minutes) without waiting for any upgrades or forks to be enabled.

src/Chainweb/Version/Guards.hs
```diff
@@ -60,14 +60,29 @@ import Chainweb.Transaction
...
+
+atOrAfter :: BlockHeight -> ForkHeight -> Bool
+atOrAfter bh (ForkAtBlock bh') = bh >= bh'
+atOrAfter _ ForkAtGenesis = True
+atOrAfter _ ForkNever = False
+
+before :: BlockHeight -> ForkHeight -> Bool
+before bh (ForkAtBlock bh') = bh < bh'
+before _ ForkAtGenesis = False
+before _ ForkNever = True
```
edmundnoble: I'm really tempted to delete `before` entirely and rename the forks that use it, so that we have less special behavior in here.

tools/ea/Ea.hs
```diff
@@ -101,85 +106,59 @@ main = void $ do
...
-mkPayload :: Genesis -> IO ()
-mkPayload (Genesis v tag cid c k a ns) = do
-    printf ("Generating Genesis Payload for %s on " <> show_ cid <> "...\n") $ show v
-    genPayloadModule v (tag <> sshow cid) txs
+mkPayload :: Genesis -> IO Text
+mkPayload gen@(Genesis v _ cidr@(ChainIdRange l u) c k a ns cc) = do
+    printf ("Generating Genesis Payload for %s on " <> show_ cidr <> "...\n") $ show v
+    payloadModules <- for [l..u] $ \cid ->
+        genPayloadModule v (fullGenesisTag gen) (unsafeChainId cid) =<< mkChainwebTxs txs
+    evaluate $ the payloadModules
```
edmundnoble, line 139: This is a check that, if we want the genesis payload to work on multiple chains, it must have the same result on all of those chains.

tools/ea/Ea/Genesis.hs
```diff
@@ -49,87 +53,51 @@ module Ea.Genesis
...
-    KAD -> "KAD"
+-- A range of chain IDs [l,r]
+data ChainIdRange
+    = ChainIdRange !Word32 !Word32
+    deriving (Eq, Ord)
+
+chainIdRangeTag :: ChainIdRange -> String
+chainIdRangeTag (ChainIdRange l u)
+    | l == u = show l
+    | otherwise = show l <> "to" <> show u
```
chessai, line 78: Is e.g. `show (ChainIdRange 5 7) == "5to7"` intentional? Did you mean that it should be `"5 to 7"`? Not sure that it matters.

tools/ea/Ea/Genesis.hs
```diff
@@ -49,87 +53,51 @@ module Ea.Genesis
...
--- | A 'Lens'' into the transaction allocation filepath
--- of a genesis payload
---
-coinbase :: Lens' Genesis (Maybe FilePath)
-coinbase = lens _coinbase (\t b -> t { _coinbase = b })
+    { _version :: ChainwebVersion
+      -- ^ chainweb version (e.g. Testnet04)
+    , _tag :: Text
+      -- ^ Module name tag
+    , _txChainIds :: ChainIdRange
```
chessai, line 87: What's the meaning of this change? Why is it now a range? And what is the correspondence to the old "GChainId" type; would `Zero :: GChainId` now be `ChainIdRange 0 0`?

tools/ea/Ea/Genesis.hs
```diff
@@ -49,87 +53,51 @@ module Ea.Genesis
...
+    , _keysets :: Maybe FilePath
+      -- ^ filepath to keyset yaml
+    , _allocations :: Maybe FilePath
+      -- ^ filepath to allocation yaml
+    , _namespaces :: Maybe FilePath
+      -- ^ filepath to namespace yaml
+    , _coinContract :: [FilePath]
+    } deriving (Eq, Ord) -- Show)
+
+makeLensesFor [("_" <> fn, fn) | fn <- ["txChainIds", "coinbase", "keysets", "allocations", "namespaces", "coinContract"]] ''Genesis
```
chessai, line 100: Nice. This could one day be a helper that leverages e.g. [reifyRecord](https://hackage.haskell.org/package/th-abstraction-0.5.0.0/docs/Language-Haskell-TH-Datatype.html#v:reifyRecord)

tools/ea/Ea/Genesis.hs
```diff
@@ -287,67 +279,69 @@ mainnet0 :: Genesis
...
 mainnet3 :: Genesis
 mainnet3 = mainnet0
-    & txChainId .~ Three
+    & txChainIds .~ ChainIdRange 3 3
     & allocations .~ (Just mainnetAllocations3)
 
 mainnet4 :: Genesis
 mainnet4 = mainnet0
-    & txChainId .~ Four
+    & txChainIds .~ ChainIdRange 4 4
```
chessai, line 307: why not add a function for these singletons, eg:
```haskell
singleton :: Word32 -> ChainIdRange
singleton x = ChainIdRange x x
```

tools/ea/Ea/Genesis.hs
```diff
@@ -49,87 +53,51 @@ module Ea.Genesis
...
-    Seven -> "7"
-    Eight -> "8"
-    Nine -> "9"
-    N -> "N"
-    KAD -> "KAD"
+-- A range of chain IDs [l,r]
+data ChainIdRange
+    = ChainIdRange !Word32 !Word32
+    deriving (Eq, Ord)
+
```
chessai, line 74: should there be a smart constructor and some other utilities?:
```haskell
mkChainIdRange :: Word32 -> Word32 -> ChainIdRange
mkChainIdRange l r = ChainIdRange (min l r) (max l r)

singleton :: Word32 -> ChainIdRange
singleton x = ChainIdRange x x
```

Also, there are a lot of "magic numbers" below. They could instead be repeatably used constants.

tools/ea/Ea/Genesis.hs
edmundnoble: Yeah this is intentional, it shows up in the files that contain the genesis payloads. `Mainnet10to19.hs` for example.

tools/ea/Ea/Genesis.hs
edmundnoble: Morally this is supposed to say which chain IDs the genesis payload is for, in a way which is easy to print and include in the name of the module that we put the payload in. The old version does that in a very weird way, this is more standard. The thinking behind ranges specifically and not just sets is, in the past upgrade that brought us from 10 to 20 chains, and I think in future chain upscaling, we added multiple homogeneous chains at once that can share the payloads.

tools/ea/Ea/Genesis.hs
edmundnoble: I'd take a `mkChainIdRange` that threw an error on an invalid range. I'll take a look at those constants.

src/Chainweb/Version/FastDevelopment.hs
```diff
@@ -0,0 +1,58 @@
...
+    FastDevelopment = fastDevnet
+
+fastDevnet :: ChainwebVersion
+fastDevnet = ChainwebVersion
+    { _versionCode = ChainwebVersionCode 0x00000002
+    , _versionName = ChainwebVersionName "fast-development"
+    , _versionForks = tabulateHashMap $ \_ -> AllChains ForkAtGenesis
+    , _versionUpgrades = AllChains mempty
+    , _versionGraphs = End twentyChainGraph
+    , _versionBlockRate = BlockRate 30_000_000
```
edmundnoble, line 33: `BlockDelay`?

chessai: What does this comment mean?

tools/ea/Ea/Genesis.hs
```diff
@@ -49,87 +53,51 @@ module Ea.Genesis
...
+    , _coinbase :: Maybe FilePath
+      -- ^ filepath to coinbase yaml
+    , _keysets :: Maybe FilePath
+      -- ^ filepath to keyset yaml
+    , _allocations :: Maybe FilePath
+      -- ^ filepath to allocation yaml
+    , _namespaces :: Maybe FilePath
+      -- ^ filepath to namespace yaml
+    , _coinContract :: [FilePath]
+    } deriving (Eq, Ord) -- Show)
```
edmundnoble, line 98: Delete this commented `Show`?

src/Chainweb/Version/FastDevelopment.hs
edmundnoble: Oh, it's for considering renaming `BlockRate` to `BlockDelay` because that's what it really is.

pact/coin-contract/v2/install-coin-contract-v2.yaml
```diff
@@ -0,0 +1,3 @@
+codeFile: coin-install.pact
+nonce: coin-contract-v2-temp
```
emilypi, line 2: this changes the nonce for v2. Why?

pact/namespaces/ns-install.pact
```diff
@@ -0,0 +1,114 @@
...
+(write-registry "free" GUARD_FAILURE true)
+
+(define-namespace "kadena"
+  (keyset-ref-guard 'ns-operate-keyset)
+  (keyset-ref-guard 'ns-operate-keyset))
+
+(define-namespace "user" GUARD_SUCCESS GUARD_FAILURE)
+(define-namespace "free" GUARD_SUCCESS GUARD_FAILURE)
+;;rotate to real operate keyset
+(define-keyset 'ns-operate-keyset (read-keyset 'ns-operate-keyset))
```
emilypi, line 114: We need to go through the changes to the pact directory here - lots of stuff changed that shouldn't be.

src/Chainweb/BlockHeader.hs
```diff
@@ -424,7 +424,7 @@ slowEpoch (ParentHeader p) (BlockCreationTime ct) = actual > (expected * 5)
   where
     EpochStartTime es = _blockEpochStart p
     v = _chainwebVersion p
-    BlockRate br = _versionBlockRate v
+    BlockDelay br = _versionBlockDelay v
```
emilypi, line 427: `br` should probably change here too

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -573,18 +575,21 @@ parseVersion = constructVersion
...
+        & versionBlockDelay .~ fromMaybe (_versionBlockDelay winningVersion) br
         & versionForks %~ HM.filterWithKey (\fork _ -> fork <= fromMaybe maxBound fub)
         & versionUpgrades .~
             maybe (_versionUpgrades winningVersion) (\fub' ->
                 OnChains $ HM.mapWithKey
                     (\cid _ ->
-                        let fubHeight = winningVersion ^?! versionForks . at fub' . _Just . onChain cid
-                        in HM.filterWithKey (\bh _ -> bh <= fubHeight) (winningVersion ^?! versionUpgrades . onChain cid))
+                        case winningVersion ^?! versionForks . at fub' . _Just . onChain cid of
+                            ForkNever -> error "the fork upper bound never occurs"
```
emilypi, line 589: this is a pretty rough error to debug for a config.

src/Chainweb/Pact/PactService/ExecBlock.hs
```diff
@@ -417,7 +417,7 @@ applyPactCmd isGenesis env miner txTimeLimit cmd = StateT $ \(T2 mcache maybeBlo
     T2 result mcache' <- do
       pd <- getTxContext (publicMetaOf gasLimitedCmd)
       if isGenesis
-      then liftIO $! applyGenesisCmd logger env P.noSPVSupport gasLimitedCmd
+      then liftIO $! applyGenesisCmd logger env P.noSPVSupport pd gasLimitedCmd
```
emilypi, line 420: correcting a historical wrong: `pd` should be `txCtx` or smth because i'ts not `PublicData` or `PublicMeta` anymore

pact/coin-contract/v2/install-coin-contract-v2.yaml
edmundnoble: We already made this change for the new 10 chains' genesis blocks, on mainnet and devnet. We just did it in `Ea.gen20ChainsPayloads`, by munging the YAML programmatically. This change just makes things more uniform, and avoids the YAML munging.

src/Chainweb/Chainweb/Configuration.hs
edmundnoble: How's the new one?

tools/ea/Ea/Genesis.hs
```diff
@@ -253,16 +261,17 @@ testnet0 :: Genesis
 testnet0 = Genesis
     { _version = Testnet04
     , _tag = "Testnet"
-    , _txChainId = Zero
+    , _txChainIds = mkChainIdRange 0 0
```
chessai, line 264:
```suggestion
    , _txChainIds = onlyChainId 0
```

tools/ea/Ea/Genesis.hs
```diff
@@ -47,89 +55,66 @@ module Ea.Genesis
...
-  } deriving (Eq, Ord, Show)
-
--- | A 'Lens'' into the transaction chain id
--- of a genesis payload
---
-txChainId :: Lens' Genesis GChainId
-txChainId = lens _txChainId (\t b -> t { _txChainId = b })
+{-# COMPLETE ChainIdRange #-}
+pattern ChainIdRange :: Word32 -> Word32 -> ChainIdRange
+pattern ChainIdRange l r <- UnsafeChainIdRange l r
```
chessai, line 82: what is the point of this pattern synonym if it doesn't use `mkChainIdRange`?

src/Chainweb/Chainweb.hs
```diff
@@ -388,7 +388,7 @@ withChainwebInternal conf logger peer serviceSock rocksDb pactDbDir backupDir re
             let mcfg = validatingMempoolConfig cid v (_configBlockGasLimit conf) (_configMinGasPrice conf)
             -- NOTE: the gas limit may be set based on block height in future, so this approach may not be valid.
             let maxGasLimit = fromIntegral <$> maxBlockGasLimit v maxBound
-            when (Just (_configBlockGasLimit conf) > maxGasLimit) $
+            unless (maybe True (\limit -> _configBlockGasLimit conf <= limit) maxGasLimit) $
```
chessai, line 391: I get confused reading `unless`, and have a semi-strong preference for `when`. Would it be okay to change this back to `when`, and flip `<=` to `>`?

src/Chainweb/Version/Registry.hs
```diff
@@ -85,9 +86,7 @@ validateVersion v = do
                     , hasAllChains (_genesisTime $ _versionGenesis v)
                     ])]
             ]
-    if null errors
-    then return ()
-    else
+    unless (null errors) $
```
chessai, line 89: Same suggestion as above;
```suggestion
    when (not (null errors)) $
```

tools/ea/Ea/Genesis.hs
edmundnoble: It should be unidirectional, so just for destructuring.

src/Chainweb/Chainweb.hs
edmundnoble: I sympathize, I think this line reads kinda like crap anyway and can be nicer.

src/Chainweb/Version/Registry.hs
edmundnoble: Oof I like this one too much as `unless`.

chessai (approved):

src/Chainweb/Chainweb.hs
jwiegley: FWIW I usually prefer:
```
    forM_ maxGasLimit $ \limit ->
      unless (configBlockGasLimit conf <= limit) $
```

PR #1628, opened 2023-04-12
Trivial Cross-network SPV PoC

edmundnoble:
Cross-network SPVSupport as a new flavor of "target", `crossnet:<some string>`. The SPV proof contains the output with the continuation itself, and a Merkle proof (which is ignored for now, and not verified). What the Merkle proof contains right now isn't very important, later on we have more interesting ideas for it.

---
  - https://app.asana.com/0/0/1204273131102768


src/Chainweb/SPV/CreateProof.hs
```diff
@@ -248,7 +249,7 @@ outputProofPrefix i db payload = do
     -- 1. TX proof
     Just outs <- tableLookup blockOutputTable $ _blockPayloadOutputsHash payload
         -- TODO: use the transaction tree cache
-    let (!subj, pos, t) = bodyTree @_ @ChainwebHashTag outs i
+    let !(!subj, pos, t) = bodyTree @_ @ChainwebHashTag outs i
```
jwiegley, line 252: Why does this outer `!` have any meaning? Since you are already pattern matching the scrutinee against a tuple pattern, then it's already going to be in WHNF.

src/Chainweb/SPV/CreateProof.hs
```diff
@@ -259,6 +260,27 @@ outputProofPrefix i db payload = do
   where
     blockOutputTable = _payloadCacheBlockOutputs $ _payloadCache db
 
+-- TODO: incorporate version into crossnet: syntax?
```
jwiegley, line 263: I think this could be a good idea, but maybe @sirlensalot has a better intuition.

src/Chainweb/SPV/CreateProof.hs
```diff
@@ -259,6 +260,27 @@ outputProofPrefix i db payload = do
...
+-- TODO: if and when this is less trivial, incorporate this logic into createPayloadProof
+createCrossNetworkTransactionOutputProof :: (CanReadablePayloadCas tbl, HasCallStack) => CutDb tbl -> Text -> ChainId -> BlockHeight -> Int -> IO (TransactionOutputProof SHA512t_256)
+createCrossNetworkTransactionOutputProof cutDb tgt scid txHeight txIx = do
+    tip <- maxEntry headerDb
+    txHeader <- maybe (throwM headerNotFound) pure =<< seekAncestor headerDb tip (int txHeight)
+    Just payload <- tableLookup pDb (_blockPayloadHash txHeader)
```
jwiegley, line 269: Can we get better error reporting if the lookup fails here?

src/Chainweb/SPV.hs
```diff
@@ -60,7 +65,7 @@ data SpvException
         { _spvExceptionMsg :: !T.Text
         , _spvExceptionSourceChainId :: !ChainId
         , _spvExceptionSourceHeight :: !BlockHeight
-        , _spvExceptionTargetChainId :: !ChainId
+        , _spvExceptionTargetChainId :: !ProofTarget
```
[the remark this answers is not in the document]
edmundnoble: Agreed assuming you mean `spvExceptionProofTarget`.

src/Chainweb/SPV.hs
```diff
@@ -169,13 +174,13 @@ data TransactionProof a = TransactionProof
...
-    toJSON (TransactionProof cid p) = object $ proofProperties cid p
-    toEncoding (TransactionProof cid p) = pairs . mconcat $ proofProperties cid p
+    toJSON (TransactionProof cid p) = object $ proofProperties (ProofTargetChain cid) p
+    toEncoding (TransactionProof cid p) = pairs . mconcat $ proofProperties (ProofTargetChain cid) p
     {-# INLINE toJSON #-}
     {-# INLINE toEncoding #-}
 
 instance FromJSON (TransactionProof SHA512t_256) where
-    parseJSON = parseProof "TransactionProof" TransactionProof
+    parseJSON = parseProof "TransactionProof" (\tgt p -> do { ProofTargetChain cid <- return tgt; return (TransactionProof cid p) })
```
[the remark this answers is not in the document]
edmundnoble: This will cause that gosh-darn incomplete pattern warning.

src/Chainweb/SPV/CreateProof.hs
edmundnoble: I will add it, I did this because there's identical code elsewhere and a missing payload is a "we're totally screwed, database is corrupt" situation. But I wouldn't mind a better error anyway.

src/Chainweb/SPV/CreateProof.hs
edmundnoble: Here's an example session, in the REPL, which may explain things:
```
> let (!x, y) = undefined
-- nothing happens, hence the tuple has not been evaluated at all.
-- this is a *lazy binding*
> y
*** Exception: Prelude.undefined
-- ^ now the tuple has been evaluated

> let !(!x, y) = undefined
*** Exception: Prelude.undefined
-- ^ this is a strict binding, having a bang on the outermost level is special. 

From an older version of the GHC user guide: 

"a bang at the top level of a let or where binding makes the binding strict, regardless of the 
pattern... the bang at the top of a binding is not part of the pattern; rather it is part of the 
syntax of the binding, creating a 'bang-pattern binding''

> let (!x, y) = (undefined, 1)
-- nothing happens, `x` has not been evaluated because the entire binding is lazy
> y
*** Exception: Prelude.undefined
-- now that the tuple has been matched against with the pattern (!x, y),
-- `x` must be evaluated because of the bang pattern
```

So my reading of this code is: strictly bind `subj`, `pos` and `t` to the components of the tuple returned by `bodyTree`, and evaluate `subj` to WHNF.

The prior version of this code doesn't trigger evaluation of `subj`, `pos`, or `t`, because:
- `subj` is returned in a lazy tuple. Whereas evaluating `pos` or `t` would trigger evaluating `subj`,
- the only use of `pos` and `t` is stuffing it into a lazy tuple in the head of a lazy non-empty list, `proof`. `proof` is already in WHNF, and the bang on `proof` therefore does nothing.

So we need even more bangs, on `pos` and `t`, and for the bang on `proof` to be deleted.

edmundnoble: I'll do this in a separate PR instead and leave this be.

src/Chainweb/SPV.hs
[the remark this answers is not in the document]
jwiegley: Sure, that's even better.

src/Chainweb/SPV.hs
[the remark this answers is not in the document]
jwiegley: But you *are* doing an incomplete pattern match here, hiding it isn't much better... I'd rather have this suggested form, along with a pragma that disables the warning in this one specific place, than circumventing the warnings mechanism by relying on `MonadFail`...

src/Chainweb/SPV/CreateProof.hs
jwiegley: Sounds good. And I forgot about irrefutable matching.

src/Chainweb/SPV.hs
```diff
@@ -185,12 +190,31 @@ proofChainId = to (\(TransactionProof cid _) -> cid)
 
 -- -------------------------------------------------------------------------- --
 -- Output Proofs
+--
+-- | Targets for proofs, usually a chain ID, but in the case of L2, a network
+-- and chain identifier. TODO: determine format, if any.
+data ProofTarget = ProofTargetChain !ChainId | ProofTargetCrossNetwork !Text
+    deriving stock (Show, Eq, Ord, Generic)
+    deriving anyclass NFData
```
larskuhtz, line 198: Let's talk about this first before spending too much time on the implementation. I don't yet fully understand this design and I am not yet convinced that it makes sense. This changes the API surface of SPV and any change we do here we have to support forever.

PR #1631, opened 2023-04-20
[WIP] use base64 over base64-bytestring

chessai:
I added SIMD to base64 [(See PR)](https://github.com/emilypi/base64/pull/50).

Per profiling, chainweb-node spends about 6% of its time performing base64 decode on blocks. According to the microbenchmarks, this should hopefully offer a speedup of ~15x on decode of things that are the average block size.

The other nice thing about the `base64` library is that it has the option of being more type-safe.
All these assertions I've added here are unsafe, but, they were just implicit before.

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -325,8 +325,7 @@ library
         , attoparsec >= 0.13
         , base >= 4.12 && < 5
         , base16-bytestring >= 0.1
-        , base64-bytestring >= 1.0 && < 1.1
-            -- due to change in error message that is inlined in pact
+        , base64 >= 0.5 && < 0.6
         , binary >= 0.8
         , bytestring >= 0.10
         , case-insensitive >= 1.2
@@ -506,8 +505,7 @@ test-suite chainweb-tests
         , aeson-pretty
         , async >= 2.2
         , base >= 4.12 && < 5
-        , base64-bytestring >= 1.0 && < 1.1
-            -- due to change in error message that is inlined in pact
+        , base64 >= 0.5 && < 0.6
         , bytestring >= 0.10
         , case-insensitive >= 1.2
         , chainweb-storage >= 0.1
--- a/src/Chainweb/Mempool/Mempool.hs
+++ b/src/Chainweb/Mempool/Mempool.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -98,9 +99,11 @@ import Crypto.Hash (hash)
 import Crypto.Hash.Algorithms (SHA512t_256)
 
 import Data.Aeson
+import Data.Bifunctor (first)
 import Data.Bits (bit, shiftL, shiftR, (.&.))
 import Data.ByteArray (convert)
-import qualified Data.ByteString.Base64.URL as B64
+import qualified Data.Base64.Types as B64
+import qualified Data.ByteString.Base64.URL as B64U
 import Data.ByteString.Char8 (ByteString)
 import qualified Data.ByteString.Char8 as B
 import qualified Data.ByteString.Short as SB
@@ -703,7 +706,8 @@ mockCodec = Codec mockEncode mockDecode
 
 mockEncode :: MockTx -> ByteString
 mockEncode (MockTx nonce (GasPrice (ParsedDecimal price)) limit meta) =
-  B64.encode $
+  B64.extractBase64 $
+  B64U.encodeBase64' $
   runPutS $ do
     putWord64le $ fromIntegral nonce
     putDecimal price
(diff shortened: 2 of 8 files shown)


jwiegley:
Does this pass replay?

chessai:
> Does this pass replay?

No; I think that we actually do need to keep `base64-bytestring` for certain operations where we care about the on-chain error messages. Right now, a replay fails immediately because of one such instance.

chainweb.cabal
```diff
@@ -325,8 +325,7 @@ library
         , attoparsec >= 0.13
         , base >= 4.12 && < 5
         , base16-bytestring >= 0.1
-        , base64-bytestring >= 1.0 && < 1.1
-            -- due to change in error message that is inlined in pact
```
larskuhtz, line 329: 👍 the upper bound is tracked in the cabal.project file.

chainweb.cabal
```diff
@@ -325,8 +325,7 @@ library
...
+        , base64 >= 0.5 && < 0.6
```
larskuhtz, line 328: Is this known to break with base64 >=0.6? Otherwise please remove, since we don't manually track "soft" upper bounds.

```suggestion
        , base64 >= 0.5
```

src/Chainweb/Mempool/Mempool.hs
```diff
@@ -746,7 +750,7 @@ getDecimal = do
 
 mockDecode :: ByteString -> Either String MockTx
 mockDecode s = do
-    s' <- B64.decode s
+    s' <- first T.unpack $ B64U.decodeBase64 $ B64.assertBase64 @'B64.UrlPadded @_ s
```
larskuhtz, line 753: I am a bit concerned about `B64.UrlPadded`. Does this mean the it uses padding? All use of base64 throughout chainweb should be base64Url without padding.

src/Chainweb/Utils.hs
```diff
@@ -604,15 +606,16 @@ iso8601DateTimeFormat = iso8601DateFormat (Just "%H:%M:%SZ")
 --
 decodeB64Text :: MonadThrow m => T.Text -> m B.ByteString
 decodeB64Text = fromEitherM
-    . first (Base64DecodeException . T.pack)
-    . B64.decode
+    . first Base64DecodeException
+    . B64.decodeBase64
+    . B64.assertBase64 @'B64.StdPadded @_
```
larskuhtz, line 611: Are the functions for the unpadded (and non-url) variants used anywhere in the code? If not we should consider removing them, so that nobody uses them. This is to keep used encodings simple and consistent.

larskuhtz (changes requested):
We should double check that all use of base64 is the URL variant without padding.

Regarding replay: this should be fine since chainweb doesn't put base64 encoded things on-chain. Base64 is used only as textual representation of binary data (mostly in JSON). Merkle hashes of on-chain data are computed on the plain binary representation.

However, I am not 100% certain about the use of base64 within pact service. In theory it shouldn't affect on-chain data (in particular errors in pact service shouldn't be included on-chain), but we should double check that.

chainweb.cabal
larskuhtz, line 328: It seems that version 0.5 of base64 hasn't yet been released. The latest version on Hackage is 0.4.2.4. 0.5 is only available as source repository from Github. If 0.5 is actually required, we would have to ask the maintainer (@emilypi?) to release 0.5 on Hackage or include it as source repository package in cabal.project.

emilypi: It hasn't been released and I"m still working on the PR, which will significantly affect this PR.

emilypi:
I like your enthusiasm @chessai! However, this is jumping the gun alittle bit. The PR that is going in to serve as the basis for `base64-0.5` is going to streamline alot of the functionality changed in this PR that does alot of redundant work (adding pads, removing pads, converting to/from text etc), so I ask that we hold off on it here until I'm confident I can nail down the api. 

That said, with the addition of SIMD support, I do endorse this change, and I would like to revert from `base64-bytestring` at some point, because it is _very_ suboptimal (by design, per Bos + HVR's request). 

> Does this pass replay?

It should @jwiegley, since we're going in at the chainweb level, and the problems with replay w.r.t. Base64 are with the Pact natives wholesale committing to the error format returned by `base64-bytestring`. I'm working on reverting those changes in the latest `base64` major version so that we can switch to that in Pact, but the changes here do not affect Pact, so they shouldn't affect block hashes.

chessai:
> I like your enthusiasm @chessai! However, this is jumping the gun alittle bit. The PR that is going in to serve as the basis for `base64-0.5` is going to streamline alot of the functionality changed in this PR that does alot of redundant work (adding pads, removing pads, converting to/from text etc), so I ask that we hold off on it here until I'm confident I can nail down the api.
> 
> That said, with the addition of SIMD support, I do endorse this change, and I would like to revert from `base64-bytestring` at some point, because it is _very_ suboptimal (by design, per Bos + HVR's request).
> 
> > Does this pass replay?
> 
> It should @jwiegley, since we're going in at the chainweb level, and the problems with replay w.r.t. Base64 are with the Pact natives wholesale committing to the error format returned by `base64-bytestring`. I'm working on reverting those changes in the latest `base64` major version so that we can switch to that in Pact, but the changes here do not affect Pact, so they shouldn't affect block hashes.

Thanks,

Yeah, after talking to you yesterday/digging around in your PR/the codebase more, it seems like this is indeed jumping the gun.
I just saw those speedups and I was really excited.

I'll leave this open for now, because I do think it's something we should return to.

PR #1633, opened 2023-04-20
Error forking in pact

jmcardon:


src/Chainweb/Version.hs
```diff
@@ -1033,6 +1034,13 @@ chainweb218Pact Development = (>= 500)
 chainweb218Pact (FastTimedCPM g) | g == petersonChainGraph = (> 60)
 chainweb218Pact _ = (> 24)
 
+chainweb219Pact :: ChainwebVersion -> BlockHeight -> Bool
+chainweb219Pact Mainnet01 = error "fill" -- Todo: comment times
+chainweb219Pact Testnet04 = error "fill" -- Todo: comment times
+chainweb219Pact Development = (>= 550)
+chainweb219Pact (FastTimedCPM g) | g == petersonChainGraph = (> 70)
```
jwiegley, line 1041: How do we determine this value?

jmcardon: Usually just a couple of blocks after the latest petersen chain fork. Last fork was block 60, this could be 65 or 70. Same with devnet.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -519,7 +529,7 @@ readInitModules logger dbEnv txCtx
       refModsCmd <- liftIO $ mkCmd $ T.intercalate " " $
         [ "coin.MINIMUM_PRECISION"
         , "ns.GUARD_SUCCESS"
-        , "gas-payer-v1.GAS_PAYER"
+        , "(use gas-payer-v1)"
```
jmcardon, line 532: note this could be left as-is and we could instead fork pre 219 and post 219, but this provides a compatible solution which doesn't seem to impact blocks (pending replay verification).

edmundnoble:
Mainnet replay passes on [2441db3](https://github.com/kadena-io/chainweb-node/pull/1633/commits/2441db36e830aeef5a41610dd80f5d090a2d153e).

emilypi (commented):
Do we have regressions for this?

larskuhtz (commented):
Since it just came up in internal discussions: can you please use 4 character indent in chainweb in order to keep this consistent throughout the project?

larskuhtz:
> Since it just came up in internal discussions: can you please use 4 character indent in chainweb in order to keep this consistent throughout the project?

Hmm, it seems that in case of src/Chainweb/Pact/TransactionExec.hs the whole module uses 2 chars. So, maybe we have to just accept that some modules use different indent (although I really think indentation step width should be consistent project wide). This ship may have sailed...

jwiegley:
🚢

src/Chainweb/Pact/TransactionExec.hs
emilypi: I'm in favor of reverting until we can prove that this summons the modules in the correct way. It should, but this is a small change in an orthogonal PR, and it would be better to test on its own as a PR.

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -822,6 +825,73 @@ pact46UpgradeTest = do
...
+    mapErrTx = buildBasicGas 10000
+        $ mkExec' (mconcat
+        [ "(map (+ 1) \"a\")"
+        ])
+    nativeDetailsMsg = "native `=`  Compare alike terms for equality"
+    nativeDetailsTx = buildBasicGas 10000
+        $ mkExec' (mconcat
+        [ "="
+        ])
+    fnDetailsMsg = "(defun coin.transfer:string"
```
emilypi, line 889: for anyone else reviewing, this is not a typo it's just a prefix

src/Chainweb/Pact/TransactionExec.hs
jmcardon: It is not orthogonal. It does not pass replay without this change.

jmcardon: Also, this passed replay.

emilypi: Ah, okay great

test/Chainweb/Test/Pact/PactMultiChainTest.hs
jmcardon: :+1: Yeah. This is a prefix because the actual string is pretty long. Same with the native docs.

emilypi (approved):

jwiegley (approved):

larskuhtz (approved):
I merged this PR into the ongoing ghc-9.6 work and did some (smaller) partial replays with it. Didn't encounter any issues.

PR #1634, opened 2023-04-21
add dockerfile

larskuhtz:
TODO:

* [ ] add (optional, maybe manually triggered?) CI build

---
  - https://app.asana.com/0/0/1204907565841956


larskuhtz:
This docker files isn't used in CI -- docker images are build in CI using a different (much cheaper) method directly from the cabal builds.

At the moment this docker file is mostly useful for local development or external use. It would also replace and simplify the docker file that currently is maintained in the infra repo.

I am planning to setup a CI build for it but haven't done yet because it's low priority and also because we would have to decide how it's triggered and how much resources we want to give to this build. Also, whether it would be required or not.

CI processes are only of value if someone actually looks at the results and they are actively maintained.

edmundnoble (approved):

edmundnoble:
I've already gotten use out of this, it's great.

PR #1637, opened 2023-04-26
Apply ormolu through the code base: just out of curiosity

jwiegley:


src/Chainweb/Backup.hs
```diff
@@ -9,110 +9,109 @@
...
-  { _backupRocksDb :: !RocksDb
-  , _backupDir :: !FilePath
-  , _backupPactDbDir :: !FilePath
-  , _backupChainIds :: !(HashSet ChainId)
-  , _backupLogger :: !logger
+  { _backupRocksDb :: !RocksDb,
+    _backupDir :: !FilePath,
+    _backupPactDbDir :: !FilePath,
+    _backupChainIds :: !(HashSet ChainId),
+    _backupLogger :: !logger
```
emilypi, line 52: Ooof. Does ormolu hate leading commas?

jwiegley: Seems so. Doesn't appear to be too configurable.

chessai:
Can this be closed? I don't think we'll apply ormolu (but we might apply fourmolu, pending discussion)

PR #1638, opened 2023-05-01
nix fixes to "pact with aeson 2"

chessai:


chessai:
Addresses https://github.com/kadena-io/kadena-ethereum-bridge/pull/13

Also, haskell.nix does not ignore the cabal.project.freeze file even with a newer GHC.

larskuhtz:
yeah, cabal doesn't ignore it neither. This break master builds with different ghc on purpose. It forces external users without dev-experience to only build released "blessed" version.

Developers are expected to know that they can just delete the freeze file to do development builds with experimental/dev GHC versions and dependencies.

larskuhtz:
That strategy has been subject of discussions -- because it comes with some inconvenience for chainweb devs.

But it supposedly makes lives easier for non-chainweb-dev users of chainweb, e.g. pact developers or external users and provides them with stable builds -- assuming they use the GHC version that is used for the official release.

I delete the freeze file locally and use the following git aliases to hide the file from git and only unhide if I need to make changes.

```
[aliases]
    hide=update-index --skip-worktree
    unhide=update-index --no-skip-worktree
```

In CI we run one build that uses the officially supported GHC version and the freeze file. All other builds are considered CI/early-integration builds and delete and regenerate the freeze file. The generated freeze file (along with all other build parameters) is recorded with the artifacts of the build to make it reproducible and help with debugging.

chessai:
> yeah, cabal doesn't ignore it neither. This break master builds with different ghc on purpose. It forces external users without dev-experience to only build released "blessed" version.
> 
> Developers are expected to know that they can just delete the freeze file to do development builds with experimental/dev GHC versions and dependencies.

> In CI we run one build that uses the officially supported GHC version and the freeze file. All other builds are considered CI/early-integration builds and delete and regenerate the freeze file. The generated freeze file (along with all other build parameters) is recorded with the artifacts of the build to make it reproducible and help with debugging.

I see. This makes sense.

> I delete the freeze file locally and use the following git aliases to hide the file from git and only unhide if I need to make changes.

I've been deleting the file locally, but all my git work with this has been sneaking about in a manual way, I should automate it.

PR #1641, opened 2023-05-02
help cabal find openssl on mac

larskuhtz:


cabal.project
```diff
@@ -122,8 +124,3 @@ constraints: base64-bytestring <1.1
 -- other pact induced bounds (not relevant for on-chain semantics)
 constraints: megaparsec <9.3
```
larskuhtz, line 126:
```suggestion
-- -------------------------------------------------------------------------- --
-- direct-sqlite 2.3.27

-- TODO remove once the bounds are upgraded in pact.
allow-newer: pact:direct-sqlite
```

jmcardon (approved):

cabal.project
```diff
@@ -2,39 +2,41 @@ packages: chainweb.cabal
...
 debug-info: True
 
+-- -------------------------------------------------------------------------- --
+-- Platform specific locations of external libraries
+
+if os(darwin)
+    if arch(aarch64)
+        package *
+            extra-include-dirs:
+                /opt/homebrew/include
```
jwiegley, line 12: I think this should not be included in this file. Why should there be declarations for paths local to developer machines? Why would it use homebrew paths on M1 but not on Intel machines? These are settings that belong in `cabal.project.local`, not `cabal.project`.

larskuhtz: These settings have been there for a long time and most likely worked for 95% of all users -- including both Mac and linux users. This PR just adds support for M1/M2. Using conditionals eliminates the warning on non-mac systems, which are currently shown.

Including this just safes chainweb developers time to not have to provide support for this to each new dev. 

We can describe in more detail how setup cabal with openssl in the README, but experience tells that most developers (e.g. pact devs who need to build chainweb)  just type `cabal build` and expect things to work and will ask for support if things don't work.

larskuhtz: The <5% of devs for which these settings do not work, would have to provide system specific settings in `cabal.project.local` or would have to ask that we include there setup in `cabal.project`, too -- which I am happy to do if it is a more widely used setup (e.g. macports, windows, etc.)

jwiegley (approved):

PR #1645, opened 2023-05-03
Hide SomeExceptions in PactError

edmundnoble:
Also logs them with level WARN.


src/Chainweb/Pact/Types.hs
```diff
@@ -641,3 +644,16 @@ logError msg = view psLogger >>= \l -> logError_ l msg
 
 logDebug :: String -> PactServiceM tbl ()
 logDebug msg = view psLogger >>= \l -> logDebug_ l msg
+
+data UnknownExnPrinting = PrintsUnknownExns | CensorsUnknownExns
```
sirlensalot, line 648: naming suggestion
```suggestion
data UnexpectedErrorPrinting = PrintsUnexpectedError | CensorsUnexpectedError
```

emilypi: Seconded.

jwiegley: Thirded.

larskuhtz:
Shouldn't this be merged into master first? Or do we change branching policy that release may contain changes that are not on master?

larskuhtz (approved):

emilypi (approved):

edmundnoble:
@larskuhtz I can put it into master before putting it into the RC branch after replay passes (today), the diff just looks better with the PR targeted to the RC branch.

larskuhtz:
@edmundnoble yeah, I think, it should go to master first. I think we should only release changes that have been in master first.

Also we should review changes against master. Otherwise reviews may become less reliable.

edmundnoble:
[a988450](https://github.com/kadena-io/chainweb-node/pull/1645/commits/a98845072132a02134734d50bb79eeab85ce8bae) passes mainnet replay.

PR #1648, opened 2023-05-09
Properly cleanup memomap

larskuhtz:
Cancel sub-tasks when a validation tasks is canceled. Sub-tasks are shared and this change assumes that when one parent task is canceled the sub-tasks would be of no use to any other parent task, too. This is generally true for header validation tasks.


larskuhtz:
> The new bang patterns don't look necessary.

My take on that is "better safe than sorry". I think, I've never been bitten by a redundant bang pattern, but have been bitten many times by missing patterns. 

So, instead of spending a lot of time reasoning about redundancy of any single pattern I just put one where I want thing to be strict. There are just too many subtleties in deciding when a pattern is required. And refactoring a data type at a different place in the code can change whether a bang pattern is redundant or not later on.

chessai:
This looks like it makes good sense.

src/Data/TaskMap.hs
```diff
@@ -80,6 +81,13 @@ size (TaskMap var) = HM.size <$!> readMVar var
 null :: TaskMap k v -> IO Bool
 null (TaskMap var) = HM.null <$!> readMVar var
 
+-- | Clear the content of the map. This cancels all tasks
+--
+clear :: Eq k => Hashable k => TaskMap k v -> IO ()
+clear (TaskMap var) = modifyMVar_ var $ \m -> do
+    mapM_ (\a -> cancel a) m
```
jwiegley, line 88:
```suggestion
    mapM_ cancel m
```

src/Data/TaskMap.hs
```diff
@@ -88,14 +96,14 @@ memo
...
-        Nothing -> do
-            !a <- asyncWithUnmask $ \umask -> umask (task k) `finally` delete tm k
-            m' <- evaluate $ HM.insert k a m
-            return (m', a)
-        (Just !a) -> return (m, a)
-    wait a
+memo tm@(TaskMap var) k task = bracket query cancel wait
+  where
+    query = do
+        -- TODO: should we insert another lookup here? It depends on how optimistic
```
jwiegley, line 102: If this needs to be done, let's add an Asana ticket as well. Otherwise, it will be absorbed into the TODO blackhole...

src/Data/TaskMap.hs
```diff
@@ -88,14 +96,14 @@ memo
...
+        -- we are that the lookup is successful.
+        modifyMVarMasked var $ \m -> case HM.lookup k m of
+            Nothing -> do
+                !a <- asyncWithUnmask $ \umask -> umask (task k) `finally` delete tm k
+                !m' <- evaluate $ HM.insert k a m
```
jwiegley, line 107: Is it meaningful to use a bang pattern with the result of an `evaluate` call?

src/Data/TaskMap.hs
larskuhtz: This TODO as old as this module and survived the introduction and abandonment of previous Asana usages :-)

I understand this more as a code comment for future code maintainers. When they get directed here during profiling, debugging, or refactoring, they may reconsider this. Since this is an open source project, it's not clear whether they'll have access to the currently used Asana instance.

We could changed it into "NOTE TO FUTURE MAINTAINERS", but that's a big lengthy. Maybe just "NOTE"?

src/Data/TaskMap.hs
larskuhtz: It is probably not meaningful, though not 100% sure. After having stumbled several times over strictness annotation semantics -- and also received contradictory advice from people that I consider experts in the field -- I kinda gave up trying to be smart on this. Also for other maintainers, I think, it's better to be explicit than assume that every Haskeller can quickly identify redundant strictness annotations. Instead, I tent to just put the bang when I won't things to be strict, not matter if it's redundant or not.

larskuhtz: For instance, is this line equivalent here to `let !m' = HM.insert k a m`? It seems like it would be the same in this case. But I am completely sure.

src/Data/TaskMap.hs
larskuhtz, line 102:
```suggestion
        -- NOTE: should we insert another lookup here? It depends on how optimistic
```

src/Data/TaskMap.hs
emilypi: For hashmap, keys, values and key arguments are evaluated to WHNF before entry into the map, so the `!` here is to force an already spine-strict structure.

edmundnoble: > Is it meaningful to use a bang pattern with the result of an `evaluate` call?

No, it isn't. 

> For instance, is this line equivalent here to `let !m' = HM.insert k a m`? It seems like it would be the same in this case. But I am completely sure.

In this case yes, but not always. `evaluate` sequences evaluation with respect to execution of IO actions, whereas `!` via `seq` manufactures data dependencies. The difference is that IO actions are executed strictly sequentially and data dependencies form a partial order. For example, a "sequence" of strict bindings one after another in a `do`-block are not ordered with respect to one another unless there are data dependencies between them, so the compiler is free to reorder them with respect to one another. This can't happen with `evaluate`.

larskuhtz: I agree that it is redundant in this case. My argument is that code is about readability. In this case the bang pattern just documents that this value should be in whnf. This is the case independent of whether it is already or if it because of the bang pattern.

Adding redundant bang patterns does't make programs less readable (in my opinion). They actually can make the program easier to read, because the reader sees the intent of the code author without having to study the intrinsics of Haskells evaluation rules.

Our code should be accessible and maintainable not only by the most seasoned Haskell programmers. While we should expect a certain level of expertise from devs, I don't think that we can or should expect instantaneous judgment of redundancy of bang patterns in all cases.

larskuhtz: I propose that discussions about bang patterns in PRs focused on the question whether strictness is desirable or not for a value and not on the question whether they are redundant. I think, the former is what is important the latter is a style question.

src/Data/TaskMap.hs
```diff
@@ -88,14 +96,14 @@ memo
...
-    wait a
+memo tm@(TaskMap var) k task = bracket query cancel wait
+  where
+    query = do
+        -- NOTE: should we insert another lookup here? It depends on how optimistic
+        -- we are that the lookup is successful.
+        modifyMVarMasked var $ \m -> case HM.lookup k m of
+            Nothing -> do
+                !a <- asyncWithUnmask $ \umask -> umask (task k) `finally` delete tm k
+                !m' <- evaluate $ HM.insert k a m
```
larskuhtz, line 107:
```suggestion
                m' <- evaluate $ HM.insert k a m
```

Let's get this PR merged -- it fixes a bug and should not stall and bitrot just because of Haskell code style questions.

edmundnoble (approved):

chessai (approved):

PR #1652, opened 2023-05-12
Prevent tests from cluttering the console

larskuhtz:
Unit tests should not print directly to the console but instead use the means for logging of the test system.

Also test resources are that are not initialized within a particular test, should be considered to be part of the test setup. They are not part of the code under test. Therefore resource components should not report status, except for in exceptional cases, that would usually lead to a test abort.


test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -416,8 +416,8 @@ checkpointerTest name relational cenvIO = testCaseSteps name $ \next -> do
 -- -------------------------------------------------------------------------- --
 -- Read Row Unit Test
 
-readRowUnitTest :: Assertion
-readRowUnitTest = simpleBlockEnvInit runUnitTest
+readRowUnitTest :: (String -> IO()) -> Assertion
```
chessai, line 419:
```suggestion
readRowUnitTest :: (String -> IO ()) -> Assertion
```

chessai, line 419:
```suggestion
readRowUnitTest :: (String -> IO ()) -> Assertion
```

test/Chainweb/Test/MultiNode.hs
```diff
@@ -300,23 +299,28 @@ replayTest loglevel v n = after AllFinish "ConsensusNetwork" $ testCaseSteps nam
     withTempRocksDb "replay-test-rocks" $ \rdb ->
     withSystemTempDirectory "replay-test-pact" $ \pactDbDir -> do
         let tastylog = step . T.unpack
+#if DEBUG_MULTINODE_TEST
```
jwiegley, line 302: Do we need to keep this CPP stuff in?

larskuhtz: It helps to find the right place quickly in case the test needs debugging. That's how it originated. But, it probably hasn't been used in a long time. So, removing it would probably not cause any harm.

edmundnoble (approved):

PR #1653, opened 2023-05-12
Add rewind depth support for execLocal

ak3n:
This PR fixes the behaviour of `execLocal` functionality with a provided rewind depth parameter — it executes the transaction in the context of the block with the provided depth.

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -47,6 +47,7 @@ import qualified Data.Aeson as A
 import Data.Default (def)
 import qualified Data.DList as DL
 import Data.Either
+import Data.Maybe (fromMaybe)
 import Data.Foldable (toList)
 import qualified Data.Map.Strict as M
 import qualified Data.Set as S
@@ -91,7 +92,7 @@ import Chainweb.Payload
 import Chainweb.Payload.PayloadStore
 import Chainweb.Time
 import Chainweb.Transaction
-import Chainweb.TreeDB (lookupM)
+import Chainweb.TreeDB (lookupM, seekAncestor)
 import Chainweb.Utils hiding (check)
 import Chainweb.Version
 import Data.LogMessage
@@ -651,7 +652,15 @@ execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
           -- withCheckpointerRewind as withCurrentCheckpointer
           -- (i.e. setting rewind to 0).
           | otherwise = Just 0
-        rewindHeader = Just $ _tcParentHeader ctx
+
+    let parentBlockHeader = _parentHeader $ _tcParentHeader ctx
+    let ancestorRank = fromIntegral $ _height $ _blockHeight parentBlockHeader - fromMaybe 0 rdepth
+    ancestor <- liftIO $ seekAncestor _psBlockHeaderDb parentBlockHeader ancestorRank
+
+    let rewindHeader
+          | Just a <- ancestor = Just $ ParentHeader a
+          -- when there is no ancestor, use the current parent
+          | otherwise = Just $ _tcParentHeader ctx
 
     let execConfig = P.mkExecutionConfig $
             [ P.FlagAllowReadInLocal | _psAllowReadsInLocal ] ++
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -668,10 +668,10 @@ execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
     let ancestorRank = fromIntegral $ _height $ _blockHeight parentBlockHeader - fromMaybe 0 rdepth
     ancestor <- liftIO $ seekAncestor _psBlockHeaderDb parentBlockHeader ancestorRank
 
-    let rewindHeader
-          | Just a <- ancestor = Just $ ParentHeader a
-          -- when there is no ancestor, use the current parent
-          | otherwise = Just $ _tcParentHeader ctx
+    rewindHeader <- case ancestor of
+        Just a -> pure $ Just $ ParentHeader a
+        Nothing -> throwM $ BlockHeaderLookupFailure $
+            "failed seekAncestor of parent header with ancestorRank " <> sshow ancestorRank
 
     let execConfig = P.mkExecutionConfig $
             [ P.FlagAllowReadInLocal | _psAllowReadsInLocal ] ++
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -140,6 +140,7 @@ runPactService' ver cid chainwebLogger bhDb pdb sqlenv config act =
                     , _psGasModel = getGasModel
                     , _psMinerRewards = rs
                     , _psReorgLimit = fromIntegral $ _pactReorgLimit config
+                    , _psLocalRewindDepthLimit = fromIntegral $ _pactLocalRewindDepthLimit config
                     , _psOnFatalError = defaultOnFatalError (logFunctionText chainwebLogger)
                     , _psVersion = ver
                     , _psValidateHashesOnReplay = _pactRevalidate config
@@ -647,11 +648,14 @@ execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
     spv <- use psSpvSupport
 
     let rewindHeight
-          | Just d <- rdepth = Just d
+          | Just d <- rdepth = d
           -- when no height is defined, treat
           -- withCheckpointerRewind as withCurrentCheckpointer
           -- (i.e. setting rewind to 0).
-          | otherwise = Just 0
+          | otherwise = 0
+
+    when ((_height rewindHeight) > _psLocalRewindDepthLimit) $ do
+        throwM $ LocalRewindLimitExceeded (fromIntegral _psLocalRewindDepthLimit) rewindHeight
 
     let parentBlockHeader = _parentHeader $ _tcParentHeader ctx
     let ancestorRank = fromIntegral $ _height $ _blockHeight parentBlockHeader - fromMaybe 0 rdepth
@@ -670,7 +674,7 @@ execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
         logger = P.newLogger _psLoggers "execLocal"
         initialGas = initialGasOf $ P._cmdPayload cwtx
 
-    withCheckpointerRewind rewindHeight rewindHeader "execLocal" $
+    withCheckpointerRewind (Just rewindHeight) rewindHeader "execLocal" $
       \(PactDbEnv' pdbenv) -> do
         --
         -- if the ?preflight query parameter is set to True, we run the `applyCmd` workflow
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -638,6 +638,8 @@ execLocal
       -- ^ rewind depth (note: this is a *depth*, not an absolute height)
     -> PactServiceM tbl LocalResult
 execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
+    parent <- syncParentHeader "execLocal"
+
     PactServiceEnv{..} <- ask
 
     let !cmd = payloadObj <$> cwtx
@@ -657,7 +659,7 @@ execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
     when ((_height rewindHeight) > _psLocalRewindDepthLimit) $ do
         throwM $ LocalRewindLimitExceeded (fromIntegral _psLocalRewindDepthLimit) rewindHeight
 
-    let parentBlockHeader = _parentHeader $ _tcParentHeader ctx
+    let parentBlockHeader = _parentHeader parent
     let ancestorRank = fromIntegral $ _height $ _blockHeight parentBlockHeader - fromMaybe 0 rdepth
     ancestor <- liftIO $ seekAncestor _psBlockHeaderDb parentBlockHeader ancestorRank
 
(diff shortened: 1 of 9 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -649,7 +650,15 @@ execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
...
-        rewindHeader = Just $ _tcParentHeader ctx
+
+    let parentBlockHeader = _parentHeader $ _tcParentHeader ctx
+    let ancestorRank = fromIntegral $ _height $ _blockHeight parentBlockHeader - fromMaybe 0 rdepth
+    ancestor <- liftIO $ seekAncestor _psBlockHeaderDb parentBlockHeader ancestorRank
+
+    let rewindHeader
+          | Just a <- ancestor = Just $ ParentHeader a
+          -- when there is no ancestor, use the current parent
+          | otherwise = Just $ _tcParentHeader ctx
```
ak3n, line 661: I'm not sure about renaming the "rewind depth" to the "confirmation depth". Because the semantics is that we want to rewind the block, not to confirm something. At least that's what I have from user's POV.

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -233,6 +234,42 @@ chainweb213Test = do
         , "(create-table tbl)"
         ]
 
+pactLocalDepthTest :: PactTestM ()
```
ak3n, line 237: Maybe there should be more testing scenarios? When the depth is too big or negative for example? What do you think?

jwiegley: I think that's a great idea.

src/Chainweb/Pact/PactService.hs
jwiegley: This makes sense to me. Rewind is the action, confirmation is an interpretation.

emilypi: +1 to both of u

test/Chainweb/Test/Pact/PactMultiChainTest.hs
emilypi: @ak3n Let's make sure we have coverage for negative rewind depths, and consult @larskuhtz on what a reasonable maximum rewind depth would be good for the nodes. I would assume we don't want people walking all the way back to genesis.

edmundnoble (approved):
Approved pending selection of a maximum rewind limit. My two cents: 100 is a fine number. Exchanges may have an opinion here.

src/Chainweb/Pact/PactService.hs
larskuhtz: I think, on the external API "confirmation" is what the user wants. Internally, it's rewind.

So, user facing should be "confirmation depth" (it's a common term in Bitcoin). Rewind is an implementation details of the our checkpointer which has no meaning to external users.

test/Chainweb/Test/Pact/PactMultiChainTest.hs
larskuhtz: In particular we have to test interactions with the different rewind limits that we have in place.

Generally, `local` is used via the service API, so we can be generous. We should give the user the ability to configure that, because it's their call how much load they expect and how much they are willing to compromise and latencies for other pact requires.

In practice, I expect that more than a few hundred blocks (also depending on the chain) would probably cause significant slow down. So, we may either put a limit or warn the user if they configure a larger number.

larskuhtz: This is easy to test, because we log execution time traces for all pact calls. So we can run a node with a large limit and just plot latencies pact request latencies (also of other pact requests) over the requested confirmation depth.

larskuhtz (commented):
Looks good to me. I think, what is missing is 

*  some more tests for larger rewind depth in order to test possible interactions with existing rewind limits,

*  addition of a configuration option for setting the maximum allowed local rewind/confirmation depth ("confirmation depth" might be the better  term here since it is user facing), and

* benchmarking (possibly manual, e.g. by deploying to testnet or mainnet node) to determine a reasonable default limit.

If testing large depth is not feasible in unit tests, it could possibly be done on devnet.

src/Chainweb/Pact/PactService.hs
ak3n: I've got some doubts about the "rewind depth" because it clashes with the checkpointer's rewind mechanism. We will get two different rewind semantics — the local command's rewind of blocks and checkpointer's rewind (the first uses the latter's).

We could say that the checkpointer stuff is an internal thing and users shouldn't know about it, but it's already leaked as an error message, see `PactException(RewindLimitExceeded)`. Also there is `reorgLimit` in `execValidateBlock`, which is a limit for checkpointer's rewind but with additional semantic.

Do we want our users to know about the checkpointer and how much? If not, then there will be two rewinds — 
"rewind depth" as an external API thing and "checkpointer rewinds" as an internal meaning, which we can try to hide. If yes, then we should be more clear and accurate with the terms to avoid ambiguity (still not a fan of "confirmation depth" in this context).

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -233,6 +234,52 @@ chainweb213Test = do
...
+
+  -- the negative depth turns into 18446744073709551611 and we expect the `LocalRewindLimitExceeded` exception
+  -- since `BlockHeight` is a wrapper around `Word64`
+  handle (\(LocalRewindLimitExceeded _ _) -> return ()) $ do
+    runLocalWithDepth (Just $ BlockHeight (-5)) cid getSender00Balance >>= \_ ->
+      liftIO $ assertFailure "block succeeded"
+
+  -- depth that reaches the genesis
+  runLocalWithDepth (Just $ BlockHeight 99) cid getSender00Balance >>= \r ->
+    checkLocalResult r $ assertTxSuccess "Should get the balance at the genesis block" (pDecimal 100000000)
```
ak3n, line 270: Two tests — with a negative depth, which tests the limit, and a depth reaching the genesis block.

ak3n:
Some notes:

1. For some reason macos build fails with a dns error

```
dieVerbatim: user error (Error: cabal: '/usr/local/opt/curl/bin/curl' exited with an error:
curl: (6) Could not resolve host: objects-us-east-1.dream.io
```

2. I've discovered a bug with the checkpointer, right now `local` command changes the checkpointer's state and we should bracket it, rewind it back after running the transaction.

src/Chainweb/Pact/PactService.hs
larskuhtz: I think, `confirmation depth` is a notion only at service REST API level.  "rewind depth" is a concept at checkpoint API level.

The idea is that in the future all service API endpoints that retrieve information from the blockchain would have a common "confirmation depth" parameter. This would include latest cut, querying headers, querying payloads, local calls, and possibly more.

"Rewind" wouldn't be a meaningful notion at the service API level, because it is relative to the checkpoint state which is hidden from the service API user. It also would not be meaningful with respect to querying cuts, headers, or payloads. Instead what service API users care about is the "level of confirmation" of the retrieved data, i.e. the depth in the chain with respect to the current (local) state of consensus.

larskuhtz: I wonder if we need to define more precisely what the notion of "current block" means within the context of the checkpointer. Is it just the current state of the checkpoint, or is the the latest state of consensus (which is external to the checkpointer). Clearly both are related, but the respective invariants are currently not very well documented or reflected in the code.

src/Chainweb/Pact/Types.hs
```diff
@@ -392,6 +399,9 @@ instance HasChainId (PactServiceEnv c) where
 defaultReorgLimit :: Word64
 defaultReorgLimit = 480
 
+defaultLocalRewindDepthLimit :: Word64
+defaultLocalRewindDepthLimit = 100
```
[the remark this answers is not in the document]
ak3n: I've raised the `defaultLocalRewindDepthLimit` to `1000`, the usual response time with `1000` is about 5 seconds.

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -266,12 +266,21 @@ pactLocalDepthTest = do
...
 
+  -- depth that goes after the genesis block should trigger the `LocalRewindLimitExceeded` exception
+  handle (\case
+      { LocalRewindGenesisExceeded -> return ()
+      ; err -> liftIO $ assertFailure $ "Expected LocalRewindGenesisExceeded, but got " ++ show err}
+      ) $ do
+    runLocalWithDepth (Just $ BlockHeight 56) cid getSender00Balance >>= \r -> do
+      liftIO $ print r
+      liftIO $ assertFailure "Expected LocalRewindGenesisExceeded, but block succeeded"
+
```
emilypi, line 283: In terms of style points here, I think we can agree to put a nix on explicit braces and choose good formatting instead. Suggesting: 

```haskell
  handle 
    (\case
        LocalRewindGenesisExceeded -> return ()
        err -> liftIO $ assertFailure $ "Expected LocalRewindGenesisExceeded, but got " ++ show err)
    (liftIO $ do
      r <- runLocalWithDepth (Just $ BlockHeight 56) cid getSender00Balance
      print r
      assertFailure "Expected LocalRewindGenesisExceeded, but block succeeded")

Comment (jwiegley) [reply test/Chainweb/Test/Pact/PactMultiChainTest.hs:283]: Yes please to no explicit braces.

Comment (ak3n) [reply test/Chainweb/Test/Pact/PactMultiChainTest.hs]: Fixed, thanks!

**Comment by ak3n** [comment]
Ok, looks like I have fixed everything. The last and the only thing is how to properly name it. And if it's even possible at this moment? If it's already a part of API as `rewindDepth`?

**Comment by larskuhtz** [comment]
> Ok, looks like I have fixed everything. The last and the only thing is how to properly name it. And if it's even possible at this moment? If it's already a part of API as `rewindDepth`?

Is the depth parameter currently exposed in the API? If not we may want to add it (otherwise the feature wouldn't be available to users). Otherwise, would we have to provide backward compatibility? I think, we never announced it, so we should be fine to break it if we want.

File: src/Chainweb/Pact/Service/Types.hs [hunk 6]
```diff
@@ -168,6 +170,10 @@ data PactException
   | BuyGasFailure !GasPurchaseFailure
   | MempoolFillFailure !Text
   | BlockGasLimitExceeded !Gas
+  | LocalRewindLimitExceeded
+    { _localRewindExceededLimit :: !Natural
+    , _localRewindRequestedDepth :: !BlockHeight }
```
edmundnoble, line 175: Nit, I think either both of these should be `BlockHeight` or both of these should be `Natural`, because they're both the same "type of thing" (block height delta)

src/Chainweb/Pact/Types.hs
```diff
@@ -354,7 +356,10 @@ data PactServiceEnv tbl = PactServiceEnv
     , _psBlockHeaderDb :: !BlockHeaderDb
     , _psGasModel :: TxContext -> GasModel
     , _psMinerRewards :: !MinerRewards
+    , _psLocalRewindDepthLimit :: {-# UNPACK #-} !Word64
+    -- ^ The limit of rewind's depth in the `execLocal` command.
```
edmundnoble, line 360: There's no need to `UNPACK` this or any other word-sized field anymore, `-funbox-small-strict-fields` is the default.

src/Chainweb/Pact/PactService.hs
```diff
@@ -646,12 +650,28 @@ execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
...
+    -- we fail if the requested depth is bigger than the current parent block height
+    -- because we can't go after the genesis block
+    when (fromMaybe 0 rdepth > _blockHeight parentBlockHeader) $ throwM LocalRewindGenesisExceeded
+
+    let ancestorRank = fromIntegral $ _height $ _blockHeight parentBlockHeader - fromMaybe 0 rdepth
+    ancestor <- liftIO $ seekAncestor _psBlockHeaderDb parentBlockHeader ancestorRank
+
+    let rewindHeader
+          | Just a <- ancestor = Just $ ParentHeader a
+          -- when there is no ancestor, use the current parent
```
edmundnoble, line 673: Maybe we want to throw or log an error here if `seekAncestor` returns `Nothing`, I don't see any legal situation where that can happen.

src/Chainweb/Pact/Service/Types.hs
ak3n: I'm being consistent here with `RewindLimitExceeded`, where the limit is `Natural` and heights are `BlockHeight`. 

And semantics is a bit different — limit says how far in the past we can go, while the height starts from the bottom. Different directions.

src/Chainweb/Pact/Types.hs
ak3n: Good point 👍 , added because of `_psReorgLimit`, will remove for both of them.

src/Chainweb/Pact/PactService.hs
ak3n: Yeah, that behavior is caught by throwing `LocalRewindGenesisExceeded` above, maybe should replace the check with just throwing an exception here.

ak3n: Fixed by throwing the `BlockHeaderLookupFailure` exception because it's indeed an exceptional situation and shouldn't happen.

src/Chainweb/Pact/Service/Types.hs
edmundnoble: Aren't they both measured from the tip of the chain?

ak3n: The rewind's limit starts from the current moment, how far we can go backwards.

edmundnoble (approved):

chessai (approved):

PR #1658, opened 2023-05-20
Temporarily revert "Force less (#1617)"

larskuhtz:
This reverts commit c39ad55cb41794d8d9b13c8e83ffe6159456bf93.


sirlensalot (approved):
Based on @larskuhtz assertion that this is a minimal diff from 2.19, and given the nature of the changes, approving.

larskuhtz:
There are several changes in the original PR that we want to keep. However, for now it is safe to revert the full PR and redo it later.

@edmundnoble let's do a follow up PR where we revisit this PR and pick the changes that we want to keep. We may also split it into two PRs, one that is concerned with strictness and another that introduced safe-exceptions everywhere.

PR #1655, opened 2023-05-18
Update to pact 4.7.1

rsoeldner:
This PR is now superseded by #1661 

# TODO
don't merge this until
 -  [x] pact-4.7.1 release is merged 
 -  [ ] the pointers to pact have been updated in this PR.

---

- Bumps pact version to 4.7.1
- Removes runtime return value typechecking tests (typechecking is moved under a standalone flag (not in the `DisablePact**` family) in Pact so that chainweb may manually disable the checks.

PR #1660, opened 2023-05-20
Pact 4.7.1 + runtime RTC change + 2.19.1 redo

sirlensalot:
This PR merges onto the new `point-release/2.19` branch which was forked off the 2.19 tag. 

It also redoes #1655 instead of cherry-picks it, as it was applying the flag too often (coinbase, upgrades, genesis) and not enough (local, pre-insert check).

Coverage is provided for testing that RTC is defeated pre- and post-fork, and in `local`. However pre-insert does not have good coverage at this time. I will poke around but if we're in a rush we should just ship this and test insert manually somehow.


jwiegley:
@sirlensalot There is definitely some urgency here. We're now 12 days from the service date, and wanted to make this release available, ideally, 14 days prior to that cut-off. Since we'll all be at the offsite this next week, let's now take the most efficient route to a delivery now, and defer what work we can until the next cycle. Thank you!

sirlensalot:
Closing, will submit new PR on master

PR #1661, opened 2023-05-20
Pact 4.7.1, disable runtime TC of user return values

larskuhtz:
Upgrade to Pact 4.7.1 and instrument disable of user-function return value runtime typechecking.

Coverage is provided for testing that RTC is defeated pre- and post-fork, and in local. However pre-insert does not have good coverage at this time.


larskuhtz:
Devenet regression test runs fail with the following failure. I assume that is expected and the test-suite needs to be adapted. @sirlensalot is that assumption correct?

```
return value type error
      ✖ return value type error
              );
              t.equal(underTest, expected, description);
      ----------^
              return reqKey["requestKeys"][0];
            } else if (reqKey.requestKeys.length > 1) {

        found: null
        wanted: "Type error: expected string, found integer"
        compare: ===
        at:
          line: 107
          column: 11
          file: util/func.js
          type: Test
        stack: |
          Test.<anonymous> (util/func.js:107:11)
```

sirlensalot (approved):
This is my PR rebased by @larskuhtz so somebody else (@emilypi, @jmcardon ) should review as well

dep/pact/github.json
```diff
@@ -3,6 +3,6 @@
   "repo": "pact",
   "branch": "master",
   "private": false,
-  "rev": "83c5944991d6edcd34d79f9fbf8e537d060689c6",
-  "sha256": "0l59xi2by6l6gi10r8c437m7ics29215zr0zl1syyr3039vgmv0x"
+  "rev": "456aa65ef45596458488ed59e7167621b40e6139",
+  "sha256": "0000000000000000000000000000000000000000000000000000"
```
emilypi, line 7: Followed up in #1664

emilypi (approved):

PR #1665, opened 2023-05-23
Add use-freeze-file to artifact archive path

edmundnoble:
I'm not entirely sure that this is necessary, but it looks like without it the two artifacts may clobber one another on S3, preventing us from (safely) releasing the use-freeze-file version.


larskuhtz:
This setting is already contained in the file name. So it should not be needed in the directory name.

larskuhtz (approved):

jwiegley (approved):

PR #1664, opened 2023-05-20
update flake.lock and pact sha

larskuhtz:

PR #1667, opened 2023-05-31
Fix nix builds by ignoring freeze & bumping hackage

enobayram:
This PR is an alternative to #1664. The main difference is that we bump the `hackage` snapshot by defining it as a direct input of our flake and telling `haskellNix` to `follow` it. The we can just `nix flake lock --update-input hackage` to bump the snapshot without actually bumping `haskell.nix`. I think this has 2 main advantages:
* Bumping `haskell.nix` may introduce unrelated changes to the build infrastructure. Changes one might not be willing to deal with just to bump the hackage snapshot.
* Bumping `haskell.nix` triggers a deeper transitive dependency rebuild session compared to just bumping `hackage`, which only requires a rebuild of the project's Haskell package dependencies. This can mean a 10-hour rebuild vs. a 2-hour rebuild.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -62,7 +62,7 @@ source-repository-package
     type: git
     location: https://github.com/kadena-io/pact.git
     tag: 456aa65ef45596458488ed59e7167621b40e6139
-    --sha256: 0l59xi2by6l6gi10r8c437m7ics29215zr0zl1syyr3039vgmv0x
+    --sha256: sha256-J2Rh2EXczT4QzOM/agOzxq1UPdU6aNPDo3fnfeoVRbA=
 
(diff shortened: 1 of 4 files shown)


larskuhtz (approved):

mbwmbw1337:
Is the intent of this to fix #1668 and #1669? I tried pulling this branch and nix-building with the same errors as detailed in #1668. It very well may be my (operator) error, but I wanted to advise in case you have yet to try.

enobayram:
@mbwmbw1337 Thanks for the heads up, I've responded under #1668. This Nix-only change should be unrelated to #1669, though.

PR #1670, opened 2023-06-02
Replace legacy default.nix with flake based default.nix

enobayram:
This PR removes the legacy Nix infrastructure for building the project. The legacy infra was based on `haskell4nix` from `nixpkgs`, but our recently used `flake.nix` uses `haskell.nix` and removes most of the Nix maintenance burden. The legacy infra has been broken since the 2.19.1 release, so it's been causing trouble for users that tried to build the project with the legacy `nix-build` command, instead of `nix build`.

This PR also removes `docker.nix` that relied on the legacy infra. We'll reintroduce Nix-built docker images in a subsequent PR using the new Nix infra.


zeus.nix
```diff
@@ -1 +0,0 @@
-(import ./. { runTests = false; })
```
chessai, line 1: Do we want to remove zeus? is it used?

larskuhtz (approved):

zeus.nix
enobayram: It used to be used by Zeus as the entry point, but Zeus can't build chainweb-node with the new flake + haskell.nix infrastructure due to it having an old Nix version. In any case, we now have a github action for building and caching the results, so Zeus is obsoleted as far as chainweb-node is concerned.

jwiegley:
I've found that `nix build -f .` on both M1 and Intel Mac now segfaults after merging this PR.

PR #1671, opened 2023-06-02
Adjust catchup step size and far ahead threshold

larskuhtz:


jwiegley:
Hi @larskuhtz, can you say more to explain the reason for this change, and why these are now the right numbers to use?

enobayram:
@larskuhtz I've tried deploying this version to a personal node with an old database, but it failed to make any progress during the 20 minutes I've watched it.

chessai:
> @larskuhtz I've tried deploying this version to a personal node with an old database, but it failed to make any progress during the 20 minutes I've watched it.

Lars and I were testing this with my node, and it failed to make progress for 3 hours, unfortunately. So this is still a WIP.

chessai:
Catchup is no longer stalled. However, it seems that it's about 0.85 cuts/s slower.

chessai (approved):

PR #1673, opened 2023-06-08
Use readonly rocksdb when syncing pact

edmundnoble:
To do parallel replays we're also going to need to be able to use different Pact DBs while using the same rocksdb, which is not currently possible as far as I can tell. Running a replay on this once it's built.


larskuhtz (approved):

jwiegley (approved):

node/ChainwebNode.hs
```diff
@@ -338,22 +338,29 @@ node conf logger = do
...
-                        , runRtsMonitor (_chainwebLogger cw)
-                        , runBlockUpdateMonitor (_chainwebLogger cw) (_cutResCutDb $ _chainwebCutResources cw)
-                        , runDatabaseMonitor (_chainwebLogger cw) rocksDbDir pactDbDir
-                        ]
+    withRocksDb' <-
+        if _configOnlySyncPact cwConf
+        then
+            if _cutPruneChainDatabase (_configCuts cwConf) == GcNone
+            then withReadOnlyRocksDb <$ logFunctionText logger Info "Opening RocksDB in read-only mode"
+            else withRocksDb <$ logFunctionText logger Info "Opening RocksDB in read-write mode, if this wasn't intended, ensure that cuts.pruneChainDatabase is set to none"
```
larskuhtz, line 346: For a follow-up PR:

I think we should just move the existing pruning functionality into `cwtool`.

Merkle-tree validation should be inlined with pact replay. This is the default use case for endusers who need to initialize a node from a untrusted rocksdb. We should carefully check the code whether additional checks would have to be added to replay. For payload the current behavior (with payload hash validation enabled) should be sufficient. So, we would have to add header validation (including braiding validation), which is generally fast and should create only minimal overhead.

For other use cases the existing functionality could be used via `cwtool`.

larskuhtz (approved):
@edmundnoble I added a comment regarding a follow up PR.

chessai (approved):

edmundnoble:
https://github.com/kadena-io/integration-tests/actions/runs/5285316952/jobs/9563716306 shows a replay succeeding in read-write mode with `pruneChainDatabase: headers-checked`, 
https://github.com/kadena-io/integration-tests/actions/runs/5285313688/jobs/9563716735 shows a replay succeeding in read-only mode with `pruneChainDatabase: none`.

PR #1675, opened 2023-06-09
release 2.19.2

chessai:
- [ ] Fix Linux Builds
- [ ] Fix MacOS Builds
- [ ] Update version in cabal file


chessai:
I think the PR base should be `release-candidate/2.19.1`?

larskuhtz:
In think, this should target master. Otherwise we are going to repeat the git issues that we had during the 2.19.1 release.

larskuhtz:
Otherwise, if this is strictly just a build fix of 2.19.1, maybe we should call it 2.19.1.1?

jwiegley:
`nix build` worked on my M1 Mac.

enobayram:
Should we base this PR on top of the `2.19.1` tag instead of the `release-candidate/2.19.1/` branch in order to keep the git history simpler and to make sure that we're not pushing a `2.19.2` release that's not different from `2.19.1` in subtle and unexpected ways?

enobayram:
I actually support @larskuhtz's suggestion of releasing from `master` since it's the cleanest option history-wise, but that requires more extensive testing, right?

jwiegley:
Note that while `nix build` works for me, `nix build -f .` no longer does. It dies with a segmentation fault, on both Intel and M1 Mac. This worked not long ago.

chessai:
> Note that while `nix build` works for me, `nix build -f .` no longer does. It dies with a segmentation fault, on both Intel and M1 Mac. This worked not long ago.

@jwiegley  and I debugged this last week and the solution was pretty simple: default.nix became a function with default args, so he needed to do eg `import ./default.nix {}` instead of `import ./default.nix`. Also, default.nix before returned a single derivation, but now it returns an attribute set with the expected derivation as one of its attributes.

PR #1676, opened 2023-06-10
update hackage.nix for merkle-log 0.2

chessai:


larskuhtz (approved):

enobayram:
@chessai I've noticed that this PR [pins hackage.nix directly in the flake inputs]{https://github.com/kadena-io/chainweb-node/pull/1676/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R8} instead of leaving it unspecified and letting it get pinned in flake.lock as we used to do. That means we can't update it anymore by running `nix flake lock --update-input hackage` anymore. Is there any reason we're being extra explicit about this particular revision of hackage.nix?

chessai:
> @chessai I've noticed that this PR [pins hackage.nix directly in the flake inputs]{https://github.com/kadena-io/chainweb-node/pull/1676/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R8} instead of leaving it unspecified and letting it get pinned in flake.lock as we used to do. That means we can't update it anymore by running `nix flake lock --update-input hackage` anymore. Is there any reason we're being extra explicit about this particular revision of hackage.nix?

I mean to undo this. This was just a preliminary test.

PR #1677, opened 2023-06-12
Introduce `Depth` and `Limit` newtypes for Pact Service

ak3n:
This PR adds two newtypes (wrappers around `Word64`) to provide better semantics in the pact service:

- `Depth` — sometimes the operation goes backwards and uses `BlockHeight` which is semantically ambiguous, `Depth` should help with it.
- `Limit` — representation for a restriction value.


src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -654,3 +640,18 @@ heightProgress initialHeight ref logFun = forever $ do
...
+    => BlockHeader
+    -> Maybe Limit
+    -> BlockHeader
+    -> PactServiceM tbl ()
+failOnTooLowRequestedHeight parent (Just limit) lastHeader
+    | parentHeight + 1 + (BlockHeight $ _limit limit) < lastHeight = -- need to stick with addition because Word64
+        throwM $ RewindLimitExceeded limit parentHeight lastHeight parent
+  where
+    parentHeight = _blockHeight parent
+    lastHeight = _blockHeight lastHeader
```
emilypi, line 656: Less logic in the guards is :ok_hand: 
```suggestion
failOnTooLowRequestedHeight parent (Just limit) lastHeader
    | parentHeight + 1 + limitHeight < lastHeight = -- need to stick with addition because Word64
        throwM $ RewindLimitExceeded limit parentHeight lastHeight parent
  where
    limitHeight = BlockHeight $ _limit Limit
    parentHeight = _blockHeight parent
    lastHeight = _blockHeight lastHeader
```

src/Chainweb/Pact/Service/Types.hs
```diff
@@ -58,13 +59,25 @@ import Chainweb.Transaction
...
+newtype Limit = Limit { _limit :: Word64 }
+  deriving (Eq, Ord)
+  deriving newtype (FromJSON, ToJSON, Num, Integral, Real, Enum, Bounded)
+instance Show Limit where show (Limit l) = show l
+
+-- | Value that represents how far to go backwards.
+newtype Depth = Depth { _depth :: Word64 }
+  deriving (Eq, Ord)
+  deriving newtype (FromJSON, ToJSON, Num, Integral, Real, Enum, Bounded)
+instance Show Depth where show (Depth d) = show d
```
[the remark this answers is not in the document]
jwiegley: Very much agree with Emily that we should avoid instances we don't need.

src/Chainweb/Pact/PactService/Checkpointer.hs
ak3n: Thanks! 👍

src/Chainweb/Pact/Service/Types.hs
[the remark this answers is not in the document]
ak3n: 1. `Num` is used in some places, but I agree, wrapper should make code more explicit. Dropped the instances and adapted the code.

2. Agree as well, I thought they can be reused somewhere else (being as general wrappers) but it's better to define types for separate situations. Renamed to `RewindDepth` and `RewindLimit`.

Thanks!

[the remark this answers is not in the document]
edmundnoble: Agreed that `Num` is a risky one, though it's hard to tell sometimes. We may push more on #1624 which could make `Num` instances less of an issue. 

I actually like the name change quite a bit, `RewindDepth` and `RewindLimit`.

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -105,6 +105,7 @@ import qualified Chainweb.Mempool.Mempool as Mempool
 import Chainweb.Mempool.P2pConfig
 import Chainweb.Miner.Config
 import Chainweb.Pact.Types (defaultReorgLimit, defaultModuleCacheLimit, defaultLocalRewindDepthLimit)
+import Chainweb.Pact.Service.Types (RewindLimit(RewindLimit))
```
edmundnoble, line 108: Seems oddly specific, maybe this is a tooling thing? Ordinarily I'd write `RewindLimit(..)`.

src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -432,8 +433,8 @@ defaultChainwebConfiguration v = ChainwebConfiguration
     , _configLogGas = False
     , _configMinGasPrice = 1e-8
     , _configPactQueueSize = 2000
-    , _configReorgLimit = int defaultReorgLimit
-    , _configLocalRewindDepthLimit = int defaultLocalRewindDepthLimit
+    , _configReorgLimit = RewindLimit defaultReorgLimit
+    , _configLocalRewindDepthLimit = RewindLimit defaultLocalRewindDepthLimit
```
edmundnoble, line 437: Perhaps make `defaultReorgLimit` and `defaultLocalRewindDepthLimit` have type `RewindLimit` themselves?

tools/cwtool/TxSimulator.hs
```diff
@@ -38,7 +38,7 @@ import Chainweb.Pact.Backend.Utils
 import Chainweb.Pact.PactService
 import Chainweb.Pact.PactService.ExecBlock
 import Chainweb.Pact.RestAPI.Server
-import Chainweb.Pact.Service.Types
+import Chainweb.Pact.Service.Types as Pact
```
edmundnoble, line 41: Is this needed now?

edmundnoble (approved):

src/Chainweb/Chainweb/Configuration.hs
ak3n: Nice catch, thanks!

emilypi (approved):

chessai (approved):

PR #1679, opened 2023-06-13
Replace Pact Service Module Cache with Checkpointer Cache

ak3n:
This is a work in progress PR exploring the direction of replacing Pact Service's module cache with Checkpointer's cache.

### DIFF
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -459,7 +459,7 @@ test-suite chainweb-tests
         Chainweb.Test.P2P.Peer.BootstrapConfig
         Chainweb.Test.Pact.DbCacheTest
         Chainweb.Test.Pact.Checkpointer
-        Chainweb.Test.Pact.ModuleCacheOnRestart
+--        Chainweb.Test.Pact.ModuleCacheOnRestart
         Chainweb.Test.Pact.NoCoinbase
         Chainweb.Test.Pact.RewardsTest
         Chainweb.Test.Pact.PactExec
--- a/chainweb.cabal
+++ b/chainweb.cabal
@@ -108,7 +108,7 @@ custom-setup
 -- -------------------------------------------------------------------------- --
 
 library
-    import: warning-flags, debugging-flags
+    import: debugging-flags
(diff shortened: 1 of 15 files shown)


chainweb.cabal
```diff
@@ -425,7 +425,7 @@ library
 -- -------------------------------------------------------------------------- --
 
 test-suite chainweb-tests
-    import: warning-flags, debugging-flags
+    import: debugging-flags
```
chessai, line 428: is this warning removal just for local dev? If so, I believe you can just add a cabal.project.local with the appropriate settings.

src/Chainweb/Pact/Backend/DbCache.hs
```diff
@@ -193,12 +230,60 @@ cacheStats mc = object
...
+keysDbCache :: DbCache PersistModuleData -> [CacheAddress]
+keysDbCache DbCache{_dcStore} = HM.keys _dcStore
+
 -- -------------------------------------------------------------------------- --
 -- Internal
 
 mkAddress :: Utf8 -> ByteString -> CacheAddress
 mkAddress (Utf8 key) rowdata = CacheAddress . BS.toShort $
-    key <> ":" <> BA.convert (C.hash @_ @SHA512t_256 rowdata)
+    "hashed:" <> key <> ":" <> BA.convert (C.hash @_ @SHA512t_256 rowdata)
```
chessai, line 286: why this change?

chainweb.cabal
ak3n: Yep, that's what I do for local development, too lazy to explore `cabal.project.local`

src/Chainweb/Pact/Backend/DbCache.hs
ak3n: We have to filter out these keys as they consist of non-unicode hashes while `ModuleName` type is a `Text`, and they are very specific.

ak3n:
Closing this PR as it we've figured out a better solution that will follow after shallow nodes support.

PR #1681, opened 2023-06-13
don't recommend cabal install

chessai:


chessai:
Wrong base branch. I am messing up with git so much today.

PR #1685, opened 2023-06-14
Delete validateHashesOnReplay option and make it the default

edmundnoble:
---
  - https://app.asana.com/0/0/1204808195351188

### DIFF
--- a/src/Chainweb/Pact/PactService/Checkpointer.hs
+++ b/src/Chainweb/Pact/PactService/Checkpointer.hs
@@ -320,8 +320,6 @@ withDiscardedBatch act = do
         (liftIO $ _cpDiscardCheckpointerBatch cp)
         act
 
-data ShouldValidateHashesOnFastForward = ValidateHashesOnFastForward | DoNotValidateHashesOnFastForward
-
 -- | INTERNAL FUNCTION. USE 'withCheckpointer' instead.
 --
 -- TODO: The performance overhead is relatively low if there is no fork. We
@@ -416,6 +414,8 @@ rewindTo rewindLimit (Just (ParentHeader parent)) = do
                           & S.length_
             logInfo $ "rewindTo.playFork: replayed " <> sshow c <> " blocks"
 
+data ShouldValidateHashesOnFastForward = ValidateHashesOnFastForward | DoNotValidateHashesOnFastForward
+
 -- | INTERNAL UTILITY FUNCTION. DON'T EXPORT FROM THIS MODULE.
 --
 -- Fast forward a block within a 'rewindTo' loop.
@@ -439,6 +439,8 @@ fastForward shouldValidateHashes (target, block) =
                 <> ". Block: "<> encodeToText (ObjectEncoded block)
             Just x -> return $ payloadWithOutputsToPayloadData x
         pwo <- execBlock block payload pdbenv
+        -- As an optimization, we only validate the payload hash of a block during a
+        -- Pact replay, and not during normal operation.
         case shouldValidateHashes of
             ValidateHashesOnFastForward -> either throwM (\_ -> return ()) pwo
             DoNotValidateHashesOnFastForward -> return ()
--- a/src/Chainweb/Pact/PactService/Checkpointer.hs
+++ b/src/Chainweb/Pact/PactService/Checkpointer.hs
@@ -320,6 +320,7 @@ withDiscardedBatch act = do
         (liftIO $ _cpDiscardCheckpointerBatch cp)
         act
 
+data ShouldValidateHashesOnFastForward = ValidateHashesOnFastForward | DoNotValidateHashesOnFastForward
 
 -- | INTERNAL FUNCTION. USE 'withCheckpointer' instead.
 --
@@ -409,7 +410,7 @@ rewindTo rewindLimit (Just (ParentHeader parent)) = do
                     withAsync (heightProgress (_blockHeight commonAncestor) heightRef (logInfo_ progressLogger)) $ \_ ->
                       s
                           & S.scanM
-                              (\ !p !c -> runPact (fastForward (ParentHeader p, c)) >> writeIORef heightRef (_blockHeight c) >> return c)
+                              (\ !p !c -> runPact (fastForward DoNotValidateHashesOnFastForward (ParentHeader p, c)) >> writeIORef heightRef (_blockHeight c) >> return c)
                               (return h) -- initial parent
                               return
                           & S.length_
@@ -423,9 +424,10 @@ fastForward
     :: forall tbl
     . HasCallStack
     => CanReadablePayloadCas tbl
-    => (ParentHeader, BlockHeader)
+    => ShouldValidateHashesOnFastForward
+    -> (ParentHeader, BlockHeader)
     -> PactServiceM tbl ()
-fastForward (target, block) =
+fastForward shouldValidateHashes (target, block) =
     -- This does a restore, i.e. it rewinds the checkpointer back in
     -- history, if needed.
     withCheckpointerWithoutRewind (Just target) "fastForward" $ \pdbenv -> do
@@ -436,7 +438,10 @@ fastForward (target, block) =
                 <> ". BlockPayloadHash: " <> encodeToText bpHash
                 <> ". Block: "<> encodeToText (ObjectEncoded block)
             Just x -> return $ payloadWithOutputsToPayloadData x
-        void $ execBlock block payload pdbenv
+        pwo <- execBlock block payload pdbenv
+        case shouldValidateHashes of
+            ValidateHashesOnFastForward -> either throwM (\_ -> return ()) pwo
+            DoNotValidateHashesOnFastForward -> return ()
         return $! Save block ()
     -- double check output hash here?
   where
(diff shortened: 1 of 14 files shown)


src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -436,7 +438,12 @@ fastForward (target, block) =
                 <> ". BlockPayloadHash: " <> encodeToText bpHash
                 <> ". Block: "<> encodeToText (ObjectEncoded block)
             Just x -> return $ payloadWithOutputsToPayloadData x
-        void $ execBlock block payload pdbenv
+        pwo <- execBlock block payload pdbenv
+        -- As an optimization, we only validate the payload hash of a block during a
+        -- Pact replay, and not during normal operation.
+        case shouldValidateHashes of
+            ValidateHashesOnFastForward -> either throwM (\_ -> return ()) pwo
+            DoNotValidateHashesOnFastForward -> return ()
```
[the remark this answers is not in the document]
larskuhtz: As outlined above, I think, a better solution is to keep `_psValidateHashesOnReplay` in the environment and call `execBlock via

```haskell
local (\env -> env { _psValidateHashesOnReplay = False }) $ execBlock block payload pdbenv
```

src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -625,7 +632,7 @@ rewindToIncremental rewindLimit (Just (ParentHeader parent)) = do
                     -- transactions (withBatchIO).
                     let playChunk :: IORef BlockHeight -> BlockHeader -> Stream (Of BlockHeader) IO r -> IO (Of BlockHeader r)
                         playChunk heightRef cur s = withBatchIO runPact $ \runPactLocal -> S.foldM
-                            (\c x -> x <$ (runPactLocal (fastForward (ParentHeader c, x)) >> writeIORef heightRef (_blockHeight c)))
+                            (\c x -> x <$ (runPactLocal (fastForward ValidateHashesOnFastForward (ParentHeader c, x)) >> writeIORef heightRef (_blockHeight c)))
```
larskuhtz, line 635: This is only ever called with `ValidateHashesOnFastForward`. So why not drop the parameter?

src/Chainweb/Pact/PactService/ExecBlock.hs
```diff
@@ -148,12 +148,9 @@ execBlock currHeader plData pdbenv = do
...
 
-    -- Validate hashes if requested
-    asks _psValidateHashesOnReplay >>= \x -> when x $
-        either throwM (void . return) $!
-        validateHashes currHeader plData miner results
-
-    return $! T2 miner results
+    -- validate hashes only on evaluation of the result
+    return $
+      validateHashes currHeader plData miner results
```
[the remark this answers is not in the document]
larskuhtz: As described above, I think, it is cleaner to keep using `_psValidateHashesOnReplay` (with `True` as default value, but could also be the other way around) and set it locally in the environment as needed.

The return type could then be simplified to just `PactServiceM tbl PayloadWithOutputs`, because all calls to `validateHashes` and possibly throwing an exception would happen here in execBlock.

src/Chainweb/Pact/Types.hs
```diff
@@ -362,7 +361,6 @@ data PactServiceEnv tbl = PactServiceEnv
     -- ^ The limit of checkpointer's rewind in the `execValidationBlock` command.
     , _psOnFatalError :: forall a. PactException -> Text -> IO a
     , _psVersion :: ChainwebVersion
-    , _psValidateHashesOnReplay :: !Bool
```
larskuhtz, line 365: I propose to keep this with `True` as default. It can then be disabled locally where it is safe and beneficial.

tools/cwtool/TxSimulator.hs
```diff
@@ -155,7 +155,7 @@ simulate sc@(SimConfig dbDir txIdx' _ _ cid ver gasLog doTypecheck) = do
           withRocksDb "txsim-rocksdb" modernDefaultOptions $ \rdb ->
             withBlockHeaderDb rdb ver cid $ \bdb -> do
               let pse = PactServiceEnv Nothing cpe paydb bdb getGasModel readRewards 100 0 ferr
-                        ver True False logger gasLogger (pactLoggers cwLogger) False 1 defaultBlockGasLimit cid
+                        ver False logger gasLogger (pactLoggers cwLogger) False 1 defaultBlockGasLimit cid
```
larskuhtz, line 158: Unrelated to this PR, but I think, it would be nicer (and easier to read and maintain) if verbose record initialization syntax would be used here.

larskuhtz (changes requested):
tldr, I think, keeping `_psValidateHashesOnReplay` in the PactService environment will result in a smaller PR and cleaner code.

The value could be set to `True` as default. This will allow to get rid of the check in `execValidateBlock` and to simplify the return type of `execBlock` to just `PactServiceM tbl PayloadWithOutputs`.

During replay in fastforward, which is the only place when the check isn't performed, it can be disabled by calling `execBlock` via `local (\e -> e { _psValidateHashesOnReplay = False }) $ execBlock ...`.

tools/cwtool/TxSimulator.hs
edmundnoble: Totally agree.

src/Chainweb/Pact/PactService/ExecBlock.hs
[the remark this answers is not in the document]
edmundnoble: The invariant here is that we should not produce a `PayloadWithOutputs` without validating it, because the reason we would need a `PayloadWithOutputs` is to insert it into the database. So we have two possible contexts here: either
1. we're validating a block we've already validated, so we don't need an artifact we can insert into the database and we're doing it purely for the effect on the checkpointer and Pact state, so the result can be ignored, or
2. we're validating a new block, so we do need an artifact we can insert into the database.

To preserve this invariant, if we remove the laziness here, we can either
1. validate unconditionally, which requires that we analyze the performance of validation, or
2. stop returning a `PayloadWithOutputs` and instead return what we used to return, the pair of `(Miner, Transactions)`, which we can make a `PayloadWithOutputs` from. My worry there is that the caller can choose to just make a `PayloadWithOutputs` without validating it, and it is as far as I know never valid to make a `PayloadWithOutputs` from the output of this function without validating.

So the code shown is Pareto-optimal in a sense, it both prevents us from shooting ourselves in the foot by producing an invalid `PayloadWithOutputs`, and it prevents us from doing the work of validating that `PayloadWithOutputs` if we don't need to produce it at all.

One other thing we can do, is we can make this function return `IO (IO PayloadWithOutputs)`, and make executing the inner IO action trigger hash validation. To me, though, this seems even more confusing, because it's not obvious which effects exactly are in the inner action, whereas it's intuitive to me that an `Either PactException PayloadWithOutputs` will when evaluated do exactly the work necessary to produce a valid `PayloadWithOutputs`.

src/Chainweb/Pact/PactService/Checkpointer.hs
edmundnoble: It's called with ValidateHashesOnFastForward when we're doing a replay (i.e. `SyncToBlock`) but with `DoNotValidateHashesOnFastForward` when we're doing a fast forward as part of a rewind (i.e. the blocks have already been validated)

edmundnoble:
I think despite my above comments, I'll go with the suggested change instead, just because it's not clearly any worse than what I did and has a smaller diff.

src/Chainweb/Pact/PactService/ExecBlock.hs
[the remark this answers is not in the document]
larskuhtz: > The invariant here is that we should not produce a PayloadWithOutputs without validating it

Why not make that explicit on the type level by returning `Maybe PayloadWithOutputs` or using two functions (different semantics deserves different functions)? Using laziness is completely invisible on the type level but still affects semantics (in this time mostly perf).

larskuhtz:
> I think despite my above comments, I'll go with the suggested change instead, just because it's not clearly any worse than what I did and is a smaller change. But I do consider it a form of tech debt that we have a parameter which is morally intended only for `fastForward`, which we include in the `PactServiceEnv` to be passed everywhere.

I agree, that the returning an potentially unvalidated `PayloadWithOutputs` and relying on the caller to be aware of it is not good. 

I think a relatively simple and clean improvement would be to just provide three function `execBlockWithUnvalidatedPayloadHash`, `execBlockValidatePayloadHash`, execBlock, where the latter would return only `IO ()`. The last two would just be thin wrappers around `execBlockWithUnvalidatedPayloadHash`. Users of `execBlockWithUnvalidatedPayloadHash` would be ware that they are using an optimization and that the had to discharge the respective invariant in their context.

larskuhtz (approved):
I agree, that it would be good to add the invariant that `execBlock` returns only validated `PayloadWithOutputs` values.

I think, the PR is fine because it reflects the current behavior.

edmundnoble:
Just a note here to add this flag removal to the changelog for next release.

PR #1686, opened 2023-06-26
Add a confirmation depth parameter to /poll

ak3n:
This PR:

- adds confirmation depth parameter to /poll endpoint
- changes `execLookupPactTxs` and `_cpLookupProcessedTx` to batch processing (impacts /poll as well)
- adds test to check if confirmation depth affects the /poll behaviour


src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -296,19 +296,39 @@ doRegisterSuccessful :: Db -> PactHash -> IO ()
...
     qtext = "SELECT blockheight, hash FROM \
             \TransactionIndex INNER JOIN BlockHistory \
-            \USING (blockheight) WHERE txhash = ?;"
-    go [SInt h, SBlob blob] = do
+            \USING (blockheight) WHERE txhash IN (" <> hashesParams <> ")"
+            <> maybe "" (const " AND blockheight <= ?") confDepth
+            <> ";"
+    hashesParams = Utf8 $ intercalate "," [ "?" | _ <- V.toList hashes]
+
+    go ((SInt h):(SBlob blob):_) = do
```
ak3n, line 331: The only suspicious thing of this PR to me.

I don't know why but sometimes the results that are being returned include not two element lists, but three elements and the last one is an empty blob `SBlob ""`.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -296,19 +296,39 @@ doRegisterSuccessful :: Db -> PactHash -> IO ()
...
+        qtypes = RInt : [ RBlob | _ <- V.toList hashes ] ++ maybe [] (const [RInt]) blockheight
+
+      qry db qtext qvals qtypes >>= mapM go
+    return $ HashMap.fromList (zip (V.toList hashes) r)
   where
     qtext = "SELECT blockheight, hash FROM \
             \TransactionIndex INNER JOIN BlockHistory \
-            \USING (blockheight) WHERE txhash = ?;"
-    go [SInt h, SBlob blob] = do
+            \USING (blockheight) WHERE txhash IN (" <> hashesParams <> ")"
```
edmundnoble, line 326: No way to do this without string concatenation eh? It's ugly, but we weren't using a prepared statement here anyway, so I doubt it's a big issue.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -296,19 +296,39 @@ doRegisterSuccessful :: Db -> PactHash -> IO ()
...
+        Just (ConfirmationDepth cd) -> do
+          currentHeight <- qry_ db currentHeightQ [RInt]
+          case currentHeight of
+            [[SInt bh]] -> pure $ Just (bh - fromIntegral cd)
+            _ -> fail "impossible"
+
+      let
+        blockheightval = maybe [] (\bh -> [SInt bh]) blockheight
+        qvals = [ SBlob (BS.fromShort hash) | (TypedHash hash) <- V.toList hashes ] ++ blockheightval
+        qtypes = RInt : [ RBlob | _ <- V.toList hashes ] ++ maybe [] (const [RInt]) blockheight
```
edmundnoble, line 319: Maybe I'm missing something here but to me, this should be `[RInt, RBlob]` still.
You have `SELECT blockheight, hash` as the body of the query, that means the rows you get out are each a pair of a block height (int) and a hash (blob). And to me this list is supposed to be the list of the columns that are expected in each row returned by the query.

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -317,8 +317,8 @@ data Checkpointer = Checkpointer
     , _cpGetBlockParent :: !((BlockHeight, BlockHash) -> IO (Maybe BlockHash))
     , _cpRegisterProcessedTx :: !(P.PactHash -> IO ())
 
-      -- TODO: this would be nicer as a batch lookup :(
```
edmundnoble, line 320: :partying_face:

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
edmundnoble: This is probably due to the above `qtypes` business.

src/Chainweb/Pact/RestAPI.hs
```diff
@@ -147,6 +149,22 @@ pactLocalWithQueryApi
...
+-- -------------------------------------------------------------------------- --
+-- POST Queries for Pact Poll
+
+type PactPollWithQueryApi_
+    = "poll"
+    :> QueryParam "confirmationDepth" ConfirmationDepth
+    :> ReqBody '[JSON] Poll
+    :> Post '[JSON] PollResponses
+
+type PactPollWithQueryApi v c = PactV1ApiEndpoint v c PactPollWithQueryApi_
```
edmundnoble, line 161: Should we change this in Pact instead of here?

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -296,19 +296,39 @@ doRegisterSuccessful :: Db -> PactHash -> IO ()
...
-         qry db qtext [ SBlob (BS.fromShort hash) ] [RInt, RBlob] >>= mapM go
-    case r of
-        [] -> return Nothing
-        (!o:_) -> return (Just o)
+      -- if there is a confirmation depth, we get the current height and calculate
+      -- the block height, to look for the transactions in range [0, current block height - confirmation depth]
+      blockheight <- case confDepth of
+        Nothing -> pure Nothing
+        Just (ConfirmationDepth cd) -> do
+          currentHeight <- qry_ db currentHeightQ [RInt]
```
edmundnoble, line 311: I'm not sure how worth it this is exactly for performance, but I suspect we should try to eliminate this query. The current block height is available in the `execLookupPactTxs` function via the argument `restorePoint :: Rewind` so you can calculate the confirmation height in there and pass it here instead of making this query to calculate it here.

In the `NoRewind` case you can still do the query. It's debatable whether `NoRewind` is ever correct, it may eventually be deleted.

test/Chainweb/Test/RestAPI/Utils.hs
```diff
@@ -267,6 +282,10 @@ polling sid cenv rks pollingExpectation =
       Just cr ->  _crReqKey cr == rk && validate (_crResult cr)
       Nothing -> False
 
+getCurrentBlockHeight :: ChainwebVersion -> ClientEnv -> ChainId -> IO (Maybe BlockHeight)
+getCurrentBlockHeight сv cenv cid = do
+  cuts <- either (const $ error "failed to get cuts") id <$> runClientM (cutGetClient сv) cenv
+  return $ _bhwhHeight <$> HM.lookup cid (_cutHashes cuts)
```
edmundnoble, line 288: I think it's fine here to return `IO BlockHeight`, I don't see why we'd call this function with a `cid` that doesn't exist on the `cv` at the time.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -227,6 +230,38 @@ localContTest iot nio = testCaseSteps "local continuation test" $ \step -> do
...
+    step "check /poll responses to extract pact id for continuation"
+
+    beforePolling <- getCurrentBlockHeight v cenv sid
+    PollResponses _ <- pollingWithDepth sid cenv rks (Just $ ConfirmationDepth 70) ExpectPactResult
+    afterPolling <- getCurrentBlockHeight v cenv sid
+
+    case (beforePolling, afterPolling) of
+      (Just bf, Just af) ->
+        assertBool "the difference between heights should be no less than the confirmation depth" $ (af - bf) >= 70
+      _ -> assertFailure "couldn't get block heights"
```
edmundnoble, line 250: It took some looking but I see what you're doing here. You rely on the `pollingWithDepth` function doing a bunch of retries until the poll succeeds. So this test checks the current block height, polls until the tx goes in, and then checks again; if the confirmation depth param worked, we should've seen at least `depth` blocks pass while polling. Maybe this test could use a comment explaining the whole idea, but I think the `step` suggestions above are enough.

It does take a while, like 14 seconds. But I really like the test, so that's fine with me.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -227,6 +230,38 @@ localContTest iot nio = testCaseSteps "local continuation test" $ \step -> do
...
+  where
+    sid = unsafeChainId 0
```
edmundnoble, line 252: `s/sid/cid`

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -227,6 +230,38 @@ localContTest iot nio = testCaseSteps "local continuation test" $ \step -> do
...
+    tx =
+      "(namespace 'free)(module m G (defcap G () true) (defpact p () (step (yield { \"a\" : (+ 1 1) })) (step (resume { \"a\" := a } a))))(free.m.p)"
```
edmundnoble, line 254: We can simplify this a bit from `localContTest` above. We don't actually care what's in `tx`, as long as it completes. It can even fail. I recommend something simple like `tx = "42"`, to make it clearer that it doesn't matter what it does.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -227,6 +230,38 @@ localContTest iot nio = testCaseSteps "local continuation test" $ \step -> do
         $ mkCont
         $ mkContMsg pid 1
 
+pollingConfirmDepth :: IO (Time Micros) -> IO ChainwebNetwork -> TestTree
+pollingConfirmDepth iot nio = testCaseSteps "poll confirmation depth test" $ \step -> do
+    cenv <- _getServiceClientEnv <$> nio
+
+    step "execute /send with initial pact continuation tx"
```
edmundnoble, line 237:
```suggestion
    step "/send a transaction"
```

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -227,6 +230,38 @@ localContTest iot nio = testCaseSteps "local continuation test" $ \step -> do
...
+    cmd1 <- firstStep
+    rks <- sending sid cenv (SubmitBatch $ pure cmd1)
+
+    step "check /poll responses to extract pact id for continuation"
```
edmundnoble, line 241:
```suggestion
    step "/poll for the transaction until it appears"
```

src/Chainweb/RestAPI/Orphans.hs
```diff
@@ -195,3 +195,9 @@ instance FromHttpApiData RewindDepth where
 
 instance ToHttpApiData RewindDepth where
     toUrlPiece (RewindDepth k) = toUrlPiece k
+
+instance FromHttpApiData ConfirmationDepth where
+    parseUrlPiece = fmap ConfirmationDepth . parseUrlPiece
+
+instance ToHttpApiData ConfirmationDepth where
+    toUrlPiece (ConfirmationDepth k) = toUrlPiece k
```
edmundnoble, line 203: I didn't think of this earlier with the depth/limit PR, but how about something like:
```suggestion
deriving newtype instance FromHttpApiData ConfirmationDepth

deriving newtype instance ToHttpApiData ConfirmationDepth
```
You can replace some other instances in this file in that way if it works too.

edmundnoble (changes requested):
Believe it or not, there are (as far as I can see) *zero* tests in `RemotePactTest` for the multi-request-key case in `/poll`. And you can kinda see how we got away with it before seeing as the implementation did a query for each request key. But now we can't anymore, so I think that warrants a new test in there. I would also like a benchmark of `/poll` if we can get it, showing that this is faster, but we can talk about that.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
chessai: Apparently sqlite doesn't even support `WHERE IN ?`. The best we can do is (?, ?, ?, ..., ?). Not to mention the limitations with direct-sqlite.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
ak3n: Yep, my bad, I've interpreted this parameter as a list of types of the query arguments, but it's the list of types of expected results. Thanks!

src/Chainweb/RestAPI/Orphans.hs
ak3n: Yep, updated

test/Chainweb/Test/Pact/RemotePactTest.hs
ak3n: Agree, thanks!

test/Chainweb/Test/Pact/RemotePactTest.hs
ak3n: Yeah, I have reduced the confirmation depth to `10`, takes a few seconds instead of `14` now.

src/Chainweb/Pact/RestAPI.hs
ak3n: I'm following the same approach taken for /local.

I guess there shouldn't even be `PactServer` in `pact` repo? Because that's the scope of the chainweb-node and it's too many efforts to keep the source of truth in the `pact` repo. What's the status of pact dev server `Pact.Server.Server`?

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
ak3n: Yeah, I thought about using the block height from parent header we use for rewinding but then discovered that there is no such with `NoRewind`. And we would need to get the blockheight the same way as now but with `doGetLatest`.

I think this query is very cheap in the performance sense. The only thing that I would add — isolation, both queries should be in the transaction to be consistent.

ak3n:
>Believe it or not, there are (as far as I can see) zero tests in RemotePactTest for the multi-request-key case in /poll.

Actually there are few cases when there are multiple request keys, at least I saw them during running the testsuite with tracing.

>I would also like a benchmark of /poll if we can get it, showing that this is faster, but we can talk about that.

Hmmm, then we need to think about the scenario I suppose. There is `Chainweb.Pact.Backend.Bench` where it's possible to call `_cpLookupProcessedTx` directly, but we will just bench the sqlite calls.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
emilypi: i don't think this opens us up to injections the way we've constructed prepared statements in the past, due to it being hidden behind a the lookup function and only called on validated request keys.

src/Chainweb/Pact/RestAPI.hs
emilypi: no - we need this api to stay static for the `pact -s` use cases. Augmenting it here is more preferable imo.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -227,6 +230,35 @@ localContTest iot nio = testCaseSteps "local continuation test" $ \step -> do
...
+    firstStep = do
+      t <- toTxCreationTime <$> iot
+      buildTextCmd
+        $ set cbSigners [mkSigner' sender00 []]
+        $ set cbCreationTime t
+        $ set cbNetworkId (Just v)
+        $ set cbGasLimit 70000
+        $ mkCmd "nonce-cont-1"
+        $ mkExec' tx
+
```
emilypi, line 261: We should add coverage more than a singleton batch here

test/Chainweb/Test/RestAPI/Utils.hs
```diff
@@ -267,6 +283,10 @@ polling sid cenv rks pollingExpectation =
       Just cr ->  _crReqKey cr == rk && validate (_crResult cr)
       Nothing -> False
 
+getCurrentBlockHeight :: ChainwebVersion -> ClientEnv -> ChainId -> IO BlockHeight
+getCurrentBlockHeight сv cenv cid = do
+  cuts <- either (const $ error "failed to get cuts") id <$> runClientM (cutGetClient сv) cenv
+  return $ fromJust $ _bhwhHeight <$> HM.lookup cid (_cutHashes cuts)
```
emilypi, line 289: the error here is really harsh

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
edmundnoble: Yeah I think we're fine, it would just be nice if we could use a prepared stmt.

emilypi: For sure. Agreed

test/Chainweb/Test/RestAPI/Utils.hs
ak3n: Introduced `GetBlockHeightFailure` for this, thanks!

src/Chainweb/Pact/Service/Types.hs
```diff
@@ -71,6 +71,11 @@ newtype RewindDepth = RewindDepth { _rewindDepth :: Word64 }
   deriving newtype (FromJSON, ToJSON, Enum, Bounded)
 instance Show RewindDepth where show (RewindDepth d) = show d
 
+newtype ConfirmationDepth = ConfirmationDepth { _confirmationDepth :: Word64 }
+  deriving (Eq, Ord)
+  deriving newtype (FromJSON, ToJSON, Enum, Bounded)
+instance Show ConfirmationDepth where show (ConfirmationDepth d) = show d
+
```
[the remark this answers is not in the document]
ak3n: Fixed, thanks!

test/Chainweb/Test/Pact/RemotePactTest.hs
ak3n: Could you please elaborate? What's missing?

bench/Chainweb/Pact/Backend/Bench.hs
```diff
@@ -372,3 +375,34 @@ cpBenchSampleKeys numSampleEvents cp =
...
+    f = "f"
+    k = "k"
+
+    hash01 = BlockHash $ unsafeMerkleLogHash "0000000000000000000000000000001a"
+    hash02 = BlockHash $ unsafeMerkleLogHash "0000000000000000000000000000002a"
+
+    go CheckpointEnv{..} (NoopNFData ut) = do
+        _cpRestore _cpeCheckpointer (Just (BlockHeight 2, hash02)) >>= \case
+          PactDbEnv' db ->
+            _cpLookupProcessedTx _cpeCheckpointer Nothing (V.fromList [Pact.TypedHash "" | _ <- [1..transactionCount]])
```
ak3n, line 408: @edmundnoble I added a simple benchmark where I just pass empty txhashes because I don't know how to calculate them on this level. There are request keys but we operate on much lower level here in the `setup'`.

Here are the results:

```
benchmarking pact-backend/batchedCheckpointer/lookupProcessedTx/transactionCount=100
time                 157.0 μs   (149.1 μs .. 166.8 μs)
                     0.988 R²   (0.976 R² .. 0.999 R²)
mean                 149.4 μs   (147.8 μs .. 153.9 μs)
std dev              8.776 μs   (4.321 μs .. 16.49 μs)
variance introduced by outliers: 58% (severely inflated)

benchmarking pact-backend/batchedCheckpointer/lookupProcessedTx/transactionCount=1000
time                 914.2 μs   (900.4 μs .. 933.6 μs)
                     0.998 R²   (0.996 R² .. 0.999 R²)
mean                 912.3 μs   (905.7 μs .. 921.5 μs)
std dev              26.52 μs   (20.01 μs .. 37.84 μs)
variance introduced by outliers: 18% (moderately inflated)
```

And these results are without batching with the following implementation:

```
void $ mapM (\i -> _cpLookupProcessedTx _cpeCheckpointer Nothing (V.fromList [i])) [Pact.TypedHash "" | _ <- [1..transactionCount]]
```

```
benchmarking pact-backend/batchedCheckpointer/lookupProcessedTx/transactionCount=100
time                 2.294 ms   (2.205 ms .. 2.406 ms)
                     0.983 R²   (0.973 R² .. 0.992 R²)
mean                 2.309 ms   (2.261 ms .. 2.372 ms)
std dev              199.3 μs   (159.0 μs .. 249.0 μs)
variance introduced by outliers: 60% (severely inflated)

benchmarking pact-backend/batchedCheckpointer/lookupProcessedTx/transactionCount=1000
time                 29.24 ms   (24.26 ms .. 31.75 ms)
                     0.948 R²   (0.911 R² .. 0.981 R²)
mean                 24.60 ms   (23.48 ms .. 26.24 ms)
std dev              3.407 ms   (2.490 ms .. 4.185 ms)
variance introduced by outliers: 61% (severely inflated)
```

edmundnoble: Looks great.

edmundnoble (approved):

emilypi (approved):

test/Chainweb/Test/Pact/RemotePactTest.hs
ak3n: There are some cases when there are several items in the request keys.

ak3n: Added one more transaction to the request keys.

PR #1688, opened 2023-07-05
Small cleanups and add gas to validateBlock time telemetry

edmundnoble:
Currently the "weight" parameter of the `validateBlock` trace telemetry is the number of transactions. I don't know if we have much use for that. But we do definitely have a use for the amount of gas in the block. This also makes some small cleanups to BlockFill to make it more paranoid in the presence of mempool bugs and adds a bang pattern or two.

Asking for Stuart's review for the block fill stuff.

---
  - https://app.asana.com/0/0/1204717400575171


src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -416,17 +416,6 @@ applyCoinbase v logger dbEnv (Miner mid mks@(MinerKeys mk)) reward@(ParsedDecima
...
-          -- in Rosetta's `matchLogs` function and follow the coinv3 pattern.
-          --
-          -- Otherwise, Rosetta tooling has no idea that these upgrade transactions
-          -- occurred.
-          -- This is especially important if the transaction changes an account's balance.
-          -- Rosetta tooling will error out if an account's balance changed and it
-          -- didn't see the transaction that caused the change.
-          --
-
           logs <- use txLogs
```
edmundnoble, line 419: This comment became obsolete when version recordification went in.

src/Chainweb/Pact/Types.hs
```diff
@@ -429,7 +429,7 @@ defaultPactServiceConfig = PactServiceConfig
 -- is initialized from the @_configBlockGasLimit@ value of @ChainwebConfiguration@.
 --
 defaultBlockGasLimit :: GasLimit
-defaultBlockGasLimit = 10000
+defaultBlockGasLimit = 20000
```
edmundnoble, line 432: Tests which used testing mempools in `PactMultiChainTest` stopped working because those mempools are willing to provide blocks of transactions that exceed the block gas limit, which was now being enforced.

src/Utils/Logging/Trace.hs
```diff
@@ -73,16 +74,28 @@ instance ToJSON Trace where
 trace
     :: MonadIO m
     => ToJSON param
-    => LogFunction
+    => (LogLevel -> JsonLog Trace -> IO ())
```
larskuhtz, line 77: 👍

src/Chainweb/Pact/PactService.hs
```diff
@@ -148,6 +148,7 @@ withPactService ver cid chainwebLogger bhDb pdb sqlenv config act =
                     , _psAllowReadsInLocal = _pactAllowReadsInLocal config
                     , _psIsBatch = False
                     , _psCheckpointerDepth = 0
+                    , _psTraceLogger = logFunction chainwebLogger
```
larskuhtz, line 151: 👍

larskuhtz (approved):

src/Chainweb/Pact/PactService.hs
edmundnoble: Later we'll just put a chainweb logger into the checkpointer as suggested by @chessai.

edmundnoble:
I reverted the substantive changes to BlockFill because the way the gas limit works is just a bit too confusing for me to justify putting it in this PR. I'll make a ticket for doing this later, if we even want to.

PR #1689, opened 2023-07-05
Drop `NoRewind` and `withCurrentCheckpointer`

ak3n:
---
  - https://app.asana.com/0/0/1204936704075315


ak3n:
Substituted by #1803

PR #1691, opened 2023-07-07
use a growable vector for amortised O(1) append in block filling

chessai:


src/Chainweb/Pact/PactService.hs
```diff
@@ -538,71 +540,90 @@ execNewBlock mpAccess parent miner = do
...
 
-        (BlockFilling _ successPairs failures) <-
-          refill fetchLimit txTimeLimit pdbenv =<<
-          foldM splitResults (incCount (BlockFilling initState mempty mempty)) pairs
-
-        liftIO $ mpaBadlistTx mpAccess (V.map gasPurchaseFailureHash failures)
-
-        let !pwo = toPayloadWithOutputs miner (Transactions successPairs cb)
+        --initialBlockFilling <- liftIO $ BlockFilling initState <$> Dyna.new <*> Dyna.new
+        --
```
edmundnoble, line 544: Delete?

src/Chainweb/Pact/PactService.hs
```diff
@@ -538,71 +540,90 @@ execNewBlock mpAccess parent miner = do
...
+    refill :: Word64 -> Micros -> PactDbEnv' -> GrowableVec (ChainwebTransaction, P.CommandResult [P.TxLog A.Value]) -> GrowableVec GasPurchaseFailure -> BlockFill -> PactServiceM tbl BlockFill
+    refill fetchLimit txTimeLimit pdbenv successes failures = go
+      where
+        go :: BlockFill -> PactServiceM tbl BlockFill
+        go unchanged@bfState = do
 
-      logDebug $ describeBF unchanged
+          (goodLength, badLength) <- liftIO $ (,) <$> Dyna.length successes <*> Dyna.length failures
+          case unchanged of
+            (BlockFill g _ c) -> do
```
edmundnoble, line 567: Parens aren't needed here

cabal.project
```diff
@@ -129,6 +135,8 @@ constraints: base16-bytestring <1
 constraints: prettyprinter <1.6.1
 constraints: hashable <1.3.1
 constraints: base64-bytestring <1.1
+constraints: contiguous == 0.6.2.0
+constraints: run-st == 0.1.1.0
```
edmundnoble, line 139: This is because of GHC version issues?

edmundnoble (approved):
Looks good, I'll ask for Lars' feedback here too.

cabal.project
larskuhtz, line 139: The constraints above are required to not break pact validation. Is that also the case for these?

Otherwise, could you add a comment about why these constraints are needed.

chessai: It's not the same case. I'll add a comment and put them in their own section.

PR #1697, opened 2023-07-18
gha: verbose nix builds and --show-trace in case an error occurs

chessai:
This is so error messages don't get truncated, and we can get callstack/trace error messages

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -31,4 +31,4 @@ jobs:
       timeout-minutes: 740
       run: |
         echo Building the project and its devShell
-        nix build .#check
+        nix build -L --show-trace .#check
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -31,4 +31,4 @@ jobs:
       timeout-minutes: 740
       run: |
         echo Building the project and its devShell
-        nix build -L --show-trace .#check
+        nix build .#check --log-lines 500 --show-trace


enobayram:
@chessai I'm a little worried about the `-L` argument, particularly for the builds when we bump `nixpkgs` or `haskellNix`, because those builds tend to take many hours and I'm not sure how much github actions will like the potentially gigabytes of logs that might be generated in that case.

chessai:
> @chessai I'm a little worried about the `-L` argument, particularly for the builds when we bump `nixpkgs` or `haskellNix`, because those builds tend to take many hours and I'm not sure how much github actions will like the potentially gigabytes of logs that might be generated in that case.

Very fair. Is there a convenient way to only use `-L` if those haven't been updated?

enobayram:
@chessai I've just discovered this `nix` command argument (seems to work for all subcommands): `--log-lines`. It tells nix how many log lines to display when a derivation fails. I've tried it on a simple failing derivation like this:
``` nix
let
  nixpkgs = import <nixpkgs> {};
in

with nixpkgs;

stdenv.mkDerivation {
  name = "failing-derivation";

  builder = writeScript "failing-builder" ''
    #!/bin/sh
    for i in $(${nixpkgs.coreutils}/bin/seq 1 100); do
      echo "Line $i"
    done
    echo "This derivation is designed to fail after 100 lines!"
    exit 1
  '';

  system = builtins.currentSystem;
}
```

```
$ nix build -f failtest.nix --log-lines 30
error: builder for '/nix/store/qk1pmm7dycwd3qiyis6mlqbb02xbnd95-failing-derivation.drv' failed with exit code 1;
       last 30 log lines:
       > Line 72
       > Line 73
       > Line 74
       > Line 75
       > Line 76
       > Line 77
       > Line 78
       > Line 79
       > Line 80
       > Line 81
       > Line 82
       > Line 83
       > Line 84
       > Line 85
       > Line 86
       > Line 87
       > Line 88
       > Line 89
       > Line 90
       > Line 91
       > Line 92
       > Line 93
       > Line 94
       > Line 95
       > Line 96
       > Line 97
       > Line 98
       > Line 99
       > Line 100
       > This derivation is designed to fail after 100 lines!
       For full logs, run 'nix log /nix/store/qk1pmm7dycwd3qiyis6mlqbb02xbnd95-failing-derivation.drv'.
```

So I guess we could just add something like `--log-lines 200` to our `nix build` calls during CI.

enobayram (approved):
LGTM!

PR #1698, opened 2023-07-18
Add max gas limit configuration and check to the /local endpoint

ak3n:
This PR adds the max gas limit configuration param to limit the gas limit in the `/local` endpoint.

---
  - https://app.asana.com/0/0/1204876391903201

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -49,6 +49,7 @@ import Control.Monad.Primitive (PrimState)
 
 import Data.Default (def)
 import qualified Data.DList as DL
+import Data.Foldable (for_)
 import Data.Either
 import Data.Word (Word64)
 import Data.Maybe (fromMaybe)
@@ -153,6 +154,7 @@ withPactService ver cid chainwebLogger bhDb pdb sqlenv config act =
                     , _psReorgLimit = _pactReorgLimit config
                     , _psLocalRewindDepthLimit = _pactLocalRewindDepthLimit config
                     , _psPreInsertCheckTimeout = _pactPreInsertCheckTimeout config
+                    , _psLocalMaxGasLimit = _pactLocalMaxGasLimit config
                     , _psOnFatalError = defaultOnFatalError (logFunctionText chainwebLogger)
                     , _psVersion = ver
                     , _psAllowReadsInLocal = _pactAllowReadsInLocal config
@@ -677,6 +679,10 @@ execLocal cwtx preflight sigVerify rdepth = pactLabel "execLocal" $ withDiscarde
 
     mc <- getInitCache
     spv <- use psSpvSupport
+
+    for_ _psLocalMaxGasLimit $ \maxGasLimit ->
+        when (view cmdGasLimit cmd > maxGasLimit) $
+            throwM $ LocalGasLimitExceeded maxGasLimit (view cmdGasLimit cmd)
 
     -- when no depth is defined, treat
     -- withCheckpointerRewind as withCurrentCheckpointer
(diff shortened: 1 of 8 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -677,6 +679,10 @@ execLocal cwtx preflight sigVerify rdepth = withDiscardedBatch $ do
     ctx <- getTxContext pm
     spv <- use psSpvSupport
 
+    for_ _psLocalMaxGasLimit $ \maxGasLimit ->
+        when (view cmdGasLimit cmd > maxGasLimit) $
+            throwM $ LocalGasLimitExceeded maxGasLimit (view cmdGasLimit cmd)
+
```
edmundnoble, line 685: I'm not sure which @emilypi prefers: should we reject txs with a gas limit higher than the max, or should we run the tx with the smaller of the two gas limits?

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -237,6 +239,27 @@ chainweb213Test = do
         , "(create-table tbl)"
         ]
 
+pactLocalMaxGasLimitTest :: PactTestM ()
+pactLocalMaxGasLimitTest = do
+  runToHeight 53
```
chessai, line 244: Can you include a comment for why this is 53?

ak3n: Removed as it doesn't make any difference here. Was a leftover from copying the test.

src/Chainweb/Pact/PactService.hs
ak3n: Just a note, since @emilypi's comment is a bit outdated, the current error provides both the requested gas limit and node's limitation:

```
LocalGasLimitExceeded
    { _localMaxGasLimit :: !GasLimit
    , _localRequestedGasLimit :: !GasLimit
    }
```

emilypi: @ak3n perfect, thanks

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -1408,3 +1426,8 @@ cbResult = do
   (o,_h) <- getPWO chid
   liftIO $
     decodeStrictOrThrow @_ @(CommandResult Hash) (_coinbaseOutput $ _payloadWithOutputsCoinbase o)
+
+checkLocalResult :: Either PactException LocalResult -> (CommandResult Hash -> PactTestM ()) -> PactTestM ()
+checkLocalResult r checkResult = case r of
+  Right (LocalResultLegacy cr) -> checkResult cr
+  res -> liftIO $ assertFailure $ "Expected LocalResultLegacy, but got: " ++ show res
```
emilypi, line 1433: Optional nit: this is really checking for _legacy_ results, which is not the same as peeking into and handling local results in general. This should probably have a comment and namechange to that effect.

emilypi (approved):
Seems mergeable at this point modulo nits. @edmundnoble?

edmundnoble (changes requested):
This PR will break quite a few of our users who set unnecessarily gas limits on local calls just to avoid having to worry about the gas limit. We can't just error out if their gas limit is too high, we want to just limit what the actual gas limit that Pact sees ends up as.

emilypi:
@edmundnoble but if people want arbitrary execution and set the limit extremely high, the only endpoint they can use is the legacy local because the preflight api is limited by the node limit. In that case they're using the node as a repl in the sky, rather than actually simulating transactions. If the command limit exceeds the node limit in the preflight case, the fact that it's going to error is feedback.

edmundnoble:
> In that case they're using the node as a repl in the sky, rather than actually simulating transactions.

Certainly, the contents of the transactions show that as well, for example using `describe-module`.

> If the command limit exceeds the node limit in the preflight case, the fact that it's going to error is very useful feedback.

Oh so your idea here is that we should only throw the error in preflight mode? That makes sense to me. But in that case the `local` limit has to be the block gas limit, because otherwise it'll error when in a real tx it could've succeeded.

ak3n:
Closing this as we will resolve the original issue in a different way.

PR #1700, opened 2023-07-19
Fix the bug with TxContext and current GasModel in /local

ak3n:
This PR fixes the incorrect usage of current `TxContext`, making it use the rewinding `ParendHeader`. Also fixes the problem with using the current gas model.

### DIFF
--- a/test/Chainweb/Test/Pact/RemotePactTest.hs
+++ b/test/Chainweb/Test/Pact/RemotePactTest.hs
@@ -363,6 +363,7 @@ localPreflightSimTest iot nio = testCaseSteps "local preflight sim test" $ \step
 
     step "Execute preflight /local tx - collect warnings"
     cmd7 <- mkRawTx' mv pcid sigs0 "(+ 1 2.0)"
+
     runLocalPreflightClient sid cenv cmd7 >>= \case
       Left e -> assertFailure $ show e
       Right LocalResultLegacy{} ->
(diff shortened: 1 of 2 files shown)


test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -9,6 +9,8 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
```
ak3n, line 12: Sorry, but I decided it's irrational to spend time to please the GHC here.

edmundnoble: Where even was the issue?

edmundnoble (approved):
Looks good.

test/Chainweb/Test/Pact/RemotePactTest.hs
ak3n: Ok, adding a type hint worked properly this time.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -363,15 +363,34 @@ localPreflightSimTest iot nio = testCaseSteps "local preflight sim test" $ \step
...
     runLocalPreflightClient sid cenv cmd7 >>= \case
       Left e -> assertFailure $ show e
       Right LocalResultLegacy{} ->
         assertFailure "Preflight /local call produced legacy result"
       Right MetadataValidationFailure{} ->
         assertFailure "Preflight produced an impossible result"
       Right (LocalResultWithWarns cr' ws) -> do
-        -- check the presence of metadata
-        assertBool "Preflight result should have metadata" $ isJust $ _crMetaData cr'
+        currentBlockHeight <- getCurrentBlockHeight v cenv sid
```
edmundnoble, line 374: I like that you're using the metadata here.

jwiegley:
>>>>> "EA" == Evgenii Akentev ***@***.***> writes:

EA> Sorry, but I decided it's irrational to spend time to please the GHC here.

Saving time today that may impact future readers’ comprehension for years to
come is not an economical savings. If there is no reasonable way to satisfy
GHC, then you must include a comment as to why that directive is being added
to the code, especially if it’s being used to resolve similar issues anywhere
else.

PR #1704, opened 2023-07-21
Add requestKeys logging to execNewBlock

ak3n:
---
  - https://app.asana.com/0/0/1205043156117013

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -539,14 +539,13 @@ execNewBlock mpAccess parent miner = pactLabel "execNewBlock" $ do
           Nothing
           (Just txTimeLimit) `catch` handleTimeout
 
-        let requestKeys = map P._crReqKey $ rights $ map snd $ V.toList pairs
-        logInfo $ "execNewBlock: "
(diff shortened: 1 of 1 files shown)


edmundnoble:
When we produce a new block, the process looks like:
  * grab a set of transactions `txs` from the mempool such that `sum (map txGasLimit) txs <= blockGasLimit`
  * execute these transactions and record the actual gas used
  * then in a loop:
    * grab a set of transactions `newTxs` from the mempool such that `sum (map txGasLimit) txs + currentlyUsedGas <= blockGasLimit`
      * if this set is empty, we're done
      * if this set is not empty, execute `newTxs` and update the actual gas used to include the amount of gas used by `newTxs`

We call this process "block fill". That's just background on how it works. The trouble is that this code actually only logs the request keys in the first batch of `txs`, ignoring all of the `newTxs`.

ak3n:
@edmundnoble thanks for the explanation! I was not sure about logging the keys after the refill step.

Take a look again please, we can just reuse the refill result then.

edmundnoble (changes requested):
Actually let's also do this for execPreInsertCheck in this PR.

edmundnoble (approved):

chessai (approved):

PR #1705, opened 2023-07-24
Generate UUID for each PactService request

ak3n:
---
  - https://app.asana.com/0/0/1205075354172799

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE ImportQualifiedPost #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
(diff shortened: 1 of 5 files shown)


src/Chainweb/Pact/PactService.hs
```diff
@@ -270,17 +271,19 @@ lookupBlockHeader bhash ctx = do
 -- | Loop forever, serving Pact execution requests and reponses from the queues
 serviceRequests
     :: forall logger tbl. (Logger logger, CanReadablePayloadCas tbl)
-    => LogFunction
+    => logger
```
chessai, line 274: why are we passing in a logger, when there is a logger in the PactServiceEnv?

chessai: I believe that this is an oversight from my PactService logging PR that this is still there.

ak3n: Fixed this.

chessai (approved):

edmundnoble (approved):

chessai (approved):

PR #1707, opened 2023-07-27
run pact tests in ci

edmundnoble:
---
  - https://app.asana.com/0/0/1204876390256506


edmundnoble:
Already done in TransactionTests.

PR #1710, opened 2023-07-28
Fix double JsonLog nesting in logJsonTrace, make it impossible

edmundnoble:


chessai (approved):

src/Data/LogMessage.hs
```diff
@@ -166,8 +166,8 @@ aNoLog = ALogFunction $ \_ _ -> return ()
 
 -- | A newtype wrapper for log messages types with a 'ToJSON' instance.
 --
```
larskuhtz, line 168:
```suggestion
-- | A newtype wrapper for log messages types with a 'ToJSON' instance.
--
-- This type must not have a `ToJSON` instance.
--
```

larskuhtz (approved):
Just one suggestion

PR #1713, opened 2023-07-29
Much quieter replays

edmundnoble:
Silences non-Error messages from Pact during replays and lets the module cache stop cleaning itself up when it's already clean.

---
  - https://app.asana.com/0/0/1204717400575185


edmundnoble:
Test failures on no-freeze-file builds are caused by new hashable, which changes the hash values. Updating the freeze file and `--accept`ing the changes.

edmundnoble:
Reverting in case this causes a problem.

larskuhtz:
Arrg, updating `hashable` should not cause pact validation failures. With aeson-2 stuff merged, this should be a thing of the past. We should create an issue for investigating this.

src/Chainweb/Pact/TransactionExec.hs
```diff
@@ -1155,7 +1155,7 @@ fatal e = do
     rk <- view txRequestKey
 
     logError_ l
-      $! "critical transaction failure: "
+      $ "critical transaction failure: "
```
emilypi, line 1158: What's the reason for undoing the bangs

src/Chainweb/Pact/PactService/ExecBlock.hs
```diff
@@ -388,6 +390,14 @@ applyPactCmds isGenesis env cmds miner mc blockGas txTimeLimit = do
       evalStateT (V.mapM (applyPactCmd isGenesis env miner txTimeLimit) cmds) (T2 mc blockGas)
     return txs
 
+-- we don't want pact tx failure or unknown exception messages when we're doing a replay.
+withSilentLoggerForReplays :: Logger logger => (logger -> PactServiceM logger tbl a) -> PactServiceM logger tbl a
+withSilentLoggerForReplays f = do
+  replaying <- view psReplaying
+  logger <- view psLogger
+  if replaying then L.withLoggerLevel L.Error logger f
```
larskuhtz, line 398: Not sure if this is too broad. This would also suppress unrelated legitimate warnings logs -- assuming that those could exist.

src/Chainweb/Pact/Types.hs
```diff
@@ -286,6 +286,15 @@ moduleCacheKeys :: ModuleCache -> [ModuleName]
 moduleCacheKeys (ModuleCache a) = fst <$> LHM.toList a
 {-# INLINE moduleCacheKeys #-}
 
+-- this can't go in Chainweb.Version.Guards because it causes an import cycle
+-- it uses genesisHeight which is from BlockHeader which imports Guards
```
larskuhtz, line 290: Would it make sense to move `genesisHeight` to a different place instead?

larskuhtz (changes requested):

src/Chainweb/Pact/Types.hs
edmundnoble: Probably would cause a cycle. I tried it during recordification.

src/Chainweb/Pact/PactService/ExecBlock.hs
```diff
@@ -151,8 +152,9 @@ execBlock currHeader plData pdbenv = do
 
     modify' $ set psStateValidated $ Just currHeader
 
-    -- Validate hashes if requested
-    asks _psValidateHashesOnReplay >>= \x -> when x $
+    -- Validate hashes if not doing a replay. no need to re-validate hashes
+    -- during a replay, because these blocks have already been validated
+    view psReplaying >>= \x -> unless x $
```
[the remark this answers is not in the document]
edmundnoble: I'm probably just confused here. I'll just make it always check.

src/Chainweb/Pact/TransactionExec.hs
edmundnoble: Loggers are always strict, so the bangs aren't necessary... except that they're lazy when they're filtered such that the log message never makes it to the queue. I've seen some pretty big log messages here, so I'd like those filtered out in case that can raise throughput.

larskuhtz: yeah, log messages are deep forced before added to the queue.

Avoiding construction when their log level is disabled is a good point.

edmundnoble:
Tested on personal node and CI, it's much quieter.

chessai (approved):

src/Chainweb/Pact/PactService/ExecBlock.hs
```diff
@@ -151,10 +151,8 @@ execBlock currHeader plData pdbenv = do
 
     modify' $ set psStateValidated $ Just currHeader
 
-    -- Validate hashes if requested
-    asks _psValidateHashesOnReplay >>= \x -> when x $
-        either throwM (void . return) $!
-        validateHashes currHeader plData miner results
+    either throwM (void . return) $
+      validateHashes currHeader plData miner results
```
larskuhtz, line 155: 👍

larskuhtz (approved):

PR #1717, opened 2023-08-02
support hashable-1.4.3

larskuhtz:
The updates the golden files are caused to changes in the test harness and don't represent on-chain semantics.

Tested that the changes also work with hashable<1.4.3.


test/Chainweb/Test/Cut/TestBlockDb.hs
```diff
@@ -64,13 +63,34 @@ mkTestBlockDb cv rdb = do
...
+    -- success
+    Right (T2 _ c') -> do
+        casInsert pdb outs
+        putMVar cmv c'
+        return True
+
+    -- mining failed, probably because chain is blocked
+    Left BadAdjacents -> do
+        putMVar cmv c
+        return False
```
larskuhtz, line 90: Before the function would fail with an `userError` in this case. This change informs the caller when mining fails.

In practice there are tests where it can happen that the mining of a blocked chain is attempted. It was by pure chance that this didn't showed up before. The latest version of `hashable` changed the hash value of chain ids and thus the ordering of chains in tests, which caused this issue to surface.

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -71,7 +70,8 @@ testVersion :: ChainwebVersion
 testVersion = slowForkingCpmTestVersion peterson
 
 cid :: ChainId
-cid = someChainId testVersion
+cid = unsafeChainId 9
+    -- several tests in this file expect chain 9
```
larskuhtz, line 74: With the new hashable version the value of `someChainId` changes. This fixes the previous value.

src/Chainweb/Version/Utils.hs
```diff
@@ -223,9 +223,8 @@ someChainId v = someChainIdAt v maxBound
 -- chain ids for the chainweb at the given height.
 --
 someChainIdAt :: HasCallStack => HasChainwebVersion v => v -> BlockHeight -> ChainId
-someChainIdAt v h = head . toList $ chainIdsAt v h
-    -- 'head' is guaranteed to succeed because the empty graph isn't a valid chain
-    -- graph.
+someChainIdAt v h = minimum $ chainIdsAt v h
+    -- guaranteed to succeed because the empty graph isn't a valid chain graph.
```
larskuhtz, line 227: We could have fixed the value to value with the previous version of `hashable`, which would have avoided updating of golden files. But using the `minimum` seems like a more reasonable and intuitive default.

cabal.project.freeze
```diff
@@ -344,7 +344,7 @@ constraints: any.Boolean ==0.2.4,
              any.time-compat ==1.9.6.1,
              time-compat -old-locale,
              any.time-manager ==0.0.0,
-             any.tls ==1.7.0,
+             any.tls ==1.7.1,
```
larskuhtz, line 347: this update is unrelated, but nice to have.

test/Chainweb/Test/Cut/TestBlockDb.hs
ak3n: What do you think about applying a "if comments on github make it easy to understand the PR, they should be in the code" rule here?

ak3n (approved):
LGTM 👍

test/Chainweb/Test/Cut/TestBlockDb.hs
larskuhtz: I am not sure in case of this comments, since they talk an old version of the code and the diff with the new version. I think, PR comments my be a better place for those.

But, generally, I agree, that comments that explain the behavior of the new code should be code comments.

edmundnoble (approved):

PR #1722, opened 2023-08-16
Use ResourceT in tests (+ small cleanups in tests)

edmundnoble:
This is hopefully the first of many cleanups for the tests, probably going to be a big refactor later to make things less of a rats' nest.

TODO: eliminate more IO params and hardcoded uses of `withResource`

---
  - https://app.asana.com/0/0/1204717400575193


ak3n (commented):
Looks good but I don't see the point yet I suppose. It made the code more verbose and less clear in small details to me.

What are the main benefits?

edmundnoble:
> What are the main benefits?
Existing functions which create resources for tests are hard-coded to use tasty's `withResource` to create and destroy them, which is for passing the resource to a `TestTree` - the resources are always created before some test group starts and destroyed after it finishes. `ResourceT` here serves as a more general type for functions that create resources: we can convert a `ResourceT` into a `withResource` call using `withResourceT` allowing resources to live during a test group, and we can also use `runResourceT` to let the resources live for an arbitrary amount of time within a test.

It does also mean a little bit more boilerplate because the `withResourceT` calls have to be added to call-sites, though that's not a huge downside. To compensate I also did some light refactoring here and there to make sure this PR is at least net negative on LOC.

My hope here in the medium term is that by making the resource code just a bit more reusable it makes it a bit easier to split up the test util code and make things a bit easier to discover.

ak3n:
Maybe add this comment as a small `NOTE` to `test/Chainweb/Test/Utils.hs` close to `withResourceT `?

src/Chainweb/Cut.hs
```diff
@@ -642,7 +642,7 @@ type DiffItem a = These a a
 
 type JoinQueue a = H.Heap (H.Entry (BlockHeight, a) BlockHeader)
 
--- | This represents the Join of two cuts in an algrithmically convenient way.
+-- | This represents the Join of two cuts in an algorithmically convenient way.
```
chessai, line 645: Spelling errors give the codebase character and flair.

edmundnoble:
Came up with another use for this in chessai's compaction tests. We wanted to create a PactService to populate the db with blocks, shut it down, and start a new differently configured PactService. We had to make *two tests that run in sequence*, and then because we expected the new PactService to fail, and it's a test resource, we had to intercept the resource code... nasty.

chessai (approved):

emilypi (approved):

PR #1725, opened 2023-08-25
Read only rewind

ak3n:


src/Chainweb/Pact/PactService.hs
```diff
@@ -823,13 +827,14 @@ execHistoricalLookup
     -> PactServiceM logger tbl (Maybe (P.TxLog P.RowData))
 execHistoricalLookup bh d k = pactLabel "execHistoricalLookup" $ do
   !cp <- getCheckpointer
+  -- TODO: pass parent header to limit the visible range
```
ak3n, line 830: Hm, I think it doesn't make sense.

The limitation would make sense only in situations when the parent header got out of sync and normally we don't need to limit these queries as we don't do any rewinds here.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -174,8 +184,9 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
         -- First, check: did we create this table during this block? If so,
         -- there's no point in looking up the key.
         checkDbTableExists tableName
+        let blockLimitParam = maybe [] (\(BlockHeight bh) -> [SInt $ fromIntegral bh - 1]) mbh
         result <- lift $ callDb "doReadRow"
-                       $ \db -> qry db queryStmt [SText rowkey] [RBlob]
+                       $ \db -> qry db queryStmt ([SText rowkey] ++ blockLimitParam) [RBlob]
```
chessai, line 189:
```suggestion
                       $ \db -> qry db queryStmt (SText rowkey : blockLimitParam) [RBlob]
```

edmundnoble:
Closed in favor of #1804

PR #1729, opened 2023-08-28
new base64-bytestring and webauthn-signatures

imalsogreg:


src/Chainweb/Rosetta/Utils.hs
```diff
@@ -726,7 +719,7 @@ createSigningPayloads (EnrichedCommand cmd _ _) = map f
 
     toRosettaSigType Nothing = Just RosettaEd25519
     toRosettaSigType (Just P.ED25519) = Just RosettaEd25519
-    toRosettaSigType (Just P.ETH) = Just RosettaEcdsa -- TODO: unsupport this
+    toRosettaSigType (Just P.WebAuthn) = Just RosettaEd25519 -- TODO Ok?
```
edmundnoble, line 722: Not sure, maybe ask Linda.

src/Chainweb/Pact/SPV.hs
```diff
@@ -169,7 +177,77 @@ verifySPV bdb bh typ proof = runExceptT $ go typ proof
...
+--
+--   `Simplified` errors are a static string, which may not describe
+--   the issue with the base64-encoded string as well, but will be
+--   more stable as we upgrade base64 decoding libraries in the future.
+--   In the `Simplified` errors setting, messages rejected for using
+--   non-canonical encodings will remain as errors, because we want to
+--   begin enforcing the expected constraints on base64-encoded messages.
+--   Similarly, Simplified parsing will not add extra padding to improperly
+--   padded messages.
+data GenerateBase64ErrorMessage
```
imalsogreg, line 198: @emilypi and @edmundnoble here is a (verbose) strategy for forking the base64 decoding behavior. This code could use lots of scrutiny!

cabal.project
```diff
@@ -158,16 +158,10 @@ allow-newer: base-compat-batteries:*
...
--- these upper bounds are required in order to not break payload validation
-constraints: base16-bytestring <1
-constraints: base64-bytestring <1.1
-
--- we have to add these because the old versions that we use would prevent
--- us to use recent versions of other packages
-allow-newer: base64-bytestring:*
-allow-newer: base16-bytestring:*
+-- other pact induced bounds (not relevant for on-chain semantics)
+-- constraints: megaparsec <9.3
```
emilypi, line 162: this megaparsec bound seems like a merge artifact

src/Chainweb/Pact/SPV.hs
```diff
@@ -169,7 +177,77 @@ verifySPV bdb bh typ proof = runExceptT $ go typ proof
...
+  $ msg
+  where
+    pad t = let s = T.length t `mod` 4 in t <> T.replicate ((4 - s) `mod` 4) "="
+    base64ErrorMessage m = case errorMessageType of
+      Legacy -> base64DowngradeErrorMessage (Text.pack m)
+      Simplified -> "could not base64-decode message"
+    patchNonCanonical decodeResult = case decodeResult of
+      Right bs -> Right bs
+      Left e | "non-canonical" `T.isInfixOf` T.pack e ->
+               Right (B64U.decodeLenient (T.encodeUtf8 msg))
```
emilypi, line 222: The reasoning behind this change in particular is that the previous base64-bytestring versions allowed "non-canonical" encodings in the sense that they are not true base64, but will decode to a particular string that will not round trip. Because we can assume that canonicity errors are the result of a structurally sound base64 input, we defer to the lenient decoder, which doesn't enforce canonicity. This is *terribly inefficient*, but since it only runs once or twice on a string of max size 1kb, we're not going to suffer too much.

src/Chainweb/Pact/SPV.hs
```diff
@@ -169,7 +177,77 @@ verifySPV bdb bh typ proof = runExceptT $ go typ proof
...
+  e -> e
+  where
+    adjustedOffset :: Text -> Int
+    adjustedOffset suffix = case readMaybe (Text.unpack suffix) of
+      Nothing -> 0
+      Just offset -> let
+        endsWithThreeEquals = T.drop (T.length msg - 3) msg == "==="
+        adjustment = if endsWithThreeEquals then -1 else 0
+        in
+        offset - (offset `rem` 4) + adjustment
```
emilypi, line 250: These calculations patch the difference between the old decode messages generated from improper padding (e.g. `X===`) and the new ones which fail immediately.

src/Chainweb/Pact/SPV.hs
```diff
@@ -183,7 +261,14 @@ verifyCont
...
 verifyCont bdb bh (ContProof cp) = runExceptT $ do
-    t <- liftIO $ decodeB64UrlNoPaddingText $ T.decodeUtf8 cp
+    let errorMessageType =
+          if CW.chainweb220Pact
+             (CW._chainwebVersion bh)
+             (_blockChainId bh)
+             (_blockHeight bh)
+          then Simplified
+          else Legacy
+    t <- decodeB64UrlNoPaddingTextWithFixedErrorMessage errorMessageType $ T.decodeUtf8 cp
```
emilypi, line 271: nice 👍

src/Chainweb/Rosetta/Utils.hs
emilypi: @LindaOrtega

chainweb.cabal
```diff
@@ -514,7 +514,7 @@ test-suite chainweb-tests
         , aeson >= 1.4.3
         , async >= 2.2
         , base >= 4.12 && < 5
-        , base64-bytestring >= 1.0 && < 1.1
+        , base64-bytestring >= 1.0
             -- due to change in error message that is inlined in pact
         , bytestring >= 0.10
         , case-insensitive >= 1.2
```
emilypi, line 520: does `base16-bytestring` need to be bumped as well, or is that unused in Chainweb?

src/Chainweb/Utils.hs
```diff
@@ -283,6 +285,8 @@ import System.Timeout
 import Text.Printf (printf)
 import Text.Read (readEither)
 
+-- import Pact.Types.Util (base64DowngradeErrorMessage)
```
edmundnoble, line 288: ^ Cleanup commented lines

src/Chainweb/Version/Mainnet.hs
```diff
@@ -139,7 +139,9 @@ mainnet = ChainwebVersion
         Chainweb217Pact -> AllChains (ForkAtBlockHeight $ BlockHeight 3_250_348) -- 2022-12-02T00:00:00+00:00
         Chainweb218Pact -> AllChains (ForkAtBlockHeight $ BlockHeight 3_512_363) -- 2023-03-03 00:00:00+00:00
         Chainweb219Pact -> AllChains (ForkAtBlockHeight $ BlockHeight 3_774_423) -- 2023-06-02 00:00:00+00:00
-        Chainweb220Pact -> AllChains ForkNever
+        Chainweb220Pact -> AllChains (ForkAtBlockHeight $ BlockHeight 4_056_499) -- 2023-09-08 00:00:00+00:00
+        -- TODO: This is a placeholder!
+        Chainweb221Pact -> AllChains (ForkAtBlockHeight $ BlockHeight 4_156_499) -- 2023-10-14 00:00:00+00:00
```
edmundnoble, line 144: For this and testnet replace with `AllChains ForkNever` just to avoid accidentally leaving this in

edmundnoble (approved):
Good to merge after Pact PR is in.

src/Chainweb/Version/Development.hs
```diff
@@ -66,6 +66,7 @@ devnet = ChainwebVersion
             Chainweb218Pact -> AllChains $ ForkAtBlockHeight $ BlockHeight 500
             Chainweb219Pact -> AllChains $ ForkAtBlockHeight $ BlockHeight 550
             Chainweb220Pact -> AllChains $ ForkAtBlockHeight $ BlockHeight 560
+            Chainweb221Pact -> AllChains $ ForkAtBlockHeight $ BlockHeight 580
```
imalsogreg, line 69: @edmundnoble Should this one, too be `ForkNever`, in this PR?

src/Chainweb/Rosetta/Utils.hs
LindaOrtega, line 722:
```suggestion
    toRosettaSigType (Just P.WebAuthn) = Nothing 
    -- TODO: Linda Ortega (09/18/2023) -- Returning `Nothing` to discourage using WebAuthn for Rosetta. `sigToScheme` will eventually throw an error.
```

src/Chainweb/Version/Development.hs
edmundnoble: No, this is fine.

Testnet and mainnet are always on, so their block height is advancing and we need to make sure their fork heights roughly match up with specific times. Devnet isn't always on, so this just affects how long it takes until a fresh devnet can use forks.

chainweb.cabal
imalsogreg: Had to be bumped to >= 1.0 for compatibility with `webauthn` lib..

emilypi (commented):
I think it would be worth a negative test for Pact SPV to cover our bases.

flake.nix
```diff
@@ -2,14 +2,9 @@
   description = "Chainweb";
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs?rev=4d2b37a84fad1091b9de401eb450aae66f1a741e";
```
chessai, line 5: why?

i would much rather us pin our own nixpkgs. we have our own cache so we don't need to depend on IOG for that.

chessai (changes requested):

ak3n (approved):

chessai (approved):

emilypi (approved):

PR #1730, opened 2023-08-29
Update bootstrap nodes & readme with reference to seed download data

mbwmbw1337:
Add a reference to community-hosted projects providing chain web-node data archives.

Signed-off-by: Michael WIlliams <Michael.Williams@kaddex.xyz>


mbwmbw1337:
I've updated the URI to use the new kaddex.xyz domain.

PR #1733, opened 2023-09-08
add _cpEarliestBlock

chessai:
The opposite of `_cpLatestBlock`. Needed for multiple other PRs, so I thought it should go in on its own.


src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -297,6 +297,9 @@ data Checkpointer logger = Checkpointer
       -- ^ commits pending modifications to block, with the given blockhash
     , _cpDiscard :: !(IO ())
       -- ^ discard pending block changes
+    , _cpGetEarliestBlock :: !(IO (Maybe (BlockHeight, BlockHash)))
```
larskuhtz, line 300: Why is the the `Maybe` needed? Is it possible to have a chainweb-node without any block in the database?

larskuhtz (changes requested):

src/Chainweb/Pact/Backend/Types.hs
chessai: > Why is the the `Maybe` needed? Is it possible to have a chainweb-node without any block in the database?

I think it shouldn't be. We should probably throw an exception in that case.

larskuhtz (approved):

PR #1736, opened 2023-09-14
make PactService exit if rosetta is enabled and we don't have the full history

chessai:
Note that this targets the `compaction` branch, not `master`.

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -72,6 +72,7 @@ import Prelude hiding (lookup)
 
 import qualified Pact.Gas as P
 import Pact.Gas.Table
+import qualified Pact.JSON.Encode as J
 import qualified Pact.Interpreter as P
 import qualified Pact.Types.ChainMeta as P
 import qualified Pact.Types.Command as P
@@ -127,8 +128,7 @@ runPactService ver cid chainwebLogger reqQ mempoolAccess bhDb pdb sqlenv config 
         serviceRequests mempoolAccess reqQ
 
 withPactService
-    :: Logger logger
-    => CanReadablePayloadCas tbl
+    :: forall logger tbl a. (Logger logger, CanReadablePayloadCas tbl)
     => ChainwebVersion
     -> ChainId
     -> logger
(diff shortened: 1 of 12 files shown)


chessai:
It seems hard to test this code - trying to access SQLite from PactSingleChainTest is giving me nothing but errors. There's something I'm missing here.

src/Chainweb/Pact/PactService.hs
```diff
@@ -162,9 +164,30 @@ withPactService ver cid chainwebLogger bhDb pdb sqlenv config act =
                     , _psBlockGasLimit = _pactBlockGasLimit config
                     , _psChainId = cid
                     }
-            !pst = PactServiceState Nothing mempty initialParentHeader P.noSPVSupport
-        runPactServiceM pst pse $ do
+        let !pst = PactServiceState Nothing mempty initialParentHeader P.noSPVSupport
+
+        when (_pactRosettaEnabled config) $ do
```
ak3n, line 169: Do we really need the configuration flag for this? For efforts/configuration from the user's POV.

We could run the check before starting the rosetta HTTPServer, turning it off if there is no required data in the db for example.

src/Chainweb/Pact/PactService.hs
```diff
@@ -162,9 +164,30 @@ withPactService ver cid chainwebLogger bhDb pdb sqlenv config act =
...
+                  , _genesisHeight = gHeight
+                  }
+            let msg = J.object
+                  [ "details" J..= e
+                  , "message" J..= J.text "Your node has Rosetta enabled,\
+                      \ which requires that the full history is available.\
+                      \ However, the full history is not available.\
+                      \ Perhaps you have compacted your node?"
+                  ]
+            onFatalError e (J.encodeText msg)
```
chessai, line 188: This should not use `onFatalError`: We should throw here, and something else should be catching whatever we throw and logging there.

src/Chainweb/Pact/PactService.hs
chessai: The node should exit if the full history is required for one of its features. However @edmundnoble had the idea that instead of this being whether rosetta is enabled or not, we have a more general flag like `_pactFullHistoryRequired :: Bool`.

edmundnoble (approved):

src/Chainweb/Pact/PactService.hs
edmundnoble: We can use this later too, to determine whether to enable incremental compaction. Even in the short term if we stop writing pending rows from each tx and just write the final rows in a block, we can use this to determine whether to do that.

PR #1739, opened 2023-09-22
add direnv for flake

chessai:


jwiegley (changes requested):
I do not want to see `.envrc` files added to our repositories. Not everyone wishes to use the same version of this file.

jwiegley (approved):

PR #1740, opened 2023-09-22
Fix execValidateBlock

ak3n:


edmundnoble (approved):
We could move the validateHashes check to the smart constructor, but I'm fine with this as it is too.

src/Chainweb/Payload/PayloadStore.hs
```diff
@@ -284,14 +284,7 @@ instance (pk ~ CasKeyType (PayloadWithOutputs_ a), CanReadablePayloadCas_ a tbl)
...
             outsHash
-        return $ PayloadWithOutputs
-            { _payloadWithOutputsTransactions = V.zip (_blockTransactions txs) (_blockOutputs outs)
-            , _payloadWithOutputsMiner = _blockMinerData txs
-            , _payloadWithOutputsCoinbase = _blockCoinbaseOutput outs
-            , _payloadWithOutputsPayloadHash = k
-            , _payloadWithOutputsTransactionsHash = txsHash
-            , _payloadWithOutputsOutputsHash = outsHash
-            }
+        return $ newPayloadWithOutputs (_blockMinerData txs) (_blockCoinbaseOutput outs) (V.zip (_blockTransactions txs) (_blockOutputs outs))
```
edmundnoble, line 287: So we're recomputing the hashes now? Can we avoid that?

tools/cwtool/TxSimulator.hs
```diff
@@ -108,7 +108,9 @@ simulate sc@(SimConfig dbDir txIdx' _ _ cid ver gasLog doTypecheck) = do
       (_cpBeginCheckpointerBatch cp)
       (_cpDiscardCheckpointerBatch cp) $ case (txIdx',doTypecheck) of
         (Just txIdx,_) -> do -- single-tx simulation
-          let PayloadWithOutputs txs md _ _ _ _ :: PayloadWithOutputs = head pwos
+          let pwo :: PayloadWithOutputs = head pwos
```
edmundnoble, line 111: Idk why this type annotation is here.

edmundnoble (changes requested):

src/Chainweb/Payload/PayloadStore.hs
ak3n: Yep, by using the neighbour smart constructor `payloadWithOutputs`.

ak3n:
@edmundnoble can we merge this one as well?

edmundnoble (approved):

PR #1741, opened 2023-09-26
execNewBlock: remove the parent header as parameter and sync from the db instead

ak3n:
---
  - https://app.asana.com/0/0/1205566925523372

### DIFF
--- a/bench/Chainweb/Pact/Backend/ForkingBench.hs
+++ b/bench/Chainweb/Pact/Backend/ForkingBench.hs
@@ -214,7 +214,7 @@ playLine pdb bhdb trunkLength startingBlock pactQueue counter =
   where
     mineLine :: BlockHeader -> Word64 -> IORef Word64 -> IO [T3 ParentHeader BlockHeader PayloadWithOutputs]
     mineLine start l ncounter =
-        evalStateT (runReaderT (mapM (const go) [startHeight :: Word64 .. startHeight + l - 1]) pactQueue) start
+        evalStateT (runReaderT (mapM (const go) [startHeight :: Word64 .. startHeight + l]) pactQueue) start
       where
         startHeight :: Num a => a
         startHeight = fromIntegral $ _blockHeight start
--- a/bench/Chainweb/Pact/Backend/ForkingBench.hs
+++ b/bench/Chainweb/Pact/Backend/ForkingBench.hs
@@ -154,16 +154,28 @@ bench rdb = C.bgroup "PactService" $
     forkingBench = withResources rdb 10 Quiet DontCompact
         $ \mainLineBlocks pdb bhdb nonceCounter pactQueue _ ->
             C.bench "forkingBench"  $ C.whnfIO $ do
-              let (T3 _ join1 _) = mainLineBlocks !! 5
+              let (T3 _ join1 pwo1) = mainLineBlocks !! 5
+
+              -- we have to execValidateBlock on `join1` block height to update the parent header
+              void $ validateBlock join1 (payloadWithOutputsToPayloadData pwo1) pactQueue
+
               void $ playLine pdb bhdb 5 join1 pactQueue nonceCounter
 
     doubleForkingBench = withResources rdb 10 Quiet DontCompact
         $ \mainLineBlocks pdb bhdb nonceCounter pactQueue _ ->
             C.bench "doubleForkingBench"  $ C.whnfIO $ do
-              let (T3 _ join1 _) = mainLineBlocks !! 5
+              let (T3 _ join1 pwo1) = mainLineBlocks !! 5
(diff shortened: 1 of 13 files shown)


test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -1404,6 +1418,12 @@ getPWO chid = do
   pwo <- liftIO $ casLookupM pdb (_blockPayloadHash h)
   return (pwo,h)
 
+getPWOByHeader :: BlockHeader -> PactTestM PayloadWithOutputs
+getPWOByHeader h = do
+  (TestBlockDb _ pdb _) <- view menvBdb
```
edmundnoble, line 1423: No need for these parens

edmundnoble (commented):
This can't be merged until the release. All of my comments are pedantic stuff which barely matters, I'm extremely impressed that we can get away with this, it looks exactly like I'd expect. We need to test this on a real mining node as well before mergew.

test/Chainweb/Test/Pact/PactMultiChainTest.hs
```diff
@@ -1404,6 +1418,12 @@ getPWO chid = do
...
+  pwo <- liftIO $ casLookupM pdb (_blockPayloadHash h)
+  return pwo
```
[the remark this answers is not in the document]
edmundnoble: Hell maybe this should be a shared utility.

edmundnoble:
We actually don't really need to test this on a real mining node, fake mining with a simulated time distribution also works, as long as it's in a network with multiple mining nodes.

ak3n:
@edmundnoble will wait until the master is fixed so we can merge with green CI this PR? Or can merge already? it was green before

edmundnoble:
@ak3n I want this PR benchmarked to see how many blocks we produce with a simulated miner first. I can help with that some time when you're free.

edmundnoble:
Actually we should probably use a load test for this benchmark.

bench/Chainweb/Pact/Backend/ForkingBench.hs
```diff
@@ -129,24 +129,40 @@ bench rdb = C.bgroup "PactService"
     forkingBench = withResources rdb 10 Quiet
         $ \mainLineBlocks pdb bhdb nonceCounter pactQueue _ ->
             C.bench "forkingBench"  $ C.whnfIO $ do
-              let (T3 _ join1 _) = mainLineBlocks !! 5
+              let (T3 _ join1 pwo1) = mainLineBlocks !! 5
+
+              -- we have to execValidateBlock on `join1` block height to update the parent header
+              void $ validateBlock join1 (payloadWithOutputsToPayloadData pwo1) pactQueue
```
0xd34df00d, line 135: You might create a helper function (for the tests only!) like
```haskell
updateParentHeader join pwo queue = void $ validateBlock join (payloadWithOutputsToPayloadData pwo) queue
```
avoiding the comments in the tests and making the tests a bit more expressive.

src/Chainweb/Pact/PactService.hs
```diff
@@ -292,9 +292,9 @@ serviceRequests memPoolAccess reqQ = do
                 go
             NewBlockMsg NewBlockReq {..} -> do
                 trace logFn "Chainweb.Pact.PactService.execNewBlock"
-                    (_parentHeader _newBlockHeader) 1 $
-                    tryOne "execNewBlock" _newResultVar $
-                        execNewBlock memPoolAccess _newBlockHeader _newMiner
+                    () 1 $
+                    tryOne "execNewBlock" _newResultVar $ do
```
0xd34df00d, line 296: Tiny nit: this `do` isn't necessary.

src/Chainweb/Pact/Service/Types.hs
```diff
@@ -292,11 +292,10 @@ data RequestMsg = NewBlockMsg !NewBlockReq
 type PactExMVar t = MVar (Either PactException t)
 
 data NewBlockReq = NewBlockReq
-    { _newBlockHeader :: !ParentHeader
-    , _newMiner :: !Miner
+    { _newMiner :: !Miner
     , _newResultVar :: !(PactExMVar PayloadWithOutputs)
     }
-instance Show NewBlockReq where show NewBlockReq{..} = show (_newBlockHeader, _newMiner)
+instance Show NewBlockReq where show NewBlockReq{..} = show (_newMiner)
```
0xd34df00d, line 298: The parens are probably no longer needed.

0xd34df00d (approved):
The logic looks good as per discussion, so I only have a few minor nits.

ak3n:
Merged in #1803

PR #1744, opened 2023-09-27
Do not persist intra-block writes

edmundnoble:
This stops us writing rows to the database that are overwritten in the same block. I haven't benchmarked it, I don't know if it's faster yet, but it should at least save space.

---
  - https://app.asana.com/0/0/1205531835120463


src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -158,12 +158,11 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
...
         let deltaKey = SQLiteDeltaKey tableNameBS rowkey
         ddata <- fmap _deltaData <$> MaybeT (return $ HashMap.lookup deltaKey (_pendingWrites p))
-        if null ddata
+        case ddata of
             -- should be impossible, but we'll check this case
-            then mzero
-            -- we merge with (++) which should produce txids most-recent-first
-            -- -- we care about the most recent update to this rowkey
-            else MaybeT $ return $! decodeStrict' $ DL.head ddata
+            [] -> mzero
```
chessai, line 163: Can we get some kind of error message here?

test/Chainweb/Test/Utils.hs
```diff
@@ -1160,4 +1160,5 @@ testRetryPolicy = stepped <> limitRetries 150
       0 -> Just 20_000
       1 -> Just 50_000
       2 -> Just 100_000
-      _ -> Just 250_000
```
chessai, line 1163: why this change?

chessai (changes requested):

test/Chainweb/Test/Utils.hs
edmundnoble: I was hoping this would make the tests that wait for a cut height slightly less noisy.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
edmundnoble: Now that you mention it, I think this should be a NonEmpty so that this case is impossible anyway.

ak3n (approved):

src/Chainweb/Pact/Service/Types.hs
```diff
@@ -79,6 +79,9 @@ newtype ConfirmationDepth = ConfirmationDepth { _confirmationDepth :: Word64 }
   deriving (Eq, Ord)
   deriving newtype (Show, FromJSON, ToJSON, Enum, Bounded)
 
+data PersistIntraBlockWrites = PersistIntraBlockWrites | DoNotPersistIntraBlockWrites
+  deriving (Eq, Ord, Show)
```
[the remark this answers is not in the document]
chessai: Personally I prefer the two-constructor sum type approach. Having True and False appear are more mental overheard, to me.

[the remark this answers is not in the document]
edmundnoble: Yeah that's my thinking as well. If it wraps bool, it's not as explicit anymore.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -550,15 +549,12 @@ doGetTxLog d txid = do
                       _deltaTableName delta == tableNameBS
 
     tableName = domainTableName d
-    (Utf8 tableNameBS) = tableName
-
-    takeHead [] = []
-    takeHead (a:_) = [a]
```
edmundnoble, line 556: This was just `take 1`.

test/Chainweb/Test/Utils.hs
edmundnoble: I'm just going to revert this in favor of #1770 which makes this instant.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -240,17 +237,19 @@ recordPendingUpdate (Utf8 key) (Utf8 tn) txid v = modifyPendingData modf
   where
     !vs = J.encodeStrict v
     delta = SQLiteRowDelta tn txid key vs
-    deltaKey = SQLiteDeltaKey tn key
 
     modf = over pendingWrites upd
     {- "const" is used here because prefer the latest update of the rowkey for
```
chessai, line 242: this comment should be deleted - no longer relevant

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -190,7 +182,7 @@ type SQLitePendingTableCreations = HashSet ByteString
 type SQLitePendingSuccessfulTxs = HashSet ByteString
 
 -- | Pending writes to the pact db during a block, to be recorded in 'BlockState'.
```
chessai, line 184: i think you should document that the outer bytestring is the table name, and the inner bytestring is the rowkey. this is no longer obvious without reading the code where this is used, because you are no longer using the type that had that documentation.

thoughtpolice: Yes, agreed. I think just giving them `type` aliases alone would help tremendously.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -341,9 +340,8 @@ doKeys d = do
     pb <- use bsPendingBlock
     mptx <- use bsPendingTx
 
-    let memKeys = DL.toList $
-                  fmap (B8.unpack . _deltaRowKey) $
-                  collect pb `DL.append` maybe DL.empty collect mptx
+    let memKeys = fmap (B8.unpack . _deltaRowKey) $
+                  collect pb ++ maybe [] collect mptx
```
thoughtpolice, line 344: You can left-align the `$` here like the code right after, but just a minor consistency suggestion.

```suggestion
    let memKeys = fmap (B8.unpack . _deltaRowKey)
                $ collect pb ++ maybe [] collect mptx
```

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -550,15 +549,12 @@ doGetTxLog d txid = do
...
-
-    takeHead [] = []
-    takeHead (a:_) = [a]
+    Utf8 tableNameBS = tableName
 
     readFromPending = do
         ptx <- maybeToList <$> use bsPendingTx
         pb <- use bsPendingBlock
-        let deltas = (ptx ++ [pb]) >>= HashMap.elems . _pendingWrites >>= takeHead . DL.toList
+        let deltas = (ptx ++ [pb]) >>= map NE.toList . HashMap.elems . HashMap.unions . HashMap.elems . _pendingWrites >>= take 1
```
thoughtpolice, line 557: At this point I feel you should just write a `do` statement here rather than making this line even longer, TBH.

edmundnoble: I don't really even know if this line is correct. I'm going to try to tease it out.

chessai (approved):

edmundnoble:
Abandoning for #1893

PR #1746, opened 2023-09-28
chainChoice should not loop forever on singletonChainGraph

chessai:


src/Chainweb/Miner/Coordinator.hs
```diff
@@ -224,8 +227,14 @@ chainChoice c choice = case choice of
   where
     loop :: ChainId -> IO ChainId
     loop cid = do
-        new <- randomChainIdAt c (minChainHeight c)
-        bool (pure new) (loop cid) $ new == cid
+        if _chainGraph c == singletonChainGraph
+        then do
+            let millisecondInMicros = 1_000
+            threadDelay millisecondInMicros
```
larskuhtz, line 233: What about using another `ChainChoice` in case of `singletonChainGraph`? Or, alternatively, waiting for a new cut to appear?

edmundnoble: How about waiting until the PrimeWork TVar changes?

chessai:
closing in favour of #1758

PR #1755, opened 2023-10-12
fix formatting of merkle log tag values

larskuhtz:


emilypi (approved):

jwiegley:
Where is the test to demonstrate the correctness of these changes?

larskuhtz:
> Where is the test to demonstrate the correctness of these changes?

Since this goes into the Merkle tree, any test that involves specific block hashes would break. In particular the Genesis headers tests and ea runs. 

Also this PR just changes the formatting of the numbers by removing leading zeros.

larskuhtz:
I good test for these constants would be a "proof of coverage" to test that changing them actually causes the test suite to fail. I think, generally, we could use more "negative" tests that prove that things don't just work by accident.

PR #1756, opened 2023-10-14
compaction pre-and-post active row test

chessai:


test/Chainweb/Test/Pact/PactSingleChainTest.hs
```diff
@@ -12,6 +12,8 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ViewPatterns #-}
 
+{-# options_ghc -fno-warn-unused-imports #-}
```
chessai, line 15: remove this

src/Chainweb/Pact/Backend/Compaction.hs
```diff
@@ -70,7 +73,7 @@ import Data.LogMessage
 import Pact.Types.SQLite (SType(..), RType(..))
 import Pact.Types.SQLite qualified as Pact
 
-newtype TxId = TxId Int64
+newtype ITxId = ITxId Int64
```
edmundnoble, line 76: Whence?

src/Chainweb/Pact/Backend/Compaction.hs
```diff
@@ -626,3 +635,56 @@ compactMain = do
...
+  let excludeThese = checkpointerTables ++ compactionTables
+  let fmtTable x = "\"" <> x <> "\""
+
+  tables <- fmap sortedTableNames $ do
+    let qry =
+          "SELECT name FROM sqlite_schema \
+          \WHERE \
+          \  type = 'table' \
+          \AND \
+          \  name NOT LIKE 'sqlite_%'"
```
edmundnoble, line 652: I like that we have a blacklist for non-compacted tables rather than the other way around.

edmundnoble (approved):

src/Chainweb/Pact/Backend/Compaction.hs
chessai: there is another type in scope that represents a TxId, and it's incompatible with this one. The 'I' means 'Int'. The reason this type exists is to provide some type-safety for TxIds used with Pact's SQLite API.

src/Chainweb/Pact/Backend/Compaction.hs
chessai: So, you are saying, you prefer the code as written? Or are you suggesting a change?

PR #1764, opened 2023-10-20
compaction compare pact state tool

chessai:
compares pact state pre- and post-compaction

---
  - https://app.asana.com/0/0/1205378300255963

### DIFF
--- a/src/Chainweb/Pact/Backend/PactState.hs
+++ b/src/Chainweb/Pact/Backend/PactState.hs
@@ -39,8 +39,11 @@ module Chainweb.Pact.Backend.PactState
   where
 
 import Data.IORef (newIORef, readIORef, atomicModifyIORef')
+import Control.Exception (bracket)
 import Control.Monad (forM, forM_, when, void)
 import Control.Monad.IO.Class (MonadIO(liftIO))
+import Control.Monad.Trans.Class (lift)
+import Control.Monad.Except (ExceptT(..), runExceptT, throwError)
 import Data.Aeson (ToJSON(..), (.=))
 import Data.Aeson qualified as Aeson
 import Data.Vector (Vector)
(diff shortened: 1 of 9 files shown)


src/Chainweb/Pact/Backend/PactState.hs
```diff
@@ -0,0 +1,337 @@
...
+    -- the chain ids at the current (version,height) pair
+    latestBlockHeight <- C.withDefaultLogger Error $ \logger -> do
+      let resetDb = False
+      withSqliteDb (unsafeChainId 0) logger cfg.pactDbDir resetDb $ \(SQLiteEnv db _) -> do
+        getLatestBlockHeight db
+    pure $ List.sort $ F.toList $ chainIdsAt cfg.chainwebVersion latestBlockHeight
+
+  -- Compaction is destructive. In order to do a streaming diff, we need to make
+  -- a copy of the pact state.
+  withSystemTempDirectory "pact-compact-diff" $ \tmpPactDbDir -> do
```
chessai, line 283: there should be an argument for where the copy should go. and there should be a check that it's different from the pactDbDir. this way the post-compaction db can be inspected.

tools/cwtool/CwTool.hs
```diff
@@ -104,7 +105,11 @@ topLevelCommands =
   , CommandSpec
       "compact"
       "Compact pact database"
-      compactMain
+      Chainweb.Pact.Backend.Compaction.main
+  , CommandSpec
+      "pact-db-compare"
```
chessai, line 110: this should probably be something like "compact-compare"

ak3n (approved):
LGTM 👍 

I think it would be nice to have a comment why exactly we use `Stream`s here, at first glance they look more like an overcomplication and it's unclear if we benefit from them.

chessai:
> LGTM 👍 
> 
> I think it would be nice to have a comment why exactly we use `Stream`s here, at first glance they look more like an overcomplication and it's unclear if we benefit from them.

Sounds good to me. I'll add a note in the code. It's primarily because the memory requirements can get very high for all the combined tables.

src/Chainweb/Pact/Backend/PactState.hs
```diff
@@ -0,0 +1,345 @@
...
+        withSqliteDb cid logger cfg.pactDbDir resetDb $ \(SQLiteEnv db _) -> do
+          diffEmptyRef <- newIORef True
+
+          let diff = diffLatestPactState (getLatestPactState db) (getLatestPactState cpDb)
+          flip S.mapM_ diff $ \utd -> do
+            writeIORef diffEmptyRef False
+            loggerFunIO logger Warn $ toLogMessage $
+              TextLog $ Text.decodeUtf8 $ BSL.toStrict $ Aeson.encode utd
+
+          diffEmpty <- readIORef diffEmptyRef
```
chessai, line 304: This should be something that happens at the very end. We should have an `IORef (Map ChainId Bool)`.

src/Chainweb/Pact/Backend/PactState.hs
```diff
@@ -0,0 +1,345 @@
...
+          when (not diffEmpty) $ do
+            loggerFunIO logger Warn $ toLogMessage $
```
chessai, line 306: We should be logging to a file, not to stdout.

PR #1758, opened 2023-10-16
STM-based chain selection algo for mining

edmundnoble:

PR #1766, opened 2023-10-25
Ignore ChainChoice, and log more

edmundnoble:
This amounts to what we discussed offline for #1758, as a much more conservative change that's less likely to have negative performance implications but still gives us better logging when there are issues finding a chain to mine and performs better in weirder cases where there's very little `PrimedWork`. It also still constitutes deleting `ChainChoice` as the PR name suggests. Strictly speaking we only need to stop using `TriedLast :: ChainId -> ChainChoice` to get the benefits from this PR, but `Suggested :: ChainId -> ChainChoice` has been slated for deletion for a while now too, so we just ignore the choice in this PR.

I added a test that `singletonChainGraph` works properly in slow-tests, which is what this fixes. I slightly refactored the slow-tests code to ensure that the database directories don't overlap, so that the tests could run in parallel, but for some reason they still can't run in parallel. I added a `sequentialTestGroup` for now, that might be a better idea even in the long run.

---
  - https://app.asana.com/0/0/1205799595610145


ak3n (approved):
LGTM 👍

edmundnoble:
I've asked in discord whether anyone was using the chain suggestion feature and haven't gotten any answer thus far.

edmundnoble:
I've changed this PR to be more conservative, and still use the suggested chain ID on its first try.

test/SlowTests.hs
```diff
@@ -30,11 +35,25 @@ main = defaultMain suite
...
+        withSystemTempDirectory "multinode-tests-fasttimedcpm-single-pact" $ \pactDbDir ->
+        Chainweb.Test.MultiNode.test loglevel (fastForkingCpmTestVersion singletonChainGraph) 10 30 rdb pactDbDir step
+    , testCaseSteps "Replay - FastTimedCPM - 6 nodes" $ \step ->
+        withTempRocksDb "replay-test-fasttimedcpm-pair-rocks" $ \rdb ->
+        withSystemTempDirectory "replay-test-fasttimedcpm-pair-pact" $ \pactDbDir ->
+        Chainweb.Test.MultiNode.replayTest loglevel (fastForkingCpmTestVersion pairChainGraph) 6 rdb pactDbDir step
+    , testCaseSteps "compact-resume" $ \step ->
+        withTempRocksDb "compact-resume-test-rocks" $ \rdb ->
+        withSystemTempDirectory "compact-resume-test-pact" $ \pactDbDir -> do
+        Chainweb.Test.MultiNode.compactAndResumeTest loglevel (fastForkingCpmTestVersion pairChainGraph) 6 rdb pactDbDir step
```
chessai, line 56: try this with singletonChainGraph?
```suggestion
        Chainweb.Test.MultiNode.compactAndResumeTest loglevel (fastForkingCpmTestVersion singletonChainGraph) 6 rdb pactDbDir step
```

PR #1770, opened 2023-11-02
Instant-forking CPM version for testing

edmundnoble:
This is a new `Chainweb.Version` which enables all forks immediately, similar to fast-devnet, but for testing. Lots of tests would prefer using this over `fastForkingCpmTestVersion`. However, doing this has revealed an odd race in `RemotePactTest`. I include a failed attempt at a mitigation here, which is to wait until warp has begun serving the service and P2P APIs.


src/Chainweb/Chainweb.hs
```diff
@@ -634,15 +635,23 @@ mkThrottler e rate c = initThrottler (defaultThrottleSettings $ TimeSpec (ceilin
 -- -------------------------------------------------------------------------- --
 -- Run Chainweb
 
+data NowServing = NowServing
```
chessai, line 638: :nail_care:

src/Chainweb/Version.hs
```diff
@@ -305,6 +305,9 @@ data Upgrade = Upgrade
     deriving stock (Generic, Eq)
     deriving anyclass (NFData)
 
+instance Show Upgrade where
+    show _ = "<upgrade>"
```
chessai, line 309: why can (or should) this not derive Show?

edmundnoble: Upgrade transactions are *massive blobs* of base64-encoded text and are probably not useful. Maybe we should actually label them with names instead to get a decent show...

edmundnoble:
Closed in favor of #1866

PR #1775, opened 2023-11-03
service: return correct tx hash in txindex inner join

### LINKED ISSUE #1732: /poll endpoint omits results
When fetching multiple tx results at once using the `/poll` endpoint some of the request keys are ignored even though they are valid. I am able to retrieve the result for those missing if I use the same endpoint and pass a single request key instead of a list of them.
For example here's a list of request keys:
```["InhUsF-zHMF7tygkkh7jrm1Xns2GpnUvPjlq014MRKU",
    "1zirlzrAQi4Zlzew5t8LF1jTtkom6YZS9qT-K9IEzHI",
    "Y5y2vaI-Fwd-XGgWnUYgUSSgUBC5A_CwJQlUWioob6c",
    "uO8BGneKvtk-xCSECn5trsS7cO4YlDQhs4GBCDN_P08",
    "BiTXUV56wcCJNUJSdmtJHQsyFsMTNKbD2phaloaWK9Q",
    "S64pHyPTw98z3icozlmMeqMCiUzgXaZ9jWTrERgEsaM",
    "y3Loj83E3fNndRTTTwow_4Rqd6v0IsvaNjE-2WoiEwA",
    "D1pfnRODkhQ9q9tmoTl5m7qSM_TYkpp5SC8UVe3j02k",
    "i1KbgndGUxHR76DaPC02_Up5O-WWR0PJdw5-KloFaXM",
    "NY3DvKdQVRT_Yz-3j76Qfofp8BIm6kgUQrZIbK5czWY",
    "HLJseaI1KwaDLT1J9b8Nggfek1YEQcuPl4EoNvIMBI4",
    "Cq4eZajyo0MeGfhz_2TpFx3KsqJF4BtGdhSKHkHOmHI",
    "MdYI2HrYztPbpnAN5BXCT4IRw10XqcIU1qAPLC0l9Eo",
    "6vv0rPLrasYpC-iQQzrgIGcvcN-bTCNvJNBjeLN6EL8"]```

I'm getting results only for these:

```
   ["BiTXUV56wcCJNUJSdmtJHQsyFsMTNKbD2phaloaWK9Q",
    "6vv0rPLrasYpC-iQQzrgIGcvcN-bTCNvJNBjeLN6EL8",
    "Y5y2vaI-Fwd-XGgWnUYgUSSgUBC5A_CwJQlUWioob6c",
    "y3Loj83E3fNndRTTTwow_4Rqd6v0IsvaNjE-2WoiEwA",
    "S64pHyPTw98z3icozlmMeqMCiUzgXaZ9jWTrERgEsaM",
    "InhUsF-zHMF7tygkkh7jrm1Xns2GpnUvPjlq014MRKU",
    "HLJseaI1KwaDLT1J9b8Nggfek1YEQcuPl4EoNvIMBI4",
    "uO8BGneKvtk-xCSECn5trsS7cO4YlDQhs4GBCDN_P08"]
   ```
I suspect this is a regression. Tested on 2.19, 2.20, same issue on both versions.

thoughtpolice:
Summary: The INNER JOIN used to return the result set when looking up a transaction hash was `zip`'d without taking into account the ordering of the result set. This in turn meant that tx ID's got mapped to the wrong block IDs, resulting in lookup failures.

This could be reproduced with the following minimal `/poll` query with only two request keys:

    {"requestKeys": [
      "1zirlzrAQi4Zlzew5t8LF1jTtkom6YZS9qT-K9IEzHI",
      "MdYI2HrYztPbpnAN5BXCT4IRw10XqcIU1qAPLC0l9Eo"
    ]}

which would return the empty JSON object.

Aside from all of that, the error handling for `internalPoll` is rather poor as it exists purely within `MaybeT`, giving no chance of reporting any meaningful errors. (I threw away a small cleanup for this since it was too debugging oriented, and needs to go through another round of review.)

Closes #1732.

Change-Id: I3e6f1a98a5d867125c54cac8d47d01ba

### DIFF
--- a/src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
+++ b/src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
@@ -339,19 +339,21 @@ doLookupSuccessful dbenv confDepth hashes = runBlockEnv dbenv $ do
           blockheightval = maybe [] (\bh -> [SInt bh]) blockheight
           qvals = [ SBlob (BS.fromShort hash) | (TypedHash hash) <- V.toList hashes ] ++ blockheightval
 
-        qry db qtext qvals [RInt, RBlob] >>= mapM go
-      return $ HashMap.fromList (zip (V.toList hashes) r)
+        qry db qtext qvals [RInt, RBlob, RBlob] >>= mapM go
+      return $ HashMap.fromList (map (\(T3 blockheight blockhash txhash) -> (txhash, T2 blockheight blockhash)) r)
   where
-    qtext = "SELECT blockheight, hash FROM \
+    qtext = "SELECT blockheight, hash, txhash FROM \
             \TransactionIndex INNER JOIN BlockHistory \
             \USING (blockheight) WHERE txhash IN (" <> hashesParams <> ")"
             <> maybe "" (const " AND blockheight <= ?") confDepth
             <> ";"
     hashesParams = Utf8 $ intercalate "," [ "?" | _ <- V.toList hashes]
 
-    go ((SInt h):(SBlob blob):_) = do
-        !hsh <- either fail return $ runGetEitherS decodeBlockHash blob
-        return $! T2 (fromIntegral h) hsh
+    go :: [SType] -> IO (T3 BlockHeight BlockHash PactHash)
+    go (SInt blockheight:SBlob blockhash:SBlob txhash:_) = do
+        !blockhash' <- either fail return $ runGetEitherS decodeBlockHash blockhash
+        let !txhash' = TypedHash $ BS.toShort txhash
+        return $! T3 (fromIntegral blockheight) blockhash' txhash'
     go _ = fail "impossible"
 
 doGetBlockHistory :: Db logger -> BlockHeader -> Domain RowKey RowData -> IO BlockTxHistory
--- a/test/Chainweb/Test/Pact/RemotePactTest.hs
+++ b/test/Chainweb/Test/Pact/RemotePactTest.hs
@@ -177,6 +177,9 @@ tests rdb = testGroup "Chainweb.Test.Pact.RemotePactTest"
                 , after AllSucceed "poll bad key test" $
                     testCaseSteps "local preflight sim test" $ \step ->
                         join $ localPreflightSimTest <$> iot <*> cenv <*> pure step
+                , after AllSucceed "poll returns correct results" $
+                    testCaseSteps "poll correct results test" $ \step ->
+                        join $ pollingCorrectResults <$> iot <*> cenv <*> pure step
                 ]
     ]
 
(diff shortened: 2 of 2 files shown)


src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -339,19 +339,23 @@ doLookupSuccessful dbenv confDepth hashes = runBlockEnv dbenv $ do
           blockheightval = maybe [] (\bh -> [SInt bh]) blockheight
           qvals = [ SBlob (BS.fromShort hash) | (TypedHash hash) <- V.toList hashes ] ++ blockheightval
 
-        qry db qtext qvals [RInt, RBlob] >>= mapM go
-      return $ HashMap.fromList (zip (V.toList hashes) r)
+        qry db qtext qvals [RInt, RBlob, RBlob] >>= mapM go
+        -- HERE
```
thoughtpolice, line 343: pair programming snafu

emilypi, line 343: stale debug comment

emilypi (commented):
Can we get a regression for the ordering here?

edmundnoble (commented):
Approved pending test.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -339,19 +339,22 @@ doLookupSuccessful dbenv confDepth hashes = runBlockEnv dbenv $ do
...
     hashesParams = Utf8 $ intercalate "," [ "?" | _ <- V.toList hashes]
 
-    go ((SInt h):(SBlob blob):_) = do
-        !hsh <- either fail return $ runGetEitherS decodeBlockHash blob
-        return $! T2 (fromIntegral h) hsh
+    go :: [SType] -> IO (T3 BlockHeight BlockHash PactHash)
+    go (SInt blockheight:SBlob blockhash:SBlob txhash:_) = do
+        !blockhash' <- either fail return $ runGetEitherS decodeBlockHash blockhash
+        let !txhash' = TypedHash $ BS.toShort txhash
+        print (blockheight, blockhash', txhash')
```
ak3n, line 356: debugging leftover?

ak3n (approved):
LGTM 👍

thoughtpolice:
Now with a regression test that confirms the bug, and some refactorings on top.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -274,6 +277,42 @@ pollingConfirmDepth t cenv step = do
...
+    (tx, tx')  = ("42", "43")
+
+    stepTx transaction = do
+      buildTextCmd
+        $ set cbSigners [mkSigner' sender00 []]
+        $ set cbCreationTime t
+        $ set cbNetworkId (Just v)
+        $ set cbGasLimit 70_000
+        $ mkCmd "nonce-cont-2"
+        $ mkExec' transaction
```
emilypi, line 314: This is really clear :+1:

emilypi (approved):

PR #1780, opened 2023-11-15
compaction test txlogs

chessai:
---
  - https://app.asana.com/0/0/1205378300254932


test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -135,6 +147,12 @@ tests rdb = testGroup "Chainweb.Test.Pact.RemotePactTest"
         withResource' getCurrentTimeIntegral $ \(iotm :: IO (Time Micros)) ->
             let cenv = _getServiceClientEnv <$> net
                 iot = toTxCreationTime <$> iotm
+                pactDir = do
+                  m <- _getNodeDbDirs <$> net
+                  case M.lookup 0 m of
```
chessai, line 152: This looks up the pactDbDir for node 0. This is kind of a hack, because there is only one node in this test. However, it doesn't matter much, because we are dealing with both submitting `/local` txs and compaction, so picking an arbitrary node to run these two operations on is fine.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -93,11 +106,10 @@ import Chainweb.Version
 import Chainweb.Version.Mainnet
 import Chainweb.Storage.Table.RocksDB
 
-
 -- -------------------------------------------------------------------------- --
 -- Global Settings
 
-nNodes :: Natural
+nNodes :: Word
```
chessai, line 112: you'll see a few similar changes throughout this PR. there's no reason for this to be a `Natural`. we're never going to run greater than `maxBound :: Word` nodes on a single machine, or even come close to approaching it.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -186,10 +207,228 @@ tests rdb = testGroup "Chainweb.Test.Pact.RemotePactTest"
...
+                [ "(namespace 'free)"
+                , "(module m" <> sshow n <> " G"
+                , "  (defcap G () true)"
+                , "  (defun test-write (id name age) (m0.write-persons id name age))"
+                , ")"
+                , "(test-write \"C\" \"chessai\" 69)"
+                ]
+          buildTextCmd
+            $ set cbSigners [mkSigner' sender00 []]
+            $ set cbGasLimit 100_000 --300_000
```
chessai, line 362:
```suggestion
            $ set cbGasLimit 100_000
```
delete comment

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -948,5 +1187,5 @@ pactDeadBeef = let (TransactionHash b) = deadbeef
                in RequestKey $ Hash b
 
 -- avoiding `scientific` dep here
-getBlockHeight :: CommandResult a -> Maybe Word64
-getBlockHeight = preview (crMetaData . _Just . key "blockHeight" . _Number . to ((fromIntegral :: Integer -> Word64 ) . round . toRational))
+crGetBlockHeight :: CommandResult a -> Maybe Word64
```
chessai, line 1190: `getBlockHeight` is a bit ambiguously named. Get it from what?
The `cr` prefix makes things a bit clearer, IMO.
I made this change initially because I had defined a function for testing called `getBlockHeight`, and was surprised to see a conflict for a function doing something entirely different.

test/Chainweb/Test/Utils.hs
```diff
@@ -1043,6 +1054,33 @@ node testLabel rdb rawLogger peerInfoVar conf nid = do
         crs = map snd $ HashMap.toList $ view chainwebChains cw
         poison cr = mempoolAddToBadList (view chainResMempool cr) (V.singleton deadbeef)
 
+withDbDirs :: Word -> ResourceT IO (Map Word (FilePath, FilePath))
```
chessai, line 1057: the code to create+destroy the temporary dirs was copied mostly from [System.IO.Temp.withSystemTempDirectory](https://hackage.haskell.org/package/temporary-1.3/docs/src/System.IO.Temp.html#withSystemTempDirectory), but we needed a Map version.

thoughtpolice (approved):
readability: lgtm

test/Chainweb/Test/Utils.hs
```diff
@@ -1043,6 +1054,33 @@ node testLabel rdb rawLogger peerInfoVar conf nid = do
...
+      destroy m = flip foldMap m $ \(d1, d2) -> do
+        ignoringIOErrors $ do
+          removeDirectoryRecursive d1
+          removeDirectoryRecursive d2
+
+  (_, m) <- allocate create destroy
+  pure m
+
+ignoringIOErrors :: (MonadCatch m) => m () -> m ()
+ignoringIOErrors ioe = ioe `catch` (\e -> const (return ()) (e :: IOError))
```
chessai, line 1082:
```suggestion
ignoringIOErrors ioe = ioe `catch` (\e -> const (pure ()) (e :: IOError))
```
use pure instead of `return`

test/Chainweb/Test/Utils.hs
```diff
@@ -992,19 +999,22 @@ runTestNodes
...
-        let baseConf = config ver n
-        conf <- if
-            | i == 0 ->
-                return $ bootstrapConfig baseConf
-            | otherwise ->
-                setBootstrapPeerInfo <$> (fst <$> readMVar portMVar) <*> pure baseConf
-        node testLabel rdb logger portMVar conf i
+runTestNodes testLabel rdb logger ver n portMVar dbDirs = do
+    let nids = [0 .. n - 1]
+    let nidWithDirs = mapMaybe (\nid -> (nid,) <$> Map.lookup nid dbDirs) nids
```
edmundnoble, line 1009: Throw an error if there is a missing dbDir instead of using `mapMaybe`, we want to be sure that we're running as many nodes as we requested. Probably better even, you could instead pass in dbDirs as a list and stop passing in `n`, with `n` implicitly set to `length dbDirs`.

test/Chainweb/Test/Utils.hs
```diff
@@ -1013,28 +1023,29 @@ node
...
-    -> Int
+    -> FilePath
+       -- ^ pact db dir
+    -> FilePath
+       -- ^ rocksdb dir
+    -> Word
         -- ^ Unique Node Id. The node id 0 is used for the bootstrap node
     -> IO ()
-node testLabel rdb rawLogger peerInfoVar conf nid = do
+node testLabel rdb rawLogger peerInfoVar conf pactDbDir rocksDbDir nid = do
```
edmundnoble, line 1033: Idk if this is the best PR to do it, but I think `withChainweb` should not take these directories as an input. `ChainwebConfiguration` contains a database directory, and that should be the base directory for all databases. That further means that `node` stops needing to take them as input because it takes a `ChainwebConfiguration` as well.

chessai: yeah, i don't think this PR should do it. we should make a ticket.

test/Chainweb/Test/Pact/RemotePactTest.hs
chessai: I have no idea. I think this test suite (RemotePactTest and some others) were originally meant to be flexible enough to run multiple nodes, but they don't. AFAIK only the slow-tests (MultiNode.hs) actually run multiple nodes.

I don't want to make the change in this PR though.

test/Chainweb/Test/Pact/RemotePactTest.hs
```diff
@@ -186,10 +207,228 @@ tests rdb = testGroup "Chainweb.Test.Pact.RemotePactTest"
...
+          buildTextCmd
+            $ set cbSigners [mkSigner' sender00 []]
+            $ set cbGasLimit 400_000
+            $ set cbTTL defaultMaxTTL
+            $ set cbCreationTime t
+            $ set cbChainId cid
+            $ set cbNetworkId (Just v)
+            $ mkCmd "test-write"
+            $ mkExec txTxt
+            $ mkKeySetData "sender00" [sender00]
```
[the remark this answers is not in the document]
chessai: yeah, i defined a helper to deduplicate. the main differences are the gas limit, tx text, and nonce.

chainweb.cabal
```diff
@@ -509,7 +509,7 @@ test-suite chainweb-tests
 
     build-depends:
         -- internal
-          chainweb
+        , chainweb
```
edmundnoble, line 512: oof

chessai: i prefer leading commas for easy re-arranging (eg with vim's `:sort u`), but i can undo this. originally it was because i added a temporary dependency above this

PR #1785, opened 2023-11-28
add pact-breakdown tool

chessai:
This PR adds a POC cwtool (subcommand `pact-breakdown`) that takes a pact database directory and outputs a report breaking down the pact state in extreme detail. This includes size information of each table, pact value, module, etc. It also includes interesting things like "how many nulls are there, and what kind of null are they? (e.g. null info)". It also includes how much size index tables take up. Here's a prefix of the output:
```
{
  "size": "5.76 GB",
  "null_counts": {
    "bool": 463922,
    "chain_id": 61562,
    "def_body": 146,
    "docs": 1998,
    "imports": 468,
    "info": 260023,
    "meta": 2088,
    "module_hash": 785,
    "module_namespace": 232003,
    "pact_exec": 36178909,
    "yield": 280
  },
  "table_type_sizes": {
    "pact_sys": 3.99 GB,
    "user": 1.61 GB,
    "ix": 584.16 MB
  },
```
It's missing some of the size info. I haven't done a tally of what all is missing. This tool isn't critical and it's not clear if we actually need this level of granularity. Putting a pin in this for now.

---
  - https://app.asana.com/0/0/1205752706866360


chessai:
This is never meant to be merged really, I think.

PR #1788, opened 2023-11-29
cutdb: allow forcing a cut through the cut pipeline

thoughtpolice:
Summary: To introduce orphan blocks to the devnet, we need to circumvent the ordinary cut pipeline and insert into the underlying `TVar` directly.

This largely gets things in place by just changing the types to layer in the needed information.

Change-Id: Id7981a7308cd669a8d48c4443a6c88c0

### DIFF
--- a/src/Chainweb/CutDB.hs
+++ b/src/Chainweb/CutDB.hs
@@ -17,6 +17,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
 
 -- |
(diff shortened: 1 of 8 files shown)


larskuhtz:
I don't think we should add logic to the cut pipeline just for testing certain network conditions in devnet.

Orphans originate from network latencies and network partitions both of which can be easily simulated in devnet by manipulating the virtual devnet (e.g. using iptables or similar techniques). This would also more faithfully represent the real conditions under which orphans occur.

If orphans have to be generate on a single node for testing, it can be done by stopping a mining node and resetting the cut when restarting the node. To a client this would appear as a reorg.

Overall we should try to keep critical production logic free of cases that only serve testing purposes whenever possible.

PR #1790, opened 2023-12-04
poll endpoint: refactorings

thoughtpolice:
Resubmitting now that the codefreeze is over.


edmundnoble (approved):

edmundnoble (approved):
I like particularly how currentHeightQ moves next to the `qry` call and query params to avoid accidentally mismatching them.

PR #1791, opened 2023-12-13
write cuts more often

chessai:
---
  - https://app.asana.com/0/0/1206164297538869


edmundnoble:
This should fix the issue of nodes taking too long to start because they're rewinding to the latest stored cut.

edmundnoble (approved):

PR #1802, opened 2024-01-23
Read-only checkpointer + checkpointer interface change

edmundnoble:
This changes the checkpointer to allow read-only rewinds, and makes its interface somewhat safer.
<details>
<summary>Background</summary>

The checkpointer does three things: firstly, it *restores* the Pact state to an ancestor block of the user's choice, deleting any newer state and providing them with a writable snapshot of the Pact state at that block. Once *restored*, the user can *save* any changes made to that snapshot back to the database as a new block. These changes can also be *discarded*.

Note that to extend the chain from the current block, one first restores the checkpointer to the current block, writes to the Pact state snapshot, and then saves the results.

```hs
_cpRestore :: Maybe (BlockHeight, ParentHash) -> IO PactDbEnv
      -- ^ prerequisite: (BlockHeight - 1, ParentHash) is a direct ancestor of
      -- the "latest block"

_cpSave :: BlockHash -> IO ()

_cpDiscard :: IO ()
```
`_cpRestore` rewinds the checkpointer to wherever it's told (the "target") providing a database snapshot at that target, `PactDbEnv`. The input can also be `Nothing`, providing a database snapshot before the genesis block has been run. Writes to that database snapshot can be committed to the database as a brand new block using `_cpSave`, or discarded using `_cpDiscard`. 

In practice, we have wrapped these operations in a function:
```hs
-- | INTERNAL FUNCTION. USE WITH CAUTION!
--
-- Same as 'withCheckpointer' but doesn't rewinds the checkpointer state to
-- the provided target. Note that it still restores the state to the target.
--
-- In other words, this method can move the checkpointer back in time to a state
-- in the current history. It doesn't resolve forks or fast forwards the
-- checkpointer.
--
-- /Assumption:/
--
-- This function assumes that '_psParentHeader' has been updated to match the
-- latest block in the checkpointers. This is guaranteed to be the case after
-- calling any of 'rewindTo', 'syncParentHeader', 'withCheckpointerRewind',
-- 'withCheckPointerWithoutRewind', or 'withCurrentCheckpointer'.
--
-- /NOTE:/
--
--
-- Only use this function when
--
-- 1. you need the extra performance from skipping the call to 'rewindTo' and
--
withCheckpointerWithoutRewind
    :: (HasCallStack, CanReadablePayloadCas tbl, Logger logger)
    => Maybe ParentHeader
        -- ^ block height and hash of the parent header
    -> (PactDbEnv -> PactServiceM logger tbl (WithCheckpointerResult a))
    -> PactServiceM logger tbl a
withCheckpointerWithoutRewind target act = ...
```
The `WithCheckpointerResult a` type returned by the callback given to `withCheckpointerWithoutRewind` is:
```hs
-- | When using the checkpointer for evaluating pact code the caller must
-- indicate whether the result of the evaluation should be persisted or
-- discarded.
--
data WithCheckpointerResult a
    = Discard !a
    | Save BlockHeader !a
```
`withCheckpointerWithoutRewind` calls `_cpDiscard` in the case of an error, and we further wrap these calls with `withBatch`, another function that calls into the checkpointer to create a sqlite transaction, reverting the `_cpRestore` rewind on discard or in case of an error, making the checkpointer transactional. Using the database snapshot restored by `_cpRestore` after either saving or discarding is illegal. 

All uses of the checkpointer in the codebase use `withCheckpointerWithoutRewind`, directly or indirectly, and wrapped in `withBatch`, except for unit tests.

The checkpointer is used by PactService when servicing its requests. These requests fall into two categories:
 - writes, which call `_cpSave`, adding new blocks to the Pact state. These eventually call `withCheckpointerWithoutRewind` with a callback that returns `Save bh a`. There are two: firstly, `SyncToBlock`, which is used to force the Pact state to the latest cut on startup, incidentally the way Pact replays work. Secondly, `ValidateBlock`, which is used to extend the Pact chain with new blocks during catch-up.
 - reads, which call `_cpDiscard`, and only use the checkpointer to read from the Pact state. These eventually call `withCheckpointerWithoutRewind` with a callback that returns `Discard a`. These are most notably `NewBlock`, which executes a new block of transactions to be mined, `LookupPactTxs` which polls a transaction's request key to see what its result was, `Local`, which evaluates Pact code (potentially at some block in the past) for its result value, and `PreInsertCheck`, which does some preliminary checks of a transaction to see if it's "valid enough" to include in the mempool.
</details>

<details>

<summary>Implementation struggles</summary>
The straightforward way to get a read-only rewind is to create a new kind of database snapshot which has a block height limit built-in, and can't observe data written after that block height. Plumbing that into the existing checkpointer entails adding a variation of `_cpRestore` which would provide a rewound `PactDbEnv`, but which was illegal to `_cpSave`, because writing into the past doesn't make sense. Evgenii implemented most of this rewound `PactDbEnv`, and worked on adding it to the existing checkpointer, and this gave him a way to iterate on the rewound `PactDbEnv`.

This extension to the already unwieldy interface of the checkpointer, which is easy to use unsafely, makes it more confusing. We needed a new flavor of `withCheckpointerWithoutRewind`, too, as well as the things that call it, but only the ones that made sense assuming we would never call `_cpSave`. Furthermore, as the comment to `withCheckpointerWithoutRewind` hints at, there is more state than just the checkpointer involved here, contained in PactService, which needs to be managed on its own in a way that doesn't interfere with read-write calls to the checkpointer.

</details>

<details>
<summary>Redesign</summary>
Every call to `_cpRestore` must be followed by either `_cpSave` or `_cpDiscard`, but the read-only rewind makes the read-write checkpointer's `_cpDiscard` mostly useless. The checkpointer is also not safe to access outside of a sqlite transaction, so let's make the new `_cpRestore` take a callback instead, which returns the hash of the new block for saving to the database, and any value that the user wants. Here's a sketch:

```hs
data Save a = Save a BlockHash
_cpRestoreAndSave :: forall a. Maybe (BlockHeight, ParentHash) -> (PactDbEnv -> IO (Save a)) -> IO a 
-- ^ prerequisite: (BlockHeight - 1, ParentHash) is a direct ancestor of
 -- the "latest block"
```

One issue with this is that with for example in `SyncToBlock`, we play multiple blocks, but we only use one sqlite transaction. That's to ensure that if we fail we don't touch the database, and also for performance. The same happens in `ValidateBlock`; if we have to validate a block and this block is on a different fork from us, we have to first rewind to the common ancestor of the two branches, then play all of the blocks on the other fork up until the parent of the block we're validating. If that block fails to validate, we don't want to remain on that new fork, we want to be back on the branch we started on. So it's critical that we can play multiple block callbacks in a single `_cpRestore` call, which will be transactional:
```hs
_cpRestoreAndSave :: forall a. Maybe (BlockHeight, ParentHash) -> [PactDbEnv -> IO (Save a)] -> IO [a]
```
In practice, each "block" here will do almost the same thing, so we can factor it out like this:
```hs
_cpRestoreAndSave :: Maybe (BlockHeight, ParentHash) -> [blk] -> (PactDbEnv -> IO (Save a)) -> IO [a]
```
And as an optimization we can use an arbitrary monoid for output rather than a list:
```hs
_cpRestoreAndSave :: Maybe (BlockHeight, ParentHash) -> [blk] -> (PactDbEnv -> IO (Save q)) -> IO q
```
Practically, we get this block data from `TreeDB`, which provides us a `Stream`, so we should use that here for performance:
```hs
_cpRestoreAndSave :: Maybe (BlockHeight, ParentHash) -> Stream (Of blk) IO r -> (PactDbEnv -> IO (Save q)) -> IO (q, r)
```
And here we're in decent shape. We can give a similar treatment to the read-only version of `_cpRestore`, noting that it will never be used with `_cpSave`, and never needs to be used with multiple blocks at once:
```hs
_cpReadFrom :: Maybe (BlockHeight, ParentHash) -> (PactDbEnv -> IO a) -> IO a
```
This is the redesign of the checkpointer's interface. There are multiple knock-on effects this has elsewhere in PactService which I continue with.
</details>

<details>
<summary>Knock-on effects to PactServiceM</summary

That allowed us to delete `withCheckpointerWithoutRewind`, but we're not done. As mentioned earlier, `withCheckpointerWithoutRewind` was merely a member of a family of functions of varying safety doing different things to PactService. Let's look at the monad PactService runs in:
```hs
newtype PactServiceM logger tbl a = PactServiceM
  { _unPactServiceM ::
       ReaderT (PactServiceEnv logger tbl) (StateT PactServiceState IO) a
  }
data PactServiceEnv logger tbl = PactServiceEnv
    { _psMempoolAccess :: !(Maybe MemPoolAccess)
    , _psCheckpointer :: !(Checkpointer logger)
    , _psPdb :: !(PayloadDb tbl)
    , _psGasModel :: !(TxContext -> GasModel)
    , _psLocalRewindDepthLimit :: !RewindLimit
    -- ^ The limit of rewind's depth in the `execLocal` command.
    , _psPreInsertCheckTimeout :: !Micros
    -- ^ Maximum allowed execution time for the transactions validation.
    , _psReorgLimit :: !RewindLimit
    , _psOnFatalError :: !(forall a. PactException -> Text -> IO a)
    , _psVersion :: !ChainwebVersion
    , _psAllowReadsInLocal :: !Bool
    , _psLogger :: !logger
    , _psGasLogger :: !(Maybe logger)

    -- The following two fields are used to enforce invariants for using the
    -- checkpointer. These would better be enforced on the type level. But that
    -- would require changing many function signatures and is postponed for now.
    --
    -- DO NOT use these fields if you don't know what they do!
    --
    , _psIsBatch :: !Bool
        -- ^ True when within a `withBatch` or `withDiscardBatch` call.
    , _psCheckpointerDepth :: !Int
        -- ^ Number of nested checkpointer calls
    , _psBlockGasLimit :: !GasLimit
    , _psChainId :: !ChainId
    }
type ModuleInitCache = M.Map BlockHeight ModuleCache

    { _psStateValidated :: !(Maybe BlockHeader)
    , _psInitCache :: !ModuleInitCache
    , _psParentHeader :: !ParentHeader
    , _psSpvSupport :: !SPVSupport
    }    
```
That's quite a mouthful, let's start with `PactServiceEnv`, the reader context. There's a pattern here: `MemPoolAccess`, `Checkpointer`, `PayloadDb`, `BlockHeaderDb`, these are all allowing access to other components of `chainweb-node` The gas model and miner rewards are fixed; they determine block validation behavior. Everything else is either for logging or configuration... excepting those two fields with the scary warning. Respectively, the first is to prevent accidentally nesting checkpointer transactions, which is illegal, and the second is used to ensure nobody accidentally nests `withCheckpointerWithoutRewind` calls in the callback passed to `withCheckpointerWithoutRewind`, which would entail calling `_cpRestore` multiple times before saving or discarding, also illegal. We'll do away with both of these, under the assumption that they're not necessary anymore (if they ever were) with the new, safer interface.

Let's move on to `PactServiceState`. 
`_psStateValidated` is never read from, so it can be safely deleted.
`_psInitCache` is the block-height-indexed module cache, which must remain for compatibility purposes but may eventually be phased out. It contains certain modules (coin, fungible, namespace) which are used by every block, and caches them for fast loading, at their appropriate version for a given block height.
`_psParentHeader` is the "current parent header".
`_psSpvSupport` is the SPV verification implementation in chainweb, as used by Pact. In practice, outside of the tx-simulator, this is constructed from `_psParentHeader` without any time or space cost.

What we want to discuss is `_psParentHeader`. If we want to ask "what is the current parent header?" in PactService, we have two options. 1. which applies when the checkpointer has been restored, is "whatever the checkpointer has been restored to". 2. which only applies when the checkpointer has not been restored, is "whatever the latest block the checkpointer has seen is", i.e. the highest block the checkpointer can be restored to. `_psParentHeader` is *usually* set to whatever block the checkpointer has most recently been restored to, excepting that the checkpointer can be restored to `Nothing`, i.e. no block at all, to run genesis blocks; when running the genesis blocks, the parent header is set to the genesis block header, which is circular and only works for reasons we'll get to later. There is a function `syncParentHeader` whose purpose is setting `_psParentHeader` to whatever block the checkpointer has most recently seen instead, i.e. option 2:
```hs
-- | Synchronizes the parent header with the latest block of the checkpointer.
--
-- Ideally, this function would not be needed. It
--
-- 1. does a best effort to recover from a situation where '_cpGetLatestBlock'
--    '_psParentHeader' got out of sync, and
-- 2. provides visibility into the state of that invariant via the info log
--    message.
--
-- This call would fail if the latest block that is stored in the checkpointer
-- got orphaned and the '_psParentHeader' got reset to some other block
-- /without/ rewinding the checkpointer, too. That must not be possible. Hence,
-- the result of '_cpGetLatestBlock' and '_psParentHeader' must be kept in sync.
--
-- The previously described scenario /is/ possible if the service gets restarted
-- and the latest block got orphaned after the service stopped. This is
-- prevented by rewinding on pact service startup to the latest available header
-- in the block header db.
--
syncParentHeader :: (Logger logger) => PactServiceM logger tbl ParentHeader
```
In other functions we set `_psParentHeader` to other things. It's as the comment says; ideally this function would not be needed. In fact, it's not needed. 

There are two contexts in which operations in PactService execute: either
* they are running on the top level, without a restored checkpointer
* or they are running "in a block", with a restored checkpointer, i.e. in a callback passed to `_cpReadFrom` or `_cpRestoreAndSave`.
Operations running on the top level can only ask for one type of "parent header", the latest header seen by the checkpointer. Operations running "in a block" on the other hand have a more meaningful notion of "parent header," whatever the checkpointer was restored to, *and* a meaningful notion of "the current Pact state", accessible as a `PactDbEnv`. To witness this, we can add a new monad `PactBlockM`, with an environment containing the parent header and database snapshot:
```hs
data PactBlockEnv logger tbl = PactBlockEnv
  { _pactServiceEnv :: PactServiceEnv logger tbl
  , _pactParentContext :: !ParentContext
  , _pactDbEnv :: !(CurrentBlockDbEnv logger)
  }

newtype PactBlockM logger tbl a = PactBlockM
  { _unPactBlockM ::
       ReaderT (PactBlockEnv logger tbl) (StateT PactServiceState IO) a
  }

    { _psInitCache :: !ModuleInitCache
    }

-- | Run 'PactBlockM' by providing the block context, in the form of
-- a database snapshot at that block and information about the parent header.
-- It is unsafe to use this function in an argument to `liftPactServiceM`.
runPactBlockM
    :: ParentHeader -> CurrentBlockDbEnv logger
    -> PactBlockM logger tbl a -> PactServiceM logger tbl a

-- | Lifts PactServiceM to PactBlockM by forgetting about the current block.
-- It is unsafe to use `runPactBlockM` inside the argument to this function.
liftPactServiceM :: PactServiceM logger tbl a -> PactBlockM logger tbl 
```

This pattern makes it easy to ensure the parent header is correct every time, and also makes it easier to access the database through the `PactDbEnv` which no longer needs to be passed around manually, though it is still possible to nest checkpointer calls by accident. We also amend `_cpRestoreAndSave` to allow it access to the parent header:
```hs
_cpRestoreAndSave :: Maybe (BlockHeight, ParentHash) -> Stream (Of blk) IO r -> (PactDbEnv -> ParentHeader -> IO (Save q)) -> IO (q, r)
```

</details>

<details>
<summary>Knock-on effects to ParentHeader usage</summary>
I mentioned earlier that there is no ParentHeader available if the checkpointer is restored to before-genesis. This is an important case which we use when both validating and generating the genesis blocks. We validate the genesis blocks when we start a node with an empty database, and we generate genesis blocks when we add new chains and as an integration test to ensure that their outputs haven't changed. `_psParentHeader` was potentially problematic because it allowed access to a parent header even when one might not exist. But there was no bug caused by this. Why?

Executing Pact transactions requires this information from the parent header:
- the block height at which the Pact code is executing, to set fork flags. We set this to the height of the parent header plus one.
- the hash of the parent block, which it runs SPV proofs against. 
- the creation time of the parent block, to avoid running expired transactions.
When we set the parent header to the genesis header, and run the genesis block, the block height is one higher than the genesis block. That isn't going to activate any different forks compared to the genesis height. There are also no cross-chain transactions being completed in genesis payloads, so SPV proofs don't matter. And the parent of the genesis block, if it existed, would have the same creation time as the genesis block anyway. Thus, the behavior is identical.

This is still very confusing, and preferably, Pact would not have to use the genesis headers as parents for themselves, nor fabricate fake parent headers for the genesis headers. So, to service Pact, we create this datatype:
```hs
-- Context for evaluating Pact transactions, which comes from the parent header.
-- Note that this isn't actually a parent header, because genesis blocks do not
-- have parents, but genesis blocks must be evaluated in a context.
data ParentContext
  = BlockParentContext !ChainwebVersion !ChainId !BlockHash !BlockHeight !BlockCreationTime
  | GenesisParentContext !ChainwebVersion !ChainId
```
The version and chain ID are included for convenience, but not strictly necessary in most contexts, because they are contained in `PactServiceEnv`. We can calculate the "current height" as follows:
```hs
parentContextCurrentHeight :: ParentContext -> BlockHeight
parentContextCurrentHeight (BlockParentContext _ _ _ bh _) = bh + 1
parentContextCurrentHeight (GenesisParentContext v cid) = genesisHeight v ci
```
This change percolates upwards to the checkpointer:
```hs
_cpRestoreAndSave :: ParentContext -> Stream (Of blk) IO r -> (PactDbEnv -> ParentContext -> IO (q, ParentContext)) -> IO (q, r)
```
</details>
<details>
<summary>Knock-on effects to NewBlock</summary>
The NewBlock call to PactService takes two parameters: a miner, to be credited the rewards for the block, and a parent block header, on top of which to make the new block. As NewBlock is a read, it can only do a read-only rewind. What are we to do if NewBlock is told to make a block on top of a header which is on a fork? Or in the future relative to the checkpointer, but stored in the block DB? It can only fail, if it's doing a read-only rewind, because we have no `_cpReadFrom` that can play multiple blocks at once, though we could add it. We could also break our rule from before, using `_cpRestoreAndSave` but failing with an exception, but that's quite nasty and does require making writes to sqlite even if they're discarded.

Let's take a step back: why does NewBlock take a parent block header? The miner wants a block on top of the latest block header chosen by consensus, to maximize the chances of its new block being chosen by consensus. Effectively, assuming that the checkpointer's latest block is the latest block chosen by consensus, the miner would like to extend the checkpointer's chain with its new block. We don't do exactly that for safety reasons I won't elaborate on here, but it remains the case that the parent the miner wants is the same block as the checkpointer has at its tip. To do that, the miner includes the latest header it's seen as the parent when it makes a NewBlock request. Instead, we can make NewBlock use the latest header the checkpointer has seen as the parent, and return it back to the miner.

</details>

<details>
<summary>Knock-on effects to LookupPactTxs</summary>

LookupPactTxs takes an optional parameter which is a block header to rewind to. We never use it, so it's deleted. It also takes a confirmation depth parameter which is now implemented using `_cpReadFrom` and reading from the "nth latest block header".

</details>

That's most of this PR, but there are some things I didn't show, like the new implementations of PactService functionality on top of the new checkpointer interface. I also deleted the TxSimulator in the hopes that we don't need it, but that will need to change eventually, because it does have at least use, which is Pact typechecking. It does some weird stuff with SPV support which I didn't initially see how to adapt, but it's probably possible.


edmundnoble:
Closed in favor of #1803

PR #1803, opened 2024-01-23
 Read-only checkpointer + checkpointer interface change

edmundnoble:
This changes the checkpointer to allow read-only rewinds, and makes its interface somewhat safer. The motivation is to allow for most Pact requests to be serviced in parallel, and to make replays themselves parallel, as well as faster, by not requiring an initial rewind.
<details>
<summary>Background</summary>

The checkpointer does three things: firstly, it *restores* the Pact state to an ancestor block of the user's choice, deleting any newer state and providing them with a writable snapshot of the Pact state at that block. Once *restored*, the user can *save* any changes made to that snapshot back to the database as a new block. These changes can also be *discarded*.

Note that to extend the chain from the current block, one first restores the checkpointer to the current block, writes to the Pact state snapshot, and then saves the results. To rewind the chain to a past block, one restores the checkpointer to that block and then discards the state snapshot.

```hs
_cpRestore :: Maybe (BlockHeight, ParentHash) -> IO PactDbEnv
      -- ^ prerequisite: (BlockHeight - 1, ParentHash) is a direct ancestor of
      -- the "latest block"

_cpSave :: BlockHash -> IO ()

_cpDiscard :: IO ()
```

`_cpRestore` rewinds the checkpointer to wherever it's told (the "target") providing a database snapshot at that target, `PactDbEnv`. The input can also be `Nothing`, providing a database snapshot before the genesis block has been run. Writes to that database snapshot can be committed to the database as a brand new block using `_cpSave`, or discarded using `_cpDiscard`. Using the database snapshot restored by `_cpRestore` after either saving or discarding is illegal.

In practice, we have wrapped these operations in a function:
```hs
-- | INTERNAL FUNCTION. USE WITH CAUTION!
--
-- Same as 'withCheckpointer' but doesn't rewinds the checkpointer state to
-- the provided target. Note that it still restores the state to the target.
--
-- In other words, this method can move the checkpointer back in time to a state
-- in the current history. It doesn't resolve forks or fast forwards the
-- checkpointer.
--
-- /Assumption:/
--
-- This function assumes that '_psParentHeader' has been updated to match the
-- latest block in the checkpointers. This is guaranteed to be the case after
-- calling any of 'rewindTo', 'syncParentHeader', 'withCheckpointerRewind',
-- 'withCheckPointerWithoutRewind', or 'withCurrentCheckpointer'.
--
-- /NOTE:/
--
--
-- Only use this function when
--
-- 1. you need the extra performance from skipping the call to 'rewindTo' and
--
withCheckpointerWithoutRewind
    :: (HasCallStack, CanReadablePayloadCas tbl, Logger logger)
    => Maybe ParentHeader
        -- ^ block height and hash of the parent header
    -> (PactDbEnv -> PactServiceM logger tbl (WithCheckpointerResult a))
    -> PactServiceM logger tbl a
withCheckpointerWithoutRewind target act = ...
```
The `WithCheckpointerResult a` type returned by the callback given to `withCheckpointerWithoutRewind` is:
```hs
-- | When using the checkpointer for evaluating pact code the caller must
-- indicate whether the result of the evaluation should be persisted or
-- discarded.
--
data WithCheckpointerResult a
    = Discard !a
    | Save BlockHeader !a
```
`withCheckpointerWithoutRewind` calls `_cpRestore` to acquire a database snapshot at a given block, calls `_cpDiscard` in the case of an error, and uses the `WithCheckpointerResult a` value to both decide whether to discard or save the snapshot and to allow the user to extract a value using reads from the snapshot. We further wrap calls to `withCheckpointerWithoutRewind` with `withBatch`, another function that calls into the checkpointer to create a sqlite transaction, reverting the `_cpRestore` rewind on discard or in case of an error, making the checkpointer fully transactional.

All uses of the checkpointer in the codebase use `withCheckpointerWithoutRewind`, directly or indirectly, and wrapped in `withBatch`, except for unit tests.

The checkpointer is used by PactService when servicing its requests. These requests fall into two categories:
 - writes, which call `_cpSave`, adding new blocks to the Pact state. These eventually call `withCheckpointerWithoutRewind` with a callback that returns `Save bh a`. There are two: firstly, `SyncToBlock`, which is used to force the Pact state to the latest cut on startup. This is used to implement Pact replays, via deleting the Pact state first and keeping the cut database. Secondly, `ValidateBlock`, which is used to extend the Pact chain with blocks during catch-up.
 - reads, which call `_cpDiscard`, and only use the checkpointer to read from the Pact state. These eventually call `withCheckpointerWithoutRewind` with a callback that returns `Discard a`. These are most notably `NewBlock`, which executes a new block of transactions to be mined, `LookupPactTxs` which polls a transaction's request key to see what its result was, `Local`, which evaluates Pact code (potentially at some block in the past) for its result value, and `PreInsertCheck`, which does some preliminary checks of a transaction to see if it's "valid enough" to include in the mempool.
</details>

<details>

<summary>Process</summary>
Each database row in the Pact state contains a txid indicating the transaction in which it was written. txids monotonically increase and are associated via block heights via an index in the checkpointer. The straightforward way to get a read-only rewind is to create a new kind of database snapshot which can't observe data written after the rewind target block. Plumbing that into the existing checkpointer entails adding a variation of `_cpRestore` which would provide a rewound `PactDbEnv`, but which would be illegal to `_cpSave`, because writing into the past wouldn't be used. Evgenii implemented most of this rewound `PactDbEnv`, and worked on adding it to the existing checkpointer, and this gave him a way to iterate on the rewound `PactDbEnv`.

This extension to the already unwieldy interface of the checkpointer, which is easy to use unsafely, makes it more confusing. We needed a new flavor of `withCheckpointerWithoutRewind`, too, as well as the things that call it, but only the ones that made sense assuming we would never call `_cpSave`. Furthermore, as the comment to `withCheckpointerWithoutRewind` hints at, there is more state than just the checkpointer involved here, contained in PactService, which needs to be managed on its own in a way that doesn't interfere with read-write calls to the checkpointer.

</details>

<details>
<summary>Redesign</summary>
Every call to `_cpRestore` must be followed by either `_cpSave` or `_cpDiscard`, but the read-only rewind makes the read-write checkpointer's `_cpDiscard` mostly useless. The checkpointer is also not safe to access outside of a sqlite transaction, so let's make the new `_cpRestore` take a callback instead, which returns the hash of the new block for saving to the database, and any value that the user wants. Here's a sketch:

```hs
data Save a = Save a BlockHash
_cpRestoreAndSave :: forall a. Maybe (BlockHeight, ParentHash) -> (PactDbEnv -> IO (Save a)) -> IO a 
-- ^ prerequisite: (BlockHeight - 1, ParentHash) is a direct ancestor of
 -- the "latest block"
```

One issue with this is that with for example in `SyncToBlock`, we play multiple blocks, but we only use one sqlite transaction. That's to ensure that if we fail we don't touch the database, and also for performance. The same happens in `ValidateBlock`; if we have to validate a block and this block is on a different fork from us, we have to first rewind to the common ancestor of the two branches, then play all of the blocks on the other fork up until the parent of the block we're validating. If that block fails to validate, we don't want to remain on that new fork, we want to be back on the branch we started on. So it's critical that we can play multiple block callbacks in a single `_cpRestore` call, which will be transactional:
```hs
_cpRestoreAndSave :: forall a. Maybe (BlockHeight, ParentHash) -> [PactDbEnv -> IO (Save a)] -> IO [a]
```
In practice, each "block" here will do almost the same thing, so we can factor it out like this:
```hs
_cpRestoreAndSave :: Maybe (BlockHeight, ParentHash) -> [blk] -> (PactDbEnv -> blk -> IO (Save a)) -> IO [a]
```
And as an optimization we can use an arbitrary monoid for output rather than a list:
```hs
_cpRestoreAndSave :: forall q blk. Monoid q => Maybe (BlockHeight, ParentHash) -> [blk] -> (PactDbEnv -> blk -> IO (Save q)) -> IO q
```
Practically, we get this block data from `TreeDB`, which provides us a `Stream`, so we should use that here for performance:
```hs
_cpRestoreAndSave :: forall q blk. Monoid q => Maybe (BlockHeight, ParentHash) -> Stream (Of blk) IO r -> (PactDbEnv -> blk -> IO (Save q)) -> IO (q, r)
```
And here we're in decent shape. We can give a similar treatment to the read-only version of `_cpRestore`, noting that it will never be used with `_cpSave`, and never needs to be used with multiple blocks at once:
```hs
_cpReadFrom :: Maybe (BlockHeight, ParentHash) -> (PactDbEnv -> IO a) -> IO a
```
This is the redesign of the checkpointer's interface. There are multiple knock-on effects this has elsewhere in PactService which I continue with.
</details>

<details>
<summary>Knock-on effects to PactServiceM</summary

That allowed us to delete `withCheckpointerWithoutRewind`, but we're not done. As mentioned earlier, `withCheckpointerWithoutRewind` was merely a member of a family of functions of varying safety doing different things to PactService. Let's look at the monad PactService runs in:
```hs
newtype PactServiceM logger tbl a = PactServiceM
  { _unPactServiceM ::
       ReaderT (PactServiceEnv logger tbl) (StateT PactServiceState IO) a
  }
data PactServiceEnv logger tbl = PactServiceEnv
    { _psMempoolAccess :: !(Maybe MemPoolAccess)
    , _psCheckpointer :: !(Checkpointer logger)
    , _psPdb :: !(PayloadDb tbl)
    , _psGasModel :: !(TxContext -> GasModel)
    , _psLocalRewindDepthLimit :: !RewindLimit
    -- ^ The limit of rewind's depth in the `execLocal` command.
    , _psPreInsertCheckTimeout :: !Micros
    -- ^ Maximum allowed execution time for the transactions validation.
    , _psReorgLimit :: !RewindLimit
    , _psOnFatalError :: !(forall a. PactException -> Text -> IO a)
    , _psVersion :: !ChainwebVersion
    , _psAllowReadsInLocal :: !Bool
    , _psLogger :: !logger
    , _psGasLogger :: !(Maybe logger)

    -- The following two fields are used to enforce invariants for using the
    -- checkpointer. These would better be enforced on the type level. But that
    -- would require changing many function signatures and is postponed for now.
    --
    -- DO NOT use these fields if you don't know what they do!
    --
    , _psIsBatch :: !Bool
        -- ^ True when within a `withBatch` or `withDiscardBatch` call.
    , _psCheckpointerDepth :: !Int
        -- ^ Number of nested checkpointer calls
    , _psBlockGasLimit :: !GasLimit
    , _psChainId :: !ChainId
    }
type ModuleInitCache = M.Map BlockHeight ModuleCache

    { _psStateValidated :: !(Maybe BlockHeader)
    , _psInitCache :: !ModuleInitCache
    , _psParentHeader :: !ParentHeader
    , _psSpvSupport :: !SPVSupport
    }    
```
That's quite a mouthful, let's start with `PactServiceEnv`, the reader context. There's a pattern here: `MemPoolAccess`, `Checkpointer`, `PayloadDb`, `BlockHeaderDb`, these are all allowing access to other components of `chainweb-node` The gas model and miner rewards are fixed; they determine block validation behavior. Everything else is either for logging or configuration... excepting those two fields with the scary warning. Respectively, the first is to prevent accidentally nesting checkpointer transactions, which is illegal, and the second is used to ensure nobody accidentally nests `withCheckpointerWithoutRewind` calls in the callback passed to `withCheckpointerWithoutRewind`, which would entail calling `_cpRestore` multiple times before saving or discarding, also illegal. We'll do away with both of these, under the assumption that they're not necessary anymore (if they ever were) with the new, safer interface.

Let's move on to `PactServiceState`. 
`_psStateValidated` is never read from, so it can be safely deleted.
`_psInitCache` is the block-height-indexed module cache, which must remain for compatibility purposes but may eventually be phased out. It contains certain modules (coin, fungible, namespace) which are used by every block, and caches them for fast loading, at their appropriate version for a given block height.
`_psParentHeader` is the "current parent header".
`_psSpvSupport` is the SPV verification implementation in chainweb, as used by Pact. In practice, outside of the tx-simulator, this is constructed from `_psParentHeader` without any time or space cost.

What we want to discuss is `_psParentHeader`. If we want to ask "what is the current parent header?" in PactService, we have two options. 1. which applies when the checkpointer has been restored, is "whatever the checkpointer has been restored to". 2. which only applies when the checkpointer has not been restored, is "whatever the latest block the checkpointer has seen is", i.e. the highest block the checkpointer can be restored to. `_psParentHeader` is *usually* set to whatever block the checkpointer has most recently been restored to, excepting that the checkpointer can be restored to `Nothing`, i.e. no block at all, to run genesis blocks; when running the genesis blocks, the parent header is set to the genesis block header, which is circular and only works for reasons we'll get to later. There is a function `syncParentHeader` whose purpose is setting `_psParentHeader` to whatever block the checkpointer has most recently seen instead, i.e. option 2:
```hs
-- | Synchronizes the parent header with the latest block of the checkpointer.
--
-- Ideally, this function would not be needed. It
--
-- 1. does a best effort to recover from a situation where '_cpGetLatestBlock'
--    '_psParentHeader' got out of sync, and
-- 2. provides visibility into the state of that invariant via the info log
--    message.
--
-- This call would fail if the latest block that is stored in the checkpointer
-- got orphaned and the '_psParentHeader' got reset to some other block
-- /without/ rewinding the checkpointer, too. That must not be possible. Hence,
-- the result of '_cpGetLatestBlock' and '_psParentHeader' must be kept in sync.
--
-- The previously described scenario /is/ possible if the service gets restarted
-- and the latest block got orphaned after the service stopped. This is
-- prevented by rewinding on pact service startup to the latest available header
-- in the block header db.
--
syncParentHeader :: (Logger logger) => PactServiceM logger tbl ParentHeader
```
In other functions we set `_psParentHeader` to other things. It's as the comment says; ideally this function would not be needed. In fact, it's not needed. 

There are two contexts in which operations in PactService execute: either
* they are running on the top level, without a restored checkpointer
* or they are running "in a block", with a restored checkpointer, i.e. in a callback passed to `_cpReadFrom` or `_cpRestoreAndSave`.
Operations running on the top level can only ask for one type of "parent header", the latest header seen by the checkpointer. Operations running "in a block" on the other hand have a more meaningful notion of "parent header," whatever the checkpointer was restored to, *and* a meaningful notion of "the current Pact state", accessible as a `PactDbEnv`. To witness this, we can add a new monad `PactBlockM`, with an environment containing the parent header and database snapshot:
```hs
data PactBlockEnv logger tbl = PactBlockEnv
  { _pactServiceEnv :: PactServiceEnv logger tbl
  , _pactParentContext :: !ParentContext
  , _pactDbEnv :: !(CurrentBlockDbEnv logger)
  }

newtype PactBlockM logger tbl a = PactBlockM
  { _unPactBlockM ::
       ReaderT (PactBlockEnv logger tbl) (StateT PactServiceState IO) a
  }

    { _psInitCache :: !ModuleInitCache
    }

-- | Run 'PactBlockM' by providing the block context, in the form of
-- a database snapshot at that block and information about the parent header.
-- It is unsafe to use this function in an argument to `liftPactServiceM`.
runPactBlockM
    :: ParentHeader -> CurrentBlockDbEnv logger
    -> PactBlockM logger tbl a -> PactServiceM logger tbl a

-- | Lifts PactServiceM to PactBlockM by forgetting about the current block.
-- It is unsafe to use `runPactBlockM` inside the argument to this function.
liftPactServiceM :: PactServiceM logger tbl a -> PactBlockM logger tbl 
```

This pattern makes it easy to ensure the parent header is correct every time, and also makes it easier to access the database through the `PactDbEnv` which no longer needs to be passed around manually, though it is still possible to nest checkpointer calls by accident. We also amend `_cpRestoreAndSave` to allow it access to the parent header:
```hs
_cpRestoreAndSave :: Maybe (BlockHeight, ParentHash) -> Stream (Of blk) IO r -> (PactDbEnv -> ParentHeader -> IO (Save q)) -> IO (q, r)
```

</details>

<details>
<summary>Knock-on effects to ParentHeader usage</summary>
I mentioned earlier that there is no ParentHeader available if the checkpointer is restored to before-genesis. This is an important case which we use when both validating and generating the genesis blocks. We validate the genesis blocks when we start a node with an empty database, and we generate genesis blocks when we add new chains and as an integration test to ensure that their outputs haven't changed. `_psParentHeader` was potentially problematic because it allowed access to a parent header even when one might not exist. But there was no bug caused by this. Why?

Executing Pact transactions requires this information from the parent header:
- the block height at which the Pact code is executing, to set fork flags. We set this to the height of the parent header plus one.
- the hash of the parent block, which it runs SPV proofs against. 
- the creation time of the parent block, to avoid running expired transactions.
When we set the parent header to the genesis header, and run the genesis block, the block height is one higher than the genesis block. That isn't going to activate any different forks compared to the genesis height. There are also no cross-chain transactions being completed in genesis payloads, so SPV proofs don't matter. And the parent of the genesis block, if it existed, would have the same creation time as the genesis block anyway. Thus, the behavior is identical.

This is still very confusing, and preferably, Pact would not have to use the genesis headers as parents for themselves, nor fabricate fake parent headers for the genesis headers. So, to service Pact, we create this datatype:
```hs
-- Context for evaluating Pact transactions, which comes from the parent header.
-- Note that this isn't actually a parent header, because genesis blocks do not
-- have parents, but genesis blocks must be evaluated in a context.
data ParentContext
  = BlockParentContext !ChainwebVersion !ChainId !BlockHash !BlockHeight !BlockCreationTime
  | GenesisParentContext !ChainwebVersion !ChainId
```
The version and chain ID are included for convenience, but not strictly necessary in most contexts, because they are contained in `PactServiceEnv`. We can calculate the "current height" as follows:
```hs
parentContextCurrentHeight :: ParentContext -> BlockHeight
parentContextCurrentHeight (BlockParentContext _ _ _ bh _) = bh + 1
parentContextCurrentHeight (GenesisParentContext v cid) = genesisHeight v ci
```
This change percolates upwards to the checkpointer:
```hs
_cpRestoreAndSave :: ParentContext -> Stream (Of blk) IO r -> (PactDbEnv -> ParentContext -> IO (q, ParentContext)) -> IO (q, r)
```
</details>
<details>
<summary>Knock-on effects to NewBlock</summary>
The NewBlock call to PactService takes two parameters: a miner, to be credited the rewards for the block, and a parent block header, on top of which to make the new block. As NewBlock is a read, it can only do a read-only rewind. What are we to do if NewBlock is told to make a block on top of a header which is on a fork? Or in the future relative to the checkpointer, but stored in the block DB? It can only fail, if it's doing a read-only rewind, because we have no `_cpReadFrom` that can play multiple blocks at once, though we could add it. We could also break our rule from before, using `_cpRestoreAndSave` but failing with an exception, but that's quite nasty and does require making writes to sqlite even if they're discarded.

Let's take a step back: why does NewBlock take a parent block header? The miner wants a block on top of the latest block header chosen by consensus, to maximize the chances of its new block being chosen by consensus. Effectively, assuming that the checkpointer's latest block is the latest block chosen by consensus, the miner would like to extend the checkpointer's chain with its new block. We don't do exactly that for safety reasons I won't elaborate on here, but it remains the case that the parent the miner wants is the same block as the checkpointer has at its tip. To do that, the miner includes the latest header it's seen as the parent when it makes a NewBlock request. Instead, we can make NewBlock use the latest header the checkpointer has seen as the parent, and return it back to the miner.

</details>

<details>
<summary>Knock-on effects to LookupPactTxs</summary>

LookupPactTxs takes an optional parameter which is a block header to rewind to. We never use it, so it's deleted. It also takes a confirmation depth parameter which is now implemented using `_cpReadFrom` and reading from the "nth latest block header".

</details>

That's most of this PR, but there are some things I didn't show, like the new implementations of PactService functionality on top of the new checkpointer interface. I also deleted the TxSimulator in the hopes that we don't need it, but that will need to change eventually, because it does have at least use, which is Pact typechecking. It does some weird stuff with SPV support which I didn't initially see how to adapt, but it's probably possible.

---
  - https://app.asana.com/0/0/1205075354172793


src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -284,43 +310,132 @@ newtype BlockHandler logger p a = BlockHandler
...
         )
 
-newtype PactDbEnv' logger = PactDbEnv' (PactDbEnv (BlockEnv logger SQLiteEnv))
+type ChainwebPactDbEnv logger = PactDbEnv (BlockEnv logger SQLiteEnv)
 
 type ParentHash = BlockHash
 
+-- Context for evaluating Pact transactions, which comes from the parent header.
+-- Note that this isn't actually a parent header, because genesis blocks do not
+-- have parents, but genesis blocks must be evaluated in a context.
```
imalsogreg, line 319: The current wording confuses me a little because of the transition from "which comes from the parent header" to "this isn't actually a parent header.".

Is it still accurate if worded this way?

```
-- | Context for evaluating a Pact transaction.
data ParentContext =
  = BlockParentContext !CWVersion !Cid !BlockHash
  -- ^ The context has a parent, referenced by the given BlockHash.
  | GenesisParentContext !CwVersion !Cid
  -- ^ The context has no parent, the block under evaluation is therefore the genesis block.
```

src/Chainweb/Chainweb/MinerResources.hs
```diff
@@ -89,13 +90,21 @@ withMiningCoordination logger conf cdb inner
     | otherwise = do
         cut <- _cut cdb
         t <- newTVarIO mempty
-        m <- initialPayloads cut >>= newTVarIO
+        initialPw <- fmap (PrimedWork . HM.fromList) $
+            forM miners $ \miner ->
+                let mid = view minerId miner
+                in fmap ((mid,) . HM.fromList) $
+                    forM cids $ \cid -> do
+                        let bh = fromMaybe (genesisBlockHeader v cid) (HM.lookup cid (_cutMap cut))
```
imalsogreg, line 98: If some `cid` isn't in `_cutMap cut`, should that be treated as at error?

bench/Chainweb/Pact/Backend/Bench.hs
```diff
@@ -57,8 +60,33 @@ import Chainweb.Utils.Bench
...
+  :: (Monoid q)
+  => Checkpointer logger
+  -> ParentContext
+  -> [(ParentContext, ChainwebPactDbEnv logger -> IO q)]
+  -> IO q
+cpRestoreAndSave cp pc blks = snd <$> _cpRestoreAndSave cp pc
+  (traverse Stream.yield [0..length blks - 1])
+  (\dbenv _pc i -> fmap (,fst (blks !! i)) $ snd (blks !! i) (_cpPactDbEnv dbenv))
+
+childOf :: ParentContext -> BlockHash -> ParentContext
```
imalsogreg, line 85: Just an observation, not necessarily a recommendation to change something,

I notice when I read `ParentContext` I keep thinking "the context of the parent", when really the meaning is "the parent value available in the current block's context").

By contrast, `ContextParent` makes me immediately think "the parent value available in the current block's context".

imalsogreg:
> This changes the checkpointer to allow read-only rewinds

Checking my understanding - "read-only" checkpointer means that rewinds are read-only, but after rewind, the checkpointer remains read-write?

And also: deletion-based rewinding does not continue to live side-by-side with read-only rewind; rewind is completely replaced by the read-only implementation, is that right?

src/Chainweb/Pact/Backend/Types.hs
edmundnoble: I think this is a legal transformation to make given what I wrote, but that's because what I wrote is bad. Lemme think about this for a bit.

src/Chainweb/Chainweb/MinerResources.hs
edmundnoble: `_cutMap cut` is the blocks in the current cut. When we add new chains, that's done at some future height, so the current cut won't contain those chains when the node starts.

bench/Chainweb/Pact/Backend/Bench.hs
edmundnoble: This is fair, this and the earlier comment make me consider renaming `ParentContext` to `ParentHeaderInfo` or something.

edmundnoble:
> > This changes the checkpointer to allow read-only rewinds
> 
> Checking my understanding - "read-only" checkpointer means that rewinds are read-only, but after rewind, the checkpointer remains read-write?
> 
> And also: deletion-based rewinding does not continue to live side-by-side with read-only rewind; rewind is completely replaced by the read-only implementation, is that right?

No, we retain read-write rewinds. Read-only rewinds are only used for "query" requests to PactService, which do not write anything to the database, so a read-only rewind is never followed by a write.

src/Chainweb/Pact/Backend/Compaction.hs
```diff
@@ -456,7 +456,7 @@ dropNewTables bh = do
 --   properly pruning RocksDB.
 compactSystemTables :: BlockHeight -> CompactM ()
 compactSystemTables bh = do
-  let systemTables = ["BlockHistory", "VersionedTableMutation", "VersionedTableCreation"]
+  let systemTables = ["BlockHistory", "VersionedTableMutation"]
```
chessai, line 459: the `column` variable is no longer needed - for every table in the loop, the column name is just "blockheight".

edmundnoble:
Note: maybe best to just return `m (q, BlockHash)` from `cpRestoreAndSave`? It's confusing to me that it's legal to extend the chain with multiple *new* blocks despite there being no way those blocks were mined.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -313,39 +315,23 @@ doGetBlockParent v cid dbenv (bh, hash)
...
-                  <> maybe "" (const " AND blockheight <= ?") confDepth
-                  <> ";"
-          qvals 
+                  <> " AND blockheight <= ?;"
+          qvals
             -- match query params above. first, hashes
             = map (\(TypedHash h) -> SBlob $ BS.fromShort h) hss
-            -- then, the block depth, if any
-            ++ maybe [] (\bh -> [SInt bh]) blockheight
+            -- then, the block height
```
thoughtpolice, line 333: Maybe add a comment the `-1`. Or not. I think it's fine and this makes sense after thinking about it

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -118,12 +128,29 @@ getPendingData = do
 forModuleNameFix :: (Bool -> BlockHandler logger e a) -> BlockHandler logger e a
 forModuleNameFix f = use bsModuleNameFix >>= f
 
+-- TODO: speed this up, cache it?
+tableExists :: Utf8 -> BlockHeight -> BlockHandler logger SQLiteEnv Bool
```
thoughtpolice, line 132: `tableExistsInDbAtHeight`

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -137,7 +164,9 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
     (Utf8 tableNameBS) = tableName
 
     queryStmt =
-        "SELECT rowdata FROM " <> tbl tableName <> " WHERE rowkey = ? ORDER BY txid DESC LIMIT 1;"
+        "SELECT rowdata FROM " <> tbl tableName <> " WHERE rowkey = ?" <> blockLimitStmt <> " ORDER BY txid DESC LIMIT 1;"
```
thoughtpolice, line 167: Move `queryStmt` and `blockLimitStmt` together and near the actual `blockLimitParam` call; this just makes all the interpolated variables easier to line up and read.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -352,13 +393,21 @@ doKeys d = do
...
+        m <- runMaybeT $ checkDbTablePendingCreation $ Utf8 tnS
         case m of
             Nothing -> return mempty
             Just () -> do
+                case mlim of
+                    Nothing -> return ()
+                    Just (bh, _) -> do
+                        exists <- tableExists tn bh
+                        unless exists $ -- TODO: try to throw a DbError more cleanly. comment this.
+                            void $ callDb "doKeys" $ \db -> qry db "garbage query" [] []
```
thoughtpolice, line 408: Pull this out since it happens in a few places; `failIfTableDoesNotExist` or something like that

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -457,11 +506,20 @@ doCreateUserTable tn@(TableName ttxt) mn = do
...
-          _ -> False
+        case r of
+          [[SText rname]] ->
+            case mbh of
+                -- if lowercase matching, no need to check equality
+                -- (wasn't needed before either but leaving alone for replay)
+                Nothing -> return (lcTables || rname == t)
+                Just bh -> do
+                    qry db
+                        "SELECT tablename FROM VersionedTableCreation WHERE createBlockheight <= ? AND lower(tablename) = lower(?);"
```
thoughtpolice, line 517: This could be unified with the query in `tableExists`, and the `<` vs `<=` distinction isn't needed here with `-1`.

thoughtpolice: Also, this could use a comment as to why the table names need to be lower cased; which is briefly "because in the past they were case sensitive, and now they're not."

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -679,134 +737,100 @@ createVersionedTable tablename db = do
...
+-- a ParentContext.
+prepareToPlayBlock
+  :: T.Text
+  -> ChainwebVersion
+  -> ChainId
+  -> Maybe ParentHeader
+  -> BlockHandler logger SQLiteEnv ()
+prepareToPlayBlock msg v cid mph = do
+  let currentHeight = maybe (genesisHeight v cid) (succ . _blockHeight . _parentHeader) mph
+  assign bsBlockHeight currentHeight
```
thoughtpolice, line 750: `assign` was really weird to read since it's actually a getter...

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -679,134 +737,100 @@ createVersionedTable tablename db = do
...
+  bsModuleNameFix .= enableModuleNameFix v cid currentHeight
+  bsSortedKeys .= pact42 v cid currentHeight
+  bsLowerCaseTables .= chainweb217Pact v cid currentHeight
+  txid <- getEndTxId (msg <> ".prepareToPlayBlock") mph
+  assign bsTxId txid
```
thoughtpolice, line 755: same

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -679,134 +737,100 @@ createVersionedTable tablename db = do
...
+  clearPendingTxState
+
+
+-- | delete any state from the database newer than the input ParentContext.
```
thoughtpolice, line 759: `ParentContext` -> "parent header"

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -679,134 +737,100 @@ createVersionedTable tablename db = do
...
+      tblNames <- qry_ db "SELECT tablename FROM VersionedTableCreation;" [RText]
+      forM_ tblNames $ \t -> case t of
+        [SText tn] -> exec_ db ("DROP TABLE [" <> tn <> "];")
+        _ -> internalError "Something went wrong when resetting tables."
+      exec_ db "DELETE FROM VersionedTableCreation;"
+      exec_ db "DELETE FROM VersionedTableMutation;"
+      exec_ db "DELETE FROM TransactionIndex;"
+rewindDbTo (Just (ParentHeader ph)) = do
+    let currentHeight = succ (_blockHeight ph)
+    assign bsBlockHeight currentHeight
```
thoughtpolice, line 780: `assign` -> `.=`

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -679,134 +737,100 @@ createVersionedTable tablename db = do
...
+      exec_ db "DELETE FROM [SYS:Modules];"
+      exec_ db "DELETE FROM [SYS:Namespaces];"
+      exec_ db "DELETE FROM [SYS:Pacts];"
+      tblNames <- qry_ db "SELECT tablename FROM VersionedTableCreation;" [RText]
+      forM_ tblNames $ \t -> case t of
+        [SText tn] -> exec_ db ("DROP TABLE [" <> tn <> "];")
+        _ -> internalError "Something went wrong when resetting tables."
+      exec_ db "DELETE FROM VersionedTableCreation;"
+      exec_ db "DELETE FROM VersionedTableMutation;"
+      exec_ db "DELETE FROM TransactionIndex;"
```
thoughtpolice, line 777: We need to assign `bsTxId` and `bsBlockHeight` in this case both to match the previous behavior, but it should also be more robust.

thoughtpolice (changes requested):
Part 1 of group review. We'll go back and do `PactService` and testing, etc tomorrow.

src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -149,379 +86,89 @@ exitOnRewindLimitExceeded = handle $ \case
...
-    :: forall logger tbl
-    . (HasCallStack, CanReadablePayloadCas tbl, Logger logger)
-    => (ParentHeader, BlockHeader)
-    -> PactServiceM logger tbl ()
-fastForward (target, block) = do
-    -- This does a restore, i.e. it rewinds the checkpointer back in
-    -- history, if needed.
-    withCheckpointerWithoutRewind (Just target) "fastForward" $ \pdbenv -> do
-        payloadDb <- asks _psPdb
-        payload <- liftIO $ tableLookup payloadDb bpHash >>= \case
```
ak3n, line 423: We need a test in master that covers fastForward scenario and then check that the test works fine in this branch — the only way to check we don't break anything here.

src/Chainweb/Pact/PactService.hs
```diff
@@ -785,23 +687,83 @@ execValidateBlock
...
+        -- for the purpose of the rewind limit check, pre-genesis and genesis
+        -- are at the same height.
+        let !currHeight = maybe (genesisHeight v cid) _blockHeight currHeader
+        let !ancestorHeight = maybe (genesisHeight v cid) _blockHeight rewindTarget
+        let !rewindLimitSatisfied = ancestorHeight + fromIntegral reorgLimit >= currHeight
+        unless rewindLimitSatisfied $
+            throwM $ RewindLimitExceeded
+                (RewindLimit reorgLimit)
+                currHeader
+                rewindTarget
```
ak3n, line 724: I propose to surround such assertions with empty lines for better readability.

test/Chainweb/Test/Pact/PactSingleChainTest.hs
```diff
@@ -619,35 +613,6 @@ assertSender00Bal bal msg hist =
...
-  where
-
-    chainDataMemPool = mempty {
-      mpaGetBlock = \_ _ _ _ bh -> do
-          fmap V.singleton $ buildCwCmd (sshow bh) testVersion
-              $ signSender00
-              $ setFromHeader bh
-              $ set cbRPC (mkExec' "(chain-data)")
-              $ defaultCmd
-      }
```
ak3n, line 649: This test triggers `fastForward`, it seems we need to adapt it or an alternative?

edmundnoble: For some reason I thought this test didn't apply anymore, let me think.

edmundnoble: Yeah this test is testing that `newBlock` with a parent header on another fork works. Now you can't give `newBlock` a parent header at all, so I figured it's not necessary.

src/Chainweb/Pact/PactService/Checkpointer.hs
edmundnoble: I basically inlined `fastForward` and its friends into its callers.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -114,127 +114,148 @@ initRelationalCheckpointer'
...
+  latestHeader <- liftIO $ doGetLatestBlock newDbEnv
+  let
+    -- is the parent the latest header, i.e., can we get away without rewinding?
+    parentIsLatestHeader = case (latestHeader, parent) of
+      (Nothing, Nothing) -> True
+      (Just (_, latestHash), Just (ParentHeader ph)) ->
+        _blockHash ph == latestHash
+      _ -> False
+
+  (`finally` liftIO (runBlockEnv newDbEnv (rollbackSavepoint BatchSavepoint >> commitSavepoint BatchSavepoint))) $ do
```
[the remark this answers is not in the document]
edmundnoble: Fair.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
edmundnoble: I don't have the block height here unfortunately. Let's just rely on `prepareToPlayBlock` setting these and delete the setting here.

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -213,192 +227,213 @@ checkpointerTest name relational cenvIO = testCaseSteps name $ \next -> do
...
+    restoreAndSave (Just pc04)
+      [(pc05Fork, \dbEnv -> do
+        void $ runExec cenv dbEnv (Just $ ksData "5") $ defModule "5"
+        runExec cenv dbEnv Nothing "(m5.readTbl)" >>= \EvalResult{..} -> Right _erOutput @?= traverse toPactValue [tIntList [1]]
+        )]
 
     next "mini-regression test for dropping user tables (part 4) fork on the empty block"
-    _blockenv04Fork <- _cpRestore cp (Just (BlockHeight 4, hash13))
-    _cpDiscard cp
+    readFrom (Just pc13) $ \_ -> return ()
```
thoughtpolice, line 324: during review we found this was redundant, delete it

test/Chainweb/Test/Pact/Checkpointer.hs
```diff
@@ -213,192 +227,213 @@ checkpointerTest name relational cenvIO = testCaseSteps name $ \next -> do
...
-        void $ runExec cenv blockEnv4b Nothing "(m7.insertTbl 'd 3)"
-        void $ runExec cenv blockEnv4b Nothing "(m7.insertTbl 'c 4)"
-        expectException $ runExec cenv blockEnv4b Nothing "(keys m7.tbl)" >>= \EvalResult{..} -> Right _erOutput @?= traverse toPactValue [tStringList $ T.words "a b c d"]
-        _cpDiscard cp
+        readFrom (Just pc13) $ \dbEnv -> do
+          void $ runExec cenv dbEnv (Just $ ksData "7") (defModule "7")
+          void $ runExec cenv dbEnv Nothing "(m7.insertTbl 'b 2)"
+          void $ runExec cenv dbEnv Nothing "(m7.insertTbl 'd 3)"
+          void $ runExec cenv dbEnv Nothing "(m7.insertTbl 'c 4)"
+          expectException $ runExec cenv dbEnv Nothing "(keys m7.tbl)" >>= \EvalResult{..} -> Right _erOutput @?= traverse toPactValue [tStringList $ T.words "a b c d"]
```
thoughtpolice, line 436: this should just assert the false case, not fail and catch that exception

src/Chainweb/Pact/Backend/Compaction.hs
thoughtpolice: ping on this

thoughtpolice (changes requested):
group reviews, round 2

test/SlowTests.hs
```diff
@@ -31,7 +31,7 @@ loglevel = Warn
 
 -- note that because these tests run in parallel they must all use distinct rocksdb and sqlite dirs.
 suite :: TestTree
-suite = sequentialTestGroup "ChainwebSlowTests" AllSucceed
+suite = sequentialTestGroup "ChainwebSlowTests" AllFinish
```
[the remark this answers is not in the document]
edmundnoble: If one of these tests fails, I don't mind at all if the others continue running.

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -284,43 +311,153 @@ newtype BlockHandler logger p a = BlockHandler
...
+    --
+    -- TODO: Under which circumstances does this return 'Nothing'?
+  , _cpLookupBlockInCheckpointer :: !((BlockHeight, BlockHash) -> IO Bool)
+    -- ^ is the checkpointer aware of the given block?
+  , _cpGetBlockParent :: !((BlockHeight, BlockHash) -> IO (Maybe BlockHash))
+  , _cpGetBlockHistory ::
+      !(BlockHeader -> Domain RowKey RowData -> IO BlockTxHistory)
+  , _cpGetHistoricalLookup ::
+      !(BlockHeader -> Domain RowKey RowData -> RowKey -> IO (Maybe (TxLog RowData)))
+  , _cpReadLogger :: logger
```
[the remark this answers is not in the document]
edmundnoble: It's part of the read-checkpointer.

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -284,43 +299,95 @@ newtype BlockHandler logger p a = BlockHandler
...
+  }
+
+-- | A callback which writes a block's data to the input database snapshot,
+-- and knows its parent header (Nothing if it's a genesis block).
+-- Reports back its own header and some extra value.
+newtype Block logger m a = Block
+  { runBlock :: CurrentBlockDbEnv logger -> Maybe ParentHeader -> m (a, BlockHeader) }
```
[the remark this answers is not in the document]
edmundnoble: I think your name is an improvement but still not perfect, so I'll use it.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -118,12 +128,28 @@ getPendingData = do
 forModuleNameFix :: (Bool -> BlockHandler logger e a) -> BlockHandler logger e a
 forModuleNameFix f = use bsModuleNameFix >>= f
 
+-- TODO: speed this up, cache it?
```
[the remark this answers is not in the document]
edmundnoble: Two possibilities:
- maybe `NewBlock` time is delayed because this is slow. we should avoid using `rewoundPactDb` in most cases when we do `NewBlock` but rocksdb is slightly lagging the checkpointer so this happens more often than it should. 
- maybe read-only replays are slower than they could be because this is slow.

I'm fine with allocating a ticket to fixing those issues, but doing a cache is a bit heavy-duty. I think we get 99% of the way there by fixing the rocksdb race for `NewBlock` and hardcoding `coin_coin-table` into `knownTables` for read-only replay.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -173,9 +201,17 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
     lookupInDb rowkey checkCache = do
         -- First, check: did we create this table during this block? If so,
         -- there's no point in looking up the key.
-        checkDbTableExists tableName
+        checkDbTablePendingCreation tableName
+        case mlim of
+            Nothing -> return ()
```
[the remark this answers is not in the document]
edmundnoble: This is perhaps something to change later, what's happening here is actually that `checkDbTablePendingCreation` is returning `MaybeT (return Nothing)`, halting this entire computation.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -605,11 +659,12 @@ createTransactionIndexTable = callDb "createTransactionIndexTable" $ \db -> do
     exec_ db "CREATE INDEX IF NOT EXISTS \
              \ transactionIndexByBH ON TransactionIndex(blockheight)";
 
+-- register a successful transaction in the pending data for the block
 indexPactTransaction :: BS.ByteString -> BlockHandler logger SQLiteEnv ()
 indexPactTransaction h = modify' $
     over (bsPendingBlock . pendingSuccessfulTxs) $ HashSet.insert h
 
-
+-- commit the index of pending successful transactions to the database
```
thoughtpolice, line 667: haddocking all these would be nice

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -680,117 +735,100 @@ createVersionedTable tablename db = do
...
+        exec' db
+            "DELETE FROM VersionedTableCreation WHERE createBlockheight >= ?"
+            [SInt (fromIntegral bh)]
+        return tbls
+    findTablesToDropStmt =
+      "SELECT tablename FROM VersionedTableCreation WHERE createBlockheight >= ?;"
+    rewindmsg =
+      "rewindBlock:\
+      \ dropTablesAtRewind: \
+      \Couldn't resolve the name of the table to drop."
```
thoughtpolice, line 806: This can just be one line.

src/Chainweb/Pact/Backend/Types.hs
```diff
@@ -284,43 +299,95 @@ newtype BlockHandler logger p a = BlockHandler
...
         )
 
-newtype PactDbEnv' logger = PactDbEnv' (PactDbEnv (BlockEnv logger SQLiteEnv))
+type ChainwebPactDbEnv logger = PactDbEnv (BlockEnv logger SQLiteEnv)
 
 type ParentHash = BlockHash
 
+-- | The parts of the checkpointer that do not mutate the database.
+data ReadCheckpointer logger = ReadCheckpointer
+  { _cpReadFrom ::
```
thoughtpolice, line 308: Just use `IO` here

src/Chainweb/Pact/PactService.hs
```diff
@@ -418,246 +410,176 @@ serviceRequests memPoolAccess reqQ = do
...
+                BlockFill g _ c -> do
+                  (goodLength, badLength) <- liftIO $ (,) <$> Vec.length successes <*> Vec.length failures
+                  liftPactServiceM $ logDebug $ "Block fill: count=" <> sshow c
+                    <> ", gaslimit=" <> sshow g <> ", good="
+                    <> sshow goodLength <> ", bad=" <> sshow badLength
+
+                  -- LOOP INVARIANT: limit absolute recursion count
+                  when (_bfCount bfState > fetchLimit) $
+                    throwM $ MempoolFillFailure $ "Refill fetch limit exceeded (" <> sshow fetchLimit <> ")"
+
```
thoughtpolice, line 519: This should have the same semantics as last time, i.e. return `pure unchanged`

thoughtpolice (changes requested):
Team review, round 3

src/Chainweb/Pact/Backend/Types.hs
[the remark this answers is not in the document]
edmundnoble: Ok I just made it `_cpLogger`

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -173,9 +201,17 @@ doReadRow d k = forModuleNameFix $ \mnFix ->
...
         -- there's no point in looking up the key.
-        checkDbTableExists tableName
+        checkDbTablePendingCreation tableName
+        case mlim of
+            Nothing -> return ()
+            Just (bh, _) -> do
+                existsInDb <- lift $ tableExists tableName bh
+                unless existsInDb $
+                  -- TODO: try to throw a DbError more cleanly. comment this.
+                  void $ callDb "doReadRow" $ \db -> qry db "garbage query" [] []
```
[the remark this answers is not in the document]
edmundnoble: I need "callDb (doReadRow): user error (Database error: ErrorError)" on-chain but using `throwDbError ""` (similarly to your branch) gives ": Failure: Database exception: " (that starting ":" is not a typo). I'll comment this, but I don't think it's worth investigating how exactly in the morass of Pact's error code it comes up with that string to reverse-engineer it and do something cleaner than this.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -114,127 +114,148 @@ initRelationalCheckpointer'
...
+        if parentIsLatestHeader
+        then chainwebPactDb
+        else rewoundPactDb currentHeight txid
+      curBlockDbEnv = CurrentBlockDbEnv
+        { _cpPactDbEnv = PactDbEnv pactDb newDbEnv
+        , _cpRegisterProcessedTx =
+          \(TypedHash hash) -> runBlockEnv newDbEnv (indexPactTransaction $ BS.fromShort hash)
+        , _cpLookupProcessedTx = \hs ->
+          -- TODO: test that we do not see transactions from the
+          -- "current block" if we've rewound.
```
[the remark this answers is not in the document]
edmundnoble: I think this doesn't actually need a further test; we have a test for polling confirmation depth already.

src/Chainweb/Pact/Backend/ChainwebPactDb.hs
```diff
@@ -859,18 +882,25 @@ initSchema = do
...
+      "SELECT endingtxid FROM BlockHistory WHERE blockheight = ? and hash = ?;"
+      [ SInt $ fromIntegral bh
+      , SBlob $ runPutS (encodeBlockHash bhsh)
+      ]
+      [RInt]
+    case r of
+      [[SInt tid]] -> return (TxId (fromIntegral tid))
+      [] -> throwM $ BlockHeaderLookupFailure $ msg <> ".getEndTxId: not in db: " <>
+            sshow (bh, bhsh)
+      _ -> internalError $ msg <> ".getEndTxId: expected single-row int result, got " <> sshow r
```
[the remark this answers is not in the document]
edmundnoble: I don't think is unsafe per se, it'll just fail if you mismatch the hash and height. It needs to be exported because `getBlockHistory` doesn't have a block header for the block it wants the end txid for, just the hash and height.

src/Chainweb/Pact/Backend/Types.hs
edmundnoble: Deleted this type entirely in the end.

bench/Chainweb/Pact/Backend/Bench.hs
edmundnoble: Deleted this type entirely in the end.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -114,127 +114,148 @@ initRelationalCheckpointer'
...
+              "doRestoreAndSave: block with no parent, genesis block, should have genesis height but doesn't,"
+          Just (ParentHeader ph)
+            | succ (_blockHeight ph) /= _blockHeight newBh -> throwM $ PactInternalError $
+              "doRestoreAndSave: non-genesis block should be one higher than its parent. parent at "
+                <> sshow (_blockHeight ph) <> ", child height " <> sshow (_blockHeight newBh)
+          _ -> return ()
+        -- persist any changes to the database
+        liftIO $ runBlockEnv dbenv $
+          commitBlockStateToDatabase (_blockHash newBh)
+        return (m'', Just (ParentHeader newBh))
```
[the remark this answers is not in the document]
edmundnoble: We discussed this offline and @ak3n is satisfied now.

thoughtpolice (approved):
All group review comments finished; some minor nits, but I approve. We'll wait for a second approval from Emily

src/Chainweb/Pact/PactService.hs
```diff
@@ -227,44 +225,42 @@ initializeCoinContract
...
         -- cheap. We could also check the height but that would be redundant.
-        if currentBlockHash /= genesisHash
+        if _blockHash (_parentHeader currentBlockHeader) /= _blockHash genesisHeader
         then do
-          readContracts
+          logger <- view psLogger
+          !mc <- liftIO $ _cpReadFrom (_cpReadCp cp) (Just currentBlockHeader) $ \pdbenv -> do
+            let pd = TxContext currentBlockHeader def
+            readInitModules logger (_cpPactDbEnv pdbenv) pd
+          updateInitCache mc currentBlockHeader
```
emilypi, line 250: This cleans up the need for syncing the parentheader nicely.

src/Chainweb/Pact/PactService.hs
```diff
@@ -418,246 +410,177 @@ serviceRequests memPoolAccess reqQ = do
...
-    refill fetchLimit txTimeLimit pdbenv successes failures = go
+    -> PactServiceM logger tbl (T2 ParentHeader PayloadWithOutputs)
+execNewBlock mpAccess miner = do
+    latestHeader <- findLatestValidBlockHeader
+    T2 latestHeader <$> execNewBlock' latestHeader
+    where
+
+    -- | Note: The BlockHeader param here is the PARENT HEADER of the new
+    -- block-to-be
+    --
```
emilypi, line 426: this is parent because `findLatestValidBlockHeader` will retrieve the latest header from the blockheader db which is by definition the parent block of the current header being built.

src/Chainweb/Pact/PactService.hs
```diff
@@ -418,246 +410,177 @@ serviceRequests memPoolAccess reqQ = do
...
-      mpaProcessFork mpAccess $ _parentHeader parent
-      mpaSetLastHeader mpAccess $ _parentHeader parent
+        handleTimeout :: TxTimeout -> PactBlockM logger cas a
+        handleTimeout (TxTimeout h) = liftPactServiceM $ do
+          logError $ "timed out on " <> sshow h
+          liftIO $ mpaBadlistTx mpAccess (V.singleton h)
+          throwM (TxTimeout h)
+
+        !parentTime =
+          ParentCreationTime (_blockCreationTime $ _parentHeader latestHeader)
```
emilypi, line 483: at some point we should rename `_parentHeader` to `_getHeader` or something - I always doubletake until i'm reminded that it's not accessing the parent header, just the header object from `ParentHeader` the data type.

src/Chainweb/Pact/PactService/Checkpointer.hs
```diff
@@ -149,304 +88,96 @@ exitOnRewindLimitExceeded = handle $ \case
...
+  -> PactServiceM logger tbl a
+readFromNthParent n doRead = do
+    bhdb <- view psBlockHeaderDb
+    ParentHeader latest <- findLatestValidBlockHeader
+    v <- view chainwebVersion
+    cid <- view chainId
+    let parentHeight =
+            -- guarantees that the subtraction doesn't overflow
+            -- this will never give us before genesis
+            fromIntegral $ max (genesisHeight v cid + fromIntegral n) (_blockHeight latest) - fromIntegral n
```
emilypi, line 116: 👍

src/Chainweb/Pact/Types.hs
```diff
@@ -519,51 +510,23 @@ defaultOnFatalError lf pex t = do
 
 type ModuleInitCache = M.Map BlockHeight ModuleCache
 
+data PactBlockEnv logger tbl = PactBlockEnv
+  { _psServiceEnv :: PactServiceEnv logger tbl
```
emilypi, line 514: missing bangpattern, or was this intentional to avoid forcing the service env more than once

emilypi (approved):
Excellent work. I've done all I can to go over it again and again, and i'm still not 100% confident with my knowledge of the parts I don't know, but we have the entire team looking at it, and I trust their reviews as well. Once it's up to date and green, let's get this in.

chessai (approved):

src/Chainweb/Pact/Types.hs
edmundnoble: Nah I can add it with no problem.

src/Chainweb/Pact/PactService.hs
edmundnoble: Agreed.

PR #1804, opened 2024-01-29
Read-only replay

edmundnoble:
With read-only checkpointer, read-only replays are a cinch. `Chainweb.Chainweb` is getting more unwieldy, pending a ticket to make the config interface to chainweb-node nicer, or move replays over to cwtool.

The process of a read-only replay is: for each block with heights in the range (`lower`, `upper`) on our current fork, read-only rewind to that block's parent, and execute the block. It's as easy as that.

Some areas I'm not 100% sure about:
- this was written before read-only checkpointer. Maybe some of this is old-checkpointer-minded and needs to be updated.
- let's do this in parallel, for cases when there are less chains still replaying than there are capabilities. But how? I like the idea of sharing a semaphore or work pool between chains to do this, then using chunks of history as work items to execute in parallel. Of course we could also use `par` if we were feeling particularly cheeky, because that's effectively a set of work threads behind a queue.
- As I lead with, the config interface isn't nice compared to a CLI command like `read-only-replay --lower <blockheight> --upper <blockheight>`, but the parser just isn't set up for that right now. We could also move it to cwtool.


edmundnoble:
Parallelism is more annoying than I thought, because duh, we need separate sqlite connections for each chunk we do in parallel.

thoughtpolice (commented):
Group review, part 1. Mostly looks OK aside from the `limitCut` bits in `Chainweb.hs`

src/Chainweb/Cut.hs
```diff
@@ -427,6 +428,41 @@ limitCut wdb h c
...
+    go :: ChainId -> BlockHeader -> IO BlockHeader
+    go cid bh = do
+        if h >= _blockHeight bh
+        then return bh
+        else do
+            !db <- getWebBlockHeaderDb wdb cid
+            -- this is safe because it's guaranteed that the requested rank is
+            -- smaller then the block height of the argument
+            let ancestorHeight = min (int $ _blockHeight bh) (int h)
+            if ancestorHeight <= fromIntegral (genesisHeight v cid)
```
chessai, line 462: nit: genesisHeight uses genesisBlockHeader, so you can do some CSE and do `let genesis = genesisBlockHeader v cid` and then change this to `if ancestorHeight <= fromIntegral (_blockHeight genesis)`, and the call to genesisBlockHeader below can just be replaced with `genesis`.

src/Chainweb/Pact/PactService.hs
```diff
@@ -599,6 +609,70 @@ execNewGenesisBlock miner newTrans = pactLabel "execNewGenesisBlock" $ readFrom
...
+    heightProgress initialHeight ref logFun = do
+        r <- newIORef initialHeight
+        forever $ do
+          h <- readIORef r
+          h' <- readIORef ref
+          writeIORef r h'
+          logFun
+            $ "processed: " <> sshow (h' - initialHeight)
+            <> ", current height: " <> sshow h'
+            <> ", rate: " <> sshow ((h' - h) `div` 20) <> "blocks/sec"
```
chessai, line 673: make the 20 across these two lines a named constant. easy to go out of sync.

chessai (approved):

src/Chainweb/Pact/PactService.hs
```diff
@@ -599,6 +609,71 @@ execNewGenesisBlock miner newTrans = pactLabel "execNewGenesisBlock" $ readFrom
...
+execReadOnlyReplay lowerBound upperBound = pactLabel "execReadOnlyReplay" $ do
+    ParentHeader cur <- findLatestValidBlockHeader
+    logger <- view psLogger
+    bhdb <- view psBlockHeaderDb
+    pdb <- view psPdb
+    v <- view chainwebVersion
+    cid <- view chainId
+    -- lower bound must be an ancestor of upper.
+    liftIO (ancestorOf bhdb (_blockHash lowerBound) (_blockHash upperBound)) >>=
+      flip unless (error "lower bound is not an ancestor of upper bound")
```
ak3n, line 627: Let's replace with `throwM`?

src/Chainweb/Pact/PactService.hs
```diff
@@ -599,6 +609,71 @@ execNewGenesisBlock miner newTrans = pactLabel "execNewGenesisBlock" $ readFrom
...
+
+    -- upper bound must be an ancestor of latest header.
+    liftIO (ancestorOf bhdb (_blockHash upperBound) (_blockHash cur)) >>=
+      flip unless (error "upper bound is not an ancestor of latest header")
```
ak3n, line 631: same here

ak3n (approved):
LGTM 👍

PR #1806, opened 2024-02-02
Optimize buy gas not to use a defpact

edmundnoble:
This PR is meant to be reviewed commit-by-commit.

Buying gas right now requires starting and resuming a defpact, called `fund-tx`, which has two steps: buying gas (the `buy-gas` function), before the start of a transaction, debits the gas payer's account; and redeeming gas (the `redeem-gas` function), after the end of a transaction, pays the gas payer for the remaining unused gas and pays the miner for the used gas. We've discussed the storage impact of this, though now that is mostly irrelevant thanks to #1744. To summarize, each started Pact occupies a row in the Pacts table with its state; each time a Pact is resumed, this row is written to, keeping the old row for reorgs. There is also a time performance aspect. We grow the pending writes map with `PactExec`s by starting and resuming these defpacts, and we also spend time serializing and deserializing the `PactExec`s as and from JSON when we do so.

This also makes buying gas itself cost slightly less gas.

This PR calls the `buy-gas` and `redeem-gas` functions directly instead of via the `fund-tx` defpact.

---
  - https://app.asana.com/0/0/1206729978146609

### DIFF
--- a/src/Chainweb/Pact/PactService.hs
+++ b/src/Chainweb/Pact/PactService.hs
@@ -419,7 +419,6 @@ serviceRequests memPoolAccess reqQ = do
             put $! s'
             return $! r
 
-
 execNewBlock
     :: forall logger tbl. (Logger logger, CanReadablePayloadCas tbl)
     => MemPoolAccess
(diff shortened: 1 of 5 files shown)


edmundnoble:
This has been delayed to next release.

PR #1808, opened 2024-02-07
compact live node test

chessai:
Add a test that you can compact a live node, and it still makes progress.

---
  - https://app.asana.com/0/0/1204890679652603

### DIFF
--- a/src/Chainweb/Pact/Backend/Compaction.hs
+++ b/src/Chainweb/Pact/Backend/Compaction.hs
@@ -47,6 +47,7 @@ import Control.Monad.Catch (MonadCatch(catch), MonadThrow(throwM))
 import Control.Monad.IO.Class (MonadIO(liftIO))
 import Control.Monad.Reader (MonadReader, ReaderT, runReaderT, local)
 import Control.Monad.Trans.Control (MonadBaseControl, liftBaseOp)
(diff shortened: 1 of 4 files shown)


src/Chainweb/Pact/Backend/Compaction.hs
```diff
@@ -336,9 +337,11 @@ withTables ts a = do
 -- | Takes a bunch of singleton tablename rows, sorts them, returns them as
 --   @TableName@
 sortedTableNames :: [[SType]] -> [TableName]
-sortedTableNames rows = M.elems $ M.fromListWith const $ flip List.map rows $ \case
-  [SText n@(Utf8 s)] -> (Text.toLower (Text.decodeUtf8 s), TableName n)
-  _ -> error "sortedTableNames: expected text"
+sortedTableNames rows = coerce
```
chessai, line 340: I could have absolutely sworn that I had already fixed this. I am 98% sure this is not fixed as a result of a bad merge or something.

chessai: To be clear, by "fixed", I only mean that the current code in master has significantly worse readability.

edmundnoble (approved):

PR #1812, opened 2024-02-08
Make compaction of live nodes retry until the db is available

edmundnoble:
`RemotePactTest.txlogsCompactionTest` is flaky without this and actually `compact-live-node` in `SlowTests` is sort of flaky as well.

### DIFF
--- a/test/Chainweb/Test/Pact/Utils.hs
+++ b/test/Chainweb/Test/Pact/Utils.hs
@@ -109,6 +109,7 @@ module Chainweb.Test.Pact.Utils
 , zeroNoncer
 -- * Pact State
 , compact
+, compactUntilAvailable
 , PactRow(..)
 , getLatestPactState
 , getPactUserTables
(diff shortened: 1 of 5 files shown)


test/Chainweb/Test/Pact/Utils.hs
```diff
@@ -1037,3 +1041,18 @@ compact :: ()
...
+-- | Compaction function that retries until the database is available.
+compactUntilAvailable
+  :: C.TargetBlockHeight
+  -> YAL.Logger SomeLogMessage
+  -> SQLiteEnv
+  -> [C.CompactFlag]
+  -> IO ()
+compactUntilAvailable tbh logger dbEnv@(SQLiteEnv db _) flags = do
+    try (C.compact tbh logger db flags) >>= \case
+        Left (C.CompactExceptionDb (fromException ->
```
chessai, line 1054: isn't this missing the Right () case?

chessai: nvm, mobile formatting issue.

chessai (approved):

PR #1813, opened 2024-02-08
Allow the node to communicate when it's ready to serve requests

edmundnoble:
This is another attempt at fixing the racy issue that's causing `RemotePactTest` to be flaky. I ran into a segfault while I was writing this, which was caused by closing the rocksdb while the node was still running, in turn caused by incorrect use of `async`.

### DIFF
--- a/test/Chainweb/Test/Utils.hs
+++ b/test/Chainweb/Test/Utils.hs
@@ -121,7 +121,7 @@ module Chainweb.Test.Utils
 ) where
 
 import Control.Concurrent
-import Control.Concurrent.Async
+import Control.Concurrent.STM
 import Control.Lens
 import Control.Monad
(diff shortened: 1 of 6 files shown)


test/Chainweb/Test/Utils.hs
```diff
@@ -989,43 +991,53 @@ awaitBlockHeight v step cenv i = do
...
     -> RocksDb
     -> logger
+    -> TVar NowServing
     -> MVar (PeerInfo, Port)
     -> ChainwebConfiguration
     -> FilePath
        -- ^ pact db dir
     -> FilePath
-       -- ^ rocksdb dir
+       -- ^ dir for db backups
```
chessai, line 1034: nice, that makes this more clear

chessai (approved):

PR #1814, opened 2024-02-09
Raise test block gas limit to avoid spurious TxTimeouts during tests

edmundnoble:
The block gas limit is used to calculate the tx timeout, and tests can trigger it if there are noisy neighbors. This hopefully fixes that.


ak3n (approved):
👍

PR #1826, opened 2024-02-20
Try to encode block payloads and block outputs without base64

edmundnoble:
Just figured I'd push this, but it turned out to be more work than I thought.


thoughtpolice:
Superseded by #1885

PR #1864, opened 2024-03-19
validate host architecture for production mining

larskuhtz:


src/Chainweb/Miner/Config.hs
```diff
@@ -68,20 +73,34 @@ newtype MinerCount = MinerCount { _minerCount :: Natural }
...
 
-validateMinerConfig :: ConfigValidation MiningConfig []
-validateMinerConfig c = do
-    when (_nodeMiningEnabled nmc) $ tell
-        [ "In-node mining is enabled. This should only be used for testing"
-        , "In order to use in-node mining, mining-coordination must be enabled, too"
-        ]
-    when (_nodeMiningEnabled nmc && not (_coordinationEnabled cc))
-        $ throwError "In-node mining is enabled but mining coordination is disabled"
-    when (_nodeMiningEnabled nmc && view minerId (_nodeMiner nmc) == "")
```
larskuhtz, line 79: Is this check also relevant when just mining coordination is enabled (but no node-mining)?

src/Chainweb/Miner/Config.hs
```diff
@@ -68,20 +73,34 @@ newtype MinerCount = MinerCount { _minerCount :: Natural }
...
+                , " Supported architectures are " <> sshow supportedArchs
+                ]
   where
     nmc = _miningInNode c
     cc = _miningCoordination c
 
+    -- This is a heuristic and we are rather a little too restrictiv. In the
+    -- future we may also consider uname -m and/or cpuinfo (including flags) here.
+    --
+    supportedArchs = [ "x86_64" ]
```
larskuhtz, line 101: This value needs to be double checked for the release binaries, i.e. the binaries and docker images produced by the CI cabal builds once CI builds for this branch are available.

src/Chainweb/Utils.hs
```diff
@@ -1368,6 +1372,18 @@ showIpv6 ha = T.intercalate ":"
...
+
+hostArch :: String
+#if aarch64_HOST_ARCH == 1
+hostArch = "aarch64"
+#elif x86_64_HOST_ARCH == 1
+hostArch = "x86_64"
+#else
+hostArch = "unknown"
+#endif
+
```
larskuhtz, line 1386: An alternative would be to make use of the `uname` systems call at runtime. However, the resulting `utsname` structure is underspecified in POSIX (the length of the fields varies between different implementations). This would require either the implementation of c-wrappers for accessing the fields, or the use of a preprocessor (e.g. hsc) or configure script, or encoding a heuristic in C-macros similar to above code.

Overall, above macro that statically gets the value from GHC seems the simplest approach. If needed, it can be easily adjusted in the future by  adding more cases.

src/Chainweb/Miner/Config.hs
edmundnoble: Can we do a blacklist instead? I think we want to be conservative here, seeing as this prevents a node from even starting

larskuhtz: I for white listing. This only affects miners on mainnet. And for those we need strict rules.

PR #1898, opened 2024-04-05
test: make PreInsertCheckTimeoutTest do a lot more work

edmundnoble:
https://gerrit.aseipp.dev/q/Ia49614f4c7b8ea80e14a36ad42df83db45b4cb5b


chessai (approved):

edmundnoble:
Alas, this doesn't work.

ak3n:
@edmundnoble maybe we should consider to drop the test? we have tested that the timeout works fine in a real world scenario, maybe the noise this test produces doesn't worth the time to spent on it?

PR #1903, opened 2024-04-10
tests: run preInsertCheckTimeoutTest 100 times

edmundnoble:
Timeouts are tricky to trigger in GH actions. We're satisfied if it's triggered once in 100 runs. 

https://gerrit.aseipp.dev/q/I72c6f717ef13eb4ff28ed9b0e9fbbed4b782579f

### DIFF
--- a/test/Chainweb/Test/Pact/PactSingleChainTest.hs
+++ b/test/Chainweb/Test/Pact/PactSingleChainTest.hs
@@ -857,8 +857,13 @@ preInsertCheckTimeoutTest _ reqIO = testCase "preInsertCheckTimeoutTest" $ do
         $ set cbRPC (mkExec' coinV5)
         $ defaultCmd
 
-  rs <- pactPreInsertCheck (V.fromList [txCoinV3, txCoinV4, txCoinV5]) q
-  assertBool ("should be InsertErrorTimedOut but got " ++ show rs) $ V.and $ V.map (== Left InsertErrorTimedOut) rs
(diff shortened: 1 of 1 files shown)


thoughtpolice (approved):

ak3n (approved):

test/Chainweb/Test/Pact/PactSingleChainTest.hs
```diff
@@ -857,8 +857,13 @@ preInsertCheckTimeoutTest _ reqIO = testCase "preInsertCheckTimeoutTest" $ do
         $ set cbRPC (mkExec' coinV5)
         $ defaultCmd
 
-  rs <- pactPreInsertCheck (V.fromList [txCoinV3, txCoinV4, txCoinV5]) q
-  assertBool ("should be InsertErrorTimedOut but got " ++ show rs) $ V.and $ V.map (== Left InsertErrorTimedOut) rs
+  -- timeouts are tricky to trigger in GH actions.
```
chessai, line 860: I would like it if we branched this behaviour on if we are in GitHub actions or not, since doing it this way could hide unreliability.

```haskell
envVars <- getEnvironment
let inGha = flip any envVars $ \(k, v) ->
       (k == "CI" && v /= "")
       || k == "GITHUB_RUN_ID"
       
```
This should be sufficient for determining if we are in gha

edmundnoble: I don't actually mind if this is unreliable, as long as it happens at all.

PR #1917, opened 2024-05-07
nix: s/pkgconfig/pkg-config/

thoughtpolice:
This alias is deprecated and removed in newer versions of nixpkgs, so just go ahead and get rid of it and use the proper spelling.

Change-Id: I793f11948bd47be080c5cf6e17c401f168bc68b4


thoughtpolice:
Included in #1917.

PR #1920, opened 2024-05-08
[CI]: Add homebrew paths to cabal.project for the recent macos-14 runners

ak3n:
- fix haskell-actions (haskell/actions was deprecated)
- switch to macos-14-large to use x86_64 because `macOS-latest` is using arm runner which is broken for now (musl test)
- remove gnu-tar work around (was fixed in runners a long time ago)
- add new homebrew openssl paths

PR #1922, opened 2024-05-09
[CI-macos]: add caching for ghcup & cabal (reduce build time by 4 minutes)

ak3n:
based on #1920

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         ghc: ["9.6.4"]
         cabal: ["3.10"]
-        os: ["macOS-latest"]
+        os: ["macos-14-large"]
         cabalcache: ["true"]
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/macos.yaml
```diff
@@ -16,33 +16,58 @@ jobs:
       matrix:
         ghc: ["9.6.4"]
         cabal: ["3.10"]
-        os: ["macOS-latest"]
```
emilypi, line 19: we should do both - one goes through the macos github runner, but one is fixed hardware we have onhand for M1 specifically for regression.

emilypi (commented):
LGTM other than the one comment

emilypi (approved):

.github/workflows/macos.yaml
emilypi: Checked and Macos-latest is macos-14

edmundnoble:
Would this .ghcup cache also work fine for our other builds?

PR #1928, opened 2024-05-14
github: ci tweaks

thoughtpolice:
We'll be moving non-freeze file builds to nightly schedule to shed some testing load.


thoughtpolice:
Closed in favor of #1939 which includes these patches.

PR #1975, opened 2024-06-30
remove upper bound on crypton-connection

larskuhtz:
The PR removes / fixes the upper bound on `crypto-connection`. This fixes the build with the latest versions of the HTTP and TLS related packages.

Along the way the PR also cleans up some issues and inconsistency in the cabal file -- which are small and trivial changes.


chainweb.cabal
```diff
@@ -110,7 +110,7 @@ common warning-flags
```
chessai, line 110: You need to update the freeze file now.

```
rm cabal.project.freeze
cabal update
cabal freeze
```

chessai (changes requested):

chainweb.cabal
larskuhtz: Just noticed. I am also wondering why there are not non-freeze file builds. It seems that there is much less ci coverage for cabal CI builds than there used to be -- in terms of dependencies, ghc versions, and platforms? Is there any reason for that?

larskuhtz: I think, that's the reason why the breaking changes in crypton-connection went unnoticed for a while -- preventing testing of latest versions of tis and http related packages.

chessai: > I think, that's the reason why the breaking changes in crypton-connection went unnoticed for a while -- preventing testing of latest versions of tis and http related packages.

I did notice, and I wrote a PR that already fixed this - I think it got accidentally "reverted" in a bad merge.

There were non-freeze builds until recently, which is when I noticed. Non-freeze builds were causing a lot of churn for working on CI, which is why we disabled them

larskuhtz: I am not sure if those "cause" churn. I think, they rather "remind" of chores that need to be done anyways. 

Usually, those are simple tasks which can be done quickly, but can become nasty when they start to pile up.

In case those tasks are not simple, we used to add upper bounds in `cabal.project` and considered them as open issues.

larskuhtz: But, yeah, it's a matter of process. 

I think, there should be some process in place, to catch and track dependency issues early on. Otherwise, upgrades get pushed back which can cause overhead during release preparation and also reduces the time that new versions of dependency (direct or indirect) get tested.

larskuhtz:
The unit test failures look concerning. I wonder if those are related to the new version of unordered-containers. In best case it's just test code that makes illegal assumptions about ordering of the items in the container.

larskuhtz:
OK, I think, the unit test failures are independent of the code changes in this PR. The have been appearing on other branches, too. I am going to investigate those based on master.

larskuhtz:
Closing this because it got merged into #1976

PR #1976, opened 2024-06-30
support ghc-9.10

larskuhtz:
* [x] update sbv-9.2 pin to latest version of vendored 9.2 fork
* [x] update pact pin to https://github.com/kadena-io/pact/pull/1367 (ghc-9.10 support for pact)
* [x] fix compiler warnings for GHC-9.10 and `base >=4.20`
    * [x] The function `fold'` is exported in `Prelude`.
    * [x] GHC warns about the use of `head` and `tail` via the new `x-partial` warning. In Pact this addressed
        through the addition of an internal library that provides "unsafe" versions of partial functions that do not
        trigger the warning. If we use partial functions in chainweb we can import them from there.   
* [x] remove `Numeric.Additive` and `Numeric.AffineSpace` and import these from `pact-time:numeric >=0.3.0.1`
    `pact-time` now provides those modules through an internal public library. 
* [x] update CI builds
    * [x] update build matrix
    * [x] use ubuntu-22.04 for docker images (uses libssl3)
    * [x] update cabal-cache-action
    * [x] fix cabal-cache store location for GHC >= 8.10.2
* [x] update flake.lock


src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -277,8 +280,7 @@ doGetEarliestBlock db = do
     [] -> return Nothing
     (!o:_) -> return (Just o)
   where
-    qtext = "SELECT blockheight, hash FROM BlockHistory \
-            \ ORDER BY blockheight ASC LIMIT 1"
+    qtext = "SELECT blockheight, hash FROM BlockHistory ORDER BY blockheight ASC LIMIT 1"
```
larskuhtz, line 283: String literals with gaps break CPP. It's probably generally a good idea to avoid them and wait for new multi-line string literals to become available in GHC (https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0569-multiline-strings.rst).

src/Numeric/Additive.hs
larskuhtz, line 1: This file (and also 'Numeric.AffineSpace') are now available in `pact-time:numeric >=0.3.0.0`.

chessai (approved):

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
```diff
@@ -335,11 +335,13 @@ doLookupSuccessful curHeight hashes = do
     callDb "doLookupSuccessful" $ \db -> do
       let
         hss = V.toList hashes
-        params = Utf8 $ intercalate "," (map (const "?") hss)
-        qtext = "SELECT blockheight, hash, txhash FROM \
-                \TransactionIndex INNER JOIN BlockHistory \
-                \USING (blockheight) WHERE txhash IN (" <> params <> ")"
-                <> " AND blockheight <= ?;"
+        params = intercalate "," (map (const "?") hss)
+        qtext = Utf8 $ intercalate " "
```
edmundnoble, line 339: `intercalate " "` is just `unwords`

edmundnoble (changes requested):

.github/workflows/applications.yml
```diff
@@ -134,21 +134,36 @@ jobs:
...
+          "os": ["ubuntu-22.04"],
+          "use-freeze-file": ["false"],
+          "include" : [
+            {
+              "ghc": "9.6.5",
+              "cabal": "3.12",
+              "os": "ubuntu-22.04",
+              "use-freeze-file": "true"
+            }
+          ]
```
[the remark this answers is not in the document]
larskuhtz: But why? Build times are not affected by the number of builds. They are in parallel and dependencies are fully cached. 

Why not give feedback early?

.github/workflows/applications.yml
```diff
@@ -326,11 +345,12 @@ jobs:
...
-      uses: larskuhtz/cabal-cache-action@a0f263d898dd246217960262caa1d381cf066e9a
+      uses: larskuhtz/cabal-cache-action@4b537195b33898fcd9adc62cee2a44986fd7b1b6
       with:
         bucket: "kadena-cabal-cache"
         region: "us-east-1"
-        folder: "${{ matrix.os }}-${{ matrix.ghc }}-${{ matrix.use-freeze-file }}"
+        folder: "packages/${{ matrix.os }}"
+        store_path: ${{ steps.setup.outputs.cabal-store }}
         aws_access_key_id: "${{ secrets.kadena_cabal_cache_aws_access_key_id }}"
         aws_secret_access_key: "${{ secrets.kadena_cabal_cache_aws_secret_access_key }}"
```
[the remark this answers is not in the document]
larskuhtz: Frozen builds should use the same cache, because there's actually a chance of sharing builds of dependencies.

The cache handles different ghc versions internally in sub-folders. Also, it uses content based hashes, so there's no risk of conflicts.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
larskuhtz: Oh, it's Bytestring:

> Module ‘Data.ByteString’ does not export ‘unwords’.

.github/workflows/applications.yml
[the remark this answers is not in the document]
edmundnoble: I'm fine with disabling the scheduled logic here and putting unfrozen builds into the matrix here.

.github/workflows/applications.yml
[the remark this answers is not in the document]
edmundnoble: That makes sense to me.

.github/workflows/applications.yml
```diff
@@ -625,7 +645,7 @@ jobs:
         # different tags
         include:
         - ghc: "9.6.5"
-          os: "ubuntu-20.04"
```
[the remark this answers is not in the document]
edmundnoble: Right this is just the image and not the binary, got it.

src/Chainweb/Pact/Backend/RelationalCheckpointer.hs
edmundnoble: Ah I see, makes sense seeing as it's binary data.

cabal.project
```diff
@@ -123,19 +123,20 @@ source-repository-package
     tag: d8019c3404d6f3b3c0b0416e9899cfdf614ef425
     --sha256: 09msayidg23rsdz97fcfqqalm4pbawx3c1qihgab8hnlmjxby103
 
--- Patch merged into master (upcoming version 10.0). We are currently using 9.2
+-- Patch merged into master (upcoming verison 10.0). We are currently using 9.2.
```
edmundnoble, line 126:
```suggestion
-- Patch merged into master (upcoming version 10.0). We are currently using 9.2.
```

edmundnoble (approved):

PR #1977, opened 2024-07-01
stop exporting constructors for BlockHeader and PayloadData

chessai:
Change-Id: Id6200af01359d9012ea532a485ac25d2ab36d0c5


src/Chainweb/BlockHeader.hs
```diff
@@ -65,7 +33,7 @@ module Chainweb.BlockHeader
 , powTarget
 
 -- * BlockHeader
-, BlockHeader(..)
+, BlockHeader
```
larskuhtz, line 36: Would it make sense to restrict the lenses into `BlockHeader` that are exported from this module to be just getters?

chessai: Done

edmundnoble (approved):

PR #1982, opened 2024-07-31
avoid custom setup

chessai:
Change-Id: I0cfd3beb2341656d978a845af198e563ac2ba3bb


.github/workflows/applications.yml
```diff
@@ -243,6 +243,10 @@ jobs:
       run: |
         ghc --version
         cabal --version
+    - name: Install cabal-plan and confirm its installation
+      run: |
+        cabal install cabal-plan --flags="exe license-report" --constraint="cabal-plan==0.7.3.0" --allow-newer
+        cabal-plan --version
```
larskuhtz, line 249: How long does this take?

Can this be cached or installed via a package manager or a container image? Otherwise this build will run for each workflow run.

chainweb.cabal
```diff
@@ -357,9 +348,10 @@ library
         , Utils.Logging
         , Utils.Logging.Config
         , Utils.Logging.Trace
+        , Utils.PackageInfo
```
larskuhtz, line 351: Is there an away to encode the dependency on the cabal-plan tool in the cabal file? I vaguely remember that there's some stanza for tools.

.github/workflows/applications.yml
chessai: It takes a couple of minutes. I am not sure what the right way to improve this is. I can't find it in any package managers. We could cache it with a nix build using our own cache.

larskuhtz: Installing and initializing a nix environment just for that one tool seems overkill to me and is probably also slow.

The cabal-cache binary we wrap into an GitHub actions. It downloads the binary from the release page. It also injects it into the runner cache.

A more lightweight approach could be to check if there binaries on the GitHub release page and, if not, clone it and create (stripped) release binaries. Or just put binaries in an S3 bucket.

Another option would be to put a container image into ghcr.io and run it from there. Whatever is the most convenient.

larskuhtz: I vaguely remember that cabal itself has support for buidling built-tools, but not sure. If building cabal-plan would be part of the normal build, it would profit from the cabal-cache and dist-cache out of the box.

larskuhtz: One final idea: instead of calling cabal-plan as external binary, could we just use it as a library and call the required functionality directly? That would simplify the TH code (no external process) and required library functions would be just build as dependencies.

How much of cabal-plan code is actually required for the desired functionality? Maybe we could extract and vendor a minimal implementation?

larskuhtz: Well, I just noticed that cabal-plan is under GPL-2.0. So we can't link against it or copy code from it.
We could still consider implementing the functionality ourself -- if it's not too much work.

chessai:
Lars found a simpler solution here. Closing.

PR #1987, opened 2024-08-19
performance: avoid redundant decode/encode in mempool lookup

edmundnoble:


chessai (approved):

chessai:
In my testing, this reduced CPU usage of the `/mempool/lookup` endpoint by a factor of 16.

PR #2006, opened 2024-10-06
make chainweb-node application a separate cabal (sub-) package

larskuhtz:
This is a small change that allows us to build chainweb with the `simple` built type.

The directory structure remains mostly the same, which chainweb-node living in a sub-directory of the chainweb package.

I cleaner implementation would be to not nest chainweb-application in chainweb-node. Instead both projects should live in subdirectories of a common project directory. However, that would probably break more infrastructure and could also result in merge conflicts. I propose to leaf that for a dedicated follow up PR.


larskuhtz:
The nix build is still broken and I may need some help from an expert for fixing it.

chessai (approved):

PR #2022, opened 2024-11-01
Send new transactions to remote mempools immediately

edmundnoble:
Instead of waiting for the first sync, send them ~immediately to cut down on a mempool hop for ~free. This may not actually be worth doing after we start penalizing peers for unreachability.


chessai:
I said this to Edmund, but in my testing on testnet05, penalising peers properly eliminated the benefit of doing this. This testing is not entirely conclusive, but it is something to note.

PR #2029, opened 2024-11-06
update freeze file to use hashes 0.3.0.1

edmundnoble:
without this change I can't build the node on gentoo.


chessai (approved):

chessai (changes requested):
You need to add an actual lower bound in the cabal file so that `cabal freeze` command will generate this properly when re-run.

edmundnoble:
This isn't needed as of #2025

PR #2041, opened 2024-11-14
randomized peer ordering

edmundnoble:
We randomize the peer ordering by assigning a peer a random number that is used subsequently to order it with respect to other peers upon request from the /peer GET endpoint.

An alternative of generating a seed on startup that would be used to "salt" the peer ordering was considered but ran into implementation difficulties and seems equivalent.

---
  - https://app.asana.com/0/0/1208677746463529

### DIFF
--- a/src/P2P/Node/PeerDB.hs
+++ b/src/P2P/Node/PeerDB.hs
@@ -15,6 +15,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE BangPatterns #-}
 
 -- |
 -- Module: P2P.Node.PeerDB
@@ -35,6 +36,7 @@ module P2P.Node.PeerDB
 , PeerEntrySticky(..)
 , HostAddressIdx
 , hostAddressIdx
+, PeerMark(..)
 
 -- * Peer Entry
 , PeerEntry(..)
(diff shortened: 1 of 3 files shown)


larskuhtz (approved):

src/P2P/Node/PeerDB.hs
```diff
@@ -178,16 +196,16 @@ data PeerEntry = PeerEntry
         --
     }
     deriving (Show, Eq, Ord, Generic)
-    deriving anyclass (ToJSON, FromJSON, NFData)
```
larskuhtz, line 181: 👍

src/P2P/Node/PeerDB.hs
```diff
@@ -258,14 +276,13 @@ addPeerEntry b m = m & case getOne (getEQ addr m) of
...
-        , _peerEntryNetworkIds = _peerEntryNetworkIds a <> _peerEntryNetworkIds b
-        , _peerEntryActiveSessionCount = _peerEntryActiveSessionCount a + _peerEntryActiveSessionCount b
-        , _peerEntrySticky = PeerEntrySticky False
-        }
+    -- we retain the old mark for the peer
+    update a = updateIx addr $! a
+            & peerEntrySuccessiveFailures +~ _peerEntrySuccessiveFailures b
+            & peerEntryLastSuccess %~ max (_peerEntryLastSuccess b)
+            & peerEntryNetworkIds <>~ _peerEntryNetworkIds b
+            & peerEntryActiveSessionCount +~ _peerEntryActiveSessionCount b
```
larskuhtz, line 284: did you double check that this is actually as strict as the original code? The `PeerEntry` should be NF.

(I always find it hard to reason about strictness when lenses are involved)

src/P2P/Node/PeerDB.hs
```diff
@@ -436,7 +466,9 @@ updatePeerDb (PeerDb _ _ _ lock var) a f
     = withMVar lock . const . atomically . modifyTVar' var $ \s ->
         case getOne $ getEQ a s of
             Nothing -> s
-            Just x -> updateIx a (f x) s
+            Just x ->
+                let !x' = f x
+                in updateIx a x' s
```
larskuhtz, line 471: 👍

larskuhtz (approved):

larskuhtz (approved):

PR #2048, opened 2024-11-21
better error message on devnet mismatch

chessai:
Change-Id: Ife724b6b4a3d18e840f98dc5adc141b206d48209

### DIFF
--- a/src/Chainweb/Version/Registry.hs
+++ b/src/Chainweb/Version/Registry.hs
@@ -129,16 +129,20 @@ lookupVersionByCode code
         -- the case that we don't actually need the version, just the code.
         lookupVersion & versionCode .~ code
     where
+
     lookupVersion :: HasCallStack => ChainwebVersion
     lookupVersion = unsafeDupablePerformIO $ do
         m <- readIORef versionMap
         return $ fromMaybe (error notRegistered) $
             HM.lookup code m
+
     notRegistered
-        | code == _versionCode recapDevnet = "recapDevnet version used but not registered, remember to do so after it's configured"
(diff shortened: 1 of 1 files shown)


src/Chainweb/Version/Registry.hs
```diff
@@ -129,17 +129,21 @@ lookupVersionByCode code
...
     notRegistered
-        | code == _versionCode recapDevnet = "recapDevnet version used but not registered, remember to do so after it's configured"
-        | code == _versionCode devnet = "devnet version used but not registered, remember to do so after it's configured"
-        | code == _versionCode pact5Devnet = "Pact 5 devnet version used but not registered, remember to do so after it's configured"
+        | code == _versionCode recapDevnet = "recapDevnet version used but not registered, remember to do so after it's configured. " <> perhaps
+        | code == _versionCode devnet = "devnet version used but not registered, remember to do so after it's configured. " <> perhaps
+        | code == _versionCode pact5Devnet = "Pact 5 devnet version used but not registered, remember to do so after it's configured. " <> perhaps
         | otherwise = "version not registered with code " <> show code <> ", have you seen Chainweb.Test.TestVersions.testVersions?"
 
+    perhaps = "Perhaps you are attempting to run a different devnet version than a previous run, and you need to delete your db directory before restarting devnet with the new version?"
```
edmundnoble, line 145: Ah. This really shouldn't be the error we get in this case, but the version registry ensures that it is for now. This has my approval. When the registry is deleted, this message will disappear in favor of a more vanilla "block version code mismatch" error, as would happen for testnet or mainnet.

edmundnoble (approved):

PR #2054, opened 2024-12-04
delete checkpointer abstraction

edmundnoble:
Replace the checkpointer abstraction with a concrete `CheckpointerResources` type. Consider the original RelationalCheckpointer code to be the "internal" portion of the checkpointer. Use qualified imports to refer to the checkpointer from PactService so that it's clear what's owned by it.


larskuhtz:
I think the term `CheckpointerResources` is misleading. In the Chainweb code base the suffix `Resources` is usually used for ephemeral records that exist only during initialization (and are supposed to be garbage collected once the services are started). The purpose of those records is it to initialize and distribute shared backend components to top-level components in Chainweb node before starting node services.

larskuhtz:
> Yeah fair point, I can rename it to `Checkpointer`.

Alternatively, we could also move away from the current (not ideal) terminology and use a more specific term for those ephemeral records. It's just to avoid confusion.

chessai (approved):

chessai (approved):

PR #2067, opened 2024-12-13
attempt at optimizing cut meet and join

edmundnoble:


edmundnoble:
Close in favor of #2077.

PR #2069, opened 2024-12-13
Pact 5 review phase 2

edmundnoble:


src/Chainweb/Pact/Backend/PactState/GrandHash/Import.hs
```diff
@@ -212,7 +212,7 @@ pactImportMain = do
         --
         -- pact-import doesn't use this environment variable; it's for
         -- debugging and/or consumption by other tools.
-        setEnv "SNAPSHOTview blockHeight" (show snapshotBlockHeight)
```
chessai, line 215: Good catch! Bad sed by me.

PR #2071, opened 2024-12-16
Migrate to property-matchers 0.2

edmundnoble:
This comes with a few API changes, and uses a version that hackage hasn't yet published in the index, but will soon.


cabal.project
```diff
@@ -86,6 +86,12 @@ package yet-another-logger
 --
 --   nix-prefetch-git --url <location> --rev <tag>
 
+source-repository-package
+    type: git
+    location: https://gitlab.com/edmundnoble/property-matchers.git
```
rsoeldner, line 91: Do we have other gitlab reps?

edmundnoble: This is just mine :innocent: I'll delete this entry because it should be on hackage by now.

rsoeldner (approved):

PR #2084, opened 2024-12-20
optimize RemotePactTest and add new invalid tx test

edmundnoble:
This test is optimized by sharing the TLS cert and HTTP manager between tests. Otherwise the rest of the testing fixture is *not* shared.

This PR also adds some tasty steps to a few tests, which makes it clearer what they're spending their time on.

There's one extra invalid tx test as well, testing the error from sending to the wrong chain.

### DIFF
--- a/test/unit/Chainweb/Test/Pact5/CutFixture.hs
+++ b/test/unit/Chainweb/Test/Pact5/CutFixture.hs
@@ -131,6 +131,7 @@ advanceAllChains Fixture{..} = do
     let blockHeights = fmap (view blockHeight) $ latestCut ^. cutMap
     let latestBlockHeight = maximum blockHeights
 
+    -- TODO: rejig this to do parallel mining.
     (finalCut, perChainCommandResults) <- foldM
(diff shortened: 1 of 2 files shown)


test/unit/Chainweb/Test/Pact5/CutFixture.hs
```diff
@@ -131,6 +131,7 @@ advanceAllChains Fixture{..} = do
     let blockHeights = fmap (view blockHeight) $ latestCut ^. cutMap
     let latestBlockHeight = maximum blockHeights
 
+    -- TODO: rejig this to do parallel mining.
```
rsoeldner, line 134: what is rejig ?

rsoeldner (approved):

test/unit/Chainweb/Test/Pact5/CutFixture.hs
edmundnoble: Redo, alter, fix, change, rearrange, that sort of thing.

edmundnoble:
This is included in #2085.

PR #2087, opened 2024-12-23
miscelaneous small changes

larskuhtz:
A number of small unrelated changes. It is easiest to review on a per commit basis.

### DIFF
--- a/src/Chainweb/BlockHeaderDB/Internal.hs
+++ b/src/Chainweb/BlockHeaderDB/Internal.hs
@@ -29,7 +29,6 @@ module Chainweb.BlockHeaderDB.Internal
 (
 -- * Internal Types
   RankedBlockHeader(..)
-, RankedBlockHash(..)
 , BlockRank(..)
 
 -- * Chain Database Handle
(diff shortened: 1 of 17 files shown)


src/Chainweb/ChainId.hs
```diff
@@ -268,6 +268,12 @@ chainIdInt :: Integral i => ChainId -> i
 chainIdInt (ChainId cid) = int cid
 {-# INLINE chainIdInt #-}
 
+-- -------------------------------------------------------------------------- --
+-- ChainMap
+
+-- TODO: Shouldn't this type guarantee that the map is total, i.e. that there
+-- exists a value for each chain?
```
edmundnoble, line 275: Hmmmmm... maybe we can do this, yes. Please leave this comment here.

src/Chainweb/ChainId.hs
```diff
@@ -305,3 +311,9 @@ atChain :: ChainId -> Fold (ChainMap a) a
 atChain cid = folding $ \case
     OnChains m -> m ^. at cid
     AllChains a -> Just a
+
+type instance Index (ChainMap a) = ChainId
+type instance IxValue (ChainMap a) = a
+
+instance IxedGet (ChainMap a) where
+    ixg i = atChain i
```
edmundnoble, line 319: :heart:

edmundnoble: We should probably even have an `Ixed` instance.

src/Chainweb/BlockHeaderDB/Internal.hs
```diff
@@ -29,7 +29,6 @@ module Chainweb.BlockHeaderDB.Internal
 (
 -- * Internal Types
   RankedBlockHeader(..)
```
edmundnoble, line 31: Do we still need `RankedBlockHeader` with the `Ranked` newtype? I suppose if we make it wrap a `Ranked BlockHeader`, that will waste some memory by duplicating the height.

edmundnoble (dismissed):
I see that this:
- adds a `Ranked` type, which should be able to replace a couple newtypes like for `RankedBlockHash`. 
- adds some extra instances for `ChainMap` and a salient comment.
- splits `BlockPayloadHash` into its own file - of course this is required by EVM.
- adds some other utilities and missing instances.

src/Chainweb/BlockHeaderDB/Internal.hs
larskuhtz: Yeah, that's the reasoning for not using `Ranked` with `BlockHeader`.

PR #2088, opened 2024-12-23
miner rewards

larskuhtz:


src/Chainweb/MinerReward.hs
```diff
@@ -0,0 +1,250 @@
...
+-- cached.
+--
+mkMinerRewards :: HasCallStack => MinerRewardsTable
+mkMinerRewards =
+    case CSV.decode CSV.NoHeader (BL.fromStrict rawMinerRewards) of
+        Left e -> error
+            $ "cannot construct miner rewards table: " <> sshow e
+        Right vs ->
+            let rewards = M.fromList . V.toList . V.map formatRow $ vs
+            in if (minerRewardsHash rewards == expectedMinerRewardsHash)
```
edmundnoble, line 212:
```suggestion
            in if minerRewardsHash rewards == expectedMinerRewardsHash
```

src/Chainweb/MinerReward.hs
```diff
@@ -0,0 +1,250 @@
...
+                then rewards
+                else error $ "hash of miner rewards table does not match expected hash"
+  where
+    formatRow :: (Word64, CsvDecimal) -> (BlockHeight, Stu)
+    formatRow (!a,!b) = (BlockHeight $ int a, kdaToStu (Kda $ _csvDecimal b))
```
edmundnoble, line 217: I don't think this bang changes anything, as the rows are already parsed. You could add bangs to the RHS, but that's probably unnecessary, because this all gets hashed immediately anyway.

edmundnoble (approved):

src/Chainweb/MinerReward.hs
larskuhtz, line 217:
```suggestion
    formatRow (a, b) = (BlockHeight $ int a, kdaToStu (Kda $ _csvDecimal b))
```

PR #2091, opened 2025-01-07
Make versionquirks testable

edmundnoble:
Prior, we'd say "the command with request key X has gas Y". Now we say "the command on chain X with block height H and at index I has gas Y". This makes it possible to test without tying the knot anywhere, because the request key is a hash that includes the version name, and also lets us integrate it into Pact 5 later.

Change-Id: Id00000007bbb29b009a2b2e4a0cf36180c8df1a7

### DIFF
--- a/src/Chainweb/Version.hs
+++ b/src/Chainweb/Version.hs
@@ -53,6 +53,8 @@ module Chainweb.Version
     , ChainwebVersionName(..)
     , ChainwebVersion(..)
     , pact4Upgrade
+    , TxIdxInBlock(..)
+    , _TxBlockIdx
     , VersionQuirks(..)
     , noQuirks
     , quirkGasFees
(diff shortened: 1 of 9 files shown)


emilypi (approved):

src/Chainweb/Version.hs
```diff
@@ -420,19 +421,24 @@ instance NFData PactUpgrade where
...
+
+makePrisms ''TxIdxInBlock
+
 -- The type of quirks, i.e. special validation behaviors that are in some
 -- sense one-offs which can't be expressed as upgrade transactions and must be
 -- preserved.
 data VersionQuirks = VersionQuirks
-    { _quirkGasFees :: !(HashMap RequestKey Gas)
-    -- ^ Note: only works for user txs in blocks right now.
+    { _quirkGasFees :: !(ChainMap (HashMap (BlockHeight, TxIdxInBlock) Gas))
```
jmcardon, line 434: Why the migration to `(BlockHeight, TxIdxInBlock)` instead of request key? I don't understand the necessity/benefit of this change.

jmcardon (approved):

edmundnoble:
Was merged with #2092.

PR #2092, opened 2025-01-07
Add gas quirks to Pact 5

edmundnoble:
We don't have a use for these yet, but this does establish the way
they can work in future. Depends on #2091.


test/unit/Chainweb/Test/Pact4/PactMultiChainTest.hs
```diff
@@ -1367,65 +1364,31 @@ checkTransferCreate = do
...
---                   [ PactTxTest mempoolCmdBuilder $
---                     \cr -> assertTxGas "quirked gas" 1 cr
---                   ]
---             void $ setPactMempool' (Just genesisHeader) cid
---               $ PactMempool [testsToBlock quirkedTests]
---             runCut'
---             rs <- txResults
---             liftIO $ print rs
---             testCurrentBlock quirkedTests
+-- TODO: pact5.
```
jmcardon, line 1367: Remove this comment? You have `quirkSpec` in the pact 5 tests

jmcardon (approved):

PR #2093, opened 2025-01-08
add a mempool insertCheck variant that doesn't short-circuit. improve error messages in /send api

chessai:
Change-Id: I435856410fb82c59f8170d32bf12e5cea69833d1

### DIFF
--- a/src/Chainweb/Mempool/InMem.hs
+++ b/src/Chainweb/Mempool/InMem.hs
@@ -3,10 +3,12 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ImportQualifiedPost #-}
 {-# LANGUAGE NumericUnderscores #-}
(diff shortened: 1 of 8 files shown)


src/Chainweb/Mempool/InMem.hs
```diff
@@ -348,6 +333,68 @@ insertCheckInMem cfg lock txs
     hasher :: t -> TransactionHash
     hasher = txHasher (_inmemTxCfg cfg)
 
+-- |
+--
+--   This function is used when a transaction is inserted into the mempool. It is
+--   NOT used when a new block is created. For the latter more strict validation
+--   methods are used. In particular TTL validation is uses the current time as
+--   reference in the former case and the creation time of the parent header in
+--   the latter case.
```
edmundnoble, line 342: There are some more specific comments we can make, reflecting that this is only done for transactions coming in on the service API.

edmundnoble (changes requested):
CI is also broken btw.

edmundnoble (approved):

edmundnoble (approved):

PR #2095, opened 2025-01-08
Update parity replay with latest classifications

jmcardon:
Adds two new classes of discreps (`defpact-entity` and `invalid-blessed-hash`), as well as classifies a few individual discrepancies related to module deploy + event emission in the same transaction.

### DIFF
--- a/classify-diffs
+++ b/classify-diffs
@@ -54,6 +54,8 @@ mkdir -p diffs/interesting
 mkdir -p diffs/defschema-term-position
 mkdir -p diffs/invalid-governance
 mkdir -p diffs/invalid-def-defschema-position
+mkdir -p diffs/invalid-blessed-hash
+mkdir -p diffs/defpact-entity
 
 classify_diff() {
   grep -q 'Loaded module' "$1" && mv -f "$1" diffs/loaded-module-hash/ && return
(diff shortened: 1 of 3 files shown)


classify-diffs
```diff
@@ -198,6 +202,16 @@ mv -f parity-replay-diffs/s6w1VVITn* diffs/list-commas 2> /dev/null
 
 mv -f parity-replay-diffs/EMaD-w9ConQkyLfg6_RsfNtld-fS1pEclfTSwtRdNzs.md diffs/interesting 2> /dev/null
 mv -f parity-replay-diffs/bSbkJeudtFC6eiz1x9FTL5jwRkdOzO4bzwKK_DLGisI.md diffs/interesting 2> /dev/null
+mv -f parity-replay-diffs/uHM_66xiDIxpB9CQzxKoNbtYfJ6hkfl1PU_2PvAyl4E.md diffs/interesting 2> /dev/null
+mv -f parity-replay-diffs/nVq3xSmmz2DRhcRzjNdRDwo4YY10TKSTgkCS9rk5L2M.md diffs/interesting 2> /dev/null
+mv -f parity-replay-diffs/XwCeYcM4tyDNZePdBqggnzmIb_plES9p8PREl-xd7yw.md diffs/interesting 2> /dev/null
```
edmundnoble, line 207: If these are all the same class, make this a manually-classified class rather than putting them in diffs/interesting, which is for one-offs.

jmcardon: 👍

PR #2138, opened 2025-02-10
Persist pending writes/reads across txs in pre-insert check batch 

edmundnoble:
This should make larger batches cheaper to pre-insert check, because they won't have to reload the coin module.

Change-Id: Id0000000558b7a0eb9399b6d94a59a03fa14ea43


chessai (commented):
this kind of change makes PactBlockM really shine

chessai (approved):

PR #2139, opened 2025-02-12
add keccak256 back

chessai:


src/Chainweb/Pact5/Backend/ChainwebPactDb.hs
```diff
@@ -240,6 +232,8 @@ makeLensesWith
         , ("_blockHandlerUpperBoundTxId", "blockHandlerUpperBoundTxId")
         , ("_blockHandlerBlockHeight", "blockHandlerBlockHeight")
         , ("_blockHandlerAtTip", "blockHandlerAtTip")
+        , ("_blockHandlerVersion", "blockHandlerVersion")
+        , ("_blockHandlerChainId", "blockHandlerChainId")
```
edmundnoble, line 236: If you're using all of the fields, you can just do `makeLenses`

src/Chainweb/Pact5/TransactionExec.hs
```diff
@@ -303,7 +302,7 @@ applyLocal logger maybeGasLogger coreDb txCtx spvSupport cmd = do
     -- anyone's workflow
     , FlagAllowReadInLocal
     , FlagRequireKeysetNs
-    ]
+    ] `Set.union` (if guardCtx chainweb228Pact txCtx then Set.empty else Set.singleton FlagDisablePact51)
```
edmundnoble, line 305: Please make a reusable function for this that can be called by `applyCmd` and `applyLocal`, similar to how it's done in `Pact4.TransactionExec`, so that we don't miss adding this to both places in future.

src/Chainweb/Pact/PactService/Checkpointer/Internal.hs
```diff
@@ -493,7 +493,9 @@ getBlockHistory db blockHeader d = do
             [SInt s,SInt e]
             [RInt,RText,RBlob]
       forM r $ \case
-        [SInt txid, SText key, SBlob value] -> (key,fromIntegral txid,) <$> Pact5.toTxLog (Pact5.renderDomain d) key value
+        [SInt txid, SText key, SBlob value] ->
+            (key,fromIntegral txid,)
+            <$> Pact5.toTxLog (_chainwebVersion blockHeader) (_chainId blockHeader) (view blockHeight blockHeader) (Pact5.renderDomain d) key value
```
edmundnoble, line 498: I know this (and the other diffs in this file) is a bit of an unused area of the code but it looks fine.

PR #2140, opened 2025-02-13
Avoid duplicate pre-insert checks;

edmundnoble:
Store a set of currently-being-inserted txs across in the mempool. This way we can avoid doing duplicate pre-insert checks, which is good, because those could take a while depending on what PactService is busy with.

Change-Id: Id0000000759106e235d28e052289a229f49212ef


src/Chainweb/Mempool/InMem.hs
```diff
@@ -510,42 +513,54 @@ insertCheckInMem' cfg lock txs
...
-            recordRecentTransactions maxRecent newHashes
+insertInMem logger cfg pendingInsertionRef lock runCheck txs0 = do
+    pendingInsertionsDirty <- readIORef pendingInsertionRef
+    let
+      txs' = flip V.mapMaybe txs0 $ \tx ->
+        let hash = hasher tx
+        in if not (HashSet.member hash pendingInsertionsDirty)
+        then Just (T2 hash tx)
+        else Nothing
+    bracket
```
jwiegley, line 538: @edmundnoble A bracket with functions that ignore their arguments is just `bracket_`.

chessai (approved):

larskuhtz (approved):

PR #2147, opened 2025-03-24
Start hashing adjacent hash records in work headers as of some fork

edmundnoble:
This allows us to expand the chain graph later without breaking miners. 

Change-Id: Id0000000d6a8a713b68381ecb28d44bbd68b5a56

---
  - https://app.asana.com/0/0/1209609847049253


larskuhtz (changes requested):
consider adding a test case for covering that the work header now has constant size and has the correct block hash record. Also, check that the hash depends on all adjacent parents.

larskuhtz (approved):

src/Chainweb/BlockHash.hs
```diff
@@ -264,6 +269,25 @@ blockHashRecordFromVector g cid = BlockHashRecord
...
+convertBlockHashRecordForMining :: BlockHashRecord -> BlockHashRecord
+convertBlockHashRecordForMining (BlockHashRecord r) =
+    BlockHashRecord $ HM.fromList
+        $ over _Snoc (\(i, l) ->
+            ( i & mapped . _2 .~ nullBlockHash
+            , l & _2 .~ BlockHash (unsafeMerkleLogHash blockRecordHash)
+            )
+        )
+        -- this is required for compatibility; we cannot enlarge the record beyond that
+        $ take (int @Natural @Int $ degree petersonChainGraph)
```
larskuhtz, line 284:
```suggestion
        -- this is the degree of the Petersen graph. It is required for compatibility;
        -- we cannot enlarge the record beyond that
        $ take 3
```

larskuhtz (approved):

larskuhtz (approved):

PR #2160, opened 2025-05-05
Allow pow to be disabled by default in chainweb versions

edmundnoble:
Change-Id: Id0000000e1649a5b99329990bf39c9395e7cac69

### DIFF
--- a/src/Chainweb/Chainweb/Configuration.hs
+++ b/src/Chainweb/Chainweb/Configuration.hs
@@ -596,7 +596,7 @@ parseVersion = constructVersion
         )
     <*> optional (textOption @Fork (long "fork-upper-bound" <> help "(development mode only) the latest fork the node will enable"))
     <*> optional (BlockDelay <$> textOption (long "block-delay" <> help "(development mode only) the block delay in seconds per block"))
-    <*> switch (long "disable-pow" <> help "(development mode only) disable proof of work check")
+    <*> optional (switch (long "disable-pow" <> help "(development mode only) disable proof of work check"))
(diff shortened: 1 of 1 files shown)


src/Chainweb/Chainweb/Configuration.hs
```diff
@@ -596,7 +596,7 @@ parseVersion = constructVersion
         )
     <*> optional (textOption @Fork (long "fork-upper-bound" <> help "(development mode only) the latest fork the node will enable"))
     <*> optional (BlockDelay <$> textOption (long "block-delay" <> help "(development mode only) the block delay in seconds per block"))
-    <*> switch (long "disable-pow" <> help "(development mode only) disable proof of work check")
+    <*> optional (switch (long "disable-pow" <> help "(development mode only) disable proof of work check"))
```
larskuhtz, line 599: Is there a reason why `parseVersion` is not following the idiom for command lines parers from configuration-tools that is used everywhere else in Chainweb node? That idiom would prevent bugs like this one.

I guess, and configuration of `ChainwebVersion` is special and doesn't fit that pattern? Or is it just an oversight?

larskuhtz (approved):
Approved with one more general question about the command line parser for ChainwebVersion.

src/Chainweb/Chainweb/Configuration.hs
edmundnoble: Yes the weirdness here is that we parse the chainweb version, and then we parse a bunch of options that also alter this chainweb version.

PR #2161, opened 2025-05-06
Pact payload provider initialization

edmundnoble:


cabal.project.freeze
larskuhtz, line 1: I think, eventually we'll have to decide whether we want to get rid of `cabal.project.freeze` or whether we want to keep it. Otherwise we'll keep deleting it (which breaks CI) and readding it (to unbreak CI).

If we decide to get rid of it, we should do that in a separate commit to master and update CI (and the release workflows) accordingly.

PR #2164, opened 2025-05-09
Fix duplicate results from keys function

edmundnoble:


src/Chainweb/Pact5/Backend/ChainwebPactDb.hs
```diff
@@ -594,8 +594,14 @@ doKeys d = do
...
     case ordDict of
-        Dict () ->
-            return $ sort (memKeys ++ parsedKeys)
+        Dict () -> do
+            v <- view blockHandlerVersion
+            cid <- view blockHandlerChainId
+            bh <- view blockHandlerBlockHeight
+            if chainweb230Pact v cid bh
+            -- the read-cache contains duplicate keys that we need to remove.
+            then return $ fmap head $ group $ sort (memKeys ++ parsedKeys)
```
chessai, line 603: use unsafeHead to avoid GHC warnings about partial functions

chessai (changes requested):

src/Chainweb/Pact5/Backend/ChainwebPactDb.hs
```diff
@@ -82,7 +82,7 @@ import Chainweb.Pact.Backend.Utils
 import Chainweb.Utils (sshow, T2)
```
edmundnoble, line 82:
```suggestion
import Chainweb.Utils (sshow, unsafeHead, T2)
```

src/Chainweb/Pact5/Backend/ChainwebPactDb.hs
edmundnoble:
```suggestion
            then return $ fmap unsafeHead $ group $ sort (memKeys ++ parsedKeys)
```

PR #2177, opened 2025-07-03
Be lenient during rewind when reintroducing transactions into the mempool

larskuhtz:


src/Chainweb/Pact/PactService.hs
```diff
@@ -633,7 +622,7 @@ syncToFork logger serviceEnv hints forkInfo = do
     getRewoundTxs :: Parent BlockHeight -> IO (Vector Pact.Transaction)
     getRewoundTxs rewindTargetHeight = do
         rewoundPayloadHashes <- Checkpointer.getPayloadsAfter sql rewindTargetHeight
-        rewoundPayloads <- liftIO $ fmap fromJuste <$>
+        rewoundPayloads <- liftIO $ catMaybes <$>
```
larskuhtz, line 625: At the very least there should be some comment that explains the effects and reasons of this code.

There should probably also be a warning if the list contains `Nothing`, since, that should not happen under normal circumstances.

PR #2187, opened 2025-07-31
Rename chainwebVersion to networkId in config

edmundnoble:


edmundnoble:
Not 100% sure on this... this might be a step too far. We already have a concept of "network ID" internally, which is roughly "p2p network ID", because there are multiple p2p networks even on mainnet or testnet, and this is crucial for scaling.

PR #2196, opened 2025-09-15
Add RankedBlockHeaderDB and simplify TreeDB a bit

larskuhtz:
Add a `TreeDB` instance `RankedBlockHeader`. This `RankedBlockHeaderDB` is strictly more efficient than the current `BlockHeaderDB` and by implementing the latter in terms of the former, the performance of the current `BlockHeaderDB` is improved, too.

The ranked version should be preferred whenever possible, which is subject to an upcoming PR. We may even consider renaming `BlockHeaderDB` into `UnrankedBlockHeaderDB` and `RankedBlockHeaderDB` into `BlockHeaderDB`. The unranked version should be used only to provide legacy API endpoints of the service API.

This PR also reverts the type of `lookup` and `lookupRanked` back to returning a maybe `Maybe`, while retaining improved error reporting in other parts. In practice, the `Either` value was redundant for improved error handling. The left value was redundant, i.e. either empty or directly derived from the call parameters. (Database errors are reported separately via exceptions). Useful information about failure condition were added at call-site which is retained by this PR.

### DIFF
--- a/src/Chainweb/BlockHeaderDB/Internal.hs
+++ b/src/Chainweb/BlockHeaderDB/Internal.hs
@@ -11,7 +11,6 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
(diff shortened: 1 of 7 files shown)


src/Chainweb/BlockHeaderDB/Internal.hs
```diff
@@ -160,7 +201,30 @@ instance HasChainId BlockHeaderDb where
...
 
 instance (k ~ CasKeyType BlockHeader, HasVersion) => ReadableTable BlockHeaderDb k BlockHeader where
-    tableLookup db k = either (\_ -> Nothing) Just <$> lookup db k
+    tableLookup = lookup
+    {-# INLINE tableLookup #-}
+
+-- -------------------------------------------------------------------------- --
+-- RankedBlockHeaderDb
+
+-- | A rangked block header db uses the same underlying storage as a
```
edmundnoble, line 210:
```suggestion
-- | A ranged block header db uses the same underlying storage as a
```

edmundnoble (approved):
Approved with notes.

PR #2201, opened 2025-09-15
Multi-chain incremental pruning

edmundnoble:
Updates the existing pruning algorithm to operate on multiple chains in lockstep to avoid dangling references and to be possible to interrupt and later resume, as well as run periodically and incrementally.

### DIFF
--- a/src/Chainweb/WebBlockHeaderDB.hs
+++ b/src/Chainweb/WebBlockHeaderDB.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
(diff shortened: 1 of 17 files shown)


src/Chainweb/WebBlockHeaderDB.hs
```diff
@@ -100,7 +105,14 @@ webEntries db f = go (view (webBlockHeaderDb . to toList) db) mempty
     go [] s = f s
     go (h:t) s = entries h Nothing Nothing Nothing Nothing $ \x ->
         go t (() <$ S.mergeOn (view blockCreationTime) s x)
-            -- FIXME: should we include the rank in the order?
+
+-- | Returns all blocks in all block header databases.
```
larskuhtz: Add a comment why the block creation time is not working here. Maybe also add a comment about the order of blocks.

src/Chainweb/WebBlockHeaderDB.hs
```diff
@@ -19,12 +20,13 @@
 -- Collect the 'BlockHeaderDB's for all chains in a single data structure.
 --
```
larskuhtz, line 21: Add comment that this now has its own state

PR #2203, opened 2025-09-15
add minimal block header history constant

chessai:
also sorted imports with ImportQualifiedPost

### DIFF
--- a/src/Chainweb/SPV.hs
+++ b/src/Chainweb/SPV.hs
@@ -3,6 +3,8 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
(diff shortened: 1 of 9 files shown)


src/Chainweb/SPV.hs
```diff
@@ -212,3 +209,8 @@ instance FromJSON (TransactionOutputProof SHA512t_256) where
 --
 outputProofChainId :: Getter (TransactionOutputProof a) ChainId
 outputProofChainId = to (\(TransactionOutputProof cid _) -> cid)
+
+-- A reasonable block history window for an SPV header oracle
+-- to look back over.
+defaultSpvProofExpirationWindow :: Word64
+defaultSpvProofExpirationWindow = 20_000
```
edmundnoble, line 216: I'd like this moved out even further, maybe even to its own file if necessary, as the minimum header oracle size.

edmundnoble (approved):

PR #2220, opened 2025-10-03
Fix ForkInfo Sync

larskuhtz:


src/Chainweb/CutDB.hs
```diff
@@ -481,110 +480,64 @@ startCutDb config logger headerStore providers cutHashesStore = mask_ $ do
...
     :: (Logger logger, HasVersion)
-    => logger -> WebBlockHeaderDb -> ChainMap ConfiguredPayloadProvider -> Cut -> IO Cut
+    => logger
+    -> WebBlockHeaderDb
+    -> ChainMap ConfiguredPayloadProvider
+    -> Cut
+    -> IO Cut
 synchronizeProviders logger wbh providers c = do
     let startHeaders = HM.unionWith
-            (\startHeader _genesisHeader -> startHeader)
```
rsoeldner, line 489: this should be equal to `HM.union (_cutHeaders c) $ (imap (\cid () -> genesisBlockHeader cid) (HS.toMap chainIds)`

src/Chainweb/CutDB.hs
```diff
@@ -481,110 +480,64 @@ startCutDb config logger headerStore providers cutHashesStore = mask_ $ do
...
-    else do
+      then do
+        return c
+      else do
         -- try to recover from the fork automatically by removing ~`diameter`
         -- blocks from the cut
         let recoveryHeight =
                 max (int (diameter (chainGraphAt maxBound))) (_cutMinHeight c) - int (diameter (chainGraphAt maxBound))
         recoveryCut <- limitCut wbh recoveryHeight c
         let recoveryHeaders = HM.unionWith
```
rsoeldner, line 511: same here

src/Chainweb/Pact/PactService.hs
```diff
@@ -509,112 +509,129 @@ syncToFork logger serviceEnv hints forkInfo = do
...
+            --
+            latestBlockRewindable <-
+                isJust <$> Checkpointer.lookupBlockHash sql (_latestBlockHash forkInfo._forkInfoTargetState)
+
+            if latestBlockRewindable
+              then do
+                -- we just have to rewind and set the final + safe blocks
+                -- TODO PP: disallow rewinding final?
+                logFunctionText logger Debug $ "pure rewind to " <> brief forkInfo._forkInfoTargetState
+                rewoundTxs <- getRewoundTxs (Parent forkInfo._forkInfoTargetState._consensusStateLatest._syncStateHeight)
```
rsoeldner, line 535: I actually really dislike `OverloadedRecordDot`, can we use lenses here?

src/Chainweb/Pact/PactService.hs
```diff
@@ -509,112 +509,129 @@ syncToFork logger serviceEnv hints forkInfo = do
...
+                        rewoundTxs <- getRewoundTxs (Parent forkInfo._forkInfoTargetState._consensusStateLatest._syncStateHeight)
+                        -- the happy case: we can find a way to get to the target block
+                        -- look up all of the payloads to see if we've run them before
+                        -- even then we still have to run them, because they aren't in the checkpointer
+                        knownPayloads <- liftIO $
+                            tableLookupBatch' pdb (each . _2) ((\e -> (e, _evaluationCtxRankedPayloadHash $ fst e)) <$> forkChainBottomToTop)
+
+                        let unknownPayloads = NEL.filter (isNothing . snd) knownPayloads
+                        unless (null unknownPayloads)
+                            $ logFunctionText logger Debug $ "unknown blocks in context: " <> sshow (length unknownPayloads)
```
rsoeldner, line 577:
```suggestion
                            $ logFunctionText logger Debug $ "unknown blocks in context: " <> sshow (length unknownPayloads) <> " " <> sshow unkownPayloads
```

src/Chainweb/ChainValue.hs
```diff
@@ -63,6 +58,12 @@ instance TraversableWithIndex ChainId ChainValue where
 instance FoldableWithIndex ChainId ChainValue
 instance FunctorWithIndex ChainId ChainValue
 
+instance HasTextRepresentation a => HasTextRepresentation (ChainValue a) where
+    toText (ChainValue cid a) = toText cid <> ":" <> toText a
+    fromText t = case T.breakOn ":" t of
+        (c, r) -> ChainValue <$> fromText c <*> fromText (T.drop 1 r)
```
rsoeldner, line 64:
```suggestion
        (c, r) -> ChainValue <$> fromText c <*> fromText (T.tail r)
```

src/Chainweb/Pact/PactService.hs
larskuhtz: It's used a lot in that module. It also seems to break HLS in my editor. Yeah, maybe we should replace it by "normal" function composition via ` . `.

edmundnoble (changes requested):
I can see these pieces in this PR so far:
- resolveForkInfo added and used in CutDB and WebBlockHeaderStore
  - includes ReadableBlockCas
- payload batch fetching
- payload prefetching in catchup, using batch fetching
- forkInfoForHeader has been cleaned up a bit
- candidate header store is now used in getBlockHeader
- prefetching is now done during syncToBlock for large traces
- potential fix to EVM PP sync issues

src/Chainweb/BlockHash.hs
```diff
@@ -355,3 +355,4 @@ instance FromJSON RankedBlockHash where
         <$> o .: "height"
         <*> o .: "hash"
     {-# INLINE parseJSON #-}
+
```
larskuhtz, line 358:
```suggestion
```

src/Chainweb/PayloadProvider/EVM.hs
```diff
@@ -822,11 +835,20 @@ newPayload
...
+            lf Warn "newPayload timed out while EVM is syncing"
+
+            -- FIXME FIXME FIXME
+            -- this is an experiment. Check comment in validatePayload for
+            -- details. It seems that when reth returns SYNCING we are actually
+            -- stuck (it does not resolve itself). So, we try to force it to
+            -- sync and hope that when we come back the newPaylaod call is
+            -- successful.
+            -- updateEvm p trgState Nothing []
+
```
larskuhtz, line 851:
```suggestion
```

src/Chainweb/PayloadProvider/Minimal.hs
```diff
@@ -385,8 +427,10 @@ minimalPrefetchPayloads
     -> [Ranked ConsensusPayload]
     -> IO ()
 minimalPrefetchPayloads p h ps = do
-    logg p Debug "prefetch payloads"
-    mapConcurrently_ (try @_ @TaskException . getPayloadForContext p h) ps
+    logg p Info $ "prefetch payloads: query " <> sshow (length ps)
+    -- mapConcurrently_ (try @_ @TaskException . getPayloadForContext p h) ps
```
larskuhtz, line 431:
```suggestion
```

src/Chainweb/Ranked.hs
```diff
@@ -84,6 +83,15 @@ decodeRanked decodeA = Ranked
     <*> decodeA
 {-# INLINE decodeRanked #-}
 
+instance HasTextRepresentation a => HasTextRepresentation (Ranked a) where
+    toText a = toText (_rankedHeight a) <> "." <> toText (_ranked a)
+    fromText t = case T.break (== '.') t of
```
larskuhtz, line 88:
```suggestion
    fromText t = case T.breakOn "." t of
```

src/Chainweb/Sync/ForkInfo.hs
```diff
@@ -0,0 +1,288 @@
...
+                -- to the payload provider protocol, the payload provider may
+                -- process the ForkInfo only partially, in which case the result
+                -- is our branch and the forkpoint is forwarded. If we find that
+                -- the result is on a fork it means that there is interference.
+                -- We may check whether we still made progress, in which case we
+                -- are fine. If we find that we do not make progress we raise an
+                -- error.
+
+                (forkBlocksDescendingStream S.:> forkPoint) <- S.toList
+                    $ branchDiff_ bhdb ppBlock hdr
```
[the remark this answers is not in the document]
larskuhtz: At the moment it is guaranteed that the parent is in the header db -- because we never process a header which has a parent that is not yet validated.

For now it might be the best to just add a check and throw an informative error when this condition in violated -- which should only happen when we refactor cut processing, which would probably include refactoring this module anyways.

src/Chainweb/PayloadProvider/EVM.hs
```diff
@@ -1321,28 +1343,127 @@ evmSyncToBlock p hints forkInfo = withLock (_evmLock p) $ do
...
 
                     -- fetch all unknown payloads
                     --
                     -- FIXME do the right thing here. Ideally, fetch all
                     -- unknowns in batches without redundant local lookups. Then
                     -- validate all payloads together before sending them to the
                     -- EVM and inserting them into the DB.
+
+
+                    -- Before we proceed to the validation we need to make sure
```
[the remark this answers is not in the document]
larskuhtz: I guess the comment in the code focuses too much on `newPayload` -- probably it was the how the issue manifested itself and helped understanding the issue during debugging.

The underlying issue is that the EVM and the Payload Provider can get out of sync. This is a rare, yet real, corner case scenario and it means that fork point does not reflect the actual state of the EVM.

The call to `newPayload` is just a concrete example where this invalid assumption about the state of the EVM leads to unexpected behavior when interacting with the EVM. But it is possible that this can cause other issues, too.

It probably is generally a good idea to detect the condition when the EVM is out of sync and correct it with a simple   (and relatively cheap) `updateEvm` call.

the `updateEvm` call is guarded by an inexpensive check, so that that the inner code is executed only in the rare cases when the EVM is actually out of sync.

PR #2273, opened 2025-10-21
remove service date

chessai:


edmundnoble (dismissed):

jmcardon (dismissed):

larskuhtz:
I propose that we hold merging (or at least releasing) this until #2272 is merged, too.

It would be difficult to introduce #2272 without a coordinated fork.

larskuhtz (changes requested):
While I approve the change, I propose that we don't merge that just yet.

If we decide to move ahead with #2272, we should make sure that we release and roll out both changes at the same time.

